PR #4, opened 2022-03-29
example async/await refactor - UserController

scottyeck:


api/controllers/UserController.js
```diff
@@ -12,96 +12,85 @@
...
-
-        return res.ok();
-      })
-      .catch(err => {
-        return res.badRequest(err);
+export async function signup(req, res) {
+  // Request was missing data
+  if (!req.body.password && !req.body.username) {
+    return res.badRequest('You did not submit a username or password');
+  }
```
itsalaidbacklife, line 23: I refactored this yesterday and am only now realizing:

1. The conditional is wrong because the error was intended to fire if either field was missing (should be `||`)
2. This conditional is actually altogether unnecessary because the `hasValidUsername` and `hasValidPassword` policies run before this action is called so this code is actually unreachable haha

itsalaidbacklife: Removed

itsalaidbacklife:
This has been abandoned as a part of the typescript implementation

PR #22, opened 2022-05-26
Update: TargetSelectionOverlay now contains an X button at the top ri…

brianEtrials:
image

image

e2e tests [X]
unit tests []
Tested myself [X]
Tested by others []

### DIFF
--- a/client/js/components/GameView/TargetSelectionOverlay.vue
+++ b/client/js/components/GameView/TargetSelectionOverlay.vue
@@ -17,13 +17,10 @@
       </v-btn>
       
     </div>
-    <v-btn
-        color="primary"
-        text
-        data-cy="cancel-target-mobile"
-        @click="$emit('cancel')"
-      >
-        X
+    <v-btn icon data-cy="cancel-target-mobile" @click="$emit('cancel')">
+      <v-icon x-large>
+        mdi-close
+      </v-icon>
     </v-btn>
   </div>
 </template>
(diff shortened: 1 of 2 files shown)


client/js/components/GameView/TargetSelectionOverlay.vue
```diff
@@ -15,7 +15,16 @@
...
+      
     </div>
+    <v-btn
+        color="primary"
+        text
+        data-cy="cancel-target-mobile"
+        @click="$emit('cancel')"
+      >
+        X
+    </v-btn>
```
itsalaidbacklife, line 27: This works, but let's make it more consistent with the X button in the `MoveChoiceOverlay` by making it an icon button and using the `mdi-close` icon instead of the letter X. Let's also make the icon extra large:

```
      <v-btn icon data-cy="cancel-target-mobile" @click="$emit('cancel')">
        <v-icon x-large>
          mdi-close
        </v-icon>
      </v-btn>
```

itsalaidbacklife (changes requested):
Nice! Everything works and your test is actually more thorough than I'd originally imagined, in a good way! I'm requesting two changes:
1. Update the button itself to be an icon which will make it look better
2. Remove a chunk of the new test code that checks canceling your move-choice (not targeting) because that's the exact same as the original test so we don't need to do it twice

itsalaidbacklife (approved):
Nice work. You're first official cuttle contribution :)

PR #37, opened 2022-06-03
Fix ESLint Configuration, resolves #36

### LINKED ISSUE #36: [Bug]: Fix ESLint configuration
### Bug Summary

ESLint has unpredicatable results when run on SFCs in VSCode

### Detailed Description

When running ESLint from VSCode each run will format slightly differently. I think that the linters are fighting with one another.

### Current Behavior

image

### Expected Behavior

Something like this:
image

### Steps to Reproduce

1. Pull down the repo
2. Run npm install locally
3. Open `client/js/views/Home.vue` and try linting from VSCode (Command+P, then `>eslint`)

seriouslysean:
⚠️ This is still a WIP for now, comments inline with each of the changes ⚠️ 

@itsalaidbacklife If you pull this branch and run `npm ci`, then restart your vscode eslint server, all of the linting should work again inside of VSCode. I was testing on the `client/js/views/Home.vue` file.

I haven't fixed any of the linting issues yet, that should probably be a separate PR off of this one, maybe convert all errors to warnings for now to limit the changeset.

Without any other changes, we're looking at the following output:
```
✖ 283 problems (278 errors, 5 warnings)
  264 errors and 0 warnings potentially fixable with the `--fix` option.
  ```

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -1,11 +1,27 @@
 {
-  "name": "Cuttle",
-  "private": true,
+  "name": "cuttle",
```
seriouslysean, line 2: This goes against the `package.json` naming convention for the `name` property, so I lowercased it while I was in there.

package.json
```diff
@@ -1,11 +1,27 @@
...
   "version": "3.0.2",
+  "private": true,
```
seriouslysean, line 4: Part of running `vue add eslint` reorganizes the package.json.

package.json
```diff
@@ -1,11 +1,27 @@
...
+    "start:server": "node app.js",
+    "- lint": "# --------------------------",
+    "lint": "vue-cli-service lint --no-fix",
+    "lint:fix": "vue-cli-service lint",
+    "- format": "# ------------------------",
+    "format:js": "./scripts/format-project-js",
+    "- e2e": "# ---------------------------",
+    "e2e:client": "npx cypress run",
+    "e2e:server": "npx cypress run --config baseUrl=http://localhost:1337",
+    "e2e:gui": "npx cypress open"
```
seriouslysean, line 22: You probably will want to redo these at some point. The line breaks go against convention since those aren't actually scripts. I would opt for the prefix approach you have and just normalize it. Maybe change `start:client` to `client:start` and have them all prefixed with what they do?

The rest already seem to follow that pattern.

package.json
```diff
@@ -28,46 +44,33 @@
...
     "chai": "^3.5.0",
     "chai-as-promised": "^5.3.0",
     "cypress": "^9.5.2",
-    "eslint": "^6.7.2",
+    "eslint": "^7.32.0",
+    "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-cypress": "^2.12.1",
-    "eslint-plugin-prettier": "^3.1.3",
-    "eslint-plugin-vue": "^6.2.2",
-    "prettier": "^1.19.1",
```
seriouslysean, line 46: I started by removing every eslint related plugin and just re-added them by using `vue add eslint` and configuring it from the command line to make sure I wasn't adding any sort of extraneous package.

.eslintrc.json
```diff
@@ -31,13 +45,42 @@
   },
   "overrides": [
     {
-      "files": ["**/__tests__/*.{j,t}s?(x)", "**/tests/unit/**/*.spec.{j,t}s?(x)"],
+      "files": [
+        "**/tests/*.{j,t}s?(x)",
+        "**/tests/unit/**/*.{j,t}s?(x)"
+      ],
```
seriouslysean, line 51: I changed this to only target the unit tests for now.

.eslintrc.json
```diff
@@ -31,13 +45,42 @@
...
+      "globals": {
+        "badRequest": true,
+        "cardService": true,
+        "io": true,
+        "ready1": true,
+        "request": true,
+        "socket1": true,
+        "socket2": true,
+        "socket3": true
+      },
```
seriouslysean, line 64: Due to the way you instantiate your tests with the `bootstrap.test.js` file I can not figure out a better way to make these globals not throw a `no-undef` error so I just opted to add them here as _known_ globals for these files.

.eslintrc.json
```diff
@@ -31,13 +45,42 @@
...
+      "rules": {
+        "import/no-unresolved": [
+          "off"
+        ]
+      }
+    },
+    {
+      "files": [
+        "**/tests/e2e/**/*.{j,t}s?(x)"
+      ],
```
seriouslysean, line 74: Also added a second entry for the e2e folder which would needed cypress specific additions (`env` and `plugin` which already exists above).

.eslintrc.json
itsalaidbacklife: Okay interesting. The unit tests with supertest aren't actually even used in CI so it may be best to simply throw them out altogether and rely entirely on e2e with cypress. I'll do some digging

seriouslysean:
@itsalaidbacklife repointed to `feat/linting`. This can probably go in as-is now and then you and I can talk next steps. My suggestion would be something like this:
1. Remove unused tests (or outdated/invalid tests)
2. Refactor the testing setup for unit/functional tests (do we need the bootstrap file, etc)
3. Run linting with the `--fix` flag
4. Fix any remaining issues
5. Merge each of these solves in to the new `feat/linting` branch
6. Merge the `feat/linting` branch in to the `main` branch

.eslintrc.json
```diff
@@ -4,11 +4,21 @@
...
-  "extends": ["plugin:vue/recommended"],
-  "plugins": ["prettier"],
+  "extends": [
+    "plugin:vue/recommended",
+    "plugin:vue/essential",
+    "eslint:recommended",
+    "plugin:prettier/recommended"
+  ],
+  "plugins": [
+    "jest",
```
itsalaidbacklife, line 14: Ha this highlights a stale dependency. There are no jest unit tests so we can omit this and actually remove jest as a dev dependency altogether

package.json
itsalaidbacklife: Definitely fine to remove the line breaks. My naming convention for the individual scripts has been about which type of command it is and then which application to do it to, if that makes sense?

so `e2e:client` means run the e2e suite against port 8080 b/c that's where the client runs
and `e2e:server` runs the e2e suite against port 1337 b/c that's the (backend) server's port

the same goes for `start`; `start:server` boots the server (on port 1337) and `start:client` boots the client (on port 8080). Is that inconsistent?

package.json
itsalaidbacklife: Great! We should be able to remove the jest one again because that's a stale dev dependency. When I get to unit testing again I can do it with mocha which already has some tests written

package.json
```diff
@@ -28,46 +44,33 @@
     "vuex": "^3.5.1"
   },
   "devDependencies": {
+    "@babel/core": "^7.12.16",
+    "@babel/eslint-parser": "^7.12.16",
     "@vue/cli-plugin-babel": "~4.5.0",
-    "@vue/cli-plugin-eslint": "^4.5.4",
+    "@vue/cli-plugin-eslint": "~5.0.0",
     "@vue/cli-plugin-router": "^4.5.4",
     "@vue/cli-plugin-unit-jest": "^4.5.6",
```
itsalaidbacklife, line 52: Let's remove this dep

package.json
```diff
@@ -28,46 +44,33 @@
...
     "chai-as-promised": "^5.3.0",
     "cypress": "^9.5.2",
-    "eslint": "^6.7.2",
+    "eslint": "^7.32.0",
+    "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-cypress": "^2.12.1",
-    "eslint-plugin-prettier": "^3.1.3",
-    "eslint-plugin-vue": "^6.2.2",
-    "prettier": "^1.19.1",
+    "eslint-plugin-jest": "^26.4.6",
```
itsalaidbacklife, line 62: Same; let's remove

package.json
itsalaidbacklife: Nevermind; we'll revisit this when we fix up the unit tests

package.json
itsalaidbacklife: nevermind; revisit this later

itsalaidbacklife (approved):
Great start. Next up we'll wrangle the tests?

PR #40, opened 2022-06-15
Add standalone Vue DevTools for debugging purposes

seriouslysean:
## Please check the following

- For New Features:

Adds Vue Devtools package that can be run with `dev:devtools`, and adds a script to the top of the dev build using the webpack html loader plugin.

image

Screenshot of the devtools running locally 🎉 

-----

@itsalaidbacklife will want your 👀 on this to verify I didn't break anything, and to double check what the prod version will look like since I'm not 100% familiar with the tooling yet.


client/public/index.html
```diff
@@ -14,6 +14,7 @@
       rel="stylesheet"
       href="https://cdn.jsdelivr.net/npm/@mdi/font@latest/css/materialdesignicons.min.css"
     />
+    <%= htmlWebpackPlugin.options.vueDevToolsScript %>
```
seriouslysean, line 17: New key that will add the devtools to the dev build

package.json
```diff
@@ -67,7 +68,8 @@
     "- e2e": "# ---------------------------",
     "e2e:client": "npx cypress run",
     "e2e:server": "npx cypress run --config baseUrl=http://localhost:1337",
-    "e2e:gui": "npx cypress open"
+    "e2e:gui": "npx cypress open",
+    "dev:devtools": "vue-devtools"
```
seriouslysean, line 72: Command to start up the standalone Vue devtools app

vue.config.js
```diff
@@ -1,24 +1,30 @@
...
-  outputDir: 'assets',
   pages: {
     index: {
       entry: 'client/js/main.js',
       template: 'client/public/index.html',
       filename: 'index.html',
+      // https://devtools.vuejs.org/guide/installation.html#standalone
+      ...(process.env.NODE_ENV != 'production' ? {
+        vueDevToolsScript: '<script src="http://localhost:8098"></script>',
+      } : {}),
```
seriouslysean, line 19: Only add the script on non prod builds (local/dev/staging)

vue.config.js
```diff
@@ -1,24 +1,30 @@
...
   transpileDependencies: ['vuetify'],
-  configureWebpack: {
-    devtool: 'source-map',
-  },
+
   outputDir: process.env.VUE_APP_BUILD_OUTPUT || 'dist',
 
   lintOnSave: false,
 
-  outputDir: 'assets',
```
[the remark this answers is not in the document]
seriouslysean: Nevermind, I fixed it in the latest push. It was a duplicate key issue but I figured out what you were intending and instead created a separate `prod` script that will set the proper output directory environment variable and then build and serve the app properly.

vue.config.js
```diff
@@ -1,24 +1,34 @@
...
     },
   },
 
   configureWebpack: {
+    // https://blog.scottlogic.com/2017/11/01/webpack-source-map-options-quick-guide.html
+    ...(process.env.NODE_ENV != 'production'
+      ? {
+          devtool: 'cheap-module-eval-source-map',
+        }
+      : {}),
```
seriouslysean, line 31: No value for prod builds mean they won't get sourcemaps at all.

client/js/components/HelloWorld.vue
```diff
@@ -1,134 +0,0 @@
-<template>
```
seriouslysean, line 1: This was never being used and was a leftover file from the Vue CLI 🙃

package.json
```diff
@@ -54,7 +55,7 @@
   "scripts": {
     "- client": "# ------------------------",
     "start:client": "vue-cli-service serve",
-    "build": "vue-cli-service build;cp ./client/public/favicon.ico ./assets",
+    "build": "VUE_APP_BUILD_OUTPUT=assets vue-cli-service build && cp ./client/public/favicon.ico ./assets",
```
seriouslysean, line 58: This fixes that duplicate key issue I noticed, removed the dupe key and just set the proper env var here instead.

tests/e2e/specs/out_of_game/home.spec.js
```diff
@@ -18,6 +18,18 @@ function assertSuccessfulJoin(gameState) {
   cy.contains('h1', `Lobby for ${gameState.name}`);
 }
 
+describe('Vue Devtools', () => {
```
seriouslysean, line 21: I added an issue for this, https://github.com/cuttle-cards/cuttle/issues/41. I think that there needs to be a standard around when e2e tests are run. It feels like only against a prod build makes sense so we can properly assert here that the dev script doesn't exist.

That said, as of today you can run either a dev build OR a prod build and run e2e tests against either. I know you use it as a way to jump in to a given scenario so maybe there's a middleground here I'm not thinking of?

cypress.config.js
```diff
@@ -0,0 +1,11 @@
+const { defineConfig } = require('cypress');
```
seriouslysean, line 1: Cypress docs suggest moving from the json file to `cypress.config.js` or `cypress.config.ts`. I needed to do this so we could pipe down the `NODE_ENV` environment variable to the Cypress runner.

package.json
```diff
@@ -65,9 +66,11 @@
     "- format": "# ------------------------",
     "format:js": "./scripts/format-project-js",
     "- e2e": "# ---------------------------",
-    "e2e:client": "npx cypress run",
-    "e2e:server": "npx cypress run --config baseUrl=http://localhost:1337",
-    "e2e:gui": "npx cypress open"
+    "e2e:client": "cypress run",
+    "e2e:server": "NODE_ENV=production cypress run --config baseUrl=http://localhost:1337",
```
seriouslysean, line 70: This ensures that if you run Cypress against the prod build it sets NODE_ENV internally so that the Vue Devtools test works properly.

tests/e2e/specs/out_of_game/home.spec.js
```diff
@@ -18,6 +18,20 @@ function assertSuccessfulJoin(gameState) {
...
+    beforeEach(setup);
+
+    if (Cypress.env('NODE_ENV') === 'production') {
+      it('Should not include Vue Devtools script on prod builds', () => {
+        cy.get('[data-vue-devtools]').should('have.length', 0);
+      });
```
seriouslysean, line 27:
```
prod:start
npm run e2e:server -- --spec tests/e2e/specs/out_of_game/home.spec.js
```
Screen Shot 2022-06-15 at 3 44 57 PM

tests/e2e/specs/out_of_game/home.spec.js
```diff
@@ -18,6 +18,20 @@ function assertSuccessfulJoin(gameState) {
...
+    } else {
+      it('Should include Vue Devtools script on dev builds', () => {
+        cy.get('[data-vue-devtools]').should('have.length', 1);
+      });
```
seriouslysean, line 31:
```
npx concurrently "npm run start:server" "npm run start:client"
npm run e2e:client --spec tests/e2e/specs/out_of_game/home.spec.js
```
Screen Shot 2022-06-15 at 3 38 58 PM

cypress.json
```diff
@@ -1,6 +0,0 @@
-{
-  "pluginsFile": "tests/e2e/plugins/index.js",
-  "baseUrl": "http://localhost:8080",
-  "testFiles": ["out_of_game/**/*.js", "in_game/**/*.js"],
-  "video": false
-}
```
itsalaidbacklife, line 6: We'll need to be smarter about telling cypress where to load the config file from in its github action. From the looks of the logs, CI failed because cypress failed to find the config file when looking for cypress.json

package.json
itsalaidbacklife: Would love to discuss this further. Some weird consequences of node env settings I want to touch base on. In particular, currently, App.vue() adds the root Vue instance to the window object during testing (when `if(winddow.Cypress) {}...`) without checking node env, but I'd been planning to only expose the app if not in prod, which would break many of the tests, as they rely heavily on access to the vuex store through the root application instance

tests/e2e/specs/out_of_game/home.spec.js
itsalaidbacklife: I think you're right and this is speaking to the value of programatic environment-specific test configuration which runs different tests depending on the environment. This is tricky stuff we'll want to think hard on and I'm really excited to dig into it together

vue.config.js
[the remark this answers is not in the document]
itsalaidbacklife: yeeeeeah my webpack foo is pretty weak 😅

vue.config.js
itsalaidbacklife: nice ty

package.json
```diff
@@ -55,16 +56,19 @@
...
+    "e2e:server": "VUE_APP_API_URL=http://localhost:1337 npx cypress run",
     "format:js": "./scripts/format-project-js",
     "lint:fix": "vue-cli-service lint",
     "lint": "vue-cli-service lint --no-fix",
     "start": "node app.js",
-    "start:client": "vue-cli-service serve",
+    "start:client": "ENABLE_VUE_DEVTOOLS=true vue-cli-service serve",
     "start:server": "node app.js",
-    "start:dev": "concurrently \"npm run start:server\" \"npm run start:client\""
+    "start:dev": "concurrently \"npm run start:server\" \"npm run start:client\"",
```
seriouslysean, line 69: This will implicitly get `ENABLE_VUE_DEVTOOLS=true` from the `npm run start:client` script.

package.json
```diff
@@ -55,16 +56,19 @@
...
+    "prod": "npm run build && npm run start:server",
+    "prod:start": "npm run build && npm run start:server"
```
seriouslysean, line 71: These will implicitly get `ENABLE_VUE_DEVTOOLS=false` since the value is not set for these runs.

tests/e2e/specs/out_of_game/home.spec.js
```diff
@@ -18,6 +18,26 @@ function assertSuccessfulJoin(gameState) {
...
+  if (Cypress.env('ENABLE_VUE_DEVTOOLS') === 'true') {
+    it('Should include Vue Devtools when ENABLE_VUE_DEVTOOLS is true', () => {
+      cy.get('[data-vue-devtools]').should('have.length', 1);
+    });
+  } else {
+    it('Should not include Vue Devtools when ENABLE_VUE_DEVTOOLS is not true', () => {
+      cy.get('[data-vue-devtools]').should('have.length', 0);
+    });
+  }
+});
```
seriouslysean, line 39: Tried to explain as best as I could, but we still need to fork the logic in these tests due to the env being build time and related to how your index.js file is being generated.

That said, you can see the successful runs here:

| GUI | Client | Server |
|-----|-------|-------|
| Screen Shot 2022-06-22 at 3 44 29 PM | Screen Shot 2022-06-22 at 3 48 14 PM |  Screen Shot 2022-06-22 at 3 46 42 PM |

vue.config.js
```diff
@@ -1,24 +1,36 @@
...
+const htmlWebpackPlugin = require('html-webpack-plugin');
 
 module.exports = {
   transpileDependencies: ['vuetify'],
-  configureWebpack: {
-    devtool: 'source-map',
-  },
+
+  // `assets` is set in the build script, and in the `.env` file as a default
+  // `dist` is the vue cli default value
```
seriouslysean, line 8: This explains the usage better 🙃

vue.config.js
```diff
@@ -1,24 +1,36 @@
...
     index: {
       entry: 'client/js/main.js',
       template: 'client/public/index.html',
       filename: 'index.html',
+      // https://devtools.vuejs.org/guide/installation.html#standalone
+      ...(process.env.ENABLE_VUE_DEVTOOLS === 'true'
+        ? {
+            vueDevToolsScript: '<script src="http://localhost:8098" data-vue-devtools></script>',
+          }
+        : {}),
```
seriouslysean, line 23: Since the vue cli is interpolating this in to the index.html file, it's webpack under the hook and inherently build time. We can probably look in to moving this stuff to client side eventually, but I'll need to do more research around how/where to do it.

itsalaidbacklife (changes requested):
Awesome QoL improvement. Requesting some minor cleanup. I'd still like to put our heads together further on the conditional testing/environment-specific testing issue

.env
```diff
@@ -1 +1,4 @@
+CUTTLE_ENV=dev
```
seriouslysean, line 1: Moved to a new `CUTTLE_ENV` env variable which makes more sense since it's not going to put sails in actual prod mode, but lets us do prod-like stuff.

tests/e2e/specs/global/tools.spec.prod.js
```diff
@@ -0,0 +1,12 @@
+function setup() {
```
seriouslysean, line 1: 🎉

tests/unit/utils/config-utils.spec.js
```diff
@@ -0,0 +1,57 @@
+describe('isProd', () => {
```
seriouslysean, line 1: I deprecated all the current unit tests since they were't being used as far as I can tell in any of your tests. We can put them back of course but I wanted to check with you first.

That said, I updated the jest implementation so that I could add some tests for my `isProd` utility. You can test it with `npm run test:unit`.

utils/config-utils.js
```diff
@@ -0,0 +1,10 @@
+// `CUTTLE_ENV` is needed to test prod specific files without actually putting the app in to
+// a production state, which requires an actual database connection
+const isProd =
+  process.env.CUTTLE_ENV !== undefined
+    ? process.env.CUTTLE_ENV === 'production'
+    : process.env.NODE_ENV === 'production';
```
seriouslysean, line 6: This is a new utility I added for checking for a prod environment, it will prefer the `CUTTLE_ENV` if set, otherwise if it is not set it will default to whatever value is in the NODE_ENV. Once we figure out all of the `NODE_ENV` stuff we can probably get rid of this abstraction.

.github/workflows/lint-build-test.yml
```diff
@@ -15,18 +15,25 @@ jobs:
         uses: actions/checkout@v2
       # Install NPM dependencies, cache them correctly
       - name: Install dependencies
-        uses: cypress-io/github-action@v2
+        uses: cypress-io/github-action@v4
```
itsalaidbacklife, line 18: Shall we run the unit tests before e2e in CI?

itsalaidbacklife: Should be as simple as

`run: npm run test:unit`

(I'm not 100% whether we need to escape the : in test:unit)

itsalaidbacklife (approved):
Great work. This will be such a huge QoL improvement. Before we merge, would you please:

1) add devtools to start:client
2) run unit tests in CI

PR #42, opened 2022-06-16
Adjust .vscode workspace settings, add suggested plugins and settings

seriouslysean:
## Please check the following

- For New Features:
  - [x] ~Have tests been added to cover any new features or fixes?~
  - [x] ~Has the documentation been updated accordingly?~

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,8 @@
+{
+  "recommendations": [
+    "vue.volar",
+    "vue.vscode-typescript-vue-plugin",
+    "esbenp.prettier-vscode",
+    "dbaeumer.vscode-eslint"
+  ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,7 @@
 {
   "editor.tabSize": 2,
-  "vetur.format.options.useTabs": true,
-  "prettier.enable": false,
-  "prettier.useTabs": true
+  "prettier.enable": true,
+  "prettier.useTabs": true,
+  "files.insertFinalNewline": true,
+  "files.trimFinalNewlines": true
 }
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "@vue/cli-plugin-unit-jest": "^4.5.6",
     "@vue/cli-service": "~4.5.0",
     "@vue/eslint-config-prettier": "^6.0.0",
+    "@vue/runtime-dom": "^3.2.37",
     "@vue/test-utils": "^1.0.3",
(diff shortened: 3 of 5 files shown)


.vscode/extensions.json
```diff
@@ -0,0 +1,8 @@
+{
```
seriouslysean, line 1: These are my best guess on suggested extensions. Vetur might be better at this point, though I did make a few changes so that Volar would work (which seems to be the preferred extension now).

.vscode/settings.json
```diff
@@ -1,6 +1,7 @@
 {
   "editor.tabSize": 2,
-  "vetur.format.options.useTabs": true,
```
seriouslysean, line 3: In moving to Volar, this would be deprecated.

.vscode/settings.json
```diff
@@ -1,6 +1,7 @@
...
-  "prettier.enable": false,
-  "prettier.useTabs": true
+  "prettier.enable": true,
+  "prettier.useTabs": true,
+  "files.insertFinalNewline": true,
+  "files.trimFinalNewlines": true
```
seriouslysean, line 6: I use these and am personally a fan, though I can see now wanting the editor to modify the files. If we don't go this route, it might make sense to have eslint rules to enforce the standard instead.

package.json
```diff
@@ -34,6 +34,7 @@
     "@vue/cli-plugin-unit-jest": "^4.5.6",
     "@vue/cli-service": "~4.5.0",
     "@vue/eslint-config-prettier": "^6.0.0",
+    "@vue/runtime-dom": "^3.2.37",
```
seriouslysean, line 37: This is required by Volar for built in typing (even though we don't use it yet).

tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
```
seriouslysean, line 1: Required by Volar even though we aren't using the majority of the features.

seriouslysean (commented):
@itsalaidbacklife it's probably worth letting this sit until the linting PR is done so that we can properly test the linting to make sure it's fully working.

.vscode/settings.json
itsalaidbacklife: Conflicted on this point if it means locking people into specific editors

seriouslysean: It does not lock anyone to VSCode, just makes anyone using VSCode have a better experience 👍🏼

seriouslysean:
@itsalaidbacklife closing this since you have some adjustments in your leaderboard feature coming. We can revisit once the rest of the linting and stuff is taken care of.

PR #43, opened 2022-06-24
Feat/leaderboard

itsalaidbacklife:
## Please check the following

- For New Features:

Adds visualization of ranking system. **Does not yet automatically track rankings in an ongoing fashion**

**### Requires database migrations**
See updated [ER diagram](https://lucid.app/lucidchart/0871fc48-423c-456e-9426-2a53e7c93e51/edit?from_internal=true)
image


.gitignore
```diff
@@ -127,3 +127,5 @@ npm-debug.log
 nbproject
 .idea
 .node_history
+tests/e2e/screenshots
+tests/e2e/videos
```
itsalaidbacklife, line 131: This prevents cypress from cluttering up git history with additional files for videos and screenshots

.vscode/settings.json
```diff
@@ -1,6 +1,6 @@
 {
   "editor.tabSize": 2,
-  "vetur.format.options.useTabs": true,
+  "vetur.format.options.useTabs": false,
   "prettier.enable": false,
-  "prettier.useTabs": true
+  "prettier.useTabs": false
```
itsalaidbacklife, line 5: May be irrelevant in light of the WIP linting work but these settings were inconsistent with eslint

client/js/components/NavigationDrawer.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+  <v-navigation-drawer
+    :value="showNav"
+    class="primary"
+    dark
+    app
+    data-cy="nav-drawer"
+    :permanent="showNav"
+    :mini-variant="isCollapsed"
```
itsalaidbacklife, line 9: Always collapsed on mobile and toggleable on desktop

client/js/components/NavigationDrawer.vue
```diff
@@ -0,0 +1,86 @@
...
+  data() {
+    return {
+      userHasCollapsed: false,
+    };
+  },
+  computed: {
+    authenticated() {
+      return this.$store.state.auth.authenticated;
+    },
+    pageLinks() {
```
itsalaidbacklife, line 49: Always shows

1. Login/Logout
2. Rules
Then if logged in, show

3. Play
4. Stats

client/js/components/StatsLeaderboard.vue
```diff
@@ -0,0 +1,359 @@
...
+        class="flex-grow-1"
+        menu-props="data-week-select-menu"
+        multiple
+        chips
+        deletable-chips
+      />
+    </div>
+    <v-data-table
+      :items="tableRows"
+      :headers="tableColumns"
```
itsalaidbacklife, line 25: See [vuetify docs for data tables](https://vuetifyjs.com/en/api/v-data-table/). 

1. Expects `headers` to be array of objects with the `text` to display and the `value` to compare to
2. Expects `items` to be an array of objects with keys that match the header `values` and values for what to display in the table

The `tableRows` computed prop adds additional key/value pairs to each object which are not displayed in order for the templates inside the cells to refer to information about both the points and the wins for the given week.

client/js/components/StatsLeaderboard.vue
```diff
@@ -0,0 +1,359 @@
...
+      :sort-by.sync="sortBy"
+      :loading="loading"
+      :item-class="tableRowClass"
+    >
+      <template #[`item.rank`]="{item, value}">
+        <span :data-rank="item.username">{{ value }} </span>
+      </template>
```
itsalaidbacklife, line 32: Checkout the docs for the `item.name` slot in data tables: https://vuetifyjs.com/en/api/v-data-table/#slots-item.%3Cname%3E

This customizes the cells in the `rank` column just so I can apply the `data-rank` attribute for selection during testing

client/js/components/StatsLeaderboard.vue
```diff
@@ -0,0 +1,359 @@
...
+        <v-tooltip :key="`${item.username}_week_${week}_wins`" top v-if="value">
+          <template #activator="{on, attrs}">
+            <v-chip
+              :key="`${item.username}_week_${week}`"
+              :color="colorForScoreByWeek(week, value)"
+              dark
+              :outlined="['primary', '#000'].includes(colorForScoreByWeek(week, value))"
+              v-bind="{
+                ...attrs,
+                ...dataAttribute(item.username, week),
```
itsalaidbacklife, line 43: Data attribute is dynamically generated html attribute for selection during testing e.g. `week-1="player1"`

client/js/components/StatsLeaderboard.vue
```diff
@@ -0,0 +1,359 @@
...
+      :items="tableRows"
+      :headers="tableColumns"
+      :sort-by.sync="sortBy"
+      :loading="loading"
+      :item-class="tableRowClass"
+    >
+      <template #[`item.rank`]="{item, value}">
+        <span :data-rank="item.username">{{ value }} </span>
+      </template>
+      <template v-for="week in ['total', ...selectedWeeks]" #[`item.week_${week}`]="{item, value}">
```
itsalaidbacklife, line 33: Bit of fancy footwork here. This customizes each column with the name `week_total`, or `week_n`. `item` refers to the entire row being rendered and `value` is the value associated with the specific cell being customized, which will be the number of points the player received for this week (that's what's used for sorting purposes)

client/js/components/StatsLeaderboard.vue
```diff
@@ -0,0 +1,359 @@
...
+              v-bind="{
+                ...attrs,
+                ...dataAttribute(item.username, week),
+              }"
+              v-on="on"
+            >
+              {{ tableCell(item, week) }}
+            </v-chip>
+          </template>
+          {{ playersBeaten(item.username, week) }}
```
itsalaidbacklife, line 50: This is the text that displays in the tooltip

client/js/components/StatsLeaderboard.vue
```diff
@@ -0,0 +1,359 @@
...
+        { text: 'Week 9', value: 9 },
+        { text: 'Week 10', value: 10 },
+        { text: 'Week 11', value: 11 },
+        { text: 'Week 12', value: 12 },
+        { text: 'Week 13', value: 13 },
+      ],
+      selectedWeeks: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13],
+    };
+  },
+  computed: {
```
itsalaidbacklife, line 95: This entire computed block is a bit of a jumble and I'm very open to finding ways to cleanup or reorganize it

client/js/components/StatsLeaderboard.vue
```diff
@@ -0,0 +1,359 @@
...
+          res.third = playerScore.total;
+        }
+      }
+      return res;
+    },
+    totalScoresSorted() {
+      return this.playerScores.map(playerStats => playerStats.total).sort((a, b) => b - a);
+    },
+  },
+  methods: {
```
itsalaidbacklife, line 246: Same goes for methods

client/js/components/StatsLeaderboard.vue
```diff
@@ -0,0 +1,359 @@
...
+<style scoped>
+::v-deep .active-user-stats {
+  background-color: var(--v-accent-lighten3);
+}
+::v-deep .v-select.fit {
+  width: min-content;
+}
+::v-deep .v-select.fit .v-select__selection--comma {
+  text-overflow: unset;
+}
```
itsalaidbacklife, line 358: v-deep is used to customize the styles children of the dynamically injected vuetify components

client/js/router/index.js
```diff
@@ -5,22 +5,25 @@ import LoginSignup from '../views/LoginSignup.vue';
...
 
 Vue.use(VueRouter);
 
+const mustBeAuthenticated = (to, from, next) => {
+  if (store.state.auth.authenticated) {
+    next();
+  } else {
+    next('/login');
+  }
+};
```
itsalaidbacklife, line 19: Moved this func to outside the Home component for general reuse. Could consider a separate utility file for it, but it's still a small list of routes so I just put it at the top

client/js/store/modules/auth.js
```diff
@@ -72,6 +72,7 @@ export default {
       return new Promise((resolve, reject) => {
         io.socket.get('/user/logout', {}, function handleResponse(resData, jwres) {
           if (jwres.statusCode === 200) {
+            context.commit('authFailure');
```
itsalaidbacklife, line 75: Clears auth in store. Should probably rename the mutation

itsalaidbacklife (commented):
Needs rebasing off main

config/routes.js
```diff
@@ -51,6 +51,8 @@ module.exports.routes = {
   '/user/reLogin': 'userController.reLogin',
   '/user/logout': 'userController.logout',
 
+  'GET /stats': 'StatsController.getStats',
+
```
itsalaidbacklife, line 55: Delete new line

package.json
```diff
@@ -11,6 +11,7 @@
     "bluebird": "^3.4.6",
     "connect-redis": "3.0.2",
     "core-js": "^3.21.1",
+    "dayjs": "^1.11.3",
```
itsalaidbacklife, line 14: Date library for comparing and rendering timestamps in front end backend and tests

tests/e2e/specs/out_of_game/stats.spec.js
```diff
@@ -0,0 +1,199 @@
...
+    // Total counts should only consider the selected weeks
+    cy.get('[week-1=Player1]').should('contain', 'W: 3, P: 5');
+    cy.get('[week-total=Player1]').should('contain', 'W: 6, P: 10');
+  });
+
+  it('Selects different seasons to show their results', () => {
+    // Three award cards for Clubs 2022
+    cy.get('[data-tournament]').should('have.length', 3);
+    // Switch season to diamonds 2022
+    cy.get('[data-cy=season-select]').click({ force: true });
```
itsalaidbacklife, line 179: Need to force because vuetify injects attributes onto the nested input inside the select which isn't actually clickable because it's covered by other injected content. This is awkward but works well enough I think

itsalaidbacklife (commented):
Todo: Remove card and match fixtures, clean up newline

config/policies.js
```diff
@@ -17,43 +17,17 @@
...
+  },
 
   UserController: {
     '*': false,
     homepage: true,
     signup: ['hasValidUsername', 'hasValidPassword'],
     login: ['hasValidUsername', 'hasPassword'],
     reLogin: ['hasValidUsername', 'hasPassword'],
-    logout: 'isLoggedIn',
+    logout: true,
```
itsalaidbacklife, line 30: Always allow logging out so we can call it whenever necessary

client/js/components/StatsLeaderboard.vue
seriouslysean: This is all great information, part of me feels like it would be really nice to have as a comment in the template but since this is a vuetify thing it's probably part of their documentation already.

client/js/components/StatsLeaderboard.vue
```diff
@@ -0,0 +1,359 @@
...
+    >
+      <template #[`item.rank`]="{item, value}">
+        <span :data-rank="item.username">{{ value }} </span>
+      </template>
+      <template v-for="week in ['total', ...selectedWeeks]" #[`item.week_${week}`]="{item, value}">
+        <v-tooltip :key="`${item.username}_week_${week}_wins`" top v-if="value">
+          <template #activator="{on, attrs}">
+            <v-chip
+              :key="`${item.username}_week_${week}`"
+              :color="colorForScoreByWeek(week, value)"
```
seriouslysean, line 38: It's a little hard for me to follow, but I feel like this doesn't need to be a method and could instead be a straight lookup in a mapping for performance since there is a finite number of weeks. That feels really nitpicky without any sort of real example to back it up. I guess my major question would be how dynamic is this? The scores seem pretty variable but if you have the data up front you might be able to map them pretty easy.

client/js/components/StatsLeaderboard.vue
seriouslysean: Something I've done before that I find useful is to prefix any sort of test data attribute with `data-qa-*` so that I can look them up more easily in the future. Not anything you need to do here but food for thought since you aren't using it anywhere else in the code.

client/js/components/StatsLeaderboard.vue
```diff
@@ -0,0 +1,359 @@
...
+        { text: 'Week 7', value: 7 },
+        { text: 'Week 8', value: 8 },
+        { text: 'Week 9', value: 9 },
+        { text: 'Week 10', value: 10 },
+        { text: 'Week 11', value: 11 },
+        { text: 'Week 12', value: 12 },
+        { text: 'Week 13', value: 13 },
+      ],
+      selectedWeeks: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13],
+    };
```
seriouslysean, line 93: Do these values get mutated at all? If not there's no reason for them to be reactive. You should instead add them as a static prop on the component in the `created()` lifecycle method. Something like this:
```js
created() {
    this.sortBy = 'rank';
    this.metricChoices = ['Points and Wins', 'Points Only', 'Wins Only'];
    this.selectedMetric = 'Points and Wins';
    // ...
}
```
Better yet, just define them as `const`s above unless you're using them directly in the template. If they're only being used in methods and computed props there's no downside.

client/js/components/StatsLeaderboard.vue
```diff
@@ -0,0 +1,359 @@
...
+      ];
+      // Add headers for each week
+      for (const weekNum of this.selectedWeeks) {
+        res.push({
+          text: `Week ${weekNum}`,
+          value: `week_${weekNum}`,
+        });
+      }
+      return res;
+    },
```
seriouslysean, line 113: I think organizing everything by columns, rows, etc, is a good option. I'd need to spend a lot more time looking it over. I think mainly it's probably worth reducing complexity where you can and making sure the code stays readable. You can always reduce complexity by pulling out code to separate functions and such at a later date during a refactor. Something like this:
```js
    tableColumns() {
        if (!this.season || !this.season.rankings || this.season.rankings.length === 0) {
            return [];
        }
        return [
            { text: 'User', value: 'username' },
            { text: 'Rank', value: 'rank' },
            { text: 'Season Total', value: 'week_total' },
            ...(this.selectedWeeks.map(weekNum => ({
                text: `Week ${weekNum}`,
                value: `week_${weekNum}`,
            }))),
        ];
    },
```

client/js/components/StatsLeaderboard.vue
```diff
@@ -0,0 +1,359 @@
...
+    tableRows() {
+      if (!this.season || !this.season.rankings || this.season.rankings.length === 0) {
+        return [];
+      }
```
seriouslysean, line 117: This is duplicated between all the computed props, it might be worth making one `generateTable` function that starts the process, does this check and then forks the columns and rows off to separate methods. 

Alternatively you could default the prop to a prefilled set of keys with defaults. I don't really like doing that though, so maybe you need an error view in case this doesn't load. At a higher level when you include this just have a `v-if` check to see if the values are set, if not show an error state.

client/js/components/StatsLeaderboard.vue
```diff
@@ -0,0 +1,359 @@
...
+              pointsThisWeek = 3;
+            } else if (playerWins[weekNum] > 0) {
+              pointsThisWeek = 1;
+            }
+            res[weekNum] = pointsThisWeek;
+            res.total += pointsThisWeek;
+          }
+        }
+        return res;
+      });
```
seriouslysean, line 177: I really prefer to avoid nesting ifs when it can be helped, it ensures the code stays easily readable. Are you able to filter the weekNums before they go in to the loop? Something like `for (const weekNum in playerWins.filter(pw.weekNum != 'total')) {`, then you wouldn't need the nested if.

client/js/router/index.js
```diff
@@ -5,22 +5,25 @@ import LoginSignup from '../views/LoginSignup.vue';
...
 import store from '../store/store.js';
 
 Vue.use(VueRouter);
 
+const mustBeAuthenticated = (to, from, next) => {
+  if (store.state.auth.authenticated) {
+    next();
+  } else {
+    next('/login');
+  }
```
seriouslysean, line 18: I think you can do this to avoid the if/else:
```suggestion
  if (store.state.auth.authenticated) {
    return next();
  }
  return next('/login');
```

client/js/router/index.js
seriouslysean: Seems like a good candidate for a middleware utility on the vue router.

client/js/store/modules/auth.js
seriouslysean: I would definitely rename that sooner than later, it's a bit confusing.

client/js/views/LoginSignup.vue
```diff
@@ -196,7 +199,7 @@ export default {
   justify-content: center;
   margin-top: 16px;
 }
-@media (orientation: landscape) and (max-width: 979px) {
+@media (orientation: landscape) and (max-width: 960px) {
```
seriouslysean, line 204: Was this adjustment intentional? Any reason for the switch? Should it be aligned with anywhere else on the site? It feels like you should normalize breakpoints and have some css utilities to call for breakpoint stuff but it may not be that important right now.

client/js/views/Stats.vue
```diff
@@ -0,0 +1,151 @@
...
+        <span data-cy="season-start-date">{{ seasonStartString }}</span>
+        -
+        <span data-cy="season-end-date"> {{ seasonEndString }} </span>
+      </h4>
+    </section>
+    <section class="px-8">
+      <!-- Season Champions -->
+      <div class="mb-10">
+        <h2
+          v-if="
```
seriouslysean, line 29: This seems like it should be a computed prop, something like `showSeasonChampions`.

client/js/views/Stats.vue
```diff
@@ -0,0 +1,151 @@
...
+  },
+  data() {
+    return {
+      loadingData: false,
+      selectedSeason: null,
+      seasons: [],
+    };
+  },
+  computed: {
+    seasonStartString() {
```
seriouslysean, line 107: `seasonStartFormatted()`or `startSeasonFormatted()` make a bit more sense imo, I like to name variables in a way that it's easy to read what they're doing at a glance.

client/js/views/Stats.vue
```diff
@@ -0,0 +1,151 @@
...
+      if (this.selectedSeason) {
+        return dayjs(this.selectedSeason.startTime).format('YYYY/MM/DD');
+      }
+      return '';
```
seriouslysean, line 111: I would probably flip these to handle the error case first:
```suggestion
      if (!this.selectedSeason) {
        return '';
      }
      return dayjs(this.selectedSeason.startTime).format('YYYY/MM/DD');
```

Alternatively, just combine it in to one return:
```suggestion
      return !this.selectedSeason
        ? ''
        : dayjs(this.selectedSeason.startTime).format('YYYY/MM/DD');
```
Trying to figure out how prettier would format something is 🤮 lol.

client/js/views/Stats.vue
```diff
@@ -0,0 +1,151 @@
...
+    seasonEndString() {
+      if (this.selectedSeason) {
+        return dayjs(this.selectedSeason.endTime).format('YYYY/MM/DD');
+      }
+      return '';
+    },
+  },
+  created() {
+    this.loadingData = true;
+    io.socket.get('/stats', (res, jwres) => {
```
seriouslysean, line 122: Do all API calls need to be via socket, or just game calls? It feels like you might be able to have these just be api GET requests to a net new endpoint specifically for data that could be cached more easily. Then again, I don't really know enough about sockets to say for sure.

seriouslysean (commented):
Nothing blocking, just a lot of questions and thoughts.

api/controllers/StatsController.js
```diff
@@ -0,0 +1,149 @@
+const dayjs = require('dayjs');
+const isBetween = require('dayjs/plugin/isBetween');
+dayjs.extend(isBetween);
+
+const Result = {
+  WON: 1,
+  LOST: 2,
+  INCOMPLETE: -1,
+};
```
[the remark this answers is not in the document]
itsalaidbacklife: Used in front and backend so that would be helpful. Might run into a really awkward issue that the frontend uses esmodules and the backend uses commonjs/require syntax...

api/controllers/StatsController.js
```diff
@@ -0,0 +1,149 @@
...
+
+/////////////
+// Helpers //
+/////////////
```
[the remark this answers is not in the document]
itsalaidbacklife: Only used here so far as I envision. If a usecase comes up for using them elsewhere, I think that'd be the time to pull them out. Interesting point about unit testing them. Worth revisiting when we get back to unit testing as a whole during the linting rework. For now, I'm comfortable with just e2e on the stats page as the consumer of this controller and thereby these helpers

api/controllers/StatsController.js
```diff
@@ -0,0 +1,149 @@
...
+        return {
+          ...season,
+          rankings: new Map(), // playerId => PlayerMatches
+          firstPlace: season.firstPlace?.username || null,
+          secondPlace: season.secondPlace?.username || null,
+          thirdPlace: season.thirdPlace?.username || null,
+          fourthPlace: season.fourthPlace?.username || null,
+        };
+      });
+      for (const match of matches) {
```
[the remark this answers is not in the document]
itsalaidbacklife: Cool! Easy fix

client/js/views/Stats.vue
itsalaidbacklife: how's `seasonStartDateString`?

client/js/views/Stats.vue
itsalaidbacklife: Yeah I generally prefer the early return via ternary or otherwise. Something that concerns me about it though is that when we move to typescript, I'm medium confident that it's not smart enough to detect early returns and will throw a fit. 

Probably an issue worth addressing when we get there. For now I'll take the ternary

client/js/views/Stats.vue
itsalaidbacklife: Currently all api calls are expected to come in via socket. This is worth revisiting holistically because there's no particular need for that to be the case, other than centralization of config (i.e. pointing api requests to the right domain both for sockets and for the other fetch approach)

client/js/views/LoginSignup.vue
itsalaidbacklife: Definitely want to revisit these overall. I changed this here because of the introduction of the nav drawer which is a responsive component and has a breakpoint at 960px, so it looks and feels better for that to align with the media query breakpoint

client/js/views/Stats.vue
itsalaidbacklife: Agreed

client/js/router/index.js
itsalaidbacklife: Yeah agreed. Prefer the return w/o else

client/js/components/StatsLeaderboard.vue
itsalaidbacklife: Yes that's cleaner (though seemingly less performant since we loop through the playerWins an extra time. But I'm comfortable prioritizing legibility

client/js/components/StatsLeaderboard.vue
itsalaidbacklife: Yeah I've been inconsistent about that. This is actually the only place in the code I think that doesn't use data-* as a prefix. I've inconsistently used data-cy-* sometimes and just data-* othertimes when the names get long, which could use cleaning. For now, there's no good reason not to at least use data-* here so I'll add that

client/js/components/StatsLeaderboard.vue
itsalaidbacklife: Yeah it's pretty well documented in vuetify

seriouslysean (approved):
Few last comments

client/js/components/AwardCard.vue
```diff
@@ -0,0 +1,60 @@
+<template>
+  <v-card class="d-flex align-center" color="#eee" :data-tournament="`${place}${suffix}`">
```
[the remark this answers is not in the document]
itsalaidbacklife: Yes totally. Actually might be reasonable to include here because Vuetify has a theme object we could put these in

client/js/components/AwardCard.vue
```diff
@@ -0,0 +1,60 @@
...
+        case 1:
+          return '#AF9500';
+        case 2:
+          return '#B4B4B4';
+        case 3:
+          return '#6A3805';
+        default:
+          return '#000';
+      }
+    },
```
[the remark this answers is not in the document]
itsalaidbacklife: Interesting alternative. Personally I like switch cases for enum values so I'm inclined to leave it. The colors themselves would be better off in the Vuetify theme though

client/js/components/NavigationDrawer.vue
```diff
@@ -0,0 +1,86 @@
...
+    </template>
+  </v-navigation-drawer>
+</template>
+
+<script>
+export default {
+  name: 'NavigationDrawer',
+  data() {
+    return {
+      userHasCollapsed: false,
```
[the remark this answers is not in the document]
itsalaidbacklife: I want to collapse the nav of the user hits the button or if we're on mobile (always collapsed on mobile without toggle option). 

So my intent here was to have the user control a Boolean in data and then control the nav drawer with a computed prop that collapses if the data or isMobile is true. Is that a sensible approach? And can you think of a way to make that more clear?

client/js/components/NavigationDrawer.vue
```diff
@@ -0,0 +1,86 @@
...
+      else {
+        res = [
+          { text: 'Logout', icon: 'logout', page: { name: 'Login' } },
+          rules,
+          { text: 'Play', icon: 'play', page: { name: 'Home' } },
+          { text: 'Stats', icon: 'chart-bar', page: { name: 'Stats' } },
+        ];
+      }
+      return res;
+    },
```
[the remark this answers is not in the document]
itsalaidbacklife: Like right above expand/collapse?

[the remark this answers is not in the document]
seriouslysean: I’m trying to think of a good example but I’ve seen the logout/profile link pinned to the bottom of the nav which might work here.

client/js/components/NavigationDrawer.vue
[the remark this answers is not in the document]
seriouslysean: I think that makes sense. In that case isCollapsed makes more sense, which is then driven by user action/mobile breakpoint.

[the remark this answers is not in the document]
itsalaidbacklife: I have isCollapsed as the computed prop which is actually sent to the nav drawer, and isCollapsedByUser as the data the user controls. Reasonable? This actually seems like a case where composition api would be great so the data and computed prop could be next to each other instead of separated across option blocks

client/js/components/NavigationDrawer.vue
```diff
@@ -0,0 +1,86 @@
...
+      let res = [];
+      const rules = { text: 'Rules', icon: 'script-text', page: { name: 'Rules' } };
+
+      if (!this.authenticated) {
+        res = [{ text: 'Login', icon: 'login', page: { name: 'Login' } }, rules];
+      }
+      // Authenticated
+      else {
+        res = [
+          { text: 'Logout', icon: 'logout', page: { name: 'Login' } },
```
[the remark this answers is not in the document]
itsalaidbacklife: Yeah the idea is to always have a link to the login page and call it either login or logout based on your current auth status

client/js/components/NavigationDrawer.vue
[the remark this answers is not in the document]
itsalaidbacklife: It could also make sense to introduce a top navbar and put it in there

client/js/components/StatsLeaderboard.vue
itsalaidbacklife: Perhaps a docs link in comments though

client/js/components/StatsLeaderboard.vue
itsalaidbacklife: This could probably use a comment

client/js/components/StatsLeaderboard.vue
itsalaidbacklife: For any given cell the value is not dynamic. A big map would be reasonable but actually I'm thinking it'd be cleaner to simply make a component so that this value can be a computed prop inside it.

client/js/components/StatsLeaderboard.vue
itsalaidbacklife: That's an interesting move. I `weeks` and `metricChoices` could be done this way but the others are selections that should be reactive

client/js/components/StatsLeaderboard.vue
itsalaidbacklife: I drafted an error view that I haven't included yet because I wasn't sure the right way to attempt recovery. Could include it with static text though

itsalaidbacklife: I don't fully follow the geberateTable suggestion. Would it be a computed prop that returns an object with one property for the table rows and one for the columns? This is to centralize the bull checking?

client/js/components/StatsLeaderboard.vue
itsalaidbacklife: I like the use of spread and map here over the for loop

client/js/components/StatsLeaderboard.vue
itsalaidbacklife: Oh wait I think we hit a gotcha here because we're iterating over object properties with for...in not over an array with for...of

itsalaidbacklife: got it like
```
    const { total, ...weeks } = playerWins;
    for (const weekNum in { ...weeks }) {
```

client/js/components/StatsLeaderboard.vue
itsalaidbacklife: Pulled this into a component which breaks up the code and allows this to be a computed prop scoped to the one cell rather than a function call

api/controllers/StatsController.js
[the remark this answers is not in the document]
itsalaidbacklife: done

api/controllers/StatsController.js
```diff
@@ -0,0 +1,149 @@
...
+      seasons = seasons.map(season => {
+        return {
+          ...season,
+          rankings: new Map(), // playerId => PlayerMatches
+          firstPlace: season.firstPlace?.username || null,
+          secondPlace: season.secondPlace?.username || null,
+          thirdPlace: season.thirdPlace?.username || null,
+          fourthPlace: season.fourthPlace?.username || null,
+        };
+      });
```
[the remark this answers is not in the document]
itsalaidbacklife: Refactored season initialization into a sails helper: https://sailsjs.com/documentation/concepts/helpers

PR #51, opened 2022-07-03
Upgrade to Vue 2.7.0 and add initial support for TypeScript

seriouslysean:
## Please check the following

- For New Features:

- https://github.com/vuejs/vue/blob/main/CHANGELOG.md#270-2022-07-01
- https://medium.com/js-dojo/adding-typescript-to-your-existing-vuejs-2-6-app-aaa896c2d40a
- Bunch of other googling


.vscode/extensions.json
```diff
@@ -1,7 +1,8 @@
 {
   "recommendations": [
     "dbaeumer.vscode-eslint",
-    "octref.vetur",
-    "rvest.vs-code-prettier-eslint"
+    "rvest.vs-code-prettier-eslint",
+    "vue.volar",
+    "Vue.vscode-typescript-vue-plugin"
```
seriouslysean, line 6: With Vue 2.7+, Volar + TS Vue Plugin are the recommended extensions, Vetur doesn't have enough support to handle the new stuff.

.vscode/settings.json
```diff
@@ -1,8 +1,8 @@
 {
   "editor.tabSize": 2,
-  "vetur.format.options.useTabs": false,
+  "eslint.lintTask.enable": true,
```
seriouslysean, line 3: Made some changes here to align with some of our standards. wrapLineLength and rules set to 110 so we know where our end of line is.

client/js/.eslintrc.json
```diff
@@ -1,12 +0,0 @@
-{
```
seriouslysean, line 1: No longer needed as far as I can tell.

client/js/App.vue
```diff
@@ -1,22 +1,16 @@
+<script setup>
+import NavigationDrawer from './components/NavigationDrawer.vue';
```
seriouslysean, line 2: I converted this to use `<script setup>` since that's going to be the default in Vue 3 but I'm curious on your opinion here. I don't know enough about Vuetify to know if it matters that `navigation-drawer` becomes `NavigationDrawer`. I prefer the capitalized multi-word form, and it just happens to work with the way you import in script setup.

Thoughts?

client/js/components/StatsLeaderboard.vue
```diff
@@ -49,7 +49,7 @@
 import { uniq } from 'lodash';
 import StatsLeaderboardCell from '@/components/StatsLeaderboardCell.vue';
 
-const Result = require('../../../types/Result');
+import Result from '../../../types/Result';
```
seriouslysean, line 52: Double check this, I feel like we should probably move it eventually.

client/js/main.ts
```diff
@@ -5,6 +5,13 @@ import store from './store/store';
 import router from './router';
 Vue.config.productionTip = false;
 
+// TODO: Think through this, move to relevant location (post Vue 3 since it changes the TS implementation)
+interface CuttleWindow extends Window {
+  Cypress: Object;
+  app: Object;
+}
+declare let window: CuttleWindow;
```
seriouslysean, line 13: Converted main.js to main.ts for the initial support and there was an error in here for the window Object so I extended it to add your new props. I think we should figure out where to put these interfaces and then talk about a new shape for the window object.

I was thinking about something like this:
```
window: {
    cuttle: {
        app: {} // only in Cypress
        version: '3.x.x`, // on all builds
        // ...?
    },
}
```

client/js/views/GameView.vue
```diff
@@ -1354,7 +1354,7 @@ export default {
   min-width: 50%;
   height: 30vh;
   & #player-hand-cards,
-  #player-hand-targeting::v-deep {
+  :deep(#player-hand-targeting) {
```
seriouslysean, line 1357: After upgrading to Vue 2.7 this warning was being thrown in the console so I would really appreciate you double checking the styles since I'm not entirely sure how deep works in this regard.

seriouslysean:
@itsalaidbacklife we should revisit this later, so many blockers to getting this to work that I feel like just moving to Vue3 eventually is probably better in the long run in terms of time/effort spent.

PR #54, opened 2022-07-09
Convert dialogs to separate component, breaking up GameView

seriouslysean:
## Please check the following

- For New Features:

Breaking up GameView and pulling dialogs out in to their own component.

Trello mentioned also pulling out overlays, etc, but I feel like this is already big enough and that may need to be it's own ticket to reduce the cognitive load of even looking at this.

### DIFF
--- a/client/js/components/GameView/GameDialogs.vue
+++ b/client/js/components/GameView/GameDialogs.vue
@@ -1,0 +1,189 @@
+<template>
+  <div class="game-dialogs">
+    <counter-dialog
+      v-model="showCounterDialog"
+      :one-off="game.oneOff"
+      :target="game.oneOffTarget"
+      :twos-in-hand="twosInHand"
+      :twos-played="twosPlayed"
+      @resolve="resolve"
+      @counter="counter($event)"
+    />
+    <cannot-counter-dialog
+      v-model="showCannotCounterDialog"
+      :one-off="game.oneOff"
+      :opponent-queen-count="opponentQueenCount"
+      :player-two-count="playerTwoCount"
+      :twos-played="twosPlayed"
+      :target="game.oneOffTarget"
+      @resolve="resolve"
+    />
+    <four-dialog v-model="discarding" @discard="discard" />
+    <three-dialog
+      v-model="pickingFromScrap"
+      :one-off="game.oneOff"
+      :scrap="scrap"
+      @resolveThree="resolveThree($event)"
+    />
+    <seven-double-jacks-dialog
+      v-model="showSevenDoubleJacksDialog"
+      :top-card="topCard"
+      :second-card="secondCard"
+      @resolveSevenDoubleJacks="resolveSevenDoubleJacks($event)"
+    />
+    <game-over-dialog v-model="gameIsOver" :player-wins="playerWins" :stalemate="stalemate" />
+    <reauthenticate-dialog v-model="mustReauthenticate" />
+  </div>
+</template>
+
+<script>
+import { mapGetters } from 'vuex';
+
+import CannotCounterDialog from '@/components/GameView/CannotCounterDialog.vue';
+import CounterDialog from '@/components/GameView/CounterDialog.vue';
+import FourDialog from '@/components/GameView/FourDialog.vue';
+import GameOverDialog from '@/components/GameView/GameOverDialog.vue';
+import ReauthenticateDialog from '@/components/GameView/ReauthenticateDialog.vue';
+import SevenDoubleJacksDialog from '@/components/GameView/SevenDoubleJacksDialog.vue';
+import ThreeDialog from '@/components/GameView/ThreeDialog.vue';
+
+export default {
+  name: 'GameDialogs',
+  components: {
+    CannotCounterDialog,
+    CounterDialog,
+    FourDialog,
+    GameOverDialog,
+    ReauthenticateDialog,
+    SevenDoubleJacksDialog,
+    ThreeDialog,
+  },
+  props: {
+    clearSelection: {
+      type: Function,
+      required: true,
+    },
+    game: {
+      type: Object,
+      required: true,
+    },
+    handleError: {
+      type: Function,
+      required: true,
+    },
+    resolvingSeven: {
+      type: Boolean,
+      required: true,
+    },
+  },
+  computed: {
+    ...mapGetters(['player', 'opponent', 'opponentPointTotal', 'opponentQueenCount']),
+    discarding() {
+      return this.$store.state.game.discarding;
+    },
+    gameIsOver() {
+      return this.$store.state.game.gameIsOver;
+    },
+    hasTwoInHand() {
+      return this.twosInHand.length > 0;
+    },
+    myTurnToCounter() {
+      return this.game.myTurnToCounter;
+    },
+    // TODO: Refactor and combine usage in GameView.vue
+    mustReauthenticate: {
+      get() {
+        return this.$store.state.auth.mustReauthenticate;
+      },
+      set(val) {
+        this.$store.commit('setMustReauthenticate', val);
+      },
+    },
+    pickingFromScrap() {
+      return this.$store.state.game.pickingFromScrap;
+    },
+    playerTwoCount() {
+      return this.twoCount(this.player);
+    },
+    playerWins() {
+      return (
+        this.$store.state.game.gameIsOver &&
+        this.$store.state.game.winnerPNum === this.$store.state.game.myPNum
+      );
+    },
+    scrap() {
+      return this.game.scrap;
+    },
+    secondCard() {
+      return this.$store.state.game.secondCard;
+    },
+    showCannotCounterDialog() {
+      return (
+        (this.myTurnToCounter && !this.hasTwoInHand) ||
+        (this.myTurnToCounter && this.hasTwoInHand && this.opponentQueenCount > 0)
+      );
+    },
+    showCounterDialog() {
+      return this.myTurnToCounter && this.hasTwoInHand && this.opponentQueenCount === 0;
+    },
+    showSevenDoubleJacksDialog() {
+      return (
+        this.resolvingSeven &&
+        this.topCard.rank === 11 &&
+        this.secondCard.rank === 11 &&
+        (this.opponentPointTotal === 0 || this.opponentQueenCount > 0)
+      );
+    },
+    stalemate() {
+      return this.$store.state.game.gameIsOver && this.$store.state.game.winnerPNum === null;
+    },
+    topCard() {
+      return this.$store.state.game.topCard;
+    },
+    twosInHand() {
+      return this.player.hand.filter((card) => card.rank === 2);
+    },
+    twosPlayed() {
+      return this.game.twos;
+    },
+  },
+  methods: {
+    counter(twoId) {
+      this.$store
+        .dispatch('requestCounter', twoId)
+        .then(this.clearSelection)
+        .catch(this.handleError);
+    },
+    discard(cardIds) {
+      const [cardId1] = cardIds;
+      const cardId2 = cardIds.length > 1 ? cardIds[1] : null;
+      this.$store.dispatch('requestDiscard', {
+        cardId1,
+        cardId2,
+      });
+    },
+    resolve() {
+      this.$store.dispatch('requestResolve').then(this.clearSelection).catch(this.handleError);
+    },
+    resolveThree(cardId) {
+      this.$store
+        .dispatch('requestResolveThree', cardId)
+        .then(this.clearSelection)
+        .catch(this.handleError);
+    },
+    resolveSevenDoubleJacks({ cardId, index }) {
+      this.$store
+        .dispatch('requestResolveSevenDoubleJacks', { cardId, index })
+        .then(this.clearSelection)
+        .catch(this.handleError);
+    },
+    /**
+     * @returns number of queens a given player has
+     * @param player is the player object
+     */
+    twoCount(player) {
+      return player.faceCards.reduce((twoCount, card) => twoCount + (card.rank === 2 ? 1 : 0), 0);
+    },
+  },
+};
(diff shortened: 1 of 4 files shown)


client/js/store/modules/game.js
```diff
@@ -52,27 +53,56 @@ function handleGameResponse(context, jwres, resolve, reject) {
...
+  return player?.faceCards?.reduce(
+    (queenCount, card) => queenCount + (card.rank === 12 ? 1 : 0),
+    0
+  );
+}
+
 const initialState = resetState();
 export default {
+  // TODO: Look in to namespacing -- currently this will not work due to a collision
+  // with the game socket also being named `game`. Potentially rename one or the other?
```
seriouslysean, line 71: These are probably the most important changes in here. With the move to break up GameView, it seems imperative that we move any shared getters and utilities out so they can be easily reused. Typically I would namespace here but the socket logic uses the same inferred `game` name that this vuex module would use.

It's probably worth talking through some options, but for now I just opted to not namespace it and use them as-is.

client/js/store/modules/game.js
```diff
@@ -52,27 +53,56 @@ function handleGameResponse(context, jwres, resolve, reject) {
...
   getters: {
+    player(state) {
+      return state.players[state.myPNum];
+    },
+    playerPointTotal(state, getters) {
+      return getters.player?.points?.reduce((total, card) => total + card.rank, 0) || 0;
+    },
+    playerQueenCount(state, getters) {
+      return queenCount(getters.player);
+    },
```
seriouslysean, line 82: We lose a little bit of the organization here from the comments you had in GameView, @itsalaidbacklife, but I think the colocation more than makes up for it. Might be worth pulling some of those comments back here if you're really partial to them.

The big benefit here is that all components that need this data all get it from a single source of truth now and it means we don't need to do any sort of complex prop drilling.

client/js/views/GameView.vue
```diff
@@ -294,39 +294,6 @@
...
-        @resolveThree="resolveThree($event)"
-      />
-      <seven-double-jacks-dialog
-        v-model="showSevenDoubleJacksDialog"
-        :top-card="topCard"
-        :second-card="secondCard"
-        @resolveSevenDoubleJacks="resolveSevenDoubleJacks($event)"
-      />
-      <game-over-dialog v-model="gameIsOver" :player-wins="playerWins" :stalemate="stalemate" />
-      <reauthenticate-dialog v-model="mustReauthenticate" />
```
seriouslysean, line 329: All of these now live in the GameDialogs component. The only exception is `reauthenticate-dialog` since it is also used at the top of GameView with in if/else. I wonder if it might be worth abstracting that later to an app-wide reauthentication.

client/js/views/GameView.vue
```diff
@@ -340,21 +307,24 @@
         @targetedOneOff="beginTargeting($event)"
         @cancel="clearSelection"
       />
+      <game-dialogs
+        :clear-selection="clearSelection"
+        :game="game"
+        :handle-error="handleError"
+        :resolving-seven="resolvingSeven"
+      />
```
seriouslysean, line 315: Much smaller 😎

client/js/views/GameView.vue
```diff
@@ -393,6 +358,14 @@ export default {
...
   },
   computed: {
+    ...mapGetters([
+      'player',
+      'playerPointTotal',
+      'playerQueenCount',
+      'opponent',
+      'opponentPointTotal',
+      'opponentQueenCount',
+    ]),
```
seriouslysean, line 368: With the move of a lot of the shared data to the store, it's as simple as mapping the data each component needs now which is a nice change imo.

client/js/components/GameView/GameDialogs.vue
```diff
@@ -0,0 +1,189 @@
+<template>
+  <div class="game-dialogs">
```
seriouslysean, line 2: The only major shift here is with this wrapper so we have a singular parent for this component. No styles on it or anything, though.

client/js/store/modules/game.js
itsalaidbacklife: What kind of namespacing do you envision? Are you talking about pulling these shared utilities into the vuex store and namespacing it's modules? Or putting them elsewhere and implementing some kind of custom namespacing solution? Something else?

client/js/views/GameView.vue
itsalaidbacklife: It very possibly would, though app wide Reauthentication would have more conditional logic

client/js/store/modules/game.js
seriouslysean: > Are you talking about pulling these shared utilities into the vuex store and namespacing it's modules?

That's exactly it. I'd like to see all shared logic that uses the game object or any of the store values hoisted back up to the store level.

Anything that isn't exactly store related, but calculates or otherwise does some logic we should pull those out to shared utilities and test them with Jest.

As for the namespacing, I'd like to use the name `game` for the game module, but we still would need to figure out what to rename the server side stuff which currently use game also. I may be missing some of the puzzle but those are the general points.

client/js/views/GameView.vue
seriouslysean: I'd like to see that reauthentication getter and setter moved to a dispatches action so it's easier to share between components at least.

client/js/views/GameView.vue
```diff
@@ -340,21 +307,24 @@
         @targetedOneOff="beginTargeting($event)"
         @cancel="clearSelection"
       />
+      <game-dialogs
+        :clear-selection="clearSelection"
+        :game="game"
+        :handle-error="handleError"
+        :resolving-seven="resolvingSeven"
```
itsalaidbacklife, line 314: Because both these methods directly mutate data inside the gameView, I think it might be more Vue-esque to bind these as event handlers that get called by the GameView to avoid having the GameDialogs directly updating GameView data. What do you think?

itsalaidbacklife (commented):
I have two questions in the comments

client/js/store/modules/game.js
itsalaidbacklife: Looking forward to a deeper dive into this.

I'm inclined to tackle module reorganization as a part of the switch to pinia, so we don't rearchitect the vuex store then throw it out altogether. I think I see the gist of what you mean but I don't actually follow how module namespacing relates to server side code. 

I think the switch to pinia would actually subvert the need for namespacing, because each module would be individually imported as its own store and locally assigned to a variable in the component(s) consuming it, which I think would prevent collisions. Does that address your concerns?

client/js/store/modules/game.js
itsalaidbacklife: This is a solid improvement

client/js/components/GameView/GameDialogs.vue
```diff
@@ -0,0 +1,189 @@
...
+      return this.$store.state.game.topCard;
+    },
+    twosInHand() {
+      return this.player.hand.filter((card) => card.rank === 2);
+    },
+    twosPlayed() {
+      return this.game.twos;
+    },
+  },
+  methods: {
```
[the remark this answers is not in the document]
seriouslysean: I definitely agree when it comes to the `clearSelection` and `handleError` methods, I've converted those to emitters in the latest pr. As for the others that i moved down I think they should stay here until they're moved to the store as actions or something similar to keep the `GameView` a bit more slim because it is _C H O N K Y_. Does that work?

client/js/store/modules/game.js
```diff
@@ -52,27 +53,62 @@ function handleGameResponse(context, jwres, resolve, reject) {
...
+  if (!player) {
+    return null;
+  }
+  return player.faceCards.reduce((queenCount, card) => queenCount + (card.rank === 12 ? 1 : 0), 0);
+}
+
 const initialState = resetState();
 export default {
+  // TODO: Look in to namespacing -- currently this will not work due to a collision
+  // with the game socket also being named `game`. Potentially rename one or the other?
```
seriouslysean, line 71:
```suggestion
```
Since we'll be moving to Pinia eventually I think it's fine to ignore this.

client/js/views/GameView.vue
seriouslysean: Yep, moved to emitters/events in the latest push.

client/js/views/GameView.vue
```diff
@@ -340,21 +307,24 @@
         @targetedOneOff="beginTargeting($event)"
         @cancel="clearSelection"
       />
+      <game-dialogs
+        :game="game"
```
itsalaidbacklife, line 311: Should we skip defining game as a prop and just read it from the store directly? One the one hand that further couples the component with the store logic, on the other, it's already pulling a bunch of computed props directly from the store, so why not go all-in?

itsalaidbacklife (approved):
Looks great. One last wondering about removing the :game prop

client/js/components/GameView/GameDialogs.vue
```diff
@@ -0,0 +1,193 @@
...
+    mustReauthenticate: {
+      get() {
+        return this.$store.state.auth.mustReauthenticate;
+      },
+      set(val) {
+        this.$store.commit('setMustReauthenticate', val);
+      },
+    },
+    pickingFromScrap() {
+      return this.$store.state.game.pickingFromScrap;
```
itsalaidbacklife, line 101: All the references to `this.$store.state.game` could be replaced with `this.game`

client/js/views/GameView.vue
```diff
@@ -360,21 +327,23 @@
         @targetedOneOff="beginTargeting($event)"
         @cancel="clearSelection"
       />
+      <game-dialogs
+        :resolving-seven="resolvingSeven"
```
itsalaidbacklife, line 331: Any particular reason to prefer a prop to mapping the store getter inside the child?

itsalaidbacklife (approved):

client/js/views/GameView.vue
seriouslysean: Nope, just missed it in my conversion

client/js/components/GameView/GameDialogs.vue
```diff
@@ -0,0 +1,187 @@
...
+      return (
+        this.$store.state.game.gameIsOver &&
+        this.$store.state.game.winnerPNum === this.$store.state.game.myPNum
+      );
+    },
+    scrap() {
+      return this.game.scrap;
+    },
+    secondCard() {
+      return this.$store.state.game.secondCard;
```
itsalaidbacklife, line 110:
```suggestion
      return this.game.secondCard;
```

client/js/components/GameView/GameDialogs.vue
```diff
@@ -0,0 +1,187 @@
...
+    pickingFromScrap() {
+      return this.$store.state.game.pickingFromScrap;
+    },
+    playerTwoCount() {
+      return this.twoCount(this.player);
+    },
+    playerWins() {
+      return (
+        this.$store.state.game.gameIsOver &&
+        this.$store.state.game.winnerPNum === this.$store.state.game.myPNum
```
itsalaidbacklife, line 103:
```suggestion
        this.game.gameIsOver &&
        this.game.winnerPNum === this.$store.state.game.myPNum
```

client/js/components/GameView/GameDialogs.vue
```diff
@@ -0,0 +1,187 @@
...
+    mustReauthenticate: {
+      get() {
+        return this.$store.state.auth.mustReauthenticate;
+      },
+      set(val) {
+        this.$store.commit('setMustReauthenticate', val);
+      },
+    },
+    pickingFromScrap() {
+      return this.$store.state.game.pickingFromScrap;
```
itsalaidbacklife, line 95:
```suggestion
      return this.game.pickingFromScrap;
```

client/js/components/GameView/GameDialogs.vue
```diff
@@ -0,0 +1,187 @@
...
+      'resolvingSeven',
+    ]),
+    game() {
+      // TODO: Figure out a better way to do this, mapping the whole module is a
+      // bit unusual-- the usages can probably be changed to only need a subset,
+      // or moved to some sort of store method or getter
+      return this.$store.state.game;
+    },
+    gameIsOver() {
+      return this.$store.state.game.gameIsOver;
```
itsalaidbacklife, line 77:
```suggestion
      return this.game.gameIsOver;
```

client/js/components/GameView/GameDialogs.vue
```diff
@@ -0,0 +1,187 @@
...
+    showSevenDoubleJacksDialog() {
+      return (
+        this.resolvingSeven &&
+        this.topCard.rank === 11 &&
+        this.secondCard.rank === 11 &&
+        (this.opponentPointTotal === 0 || this.opponentQueenCount > 0)
+      );
+    },
+    stalemate() {
+      return this.$store.state.game.gameIsOver && this.$store.state.game.winnerPNum === null;
```
itsalaidbacklife, line 130:
```suggestion
      return this.game.gameIsOver && this.game.winnerPNum === null;
```

client/js/components/GameView/GameDialogs.vue
```diff
@@ -0,0 +1,187 @@
...
+    },
+    topCard() {
+      return this.$store.state.game.topCard;
```
itsalaidbacklife, line 133:
```suggestion
      return this.game.topCard;
```

client/js/views/GameView.vue
```diff
@@ -446,7 +420,7 @@ export default {
     // Game, Deck, Log and Scrap //
     ///////////////////////////////
     game() {
-      return this.$store.state.game;
+      return this.game;
```
itsalaidbacklife, line 423: I believe this still needs the $store reference lest it recurse

```suggestion
      return this.$store.state.game;
```

itsalaidbacklife (approved):

PR #65, opened 2022-07-28
Feat/opponent must resolve

itsalaidbacklife:
## Please check the following

- For New Features:

### DIFF
--- a/client/js/views/GameView.vue
+++ b/client/js/views/GameView.vue
@@ -291,7 +291,7 @@
         v-model="waitingForOpponentToCounter"
         opacity=".6"
       >
-        <h1>Opponent May Counter</h1>
+        <h1>{{ waitingForOpponentToCounterMessage }}</h1>
       </v-overlay>
       <v-overlay
         id="waiting-for-opponent-discard-scrim"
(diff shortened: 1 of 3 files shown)


client/js/components/GameView/GameOverlays.vue
```diff
@@ -83,7 +83,22 @@ export default {
...
+      'hasGlassesEight',
+    ]),
+    waitingForOpponetToCounterMessage() {
+      const mayCounter = 'Opponent May Counter';
+      const mustResolve = 'Opponent Must Resolve';
+      const opponentHasTwo = this.opponent.hand.some((card) => card.rank === 2);
+      if (this.playerQueenCount || (this.hasGlassesEight && !opponentHasTwo)) {
+        return mustResolve;
+      }
+      return mayCounter;
```
seriouslysean, line 100:
```suggestion
      return (this.playerQueenCount || (this.hasGlassesEight && !opponentHasTwo)) ?
        mustResolve : mayCounter;
```
Don't merge this as-is because prettier will yell, but you could also one line this. Probably cleaner with the if block though.

seriouslysean (approved):

client/js/components/GameView/GameOverlays.vue
itsalaidbacklife: Cool thanks. I was on the fence about a one-liner. It's helpful to see it written out like this. I think I'll stick with the conditional just to be explicit and obvious

PR #66, opened 2022-08-02
Move overlays to separate component

seriouslysean:
## Please check the following

- For New Features:

### DIFF
--- a/client/js/components/GameView/GameOverlays.vue
+++ b/client/js/components/GameView/GameOverlays.vue
@@ -1,0 +1,94 @@
+<template>
+  <div class="game-overlays">
+    <v-overlay
+      id="waiting-for-opponent-counter-scrim"
+      v-model="waitingForOpponentToCounter"
+      opacity=".6"
+    >
+      <h1>Opponent May Counter</h1>
+    </v-overlay>
+    <v-overlay
+      id="waiting-for-opponent-discard-scrim"
+      v-model="waitingForOpponentToDiscard"
+      opacity=".6"
+    >
+      <h1>Opponent Is Discarding</h1>
+    </v-overlay>
+    <v-overlay
+      id="waiting-for-opponent-resolve-three-scrim"
+      v-model="waitingForOpponentToPickFromScrap"
+      opacity=".6"
+    >
+      <h1>Opponent Choosing Card from Scrap</h1>
+    </v-overlay>
+    <v-overlay
+      id="waiting-for-opponent-play-from-deck-scrim"
+      v-model="waitingForOpponentToPlayFromDeck"
+      opacity=".6"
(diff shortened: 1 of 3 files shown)


client/js/components/GameView/GameOverlays.vue
```diff
@@ -0,0 +1,94 @@
...
+  },
+  computed: {
+    // Since we're not using namespacing, we need to destructure the game module
+    // off of the global state to directly access the state values
+    ...mapState({
+      waitingForOpponentToCounter: ({ game }) => game.waitingForOpponentToCounter,
+      waitingForOpponentToDiscard: ({ game }) => game.waitingForOpponentToDiscard,
+      waitingForOpponentToPickFromScrap: ({ game }) => game.waitingForOpponentToPickFromScrap,
+      waitingForOpponentToPlayFromDeck: ({ game }) => game.waitingForOpponentToPlayFromDeck,
+    }),
```
itsalaidbacklife, line 85: Is that the case? Looking at the [vuex docs](https://vuex.vuejs.org/guide/state.html#the-mapstate-helper) it looks like we might be able to just pass a string name. Would this work?

```suggestion
    ...mapState([
      'waitingForOpponentToCounter',
      'waitingForOpponentToDiscard',
      'waitingForOpponentToPickFromScrap',
      'waitingForOpponentToPlayFromDeck',
    ]),
```

client/js/views/GameView.vue
```diff
@@ -375,7 +344,11 @@ export default {
     };
   },
   computed: {
+    ...mapState({
+      waitingForOpponentToPlayFromDeck: ({ game }) => game.waitingForOpponentToPlayFromDeck,
```
itsalaidbacklife, line 348: This would be the same as above if it works.
```suggestion
      waitingForOpponentToPlayFromDeck'
```

client/js/views/GameView.vue
```diff
@@ -375,7 +344,11 @@ export default {
...
+    }),
```
itsalaidbacklife, line 349: This would be the same as above if I'm right.
```suggestion
    ...mapState([
      'waitingForOpponentToPlayFromDeck',
    ]),
```

itsalaidbacklife (changes requested):
Looks great! Solid improvement. Is the destructing necessary for mapState? I haven't worked with it but it looks like you can just pass a string?

client/js/components/GameView/GameOverlays.vue
seriouslysean: I did that first and they’re all undefined. This isn’t flat like getters seems to be.

itsalaidbacklife (approved):
Great work

PR #71, opened 2022-08-22
Feat/request stalemate

itsalaidbacklife:
## Please check the following

- For New Features:

**REQUIRES DATABASE UPDATE:** (adds two columns to game table)
Allows players to request a stalemate from their opponent which can be accepted or rejected. Stalemates must be accepted on the turn that they are requested or else the request is considered invalid.

### DIFF
--- a/api/controllers/game/stalemate-reject.js
+++ b/api/controllers/game/stalemate-reject.js
@@ -1,0 +1,30 @@
+module.exports = async function (req, res) {
+  try {
+    const game = await Game.findOne({ id: req.session.game }).populate('players');
+    let gameUpdates = {
+      turnStalemateWasRequestedByP0: -1,
+      turnStalemateWasRequestedByP1: -1,
+    };
+
+    const victory = {
+      gameOver: false,
+      winner: null,
+      conceded: false,
+    };
+
+    await Game.updateOne({ id: game.id }).set(gameUpdates);
+
+    Game.publish([game.id], {
+      verb: 'updated',
+      data: {
+        change: 'rejectStalemate',
+        game,
+        victory,
+        requestedByPNum: req.session.pNum,
+      },
+    });
+    return res.ok();
+  } catch (err) {
+    return res.badRequest(err);
+  }
+};
--- a/api/controllers/game/stalemate.js
+++ b/api/controllers/game/stalemate.js
@@ -46,6 +46,7 @@ module.exports = async function (req, res) {
         change: 'requestStalemate',
         game,
         victory,
+        requestedByPNum: req.session.pNum,
       },
     });
     return res.ok();
--- a/api/controllers/game/stalemate.js
+++ b/api/controllers/game/stalemate.js
@@ -1,35 +1,46 @@
 module.exports = async function (req, res) {
   try {
     const game = await Game.findOne({ id: req.session.game }).populate('players');
+    let gameUpdates = {};
+    const updatePromises = [];
+
     switch (req.session.pNum) {
-      case 0:
-        game.turnStalemateWasRequestedByP0 = game.turnStalemateWasRequestedByP0 ? game.turn : null;
+      case 0: {
+        const newVal = game.turnStalemateWasRequestedByP0 ? game.turn : null;
+        gameUpdates.turnStalemateWasRequestedByP0 = newVal;
+        game.turnStalemateWasRequestedByP0 = newVal;
         break;
-      case 1:
-        game.turnStalemateWasRequestedByP1 = game.turnStalemateWasRequestedByP1 ? game.turn : null;
+      }
+      case 1: {
+        const newVal = game.turnStalemateWasRequestedByP1 ? game.turn : null;
+        gameUpdates.turnStalemateWasRequestedByP1 = newVal;
+        game.turnStalemateWasRequestedByP1 = newVal;
+        break;
+      }
     }
     const victory = {
       gameOver: false,
       winner: null,
       conceded: false,
     };
+
     // End in stalemate if both players requested stalemate this turn
     if (
       game.turnStalemateWasRequestedByP0 === game.turnStalemateWasRequestedByP1 &&
       game.turnStalemateWasRequestedByP0 === game.turn
     ) {
       victory.gameOver = true;
-      const gameUpdates = {
+      gameUpdates = {
+        ...gameUpdates,
         p0: game.players[0].id,
         p1: game.players[1].id,
         result: gameService.GameResult.STALEMATE,
       };
-      await Promise.all([
-        Game.updateOne({ id: game.id }).set(gameUpdates),
-        gameService.clearGame({ userId: req.session.usr }),
-      ]);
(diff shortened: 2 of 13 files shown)


api/controllers/game/stalemate-reject.js
```diff
@@ -0,0 +1,30 @@
+module.exports = async function (req, res) {
+  try {
+    const game = await Game.findOne({ id: req.session.game }).populate('players');
+    let gameUpdates = {
+      turnStalemateWasRequestedByP0: -1,
+      turnStalemateWasRequestedByP1: -1,
```
seriouslysean, line 6: I feel like these should be `null` instead since it's a meaningful choice. Though this way you get to always know it's a number which is also fine with me.

api/controllers/game/stalemate-reject.js
```diff
@@ -0,0 +1,30 @@
...
+      data: {
+        change: 'rejectStalemate',
+        game,
+        victory,
+        requestedByPNum: req.session.pNum,
+      },
+    });
+    return res.ok();
+  } catch (err) {
+    return res.badRequest(err);
```
seriouslysean, line 28: Not for this PR, but I'd love to take a larger look at error logging across the board and get more meaningful dev logs + client feedback.

api/controllers/game/stalemate.js
```diff
@@ -0,0 +1,56 @@
+module.exports = async function (req, res) {
+  try {
+    const game = await Game.findOne({ id: req.session.game }).populate('players');
```
seriouslysean, line 3: I've been working on doing a bit more destructuring, is that something you'd be in interested in using across the codebase?
```suggestion
  try {
    const { game, pNum } = req.session; 
    const game = await Game.findOne({ id: req.session.game }).populate('players');
```

api/controllers/game/stalemate.js
```diff
@@ -0,0 +1,56 @@
...
+        game.turnStalemateWasRequestedByP0 = newVal;
+        break;
+      }
+      case 1: {
+        const newVal = game.turnStalemateWasRequestedByP1 ? game.turn : null;
+        gameUpdates.turnStalemateWasRequestedByP1 = newVal;
+        game.turnStalemateWasRequestedByP1 = newVal;
+        break;
+      }
+    }
```
seriouslysean, line 20: Switch works here, but you could probably also DRY up the logic a bit.
```suggestion
    const { pNum } = req.session;
    const turnKey = `turnStalemateWasRequestedByP${pNum}`;
    const newVal = game[turnKey] ? game.turn : null;
    gameUpdates[turnKey] = newVal;
    game[turnKey] = newVal;
```

seriouslysean: You'd also need to update the below code, so don't just add this. It's strictly a suggestion for code style.

api/models/Game.js
```diff
@@ -76,7 +76,14 @@ module.exports = {
       columnType: 'text[]',
       defaultsTo: [],
     },
-
+    turnStalemateWasRequestedByP0: {
+      type: 'number',
```
seriouslysean, line 80: Ah yeah, you do assert it's a number. Nice!

client/js/components/GameView/GameDialogs.vue
```diff
@@ -33,6 +33,7 @@
     />
     <game-over-dialog v-model="gameIsOver" :player-wins="playerWins" :stalemate="stalemate" />
     <reauthenticate-dialog v-model="mustReauthenticate" />
+    <opponent-requested-stalemate-dialog v-model="consideringOpponentStalemateRequest" />
```
seriouslysean, line 36: Look we're using the dialog component! 😎

client/js/components/GameView/GameDialogs.vue
```diff
@@ -124,6 +127,14 @@ export default {
...
       return this.$store.state.game.gameIsOver && this.$store.state.game.winnerPNum === null;
     },
+    consideringOpponentStalemateRequest: {
+      get() {
+        return this.$store.state.game.consideringOpponentStalemateRequest;
+      },
+      set(val) {
+        this.$store.commit('setConsideringOpponentStalemateRequest', val);
+      },
+    },
```
seriouslysean, line 137: I mulled over the best way to do this for a while when I was breaking these components out. But due to needing to set and get from the same value for Vuetify components I landed on this just being acceptable, haha.

client/js/components/GameView/GameMenu.vue
```diff
@@ -57,19 +60,70 @@ export default {
...
+      },
+    },
+    dialogTitle() {
+      return this.showConcedeDialog ? 'Concede' : 'Request Stalemate';
+    },
+    dialogText() {
+      return this.showConcedeDialog
+        ? 'The game will end and your opponent will win.'
+        : 'If your opponent agrees, the game will end in a stalemate. The request will cancel if the opponent declines or either player makes a move';
+    },
```
seriouslysean, line 87: We should eventually move all of these strings to a localization json file for ease of translating in the future.

client/js/plugins/sails.js
```diff
@@ -115,6 +115,15 @@ io.socket.on('game', function (evData) {
         case 'reLogin':
           store.dispatch('updateGameThenResetPNumIfNull', evData.data.game);
           break;
+        case 'requestStalemate':
+          if (evData.data.requestedByPNum !== store.state.game.myPNum) {
+            store.commit('setConsideringOpponentStalemateRequest', true);
+          }
```
seriouslysean, line 121: I'd switch this to a fast fail for readability and not needing to nest.
```suggestion
          if (evData.data.requestedByPNum === store.state.game.myPNum) {
            return;
          }
          store.commit('setConsideringOpponentStalemateRequest', true);
```

seriouslysean (approved):
Looks good to me! I left a few comments around style and future improvements, but as-is I don't have anything that is a deal breaker.

api/controllers/game/stalemate-reject.js
itsalaidbacklife: Yeah this was mostly b/c of database adapter type safety. Basically sails gets pissy using null for number columns despite the fact that it's actually fine in postgres

api/controllers/game/stalemate-reject.js
itsalaidbacklife: yeah this is a bit weak there

api/controllers/game/stalemate.js
itsalaidbacklife: Yes I think destructuring is nice. Will use (though in this case it can't be the exact suggestion due to name collision, and we'll want to use the variable in the method call

client/js/components/GameView/GameMenu.vue
itsalaidbacklife: agreed

client/js/plugins/sails.js
itsalaidbacklife: What's funky about this is that you still have the nested return. I think it's slightly more clear as-is

PR #72, opened 2022-08-22
Update deps, fix logout logic, add persistent auth status (no lobby or game logic)

seriouslysean:
## Please check the following

- For New Features:


.gitignore
```diff
@@ -39,7 +39,6 @@
 #
 ################################################
 
-config/local.js
```
seriouslysean, line 42: This seems useful to keep locally, for local only settings. I wasn't sure if you were using development for that purpose-- if so we can get rid of this.

api/controllers/UserController.js
```diff
@@ -100,8 +100,27 @@ module.exports = {
   },
 
   logout: function (req, res) {
-    delete req.session.usr;
-    req.session.loggedIn = false;
-    return res.ok();
+    req.session.destroy(function afterDestroy() {
+      return res.ok();
+    });
+  },
```
seriouslysean, line 106: This is the better logout logic so that it properly destroys the session.

api/controllers/UserController.js
```diff
@@ -100,8 +100,27 @@ module.exports = {
...
+      const { username } = await userAPI.findUser(id);
+      return res.ok({
+        id,
+        username,
+        authenticated,
+      });
+    } catch (err) {
+      return res.badRequest(err);
+    }
   },
```
seriouslysean, line 125: This is a barebones implementation of the User status check to only check whether they are authenticated or not.

client/js/components/NavigationDrawer.vue
```diff
@@ -65,7 +65,7 @@ export default {
             ...rules,
           ]
         : [
-            { text: 'Logout', icon: 'logout', page: { name: 'Login' } },
+            { text: 'Logout', icon: 'logout', page: { name: 'Logout' } },
```
seriouslysean, line 68: This prefers middleware to handle the logout process which feels much more reliable and safe.

client/js/main.js
```diff
@@ -5,14 +5,29 @@ import store from './store/store';
...
-}).$mount('#app');
+async function initApp() {
+  // Make sure we try and reestablish a player's session if one exists
+  // We do this before the app mounts to preempt the reauth/logout logic
+  try {
+    await store.dispatch('requestStatus');
+  } catch {
+    // Unable to validate player status or player not authenticated
+    // Swallow the error to let the player continue
+  }
```
seriouslysean, line 16: This will set the user's auth status so if you reload on the homepage, the stats page or the game list page it'll keep you logged in.

client/js/main.js
```diff
@@ -5,14 +5,29 @@ import store from './store/store';
...
+    router,
+    render: (h) => h(App),
+  }).$mount('#app');
+
+  // Expose app for testing
+  if (process.env.NODE_ENV === 'development' || window.Cypress) {
+    window.cuttle = {
+      app,
+    };
+  }
```
seriouslysean, line 30: Changed the namespace to keep some consistency, and enabled it on dev mode also.

client/js/router/index.js
```diff
@@ -17,6 +17,11 @@ const mustBeAuthenticated = (to, from, next) => {
   return next('/login');
 };
 
+const logoutAndRedirect = async (to, from, next) => {
+  await store.dispatch('requestLogout');
+  return next('/login');
+};
```
seriouslysean, line 23: Middleware to fire the logout action and then redirect to the login page.

client/js/router/index.js
```diff
@@ -29,6 +34,13 @@ const routes = [
     name: 'Login',
     component: LoginSignup,
   },
+  // This route is just a passthrough to make sure the user is fully logged out before putting
+  // them on the login screen
+  {
+    path: '/logout',
+    name: 'Logout',
+    beforeEnter: logoutAndRedirect,
+  },
```
seriouslysean, line 43: All logout action is magic now when you hit `/logout`.

client/js/store/modules/auth.js
```diff
@@ -107,5 +107,28 @@ export default {
...
+        if (username) {
+          context.commit('authSuccess', username);
+        }
+
+        return response;
+      } catch (err) {
+        context.commit('clearAuth');
+        throw new Error(err);
+      }
+    },
```
seriouslysean, line 132: Barebones implementation of requesting the user's status, done as a fetch request so that we get the cookie set when using the server/client instances separately.

config/blueprints.js
```diff
@@ -2,159 +2,36 @@
...
- * a model exists with the same name as the controller, Sails will respond with built-in CRUD
- * logic in the form of a JSON API, including support for sort, pagination, and filtering.
- *
- * For more information on the blueprint API, check out:
- * http://sailsjs.org/#!/documentation/reference/blueprint-api
- *
- * For more information on the settings in this file, see:
- * http://sailsjs.org/#!/documentation/reference/sails.config/sails.config.blueprints.html
+ * For background on the blueprint API in Sails, check out:
+ * https://sailsjs.com/docs/reference/blueprint-api
```
seriouslysean, line 6: You can ignore the majority of these updates. I got rid of all the old comments and updated to the latest versions from the 1.4.0 sails release and kept any changes we made.

config/bootstrap.js
```diff
@@ -1,16 +1,14 @@
...
+ * For more information on seeding your app with fake data, check out:
+ * https://sailsjs.com/config/bootstrap
  */
 
-module.exports.bootstrap = function (cb) {
-  // It's very important to trigger this callback method when you are finished
-  // with the bootstrap!  (otherwise your server will never lift, since it's waiting on the bootstrap)
-  cb();
+module.exports.bootstrap = async function () {
+  //
```
seriouslysean, line 13: As far as I can tell you no longer need the callback function.

config/env/development.js
```diff
@@ -18,4 +18,7 @@ module.exports = {
   // models: {
   //   connection: 'someMongodbServer'
   // }
+
+  // config/sockets.js
+  onlyAllowOrigins: undefined,
```
seriouslysean, line 23: We may not need this since it's explicitly everything when not set, but I wanted to be specific. Don't care if we drop it though.

config/http.js
```diff
@@ -3,85 +3,53 @@
...
+     ***************************************************************************/
+    // bodyParser: (function _configureBodyParser(){
+    //   var skipper = require('skipper');
+    //   var middlewareFn = skipper({ strict: true });
+    //   return middlewareFn;
+    // })(),
+  },
+
+  // https://sailsjs.com/documentation/reference/configuration/sails-config-http#?properties
+  trustProxy: true,
```
seriouslysean, line 54: Made sure to keep your `trustProxy` key and verified it still exists since you need it for Heroku.

config/local.js
```diff
@@ -0,0 +1,26 @@
...
+  onlyAllowOrigins: undefined,
+
+  // config/security.js
+  security: {
+    cors: {
+      allRoutes: true,
+      allowOrigins: ['http://localhost:1337', 'http://localhost:8080'],
+      allowCredentials: true,
+    },
+  },
```
seriouslysean, line 25: This seemed like we only needed it locally but wanted your feedback on if `development` env was sufficient.

config/security.js
```diff
@@ -1,69 +1,47 @@
...
+   *                                                                           *
+   * CSRF protection should be enabled for this application.                   *
+   *                                                                           *
+   * For more information, see:                                                *
+   * https://sailsjs.com/docs/concepts/security/csrf                           *
+   *                                                                           *
+   ****************************************************************************/
+
+  // TODO: We should look in to enabling this for security reasons in the future
+  csrf: false,
```
seriouslysean, line 46: The example app had this as true but I don't think we can do this yet due to the way SSL is handled?

config/session.js
```diff
@@ -24,70 +21,22 @@ module.exports.session = {
...
+   *                                                                          *
+   * https://sailsjs.com/config/session                                       *
    *                                                                          *
    ***************************************************************************/
+  // isSessionDisabled: function (req){
+  //   return !!req.path.match(req._sails.LOOKS_LIKE_ASSET_RX);
+  // },
 
+  // https://sailsjs.com/documentation/reference/configuration/sails-config-session#?properties
+  name: 'cuttle.sid',
```
seriouslysean, line 38: Use a specific cuttle namespace for the session id cookie.

package.json
```diff
@@ -14,26 +14,28 @@
     "e2e:gui": "ENABLE_VUE_DEVTOOLS=true npx cypress open --config-file tests/cypress.config.js",
     "e2e:mockprod": "CUTTLE_ENV=production npm run e2e:server",
     "e2e:server": "VUE_APP_API_URL=http://localhost:1337 npx cypress run --config-file tests/cypress.config.js",
+    "e2e:server:gui": "VUE_APP_API_URL=http://localhost:1337 npx cypress open --config-file tests/cypress.config.js",
```
seriouslysean, line 17: Kept the two new commands I added in the previous PR since they're useful.

package.json
```diff
@@ -14,26 +14,28 @@
...
     "machinepack-passwords": "^2.3.0",
-    "sails": "^1.4.3",
+    "sails": "^1.5.3",
     "sails-disk": "^2.1.1",
     "sails-hook-orm": "^2.1.1",
     "sails-hook-sockets": "^1.5.5",
     "sails-postgresql": "^4.0.0",
     "sails.io.js": "^1.2.1",
-    "socket.io-client": "^2.3.0",
+    "socket.io-client": "^2.3.1",
```
seriouslysean, line 38: Updated a few deps in here but wanted to talk with you about this. Some of these are like ~7 years old at this point but a straight upgrade does break things. It also looks like the new sails generate adds some sailshq namespaced packages which threw me off.
```
"dependencies": {
  "sails": "^1.5.3",
  "sails-hook-apianalytics": "^2.0.3",
  "sails-hook-organics": "^2.0.0",
  "sails-hook-orm": "^4.0.0",
  "sails-hook-sockets": "^2.0.0",
  "@sailshq/connect-redis": "^3.2.1",
  "@sailshq/socket.io-redis": "^5.2.0",
  "@sailshq/lodash": "^3.10.3"
},
```

vue.config.js
```diff
@@ -39,4 +39,21 @@ module.exports = {
...
+      '/user': {
+        target: 'http://localhost:1337',
+        changeOrigin: true,
+      },
+      '/test': {
+        target: 'http://localhost:1337',
+        changeOrigin: true,
+      },
+    },
+  },
```
seriouslysean, line 58: Needed for the client calls to get data from the server successfully and set the cookie.

.gitignore
itsalaidbacklife: imo local only => different on each computer. Shared settings that are the same for localhost development belong in development.js

config/local.js
itsalaidbacklife: We should use development.js for this stuff. local settings should be removed from version control so they can be machine-specific ie if you want to set up a local postgres instance instead of mocking the db in memory you could configure it here without anyone else needing your same db setup

itsalaidbacklife: We should use development.js for this stuff. local settings should be removed from version control so they can be machine-specific ie if you want to set up a local postgres instance instead of mocking the db in memory you could configure it here without anyone else needing your same db setup

config/sockets.js
```diff
@@ -2,63 +2,30 @@
...
+   * > Never change this here without also configuring `io.sails.transports`  *
+   * > in your client-side code.  If the client and the server are not using  *
+   * > the same array of transports, sockets will not work properly.          *
+   * >                                                                        *
+   * > For more info, see:                                                    *
+   * > https://sailsjs.com/docs/reference/web-sockets/socket-client           *
    *                                                                          *
    ***************************************************************************/
-  // grant3rdPartyCookie: true,
+  // transports: [ 'websocket' ],
```
itsalaidbacklife, line 28: Interesting to see this. I wonder if using the websocket transport protocol offers console visibility into traffic

vue.config.js
itsalaidbacklife: Does this not affect prod/mockprod?

package.json
itsalaidbacklife: I'll take a deeper look

config/bootstrap.js
itsalaidbacklife: I'll poke around

itsalaidbacklife (commented):
Great work. Haven't checked out locally for the full skinny but it looks good. Feedback's in comments

seriouslysean:
Closing in lieu of https://github.com/cuttle-cards/cuttle/pull/73 and downscoping again.

PR #69, opened 2022-08-13
Add persistent sessions, rewrite login/logout flow

seriouslysean:
## Please check the following

- [x] Do the tests still pass? (see [Run the Tests](https://github.com/cuttle-cards/cuttle#run-the-tests))
- [x] Is the code formatted properly? (see [Linting (Formatting)](https://github.com/cuttle-cards/cuttle#linting-formatting))
- For New Features:
  - [x] Have tests been added to cover any new features or fixes?
  - [ ] Has the documentation been updated accordingly?

## Please describe additional details for testing this change

1. Run `npm run start:mockprod:tools`
2. Create a new account and login
3. Refresh the homepage
4. Verify you stay logged in
5. Log out from the sidebar
6. Refresh the login page
7. Verify you stay logged out
8. Go to a lobby
9. Refresh and verify you stay in the lobby and the name is populated properly
10. Go to a game
11. Refresh and verify you stay in the game
12. Run the tests and verify they all pass

PR #73, opened 2022-08-26
Feature: Add Initial Session Support

seriouslysean:
# Third Times the Charm 🤩 

Spiritual successor of the previous two attempts, downscoped to do this feature in pieces:
- https://github.com/cuttle-cards/cuttle/pull/69
- https://github.com/cuttle-cards/cuttle/pull/72

There are a few gotchas with this PR that are worth knowing:
- Sessions persist across reloads now
- Vue CLI will proxy requests to the server to maintain the same domain/port
- There is a new `requestStatus` call that we need _BEFORE VUE MOUNTS_ to set the initial session cookie
- If you reload on the lobby page, it will keep your auth status and kick you to the game list
- If you reload on the game page, it will not keep your auth status and will let you login again via the reconnect dialog
- The game list view will leave any lobbies you're a part of before showing the game list-- this means if you are in a game and then hit the game list it will remove you from the game

Best viewed **commit by commit** to understand the changes more easily, https://github.com/cuttle-cards/cuttle/pull/73/commits.

## Please check the following

- For New Features:


client/js/store/modules/auth.js
```diff
@@ -107,5 +107,51 @@ export default {
...
+        if (username) {
+          context.commit('authSuccess', username);
+        }
+
+        // If this is a lobby, redirect the user to the game list so they don't have to
+        // log back in again
+        if (isLobby) {
+          location.href = '/#/';
+          return;
+        }
```
seriouslysean, line 148: If you're authenticated, but trying to rejoin a lobby it won't work. We don't need to log you out or anything, so we just send them back to the game list (which will remove them from any lobby, freeing them up to join another if needed).

client/js/views/Home.vue
```diff
@@ -119,8 +119,18 @@ export default {
...
+    // Leave any existing lobbies before getting the game list
+    // Need to make sure we do these in order so the order of operations is:
+    //     leave -> list
+    try {
+      await this.$store.dispatch('requestLeaveLobby');
+    } catch (err) {
+      // A user may not actually be in a lobby, so we need to
+      // swallow this error so the component doesn't error out
+      // TODO: Improve this logic with an isInLobby helper
+    }
```
seriouslysean, line 132: This is a _little_ weird, but I wanted to make sure I called out why I'm doing this. We need to await the calls to make sure we do them in order. We also need to make sure we swallow any error from `requestLeaveLobby` because the user may not be in a lobby at the moment.

Any error thrown here would break the component's mounting method.

config/blueprints.js
```diff
@@ -2,159 +2,36 @@
...
- * a model exists with the same name as the controller, Sails will respond with built-in CRUD
- * logic in the form of a JSON API, including support for sort, pagination, and filtering.
- *
- * For more information on the blueprint API, check out:
- * http://sailsjs.org/#!/documentation/reference/blueprint-api
- *
- * For more information on the settings in this file, see:
- * http://sailsjs.org/#!/documentation/reference/sails.config/sails.config.blueprints.html
+ * For background on the blueprint API in Sails, check out:
+ * https://sailsjs.com/docs/reference/blueprint-api
```
seriouslysean, line 6: You can ignore almost every `config/*` file. I installed the latest sails version locally and built an app, then copied over the defaults. We had a lot of out of date comments, etc, that I wanted to get rid of. I'll call out the specific areas where things have changed.

config/bootstrap.js
```diff
@@ -1,16 +1,16 @@
...
  */
 
-module.exports.bootstrap = function (cb) {
-  // It's very important to trigger this callback method when you are finished
-  // with the bootstrap!  (otherwise your server will never lift, since it's waiting on the bootstrap)
-  cb();
+module.exports.bootstrap = async function (done) {
+  // Don't forget to trigger `done()` when this bootstrap function's logic is finished.
+  // (otherwise your server will never lift, since it's waiting on the bootstrap)
+  done();
```
seriouslysean, line 15: I double checked if we need the `done()` call and it looks like we do. They don't explicitly call it out but their generated app has it in place.

config/env/development.js
```diff
@@ -18,4 +18,7 @@ module.exports = {
   // models: {
   //   connection: 'someMongodbServer'
   // }
+
+  // config/sockets.js
+  onlyAllowOrigins: undefined,
```
seriouslysean, line 23:
```suggestion
```

config/session.js
```diff
@@ -24,70 +21,22 @@ module.exports.session = {
...
+   *                                                                          *
+   * https://sailsjs.com/config/session                                       *
    *                                                                          *
    ***************************************************************************/
+  // isSessionDisabled: function (req){
+  //   return !!req.path.match(req._sails.LOOKS_LIKE_ASSET_RX);
+  // },
 
+  // https://sailsjs.com/documentation/reference/configuration/sails-config-session#?properties
+  name: 'cuttle.sid',
```
seriouslysean, line 38: Changed the name of our session cookie to use the cuttle namespace.

tests/e2e/specs/out-of-game/navigation.spec.js
```diff
@@ -47,23 +47,33 @@ describe('Navigation Drawer', () => {
...
-    // Navigate to Home (Play)
-    cy.get('[data-nav=Play]').click();
-    cy.hash().should('equal', '#/');
-    // Navigate to Stats
-    cy.get('[data-nav=Stats]').click();
-    cy.hash().should('equal', '#/stats');
-    // Log out
-    cy.get('[data-nav=Logout]').click();
-    cy.hash().should('equal', '#/login');
+  describe('Navigates to Rules, Home, Stats, and Login pages', () => {
```
seriouslysean, line 50: New test to verify that the login status persists on refresh. 🙃

vue.config.js
```diff
@@ -39,4 +39,24 @@ module.exports = {
...
+      '/user': {
+        target: 'http://localhost:1337',
+        changeOrigin: true,
+      },
+      '/test': {
+        target: 'http://localhost:1337',
+        changeOrigin: true,
+      },
+    },
+  },
```
seriouslysean, line 61: Need this for the cookie logic to work when running discrete client/server apps.

api/controllers/UserController.js
```diff
@@ -100,8 +100,35 @@ module.exports = {
...
-    req.session.loggedIn = false;
-    return res.ok();
+    // If the user isn't logged in, just get them out of here
+    if (!req.session.loggedIn) {
+      return res.ok();
+    }
+    req.session.destroy(function afterDestroy() {
+      return res.ok();
+    });
+  },
```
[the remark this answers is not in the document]
itsalaidbacklife: Man I can't find docs on this at all, which totally sucks. Do you know if `req.session.destroy()` returns a promise? Maybe we can `await` it

api/controllers/UserController.js
```diff
@@ -100,8 +100,35 @@ module.exports = {
...
 
   logout: function (req, res) {
-    delete req.session.usr;
-    req.session.loggedIn = false;
-    return res.ok();
+    // If the user isn't logged in, just get them out of here
+    if (!req.session.loggedIn) {
+      return res.ok();
+    }
+    req.session.destroy(function afterDestroy() {
```
seriouslysean, line 107:
```suggestion
    // https://github.com/expressjs/session#sessiondestroycallback
    req.session.destroy(function afterDestroy() {
```

api/controllers/UserController.js
```diff
@@ -100,8 +100,36 @@ module.exports = {
...
 
   logout: function (req, res) {
-    delete req.session.usr;
-    req.session.loggedIn = false;
-    return res.ok();
+    // If the user isn't logged in, just get them out of here
+    if (!req.session.loggedIn) {
+      return res.ok();
+    }
+    // https://github.com/expressjs/session#sessiondestroycallback
```
seriouslysean, line 107: Finally found it-- it's an express session package, https://github.com/expressjs/session#sessiondestroycallback.

seriouslysean: This doesn't appear to be a promise but we _could_ theoretically call the underlying function that `session.js` uses, https://github.com/expressjs/session/blob/master/session/session.js#L101-L113.

Something like this?
```
const destroyUserSession = (req, res) => {
        // Might need to be more defensive
        if (!req.session.id) {
            return reject('User not logged in');
        }
        req.session.destroy(function afterDestroy() {
            resolve('User logged out');
        });
    });
}

//

logout: async function (req, res) {
    await destroyUserSession(req, res);
    return res.ok();
},
```

Though I don't see a need to do it this way, we just avoided the abstraction.

itsalaidbacklife: Cool good to know; thanks for digging. I agree; no point in abstracting it into a helper just to await` here

api/controllers/UserController.js
```diff
@@ -100,8 +100,36 @@ module.exports = {
...
+      // If the user is logged in, see if we can find them first to verify they exist
+      const { username } = await userAPI.findUser(id);
+      return res.ok({
+        id,
+        username,
+        authenticated,
+      });
+    } catch (err) {
+      return res.badRequest(err);
+    }
```
itsalaidbacklife, line 133: Should we log you out in this case? Like if your session says you're user id 1337, but we can't find that user, maybe we should clear the session data so you can reauthenticate as a real user?

client/js/main.js
```diff
@@ -5,14 +5,29 @@ import store from './store/store';
...
-  window.app = app;
+  const app = new Vue({
+    vuetify,
+    store,
+    router,
+    render: (h) => h(App),
+  }).$mount('#app');
+
+  // Expose app for testing
+  if (process.env.NODE_ENV === 'development' || window.Cypress) {
```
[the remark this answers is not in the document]
itsalaidbacklife: Here's a question: am I right that the node env vars are read-only to the application? Maybe we should only check that the mode is development so that it's harder to force the application to expose its internals, since `process.env.NODE_ENV` is set during build time and theoretically can't be overwritten by the client. What do you think? This would however limit the e2e test suite to `development` mode, which to be fair is generally what we do anyway. Just food for thought

```suggestion
  if (process.env.NODE_ENV === 'development') {
```

client/js/main.js
```diff
@@ -5,14 +5,29 @@ import store from './store/store';
...
 
-const app = new Vue({
-  vuetify,
-  store,
-  router,
-  render: (h) => h(App),
-}).$mount('#app');
+async function initApp() {
+  // Make sure we try and reestablish a player's session if one exists
+  // We do this before the app mounts to preempt the reauth/logout logic
```
itsalaidbacklife, line 10: The comment says that we run this check before mount, but `initApp()` is called below the application mount as the last line of the file. What's the intention here?

client/js/store/modules/auth.js
```diff
@@ -107,5 +107,51 @@ export default {
         );
       });
     },
+    async requestStatus(context) {
+      if (!window) {
+        // Swallow error, this call is only supported client side
+        return;
+      }
```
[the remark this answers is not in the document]
itsalaidbacklife:
```suggestion
```
My thought is we leave it out until a time where it's relevant. By the time we more seriously consider SSR it's possible other things will have changed such that this will be stale. I'm not strongly opinionated about it, so I'll leave it to your judgement but my thought is we call any SSR considerations out of scope until we tackle it holistically.

client/js/store/modules/auth.js
```diff
@@ -107,5 +107,51 @@ export default {
...
+
+      const { location } = window;
+      const isLobby = location.hash.startsWith('#/lobby');
+      const isGame = location.hash.startsWith('#/game');
```
itsalaidbacklife, line 118: So I'm fuzzy on the bootstrapping lifecycle. I'm surprised the vue router won't necessarily be available at this time because we're inside a vuex action so doesn't the app have to have mounted for this function to fire? Shouldn't that guarantee that the vue router is available? Not that this is a terrible way to check the route, I'd just like to learn more about when which resources are available and why this was necessary

client/js/store/modules/auth.js
```diff
@@ -107,5 +107,51 @@ export default {
...
+        // log back in again
+        if (isLobby) {
+          location.href = '/#/';
+          return;
+        }
+
+        return response;
+      } catch (err) {
+        context.commit('clearAuth');
+        throw new Error(err);
```
itsalaidbacklife, line 153: Silly nitpick, but any reason to wrap this in a new error vs throwing the OG?
```suggestion
        throw err;
```

client/js/views/Home.vue
```diff
@@ -119,8 +119,18 @@ export default {
...
+  async mounted() {
+    // Leave any existing lobbies before getting the game list
+    // Need to make sure we do these in order so the order of operations is:
+    //     leave -> list
+    try {
+      await this.$store.dispatch('requestLeaveLobby');
+    } catch (err) {
+      // A user may not actually be in a lobby, so we need to
+      // swallow this error so the component doesn't error out
+      // TODO: Improve this logic with an isInLobby helper
```
itsalaidbacklife, line 131: We may also want to make the backend more robust here as a part of that cleanup. I think it'd make sense for the backend to just silently 200 you if you request to leave a lobby when you're not in one. Then we wouldn't expect an error here anyway

client/js/views/Home.vue
itsalaidbacklife: Why do they need to be in order? Is it just for the different error handling? Not a big deal but from a performance standpoint this seems like a potentially reasonable use-case for `Promise.all()`, with the caveat that it could theoretically result in joining a game super quickly and having the requestLeaveLobby fire late and remove you from the game you _just_ joined

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -149,9 +149,10 @@ describe('Lobby - P0 Perspective', () => {
       assertGameStarted();
     });
   });
-  it.skip('[Missing Feature] Loads lobby after page refresh', () => {
+  // TODO: Rewrite this test when auth is supported in the lobby
+  it.skip('Redirects to game list after page refresh', () => {
```
itsalaidbacklife, line 153: Is this test skipped deliberately? I thought that this feature was in-scope for this PR
```suggestion
  it('Redirects to game list after page refresh', () => {
```

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -149,9 +149,10 @@ describe('Lobby - P0 Perspective', () => {
...
     cy.reload();
-    expect(true).to.eq(false, 'Empty Test');
+    cy.hash().should('equal', '#/');
```
itsalaidbacklife, line 155: If we want to be a bit more thorough here we could assert that you can then join a game, see yourself as the only player, have an opponent join, and see that the game starts normally

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -243,7 +248,9 @@ describe('Lobby - P1 Perspective', () => {
     cy.readyOpponent();
     assertGameStarted();
   });
-  it.skip('[Feature Missing] Loads lobby after page refresh', () => {
-    expect(true).to.eq(false, 'Empty Test');
+  // TODO: Rewrite this test when auth is supported in the lobby
+  it.skip('Redirects to game list after page refresh', () => {
```
itsalaidbacklife, line 252: Same as above -- still skip?
```suggestion
  it('Redirects to game list after page refresh', () => {
```

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -243,7 +248,9 @@ describe('Lobby - P1 Perspective', () => {
...
+    cy.reload();
+    cy.hash().should('equal', '#/');
```
itsalaidbacklife, line 254: Same as above -- could confirm that you can join another game and have it work correctly

tests/e2e/specs/out-of-game/navigation.spec.js
```diff
@@ -47,23 +47,33 @@ describe('Navigation Drawer', () => {
...
+      cy.signupPlayer(playerOne.username, playerOne.password);
+      cy.vueRoute('/');
+    });
+    it('When authenticated', () => {
+      verifyAuthenticatedLinks();
+    });
+    it('When authenticated on reload', () => {
+      cy.reload();
+      verifyAuthenticatedLinks();
+    });
```
itsalaidbacklife, line 77: *Chef's kiss*

itsalaidbacklife (approved):
This is looking great. Feedback's in the comments. Huge QoL bump incoming!

client/js/store/modules/auth.js
seriouslysean: image

This is what happens if you try and pass the router to this action before we's mounted vue. I have a feeling we could potentially a lifecycle method the app itself and fire stuff there with the router but in my initial attempts I was not successful due to the order of operations. Something like this:
```
const app = new Vue({
  vuetify,
  store,
  router,
  async beforeCreate() {
    // Make sure we try and reestablish a player's session if one exists
    // We do this before the app mounts to preempt the reauth/logout logic
    try {
      await store.dispatch('requestStatus', this.$route);
    } catch {
      // Unable to validate player status or player not authenticated
      // Swallow the error to let the player continue
    }
  },
  render: (h) => h(App),
}).$mount('#app');
```

But what happens is that the vue router beforeResolve fires before the beforeCreate lifecycle method which will redirect you to the /login route when you're actually authed. Something to look into as a follow up, though, for sure.

tests/e2e/specs/out-of-game/lobby.spec.js
itsalaidbacklife: That makes sense. Not sure the right way to balance the overhead of redundant test flows against the potential confidence that that same logic works on refresh. I like your idea of checking a full happy flow after reconnect though then there is the consideration of which reconnect situations to fully check.

Either way I think this is fine for now

client/js/store/modules/auth.js
seriouslysean: The other thing we could potentially look in to doing is adding this call to vue router middleware, but I'd need to figure out how to run this once before the router did everything else.

itsalaidbacklife (approved):
Great work

client/js/store/modules/auth.js
seriouslysean: So I'm torn about this. I still don't know the best practices surrounding this, https://www.geeksforgeeks.org/difference-between-throw-errormsg-and-throw-new-errormsg/. (which is more about `Error()` new `new Error()`, but hear me out.


I'm not actually calling throw in the try part of the block, I'm relying on the calls themselves to fail for one reason or the other to get me in to the catch block, and then calling new Error manually.

I guess it depends on the way fetch is sending it. When you throw a new error it will wrap the message, something like this:
image

I used this code in the browser console:
```
(async () => {
    try {
        const response = await fetch('http://localhost:1234');
        if (!response.ok) {
            throw new Error('response not ok');
        }
        return response;
    } catch (err) {
        console.log('err (before):', err);
        const newErr = new Error(err);
        console.log('err (after):', newErr);
        throw newErr;
    };
})();
```

All this leads back to "I don't know" so I did it the way I usually do.

**This did make me take a more top down view as to what I was trying to accomplish and since I was just catching the error in the `main.js` file and silently failing made me realize we don't need to throw an error at all so I removed it.**

PR #76, opened 2022-09-03
Feature: Add Session Support to the Game Route

seriouslysean:
## Please check the following

- For New Features:


client/js/main.js
```diff
@@ -5,24 +5,16 @@ import store from './store/store';
...
-  if (process.env.NODE_ENV === 'development' || window.Cypress) {
-    window.cuttle = {
-      app,
-    };
-  }
+// Expose app for testing
+if (process.env.NODE_ENV === 'development' || window.Cypress) {
+  window.cuttle = {
+    app,
+  };
```
seriouslysean, line 19: By moving the initial auth logic to the vue router implementation, we can simplify this file again.

client/js/router/index.js
```diff
@@ -8,6 +8,9 @@ import Rules from '../views/Rules.vue';
 import Stats from '../views/Stats.vue';
 import store from '../store/store.js';
 
+export const ROUTE_NAME_LOBBY = 'Lobby';
+export const ROUTE_NAME_GAME = 'Game';
```
seriouslysean, line 12: Need these as consts to keep the references DRY as we need to use these values both here, and inside the auth status check action.

client/js/router/index.js
```diff
@@ -73,4 +76,14 @@ const router = new VueRouter({
...
 
+router.beforeEach(async (to, _from, next) => {
+  // Make sure we try and reestablish a player's session if one exists
+  // We do this before the route resolves to preempt the reauth/logout logic
+  await store.dispatch('requestStatus', {
+    router,
+    route: to,
+  });
+  next();
+});
```
seriouslysean, line 87: This is where the magic happens now. the `to` param is the current route, so we can use that to determine which logic to run. This also has the benefit of being within the Vue Router lifecycle, prior to any component level navigation guards so they can handle auth logic passed this point.

client/js/store/modules/auth.js
```diff
@@ -1,8 +1,10 @@
 import { io } from '../../plugins/sails.js';
+import { ROUTE_NAME_LOBBY, ROUTE_NAME_GAME } from '@/router';
 
 export default {
   state: {
-    authenticated: false,
+    // This value will ONLY be null on the initial load
+    authenticated: null,
```
seriouslysean, line 7: This is really important. We basically need to track 3 values here:
- `null` no auth attempt. This is a meaningful choice and purposely set by us so we can differentiate the initial load from a user being authed or not. This allows us to do auth logic once on the client side and avoid making the same request (or nooping) every route.
- `false` user is not authed (`clearAuth` also sets this)
- `true` user is authed (`authSuccess` also sets this)

client/js/store/modules/auth.js
```diff
@@ -107,10 +109,16 @@ export default {
...
-    async requestStatus(context) {
-      const { location } = window;
-      const isLobby = location.hash.startsWith('#/lobby');
-      const isGame = location.hash.startsWith('#/game');
+    async requestStatus(context, { router, route }) {
+      const { state } = context;
+
+      if (!router || !route || state.authenticated !== null) {
+        return;
+      }
```
seriouslysean, line 117: Need a new check here to make sure router & route exist, and that the user's auth status isn't anything but `null`.

client/js/store/modules/auth.js
```diff
@@ -139,11 +148,10 @@ export default {
         // If this is a lobby, redirect the user to the game list so they don't have to
         // log back in again
         if (isLobby) {
-          location.href = '/#/';
-          return;
+          return router.push('/');
```
seriouslysean, line 151: We can now use the router methods 😎

client/js/store/modules/auth.js
```diff
@@ -128,7 +136,8 @@ export default {
 
         // If the user is not authenticated, we're done here
         if (!authenticated) {
-          return response;
+          context.commit('clearAuth');
```
seriouslysean, line 139: If the user isn't authed, this will explicitly set false.

client/js/store/modules/auth.js
```diff
@@ -107,10 +109,16 @@ export default {
...
+    async requestStatus(context, { router, route }) {
+      const { state } = context;
+
+      if (!router || !route || state.authenticated !== null) {
+        return;
+      }
+
+      const { name } = route;
+      const isLobby = name === ROUTE_NAME_LOBBY;
+      const isGame = name === ROUTE_NAME_GAME;
```
seriouslysean, line 121: We can reference the route names by const now, which should dry this up a bit and prevent errors.

seriouslysean (commented):
Still a lot more to go but I figured out some of the problems plaguing the previous session PR.

client/js/store/modules/auth.js
itsalaidbacklife: How might we call this before router has loaded and what happens in that case when we return early? Does the page load later and detect the null auth?

itsalaidbacklife:
Looks great. A nice cleanup of the foundations of session recovery

api/controllers/UserController.js
```diff
@@ -71,7 +67,11 @@ module.exports = {
     userAPI
       .findUserByUsername(req.body.username)
       .then(function gotUser(user) {
-        const checkPass = passwordAPI.checkPass(req.body.password, user.encryptedPassword);
+        console.log();
```
itsalaidbacklife, line 70: teehee
```suggestion
```

itsalaidbacklife: I saw this when running the tests and was like wtf?
empty console gif

api/controllers/UserController.js
```diff
@@ -71,7 +67,11 @@ module.exports = {
...
+        const checkPass =
+          req.session.loggedIn !== true && req.body.password
```
itsalaidbacklife, line 72: Wondering about a falsey eval of req.session.loggedIn
```suggestion
          !req.session.loggedIn && req.body.password
```

itsalaidbacklife: I guess the null case is handled differently on purpose?

api/controllers/UserController.js
```diff
@@ -105,7 +105,7 @@ module.exports = {
   },
 
   status: async function (req, res) {
-    const { usr: id, loggedIn: authenticated } = req.session;
+    const { usr: id, loggedIn: authenticated, game } = req.session;
```
itsalaidbacklife, line 108: This might be clearer if it were called `gameId`. I know that's coming directly from the session but it's a little funky seeing it here without typescript telling us whether that's a number or obj

api/controllers/UserController.js
```diff
@@ -117,10 +117,17 @@ module.exports = {
     try {
       // If the user is logged in, see if we can find them first to verify they exist
       const { username } = await userAPI.findUser(id);
+
+      // If the user is currently in a game, we need to populate the game
+      if (game) {
+        gameService.populateGame({ gameId: game });
+      }
```
itsalaidbacklife, line 124: Doesn't this throw the value away? `gameService.populateGame` returns a value wrapped in a promise; it shouldn't have any side effects. What's this doing?

itsalaidbacklife: This sorta speaks to my confusion about the type of `game` in the obj param above because from this I expect it to be an int, but below you're returning the original `game` instead of the one that's populated by this call

itsalaidbacklife (commented):
WIP review. Submitting now so you can check my sweet gif

api/controllers/UserController.js
seriouslysean: I originally had it that way, I kind of wish the schema called it that too.

app.js
```diff
@@ -61,6 +63,15 @@ process.chdir(__dirname);
...
   }
 
+  // In dev mode, ensure unhandles promises log to the console with the full stack
+  // See https://stackoverflow.com/a/46661020
+  if (!isProd) {
+    process.on('unhandledRejection', (err) => {
+      console.error(err);
+      throw err;
+    });
+  }
```
seriouslysean, line 73: This was the key to debugging that unhandled promise. This will force it to console log as the full stack trace. 😎

api/controllers/UserController.js
seriouslysean: Fixed, thanks for calling that out. I actually had an extra populateGame call that I didn't even need.

seriouslysean:
@itsalaidbacklife the last push passed but I had to rebase back on to main again, haha.

itsalaidbacklife (approved):
Huge win with this one

PR #84, opened 2022-09-14
Gameview grid implementation

peterancourt:
## Please check the following

- For New Features:

Lots of changes here!
* Converted GameView.vue to use a grid layout using grid-template-areas
* Added overrides for styling on mobile devices:
    * Sliders for cards when a players hand is too wide
    * Moves draw/scrap piles to between players hand and game field
    * Moves history to a navigation drawer (scrollable)
    * Changes card sizes to account for smaller viewport

**Testing**
* Play around with sizing, the game view should be responsive and look as expected until 600px wide
* Mobile devices should have a new layout when in portrait mode

Screen Shot 2022-09-18 at 9 06 38 AM


client/js/App.vue
```diff
@@ -34,4 +34,10 @@ export default {
   -webkit-background-clip: text;
   -webkit-text-fill-color: transparent;
 }
+
+@media (max-width: 600px) {
+  div > .v-application--wrap {
+    min-height: 100%;
+  }
+}
```
seriouslysean, line 42: Need a comment above this with the vuetify documentation on using vh and maybe a link to where it doesn't work on mobile so we have some context.

client/js/components/GameView/Card.vue
```diff
@@ -194,4 +194,15 @@ export default {
     opacity: 0;
   }
 }
+
+@media (max-width: 600px) {
```
seriouslysean, line 198: I would love to see us standardize these widths for our breakpoints and move to a mixin style approach with name based widths (small, medium, large) or similar.

client/js/components/GameView/Card.vue
```diff
@@ -194,4 +194,15 @@ export default {
...
+  .player-card {
+    max-height: 10vh;
+    max-width: calc(10vh / 1.45);
+    &.glasses {
+      max-width: 10vh;
+      height: calc(10vh / 1.45);
+    }
+  }
```
seriouslysean, line 206: I would like to move cards to a width based implementation, using height feels weird though there may have been a reason for it. @itsalaidbacklife any idea?

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -245,6 +245,9 @@ export default {
       });
     },
     cardWidth() {
+      if (this.$vuetify.breakpoint.name === 'xs') {
+        return '100%';
+      }
```
seriouslysean, line 250: This feels really weird to target it in this way. If we're going to keep vuetify I would want to abstract this in to global vue helpers that work something like this:
```js
this.$isSmall = ['xs', 's'].includes(this.$vuetify.breakpoint.name);
this.$isMedium = this.$vuetify.breakpoint.name === 'md';
this.$isLarge = ['lg', 'xl'].includes(this.$vuetify.breakpoint.name);
```
or whatever the vuetify values would be so we have a single utility to update if they ever change.

seriouslysean (changes requested):
Really nice work, @peterancourt! Few suggestions before this gets merged, though.

client/js/components/GameView/Card.vue
itsalaidbacklife: Seconded. I think that can reasonably be a separate issue. Thoughts @seriouslysean?

seriouslysean: No need in this PR. Out of scope :)

client/js/components/GameView/MoveChoiceOverlay.vue
itsalaidbacklife: My issue with global mixins is they feel magical. It can be opaque where they come since they're not explicitly imported functions. The vuetify [breakpoint service](https://vuetifyjs.com/en/features/breakpoints/#breakpoint-service-object) actually has most of these utilities. `name` exists as a string basically so you can use `switch`statements with them, but there are also boolean values for whether the current breakpoint matches a specific size or whether it is less-than-or-equal-to it, etc.

So we could just
```suggestion
      if (this.$vuetify.breakpoint.xs) {
        return '100%';
      }
```

itsalaidbacklife: I think the breakpoint service has all the utilities/shortcuts we'd want here without needing to write it as a global mixin

client/js/components/GameView/Card.vue
itsalaidbacklife: Yeah I'm open to that. I think the general idea was to ensure that the grid took up the full height of the viewport, but there's no good reason why that really needs to apply to the cards specifically

seriouslysean (approved):
No deal breakers imo, few follow ups for later though.

client/js/components/GameView/MoveChoiceOverlay.vue
seriouslysean: That's fair, though I think some magic is ok when it eases developmental concerns. Same way that you can have something like `this.$cookie`, or `this.$logger`, etc. I think more appropriately we shouldn't need to do much at the JS level and it should be handled by the CSS side of things with some additional planning.

client/js/components/GameView/Card.vue
peterancourt: Also agreed, I can write an issue for that this week to take care of it

client/js/components/GameView/Card.vue
peterancourt: Another issue I can write up this week! Good call

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -245,6 +245,9 @@ export default {
       });
     },
     cardWidth() {
+      if (this.$vuetify.breakpoint.name === 'xs') {
```
itsalaidbacklife, line 248: More concise alternative ([docs](https://vuetifyjs.com/en/features/breakpoints/#breakpoint-service-object))
```suggestion
      if (this.$vuetify.breakpoint.xs) {
```

client/js/views/GameView.vue
```diff
@@ -6,69 +6,132 @@
     </template>
     <!-- Authenticated View -->
     <template v-else>
-      <div id="game-menu-wrapper">
+      <div id="game-menu-wrapper" class="d-flex flex-column">
         <game-menu />
+        <v-icon
+          v-if="$vuetify.breakpoint.name === 'xs'"
```
itsalaidbacklife, line 12:
```suggestion
          v-if="$vuetify.breakpoint.xs"
```

client/js/views/GameView.vue
```diff
@@ -6,69 +6,132 @@
...
+        </v-icon>
       </div>
 
-      <!-- Opponent Hand -->
-      <div
-        id="opponent-hand"
-        class="d-flex flex-column justify-start align-center px-2 pb-2 mx-auto"
+      <!-- Mobile History Drawer -->
+      <v-navigation-drawer
+        v-if="$vuetify.breakpoint.name === 'xs'"
```
itsalaidbacklife, line 23:
```suggestion
        v-if="$vuetify.breakpoint.xs"
```

client/js/views/GameView.vue
```diff
@@ -6,69 +6,132 @@
...
+                    name="slide-below"
                   >
-                    <v-card class="opponent-card-back" data-opponent-hand-card>
-                      <v-img :src="require('../img/logo_head.svg')" contain />
-                    </v-card>
-                  </div>
-                </transition-group>
-              </transition>
+                    <v-slide-group
+                      v-if="$vuetify.breakpoint.name === 'xs'"
```
itsalaidbacklife, line 74:
```suggestion
                      v-if="$vuetify.breakpoint.xs"
```

client/js/views/GameView.vue
```diff
@@ -115,88 +178,98 @@
...
-              />
-            </transition-group>
+              </transition-group>
+            </div>
           </div>
         </div>
+      </div>
+
+      <!-- History -->
+      <div v-if="$vuetify.breakpoint.name !== 'xs'" class="history-container">
```
itsalaidbacklife, line 272:
```suggestion
      <div v-if="$vuetify.breakpoint.xs" class="history-container">
```

client/js/views/GameView.vue
```diff
@@ -214,66 +287,92 @@
...
-                :rank="card.rank"
-                :is-selected="selectedCard && card.id === selectedCard.id"
-                :is-frozen="player.frozenId === card.id"
-                class="mt-8 transition-all"
-                :data-player-hand-card="`${card.rank}-${card.suit}`"
-                @click="selectCard(index)"
-              />
-            </transition-group>
+            <username-tool-tip
+              v-if="$vuetify.breakpoint.name !== 'xs'"
```
itsalaidbacklife, line 316:
```suggestion
              v-if="$vuetify.breakpoint.xs"
```

client/js/views/GameView.vue
```diff
@@ -214,66 +287,92 @@
...
+            />
+            <div class="player-cards-container">
+              <transition-group
+                tag="div"
+                name="slide-above"
+                class="d-flex justify-center align-start player-cards-mobile-overrides"
+                :class="{ 'my-turn': isPlayersTurn }"
+              >
+                <v-slide-group
+                  v-if="$vuetify.breakpoint.name === 'xs'"
```
itsalaidbacklife, line 330:
```suggestion
                  v-if="$vuetify.breakpoint.xs"
```

client/js/views/GameView.vue
```diff
@@ -6,69 +6,132 @@
...
+                    tag="div"
+                    name="slide-below"
                   >
-                    <v-card class="opponent-card-back" data-opponent-hand-card>
-                      <v-img :src="require('../img/logo_head.svg')" contain />
-                    </v-card>
-                  </div>
-                </transition-group>
-              </transition>
+                    <v-slide-group
```
itsalaidbacklife, line 73: It seems a little weird to nest the `v-slide-group` inside the regular vue `<transition-group>` becuase I assume that vuetify injects a `<transition-group>` for its implementation of `<v-slide-group>`. Does this work okay?

itsalaidbacklife (approved):
Great work. A couple small cleanup requests in the comments

client/js/views/GameView.vue
peterancourt: Oh interesting I didn't know that, It does seem to work without issue

peterancourt: @itsalaidbacklife in trying move the `v-slide-group` out of the `transition-group`, the layout started acting a bit goofy actually. If you're cool with it, maybe we can revisit this since it's currently working as expected?

itsalaidbacklife: Yeah that works for me

Issue #87, opened 2022-09-16
[Feature]: View effects of cards in play


itsalaidbacklife:
### Feature Summary

Want to be able to click a card in play to see what it does

### Detailed Description
Blocked by #103 

Players should be able to click cards to see what they do, even when those cards are on the field, or revealed in opponent's hand (when you have glasses).

Should consider the design challenge of how to communicate whether the displayed card is in your hand and therefore playable, or anywhere else

PR #95, opened 2022-09-17
[Feature]: View effects of cards in play

seriouslysean:
## Please check the following

- For New Features:


utils/game-utils.js
```diff
@@ -1,11 +1,26 @@
+/**
+ * @returns int <= 0 if card1 is lower rank or same rank & lower suit
+ * @param {rank: number, suit: number} card1
+ * @param {rank: number, suit: number} card2
+ */
+function compareByRankThenSuit(card1, card2) {
+  // Compare by suit (clubs->diamonds->hearts->spades)
+  // Then compare by rank (1,2...12,13)
+  return card1.suit - card2.suit || card1.rank - card2.rank;
+}
```
seriouslysean, line 10: Moved the card compare util here and improved the logic.

utils/game-utils.js
```diff
@@ -1,11 +1,26 @@
...
+
 const getPlayerPnumByUsername = (players, username) => {
+  if (!Array.isArray(players)) {
+    return null;
+  }
```
seriouslysean, line 15: Made this a bit more defensive just in case (also for testing expectations).

tests/unit/utils/game-utils.spec.js
```diff
@@ -0,0 +1,72 @@
+const {
```
seriouslysean, line 1: Test coverage for all the new game utils we added over the last few PRs.

client/js/store/modules/game.js
```diff
@@ -71,6 +72,33 @@ const initialState = resetState();
...
+        ...getters.opponent.faceCards,
+      ]
+        .map(({ id, rank, suit, name, ruleText }) => ({
+          id,
+          rank,
+          suit,
+          name,
+          ruleText,
+        }))
+        .sort(compareByRankThenSuit);
```
seriouslysean, line 100: @itsalaidbacklife this is what I was referring to when I said I needed a full set of cards in order to get the information I need to show an overlay. I'm still thinking through the all the things this touches but the idea is to use this array instead of the players hand when looking at a card that is ouside of it (field, royals, opponents hand, etc).

There also need to be some other assumptions like:
- Can't view overlay on cards in the scrap (for now)
- What to do with the click events for cards that already have them
    - `@click="targetOpponentPointCard(index)"`
    - `@click="targetOpponentFaceCard(-index - 1)"`
    - `@click="targetOpponentFaceCard(index)"`
- Logic changes from changing from an index to the need to do a lookup by id or similar

api/services/gameService.js
```diff
@@ -33,9 +21,9 @@ function tempUser(usr, points) {
   this.id = usr.id;
   this.username = usr.username;
 
-  this.hand.sort(comapreByRankThenSuit);
-  this.points.sort(comapreByRankThenSuit);
-  this.faceCards.sort(comapreByRankThenSuit);
+  this.hand.sort(compareByRankThenSuit);
+  this.points.sort(compareByRankThenSuit);
+  this.faceCards.sort(compareByRankThenSuit);
```
seriouslysean, line 26: Typo :)

seriouslysean (commented):
Still a lot to do, but this is the start of preparing the overlays on any card click.

tests/unit/utils/game-utils.spec.js
```diff
@@ -0,0 +1,72 @@
...
+    suit: 0,
+  };
+  const aceOfHearts = {
+    rank: 1,
+    suit: 2,
+  };
+  const jackOfHearts = {
+    rank: 11,
+    suit: 2,
+  };
```
itsalaidbacklife, line 23: We could DRY this up by reusing the Card object that's defined at the bottom of tests/e2e/support/helpers.js

That already has all the cards defined using the suit and rank so then you could use `Card.ACE_OF_HEARTS` in the same way the e2e tests do. That said, we'd probably want to move the Card object into something shared by both e2e and unit tests to do so

utils/game-utils.js
```diff
@@ -1,11 +1,26 @@
+/**
+ * @returns int <= 0 if card1 is lower rank or same rank & lower suit
+ * @param {rank: number, suit: number} card1
+ * @param {rank: number, suit: number} card2
+ */
+function compareByRankThenSuit(card1, card2) {
+  // Compare by suit (clubs->diamonds->hearts->spades)
+  // Then compare by rank (1,2...12,13)
```
itsalaidbacklife, line 8: Comment is backwards--rank is considered first

itsalaidbacklife: Wait actually I'm confused because the return statement does appear to consider suit first but the function name implies rank is first. I'm not sure the full context yet but for scuttling purposes we sort by rank first and use suit to break a tie. Something's inconsistent here but I'm not sure which way it should go

client/js/store/modules/game.js
```diff
@@ -71,6 +72,33 @@ const initialState = resetState();
 export default {
   state: initialState,
   getters: {
+    // TODO: Replace this with an actual deck array :)
+    cards(state, getters) {
```
itsalaidbacklife, line 76: Let's further discuss the intended consumption of this getter. My instinct is that a dictionary/map would be more appropriate than an array to look up by id for example in constant time, but it depends on the usage. 

That said, I'm still of the opinion that the full list of cards may be unnecessary for the selection process, because the data you need in order to display a card and to determine what moves it can do is the cards suit and rank, both of which are already available on the individual card component that gets clicked. So it's not fully clear to me what we'll do with this full list. Let's discuss

utils/game-utils.js
itsalaidbacklife: Yeah okay I think the logic and the tests should switch to match the function name (rank then suit) because this function controls the order of cards in your hand and it makes sense to group them by rank more than by suit

itsalaidbacklife (commented):
Cool start. We should discuss the intended consumption of the new store getter before proceeding

client/js/views/GameView.vue
```diff
@@ -35,6 +35,7 @@
                     :rank="card.rank"
                     :data-opponent-hand-card="`${card.rank}-${card.suit}`"
                     class="transition-all opponent-hand-card-revealed"
+                    @click="selectCard(card)"
```
seriouslysean, line 38: It's going to be a little tricky to allow this on the face cards since you have targeting rules on click, @itsalaidbacklife. Could combine them perhaps, but might be worth taking another look at how those various rules work to think through a refactor.

client/js/views/GameView.vue
```diff
@@ -330,10 +333,11 @@ export default {
   },
   data() {
     return {
+      observedCard: null,
       showSnack: false,
       snackMessage: '',
       snackColor: 'error',
-      selectionIndex: null, // when select a card set this value
```
seriouslysean, line 336: No more need to selectedIndex, can directly use `selectedCard`.

client/js/views/GameView.vue
```diff
@@ -645,20 +648,13 @@ export default {
...
       this.targetingMoveDisplayName = null;
       this.clearOverlays();
     },
-    selectCard(index) {
-      if (index === this.selectionIndex) {
-        this.clearSelection();
-      } else {
-        this.selectionIndex = index;
-      }
-    },
```
seriouslysean, line 661: Changing this has unintended side effects. It breaks card targeting as those work by index (see `targetOpponentPointCard` and `targetOpponentFaceCard`). There's also another unintended side effect where the card overlay pops right before trying to target a card and I'm certain it's due to the order of operations.

I likely need to reconfigure the logic in the card overlay to know when it is and is not valid for showing or hiding now that we're saying it can show anytime. There are certain states in the game we'll need to track more transparently. cc @itsalaidbacklife

itsalaidbacklife:
Closing to separate into #103 and #87

PR #99, opened 2022-09-20
Add frozen disabled text to MoveChoiceCard and disable MoveChoiceCard…

TommySchmid:
… when frozen. Tests in oneOff.spec and sevens.spec have been updated

## Please check the following

- For New Features:

### DIFF
--- a/client/js/components/GameView/MoveChoiceOverlay.vue
+++ b/client/js/components/GameView/MoveChoiceOverlay.vue
@@ -86,14 +86,24 @@ export default {
         cardRank = this.selectedCard.rank;
       }
       // Re-usable move-objects
+      const allMovesAreDisabled = !this.isPlayersTurn || this.frozenId === this.selectedCard.id;
+      const allMovesDisabledExplanation = !this.isPlayersTurn
+        ? "It's not your turn"
+        : 'This card is frozen';
+
+      let disabledText;
+      if (allMovesAreDisabled) {
+        disabledText = allMovesDisabledExplanation;
+      }
+
       // Points
       const pointsDescription = `Gain ${cardRank} point` + (cardRank === 1 ? '' : 's');
       const pointsMove = {
         displayName: 'Points',
         eventName: 'points',
         moveDescription: pointsDescription,
-        disabled: !this.isPlayersTurn,
-        disabledExplanation: "It's not your turn",
+        disabled: allMovesAreDisabled,
+        disabledExplanation: disabledText,
       };
       // Scuttling
       const scuttleDisabled = !this.isPlayersTurn || !this.hasValidScuttleTarget;
@@ -101,16 +111,17 @@ export default {
       if (this.$store.getters.opponent.points.length === 0) {
         scuttleDisabledExplanation = 'Your opponent has no point cards to scuttle';
       }
-      if (!this.isPlayersTurn) {
-        scuttleDisabledExplanation = "It's not your turn";
+      if (allMovesAreDisabled) {
+        scuttleDisabledExplanation = allMovesDisabledExplanation;
       }
       const scuttleMove = {
         displayName: 'Scuttle',
         eventName: 'scuttle',
         moveDescription: 'Scrap a lower point card',
-        disabled: scuttleDisabled,
+        disabled: scuttleDisabled || allMovesAreDisabled,
         disabledExplanation: scuttleDisabledExplanation,
       };
+
       switch (cardRank) {
         case 1:
         case 3:
@@ -126,8 +137,8 @@ export default {
               displayName: 'One-Off',
               eventName: 'oneOff',
               moveDescription: this.selectedCard.ruleText,
-              disabled: !this.isPlayersTurn,
-              disabledExplanation: "It's not your turn",
+              disabled: allMovesAreDisabled,
+              disabledExplanation: disabledText,
             },
           ];
           break;
@@ -152,7 +163,9 @@ export default {
                 );
                 const numTotalTargets = numOpFaceCards + numOpJacks;
                 validTargetExists = numTotalTargets >= 1;
-                if (!validTargetExists) {
+                if (allMovesAreDisabled) {
+                  oneOffDisabledExplanation = allMovesDisabledExplanation;
+                } else if (!validTargetExists) {
                   oneOffDisabled = true;
                   oneOffDisabledExplanation = 'There are no Royals to target';
                 }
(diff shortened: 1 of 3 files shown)


itsalaidbacklife (changes requested):
It looks like the `disabledExplanation` without setting the `disabled` state itself. Let's try an additional  test about a number card to confirm that its move choices are also disabled

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -86,23 +86,33 @@ export default {
...
+        disabledText = allMovesDisabledExplanation;
+      }
+
       // Points
       const pointsDescription = `Gain ${cardRank} point` + (cardRank === 1 ? '' : 's');
       const pointsMove = {
         displayName: 'Points',
         eventName: 'points',
         moveDescription: pointsDescription,
         disabled: !this.isPlayersTurn,
```
seriouslysean, line 105: Seems like you could do this
```suggestion
        disabled: !this.isPlayersTurn || allMovesAreDisabled,
```

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -86,23 +86,33 @@ export default {
         cardRank = this.selectedCard.rank;
       }
       // Re-usable move-objects
+      const allMovesAreDisabled = !this.isPlayersTurn || this.frozenId === this.selectedCard.id;
```
seriouslysean, line 89: Longer term, I think this whole function needs to be refactored. It's a bit of a mess and it's making it hard to understand what is actually happening throughout the logic chains which I'm sure isn't helping your understanding of it all.

seriouslysean (changes requested):
Few more changes before this is ready to go.

client/js/components/GameView/MoveChoiceOverlay.vue
TommySchmid: It makes sense to me now, but im sure coming back to this or someone coming into it would certainly be a little confused :/

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -155,6 +166,8 @@ export default {
                 if (!validTargetExists) {
                   oneOffDisabled = true;
                   oneOffDisabledExplanation = 'There are no Royals to target';
+                } else if (allMovesAreDisabled) {
+                  oneOffDisabledExplanation = allMovesDisabledExplanation;
```
itsalaidbacklife, line 170: Interesting UX question. If all moves are disabled and also this specific move is disabled (like it's not your turn but even if it were your turn there's no valid target for your one-off), what message should we show?

I had imagined we would show the global one e.g. "It's not your turn", but this would prioritize the specific one "there's no valid target" instead. What do you think would be preferable?

tests/e2e/specs/in-game/sevens.spec.js
```diff
@@ -1241,7 +1241,7 @@ describe('Opponent playing SEVENS', () => {
       });
     }); // End Opponent NINE from seven
 
-    it('Opponent plays NINE on jacks from seven (second card)', () => {
+    it.only('Opponent plays NINE on jacks from seven (second card)', () => {
```
itsalaidbacklife, line 1244:
```suggestion
    it('Opponent plays NINE on jacks from seven (second card)', () => {
```

itsalaidbacklife (approved):
Looking good. Nice work on the new test case. 

There's one .only() that should be removed and I have a question about the priority of disabled explanations when multiple reasons apply simultaneously. With those resolved, we'll be good to merge

client/js/components/GameView/MoveChoiceOverlay.vue
itsalaidbacklife: If we opt to switch the priority, which I'm leaning towards, all you'll need to do is swap the if and else if statements

seriouslysean: I also suggest swapping them fwiw.

TommySchmid: Yeah I think "It's not your turn" should have the highest priority. When it's not a players turn, that's all that really matters imo.

itsalaidbacklife: Sounds good. I think with this and the .only() fix we're good to go

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -161,7 +174,9 @@ export default {
                 const numValidTargets =
                   this.$store.getters.opponent.points.length +
                   this.$store.getters.opponent.faceCards.length;
-                if (numValidTargets === 0) {
+                if (allMovesAreDisabled) {
+                  oneOffDisabledExplanation = allMovesDisabledExplanation;
+                } else if (numValidTargets === 0) {
```
TommySchmid, line 179: Swapped priority of disabled text to favor "It's not your turn" here as well

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -86,31 +86,42 @@ export default {
...
       const scuttleDisabled = !this.isPlayersTurn || !this.hasValidScuttleTarget;
       let scuttleDisabledExplanation = 'You can only scuttle smaller point cards';
       if (this.$store.getters.opponent.points.length === 0) {
         scuttleDisabledExplanation = 'Your opponent has no point cards to scuttle';
       }
-      if (!this.isPlayersTurn) {
-        scuttleDisabledExplanation = "It's not your turn";
+      if (allMovesAreDisabled) {
+        scuttleDisabledExplanation = allMovesDisabledExplanation;
       }
```
TommySchmid, line 116: I feel like we could clean this up a bit too to prioritize the "It's not your turn"/"Card is frozen" text for scuttle moves. Im thinking:
`if (allMovesAreDisabled) {
        scuttleDisabledExplanation = allMovesDisabledExplanation;
      } else if (this.$store.getters.opponent.points.length === 0) {
        scuttleDisabledExplanation = 'Your opponent has no point cards to scuttle';
      }`

itsalaidbacklife: I'm open to it. I'd like to open a new issue to refactor this function so every move is it's own computed property and the `moves()` computed prop just combines them all. I think that'd be cleaner. Thoughts @TommySchmid and @seriouslysean? If so then we could handle that change in the file-wide rework

TommySchmid: Works for me!

PR #100, opened 2022-09-22
Feat/automated rankings

itsalaidbacklife:
**DO NOT MERGE WITHOUT DATABASE MIGRATION**

## Please check the following

- For New Features:

Add tests  to check happy flow for match data creation and updates as ranked games are created, and updates homepage tests to accommodate ranked flows.

## Enable create ranked games which track match data automatically
Create-Ranked-Game


api/models/Match.js
```diff
@@ -25,7 +25,11 @@ module.exports = {
     },
     endTime: {
       type: 'number',
-      required: true,
+      required: false,
+    },
+    games: {
+      collection: 'Game',
+      via: 'match',
```
itsalaidbacklife, line 32: This lets us query for all the games associated with a specific match, which is used when adding a game to a match to determine if the match should now be over

api/services/gameService.js
```diff
@@ -55,6 +55,8 @@ function tempGame(game, p0, p1) {
   this.passes = game.passes;
   this.resolving = game.resolving;
   this.lastEvent = game.lastEvent;
+  this.result = game.result;
+  this.ranked = game.ranked;
```
itsalaidbacklife, line 59: Adding these to the return value of `.populateGame()` so they can be read off it when we pass a populated game as an arg to the new helper

client/js/store/modules/gameList.js
```diff
@@ -7,6 +7,7 @@ class GameSummary {
     this.name = obj.name ? obj.name : null;
     this.numPlayers = Object.prototype.hasOwnProperty.call(obj, 'players') ? obj.players.length : 0;
     this.status = Object.prototype.hasOwnProperty.call(obj, 'status') ? obj.status : false;
+    this.ranked = Object.prototype.hasOwnProperty.call(obj, 'ranked') ? obj.ranked : false;
```
itsalaidbacklife, line 10: This isn't used yet but adding this to the store will let us display games differently in the list if they are ranked

client/js/views/Home.vue
```diff
@@ -8,8 +8,8 @@
               <v-col cols="4">
                 <h1 id="home-card-title">Games</h1>
               </v-col>
-              <v-col cols="8">
-                <v-row id="add-new-game">
+              <!-- <v-col cols="8"> -->
+              <!-- <v-row id="add-new-game">
```
itsalaidbacklife, line 12: Layout is WIP

client/js/views/Home.vue
```diff
@@ -85,6 +85,61 @@
...
+                  :loading="creatingGame"
+                  color="primary"
+                  depressed
+                  @click="submitNewGame"
+                >
+                  Create Game
+                </v-btn>
+              </v-card-actions>
+            </v-card>
+          </v-dialog>
```
itsalaidbacklife, line 141: Planning to pull this into a component at the end

config/env/development.js
```diff
@@ -18,4 +18,7 @@ module.exports = {
   // models: {
   //   connection: 'someMongodbServer'
   // }
+  blueprints: {
+    rest: true,
+  },
```
itsalaidbacklife, line 23: Enabling blueprints in development so we can test more easily. This lets us request localhost:1337/match to get all data for all matches, which I'm using in testing. Saves us the trouble of implementing dev-only endpoints and generally makes it easier to see into the database while mocking

tests/e2e/specs/in-game/ranked-matches.spec.js
```diff
@@ -0,0 +1,82 @@
+const dayjs = require('dayjs');
+import { setupGameAsP0, setupGameAsP1, Card } from '../../support/helpers';
+import { seasonFixtures } from '../../fixtures/statsFixtures';
+import { playerOne, playerTwo } from '../../fixtures/userFixtures';
+
+describe('Creating And Updating Ranked Matches', () => {
```
itsalaidbacklife, line 6: I plan to move these tests into the gameOver.spec, which I think best captures the context of these tests anyway

tests/e2e/specs/in-game/ranked-matches.spec.js
```diff
@@ -0,0 +1,82 @@
...
+    // Log in as playerOne
+    cy.loginPlayer(playerOne.username, playerOne.password);
+    setupGameAsP0(true, true);
+    // Set up season
+    const [_clubsSeason, diamondsSeason] = seasonFixtures;
+    diamondsSeason.startTime = dayjs().subtract(2, 'week').subtract(1, 'day').valueOf();
+    diamondsSeason.endTime = dayjs().add(11, 'weeks').valueOf();
+    cy.loadSeasonFixture([diamondsSeason]);
+  });
+  it.only('Creates a match when two players play a ranked game for the first time this week', function () {
```
itsalaidbacklife, line 22: Initially I planned on making many tests for this but now I'm thinking one happy flow test (this) plus a couple edge cases like skipping unranked games or completed matches

tests/e2e/specs/in-game/ranked-matches.spec.js
```diff
@@ -0,0 +1,82 @@
...
+      expect(match.player2.id).to.eq(this.playerTwoId);
+      expect(match.startTime).to.be.greaterThan(0);
+      expect(match.endTime).to.eq(0);
+      expect(match.games.length).to.eq(2);
+      expect(match.games[0].result).to.eq(0); // P0 should have won the first game
+      cy.log('Match data is correct after second game', res);
+    });
+  });
+
+  it('Adds games to the existing match for this week between players and sets winner appropriately', () => {});
```
itsalaidbacklife, line 78:
```suggestion
```

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -153,7 +153,7 @@ describe('Home - Game List', () => {
 
 describe('Home - Create Game', () => {
   beforeEach(setup);
-  it('Creates a new game by hitting enter in text field', () => {
+  it.skip('Creates a new game by hitting enter in text field', () => {
```
itsalaidbacklife, line 156: Will probably remove this test

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -169,9 +169,16 @@ describe('Home - Create Game', () => {
       });
   });
 
-  it('Creates a new game by hitting the submit button', () => {
-    cy.get('[data-cy=create-game-input]').type('test game');
+  it.only('Creates a new unranked game by hitting the submit button', () => {
```
itsalaidbacklife, line 172:
```suggestion
  it('Creates a new unranked game by hitting the submit button', () => {
```

tests/e2e/support/helpers.js
```diff
@@ -7,13 +7,13 @@ export const opponentPassword = 'deviousTrickery';
  * Signs up two players, navigates home, creates game, subscribes, ready's up
  * @param {boolean} alreadyAuthenticated: skips setup steps: db wipe, signup, navigate /
  */
-export function setupGameAsP0(alreadyAuthenticated = false) {
+export function setupGameAsP0(alreadyAuthenticated = false, ranked = false) {
```
itsalaidbacklife, line 10: Updated these to allow optionally setting up ranked games

itsalaidbacklife (commented):
WIP. Need to fix a couple bugs, clean up a few areas, and update the home page to show which games are ranked

api/controllers/game/concede.js
```diff
@@ -1,30 +1,28 @@
...
-          change: 'concede',
-          game,
-          victory,
-        },
-      });
-      return res.ok();
-    })
-    .catch(function failed(err) {
-      return res.badRequest(err);
+module.exports = async function (req, res) {
```
seriouslysean, line 1: Good switch to async/await 👍🏼

api/controllers/game/concede.js
```diff
@@ -1,30 +1,28 @@
...
+  try {
+    const winner = (req.session.pNum + 1) % 2;
```
seriouslysean, line 3: I wonder if we need a helper to manage pNum eventually. There's a lot of manual math to get the pnum and forgetting to add one and/or modulus it seems like an area that may be error prone.

api/controllers/game/create.js
```diff
@@ -1,14 +1,16 @@
 var gameAPI = sails.hooks['customgamehook'];
 
 module.exports = function (req, res) {
```
seriouslysean, line 3: I feel like this should be updated to async await also, but it's not a blocker by any means.

api/controllers/game/create.js
```diff
@@ -1,14 +1,16 @@
...
+  const { gameName, ranked } = req.body;
```
seriouslysean, line 4: should this have a logical default? 
```suggestion
  const { gameName, ranked = false } = req.body;
```

api/controllers/game/create.js
```diff
@@ -1,14 +1,16 @@
...
   if (req.body.gameName) {
     gameAPI
-      .createGame(req.body.gameName)
+      .createGame(gameName, ranked)
```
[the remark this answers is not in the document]
seriouslysean: Same here, should createGame default ranked if it's not passed?
```js
createGame(gameName, ranked = false) {}
```

api/helpers/add-game-to-match.js
```diff
@@ -0,0 +1,68 @@
+module.exports = {
```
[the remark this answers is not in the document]
seriouslysean: A future update might be nice to have the lobby alert you if you've played a ranked game already and downgrade it to a non ranked game (alerting the users so they can choose to play or exit the lobby).

api/helpers/add-game-to-match.js
```diff
@@ -0,0 +1,68 @@
...
+      required: true,
+    },
+  },
+  fn: async ({ game }, exits) => {
+    try {
+      const [player1, player2] = game.players;
+      let relevantMatch = await sails.helpers.findOrCreateCurrentMatch(player1.id, player2.id);
+      if (!relevantMatch) {
+        return exits.error('Could not add game to match');
+      }
```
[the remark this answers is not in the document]
seriouslysean: This this always guaranteed to have both players? That could be part of the reason for the error. A snack bar sounds fine and goes along with my above suggestion. I think I need to know more about matches. A game is technically a match or are we saying a specific player<>player scenario is a match so we can better track it?

api/helpers/add-game-to-match.js
```diff
@@ -0,0 +1,68 @@
...
+      if (relevantMatch.endTime || relevantMatch.winner) {
+        return exits.success(null);
```
seriouslysean, line 23: Why pass null?

api/helpers/add-game-to-match.js
```diff
@@ -0,0 +1,68 @@
...
+      }
```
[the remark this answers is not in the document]
seriouslysean: Does this have any sort of UX implication or way to notify the players it isn't going to count?

api/helpers/add-game-to-match.js
```diff
@@ -0,0 +1,68 @@
...
+        if (priorGame.result === 0) {
+          numPlayer1Wins++;
+        } else if (priorGame.result === 1) {
+          numPlayer2Wins++;
+        }
+      }
+      // Add game to match
+      await Match.addToCollection(relevantMatch.id, 'games').members([game.id]);
+      // End match if this game clinches it
+      switch (game.result) {
```
seriouslysean, line 40: This switch feels weird to me and I'm having trouble following the logic. You're switching between the game result, but what is result usually. It looks like `0` or `1` but I don't know what those are mapping to. The number of wins also seems kinda weird since you could hit a case and not run the resulting code in the switch block?

Is there a better way to format this for easier readability?

api/helpers/find-or-create-current-match.js
```diff
@@ -0,0 +1,70 @@
...
+    },
+    player2: {
+      type: 'number',
+      descritiption: 'The database id of the second player in the desired match',
+      example: 45,
+      required: true,
+    },
+  },
+
+  fn: async ({ player1, player2 }, exits) => {
```
seriouslysean, line 24: This helper answers one of my above questions, haha. Should it fast fail if both players don't exist?

api/helpers/find-or-create-current-match.js
```diff
@@ -0,0 +1,70 @@
...
+      });
+      // FIXME: Handle missing season gracefully
+      if (!currentSeason) {
+        return exits.success(null);
+      }
+      // Find relevant match between specified players for current week
+      const seasonStartTime = dayjs(currentSeason.startTime);
+      const weeksSinceSeasonStart = currentTime.diff(seasonStartTime, 'week');
+      const currentWeekStartTime = seasonStartTime.add(weeksSinceSeasonStart, 'week').valueOf();
+      const currentWeekEndTime = seasonStartTime.add(weeksSinceSeasonStart + 1, 'week').valueOf();
```
seriouslysean, line 39: This feels like maybe it should be a utility that can be tested functionally.

api/hooks/customGameHook/index.js
```diff
@@ -3,11 +3,13 @@ module.exports = function gameHook() {
   // Game API //
   //////////////
   return {
-    createGame: function (gameName) {
+    createGame: function (gameName, isRanked = false) {
```
seriouslysean, line 6: I asked about this earlier so it's good to see it has a logical default for `isRanked`.

api/hooks/customGameHook/index.js
```diff
@@ -3,11 +3,13 @@ module.exports = function gameHook() {
...
       return new Promise(function (resolve, reject) {
+        const ranked = isRanked ?? false;
```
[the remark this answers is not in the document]
seriouslysean: Yeah, haha. I'd agree you could use it directly.

api/hooks/customGameHook/index.js
```diff
@@ -3,11 +3,13 @@ module.exports = function gameHook() {
...
         Game.create({
           name: gameName,
           status: true,
+          ranked,
```
seriouslysean, line 12: That said, you may want to use it directly and coerce it to be a boolean though.
```suggestion
          ranked: ranked === true,
```

api/services/gameService.js
seriouslysean: I wonder if it's worth refactoring this to assign them all in a loop instead of one by one?
```
Object.entries(game).forEach(([key, val]) => {
    if (key === 'players') {
        this[key] = [p0, p1];
        return;
    }
    this[key] = val;
});
```

api/services/gameService.js
```diff
@@ -175,6 +177,9 @@ module.exports = {
         gameUpdates.result = GameResult.P1_WINS;
       }
       await Game.updateOne({ id: game.id }).set(gameUpdates);
+      if (game.ranked) {
+        await sails.helpers.addGameToMatch(game);
+      }
```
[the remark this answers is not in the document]
seriouslysean: Fast fail makes more sense here imo:
```suggestion
      if (!game.ranked) {
          return res;
      }
      await sails.helpers.addGameToMatch(game);
```

seriouslysean (commented):
First half of my review, looking good though 👍🏼

api/controllers/game/create.js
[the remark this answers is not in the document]
itsalaidbacklife: Yeah I think that's reasonable

api/helpers/add-game-to-match.js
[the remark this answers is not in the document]
itsalaidbacklife: Agreed

api/controllers/game/concede.js
itsalaidbacklife: Open to it. Not sure what it'd look like. A helper to return your opponent's pNum maybe?

api/controllers/game/create.js
itsalaidbacklife: Yeah I agree. Trying to minimize current changes but we could update the entire backend to switch I'd imagine

api/controllers/game/create.js
itsalaidbacklife: Yeah I think that's reasonable. might let us skip updating some places where I didn't want to include ranked anyway

api/helpers/add-game-to-match.js
itsalaidbacklife: No good reason--this is half-baked. This is the case where the match is already over and we shouldn't add the new game to the match so it returns early. Could maybe skip null and just `return exits.success()` or even make a specific exit for this case so it can be handled more particularly by the consumer of this method

api/helpers/add-game-to-match.js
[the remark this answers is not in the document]
itsalaidbacklife: I definitely want to make one, but I was thinking this could be a potential scope cut for the initial rollout. The idea being that in the beginning, we'll just tell people that ranked games now automatically track the scores and let people create ranked games and see the results of their matches in the stats page. Then we can update the GameOverDialog and some other areas to for example show you the state of your current match against your opponent and let you know the score and whether it was finished before the current game started

api/helpers/add-game-to-match.js
itsalaidbacklife: Agreed this part feels awkward and it's also literally broken/the next thing I need to fix, though I'm not currently sure how.

The game's `result` is the pNum of which player won (or -1 if the game was never completed, or 2 if the game ended in a stalemate).

So the games will have a p0 and p1 attribute which is the player id of p0 and p1, and then it will have a `result` which tells us which of those players won.

api/helpers/find-or-create-current-match.js
itsalaidbacklife: Like look up if users with those ids don't exist? An issue with that is that currently we don't really need to look up the users themselves, so it adds an extra database query for each player in order to fail fast, which feels not fast by comparison. I'm open to it, but we'd likely want to rework the query then to avoid the use of `.populate()` in order to prevent redundant queries

api/helpers/find-or-create-current-match.js
itsalaidbacklife: Yes I agree, but I think it needs to be done against a booted backend because it's query for data in a database (or in mocked-in-memory sails-disk), so it's not a pure function

client/js/store/modules/gameList.js
```diff
@@ -57,12 +58,13 @@ export default {
         });
       });
     },
-    requestCreateGame(context, newGameName) {
+    requestCreateGame(context, { gameName, ranked }) {
```
seriouslysean, line 61: Same thing here, does ranked need to be defaulted? If not, does it make more sense to fast fail if it isn't passed?

client/js/views/Home.vue
```diff
@@ -109,9 +164,11 @@ export default {
   data() {
     return {
       newGameName: '',
+      newGameIsRanked: false,
```
seriouslysean, line 167: What about making newGame an object with two keys, `name` and `ranked`?

tests/e2e/specs/in-game/ranked-matches.spec.js
seriouslysean: This is where I landed with a lot of the test cases. Generally one happy path makes the most sense to me, with a few known edge-cases sprinkled in to keep down complexity. Might be worth looking in to this more when we move to Cypress 10.

seriouslysean (commented):
Second review pass

api/helpers/add-game-to-match.js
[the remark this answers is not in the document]
itsalaidbacklife: Created #108

api/services/gameService.js
itsalaidbacklife: Yes I think so, but in the name of scope let's not now. Part of why it was set up this way was to allow deliberate omission of fields from the data sent to users in the socket payload, but that's more important for the user objects (e.g. scrubbing `encryptedPassword`) than for the game

tests/e2e/specs/out-of-game/home.spec.js
itsalaidbacklife: Opted to make this work and keep the test. Can submit new game form with enter key

api/helpers/add-game-to-match.js
itsalaidbacklife: Worth another look, but the core fix for a related bug did remove this switch

client/js/views/Home.vue
```diff
@@ -136,13 +112,22 @@ export default {
...
           this.creatingGame = false;
+          this.showCreateGameDialog = false;
         })
         .catch(this.handleError);
     },
+    cancelCreateGame() {
+      this.newGameName = '';
+      this.creatingGame = false;
+      this.showCreateGameDialog = false;
+    },
```
itsalaidbacklife, line 130: Todo: remove stale methods

client/js/views/Home.vue
```diff
@@ -85,6 +58,9 @@
             </v-btn>
           </v-col>
         </v-row>
+        <v-row class="d-flex justify-end mt-8">
+          <create-game-dialog @error="handleError" />
```
itsalaidbacklife, line 62: I went back and forth on whether to make this a "dumb" presentational component that used props and events to manage creating a game from the Home page.

I opted to make the component talk to the store and manage game creation itself to DRY it up. It seems simpler this way to me

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -388,3 +392,233 @@ describe('Stalemates', () => {
...
+      expect(currentMatch.startTime).to.be.greaterThan(0);
+      expect(currentMatch.games.length).to.eq(4);
+      expect(currentMatch.games[3].result).to.eq(2);
+      // Match is incomplete
+      expect(currentMatch.endTime).to.eq(0);
+      expect(currentMatch.winner).to.eq(null);
+      cy.log('Match data is correct after fourth game', res);
+    });
+
+    // 5th Game: player wins via points and wins match
```
itsalaidbacklife, line 554: Todo: play an unranked game and confirm it doesn't get associated with the match

itsalaidbacklife (commented):
Commenting my todos and one note

api/services/gameService.js
itsalaidbacklife: Created issue #111 to refactor `populateGame`

api/services/gameService.js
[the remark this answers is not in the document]
itsalaidbacklife: Interesting tradeoff I'd never really considered, but early return adds a line (because the same return statement is duplicated), but the late-return involves another level of indentation for the bulk of the function body. Low FIG score for me, but I'll take your suggestion and return early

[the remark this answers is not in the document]
itsalaidbacklife: Wait actually I take that back. Because there's only one line inside the conditional block either way, the first way is the same indentation for the same blocks of code, but is one line shorter. Why return early?

client/js/views/Home.vue
itsalaidbacklife: Why? Same issue could apply in the new `CreateGameDialog` -- not clear to me what the pros and cons are of one object vs separate params.

client/js/views/Home.vue
seriouslysean: I think connecting to the store makes the most sense. There's a time and a place for presentational components and logically I think this is easier to understand.

api/models/Game.js
```diff
@@ -103,6 +103,9 @@ module.exports = {
       type: 'json',
       defaultsTo: {},
     },
+    match: {
+      model: 'match',
+    },
```
itsalaidbacklife, line 108: This is the change that requires the migration. Need to alter the `Game` table to add a `match` column that's a foreign key to `Match(id)`

itsalaidbacklife (commented):
Commented my final thoughts. Imo this is ready to merge with the caveat that it needs a migration run simultaneously

api/controllers/game/create.js
seriouslysean: Yep, tech debt 👍🏼

api/helpers/add-game-to-match.js
```diff
@@ -0,0 +1,73 @@
...
+        relevantMatch = await Match.updateOne(relevantMatch.id).set({
+          endTime: dayjs().valueOf(),
+          winner: game.players[0].id,
+        });
+      } else if (numPlayer2Wins >= 2) {
+        relevantMatch = await Match.updateOne(relevantMatch).set({
+          endTime: dayjs().valueOf(),
+          winner: game.players[1].id,
+        });
+      }
```
seriouslysean, line 67: Since we always do one or the other, does it make sense to limit how much we're repeating?
```suggestion
    let winner;
    if (numPlayer1Wins >= 2) {
      winner = game.players[0].id;
    } else if (numPlayer2Wins >= 2) {
      winner = game.players[1].id;
    }

    relevantMatch = await Match.updateOne(relevantMatch).set({
      endTime: dayjs().valueOf(),
      winner,
    });
```

api/helpers/add-game-to-match.js
```diff
@@ -0,0 +1,73 @@
...
+      await Match.addToCollection(relevantMatch.id, 'games').members([game.id]);
+
+      // End match if this game clinches it
+      if (numPlayer1Wins >= 2) {
+        relevantMatch = await Match.updateOne(relevantMatch.id).set({
+          endTime: dayjs().valueOf(),
+          winner: game.players[0].id,
+        });
+      } else if (numPlayer2Wins >= 2) {
+        relevantMatch = await Match.updateOne(relevantMatch).set({
```
seriouslysean, line 63: Is this a typo or am I missing something?
```suggestion
        relevantMatch = await Match.updateOne(relevantMatch.id).set({
```
I think `.id` might need to be added here.

api/models/Game.js
itsalaidbacklife: I need to confirm, but I believe the necessary migration is 
```
alter table "Game"
add column match int4
```

seriouslysean (approved):

api/helpers/add-game-to-match.js
itsalaidbacklife: The funky bit is that we only need to update the `relevantMatch` if either player wins (in that case we set the `winner` and `endTime` fields), but if this is say the 2nd game and the players are tied 1-1, then we don't actually want to set an `endTime` or a winner. We could set it up so that we set variables (defaulting to `null`) for winner and endTime and then always update the match at the end (with possibly null values), but this does result in unnecessary queries for games where the match doesn't actually need an update b/c it's not over yet.

Thoughts?

api/helpers/add-game-to-match.js
itsalaidbacklife: 100%. Great catch. This clearly means the testing isn't as thorough as it should be as well. Fixing for now and I think it further indicates the value of setting up backend unit tests against a booted server

api/services/gameService.js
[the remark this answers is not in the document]
itsalaidbacklife: I think that's reasonable, but because the entire rest of the function is just closing braces and the same return statement, I'm gonna keep it this way atm

client/js/components/GameListItem.vue
```diff
@@ -73,6 +80,9 @@ export default {
     readyText() {
       return `${this.numPlayers} / 2`;
     },
+    buttonText() {
+      return this.isRanked ? 'Play Ranked' : 'Play';
+    },
```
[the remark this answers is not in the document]
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/113

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cuttle",
-  "version": "3.0.2",
+  "version": "3.1.0",
   "private": true,
```
seriouslysean, line 4: Finally! 😂

seriouslysean:
@itsalaidbacklife branched off of this PR and upgraded Cypress to version 10 to try and fix some of the intermittent errors we've been seeing. https://github.com/cuttle-cards/cuttle/pull/117

seriouslysean:
Just recapping our last chat over Discord. You're good to merge anytime, the intermittent failure appears to be some sort of race condition we can chase down separately.

PR #120, opened 2022-10-04
Create computed properties for each move type. Create computed proper…

TommySchmid:
…ties for one offs two and nine. Create computed properties for one off jacks.'

## Please check the following

- For New Features:

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -58,8 +58,8 @@ jobs:
     name: "Test: End to End"
     runs-on: ubuntu-latest
     needs: test-unit
-    # stop the job if it runs over 20 minutes to prevent a hanging process from eating minutes
-    timeout-minutes: 20
+    # stop the job if it runs over 60 minutes to prevent a hanging process from eating minutes
+    timeout-minutes: 60
     steps:
       # Checkout and node setup are shared in both `pr-actions.yml` and `run-npm-script.yml`
       # If you change one, please change the other
--- a/client/js/components/GameView/GameOverlays.vue
+++ b/client/js/components/GameView/GameOverlays.vue
@@ -36,6 +36,7 @@
       <h1>Opponent Considering Stalemate Request</h1>
     </v-overlay>
     <move-choice-overlay
+      v-if="selectedCard || cardSelectedFromDeck"
       :value="!targeting && (!!selectedCard || !!cardSelectedFromDeck)"
       :selected-card="selectedCard || cardSelectedFromDeck"
       :is-players-turn="isPlayersTurn"
--- a/client/js/components/GameView/MoveChoiceOverlay.vue
+++ b/client/js/components/GameView/MoveChoiceOverlay.vue
@@ -54,7 +54,7 @@ export default {
     },
     selectedCard: {
       type: Object,
-      default: null,
+      required: true,
     },
     isPlayersTurn: {
       type: Boolean,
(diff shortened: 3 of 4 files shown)


client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -70,166 +70,175 @@ export default {
...
             // Glasses
             {
               displayName: 'Glasses',
               eventName: 'faceCard',
               moveDescription: 'Your opponent plays open handed',
-              disabled: allMovesAreDisabled,
-              disabledExplanation: disabledText,
+              disabled: this.allMovesAreDisabled,
+              disabledExplanation: this.disabledText,
             },
```
TommySchmid, line 234: I left the G8 one off object as is (same with queens and kings) since there is no logic being executed. I figured it might be cleaner to leave these in line since they're small, as opposed to creating two more computed properties.

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -238,8 +247,8 @@ export default {
               displayName: 'Royal',
               eventName: 'faceCard',
               moveDescription: this.selectedCard.ruleText,
-              disabled: allMovesAreDisabled,
-              disabledExplanation: disabledText,
+              disabled: this.allMovesAreDisabled,
+              disabledExplanation: this.disabledText,
             },
```
TommySchmid, line 252: Same question as G8 oneOff.

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -70,166 +70,175 @@ export default {
...
+        disabledExplanation,
+      };
+      return oneOffJack;
+    },
+
+    /**
+     * Returns list of objects representing the available moves,
+     * based on the selected card
+     */
+    moveChoices() {
```
itsalaidbacklife, line 206: Let's remove the `res` variable and the `break` statements and just `return` the array that were assigning to `res` in this function

itsalaidbacklife (changes requested):
Good stuff. One general code style point I'm requesting changes on is to avoid capturing variables just so we can return them. We can return the value directly instead. Other change reqs are in the comments. It's already an improvement and it's close to ready

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -70,181 +70,173 @@ export default {
...
+        return !this.isPlayersTurn ? "It's not your turn" : 'This card is frozen';
       }
-
-      // Points
-      const pointsDescription = `Gain ${cardRank} point` + (cardRank === 1 ? '' : 's');
-      const pointsMove = {
+      return '';
+    },
+    pointsMove() {
+      if (!this.selectedCard) return [];
```
itsalaidbacklife, line 85: Because this computed prop returns an object, the 'empty' case should be an empty object, or null, rather than an array.
```suggestion
      if (!this.selectedCard) return null;
```

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -70,181 +70,173 @@ export default {
...
         moveDescription: pointsDescription,
-        disabled: allMovesAreDisabled,
-        disabledExplanation: disabledText,
+        disabled: this.allMovesAreDisabled,
+        disabledExplanation: this.disabledText,
       };
-      // Scuttling
+    },
+    scuttleMove() {
+      if (!this.selectedCard) return [];
```
itsalaidbacklife, line 97: Same here
```suggestion
      if (!this.selectedCard) return null;
```

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -70,181 +70,173 @@ export default {
...
         displayName: 'Scuttle',
         eventName: 'scuttle',
         moveDescription: 'Scrap a lower point card',
-        disabled: scuttleDisabled || allMovesAreDisabled,
+        disabled: scuttleDisabled || this.allMovesAreDisabled,
         disabledExplanation: scuttleDisabledExplanation,
       };
+    },
+    oneOffMove() {
+      if (!this.selectedCard) return [];
```
itsalaidbacklife, line 115: And here
```suggestion
      if (!this.selectedCard) return null;
```

client/js/components/GameView/MoveChoiceOverlay.vue
TommySchmid: Should the `if (!this.selectedCard) return null;` check be added to each computed prop? I remember you saying the check should be required. I added it to a few as you can see, but im not sure I fully though through which ones to add/not add it to.

And, if we are adding to all of them, is there a better way we can do that instead of having it written in each one?

TommySchmid: I guess specifically, `allMovesAreDisabled()` and `disabledText()`

itsalaidbacklife: Good point. This is a matter of type-safety. Essentially any time we reference a property on an object, we should first be sure that the object exists. I wrote earlier that we could assume there was always a selected card, and here I'm kindof waffling on that point. 

We either need to handle that possibility in every computed prop that references `this.selectedCard`, or we need to make the `selectedCard` prop required, and then update the `GameOverlays` so that this component is `v-if`'d to only appear when a card is selected. I think the second approach is actually cleaner. We can completely assume inside this component that `this.selectedCard` is defined and is a card, and therefore remove _all_ these null checks. The only thing is that we would still need all the computed props to return a value in all circumstances, so we'd just need to ensure there are no functions that only return inside a conditional like

```
someComputedProp() {
  if (something) {
    return whatever;
  }
}
```

That should instead be
```
someComputedProp() {
  if (something) {
    return whatever;
  }
  return whateverMakesSense;
}
```

client/js/components/GameView/MoveChoiceOverlay.vue
itsalaidbacklife: per our conversation, let's actually make `selectedCard` a required prop and remove these null checks instead

client/js/components/GameView/GameOverlays.vue
```diff
@@ -36,6 +36,7 @@
       <h1>Opponent Considering Stalemate Request</h1>
     </v-overlay>
     <move-choice-overlay
+      v-if="(selectedCard || cardSelectedFromDeck)"
```
TommySchmid, line 39: sevens.spec didn't work at all before including `cardSelectedFromDeck`. Makes sense.

tests/e2e/specs/in-game/basicMoves.spec.js
```diff
@@ -373,7 +373,7 @@ describe('Game Basic Moves - P0 Perspective', () => {
     // Should have 3 move options
     cy.get('[data-move-choice]').should('have.length', 3);
     cy.get('[data-cy=cancel-move]').click();
-    cy.get('#move-choice-overlay').should('not.be.visible');
+    cy.get('#move-choice-overlay').should('not.exist');
```
TommySchmid, line 376: When the overlay was closed, `#move-choice-overlay` was still expected to be found, just not visible, but with the requirement change, it's not present at all. 

This was the only instance of `cy.get('#move-choice-overlay').should('not.be.visible');` in the whole testing suite.

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -70,181 +70,167 @@ export default {
...
         moveDescription: pointsDescription,
-        disabled: allMovesAreDisabled,
-        disabledExplanation: disabledText,
+        disabled: this.allMovesAreDisabled,
+        disabledExplanation: this.disabledText,
       };
-      // Scuttling
+    },
+    scuttleMove() {
       const scuttleDisabled = !this.isPlayersTurn || !this.hasValidScuttleTarget;
```
itsalaidbacklife, line 96: Let's clean this up while we're in here. 
```suggestion
      const scuttleDisabled = !this.allMovesAreDisabled || !this.hasValidScuttleTarget;
```

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -70,181 +70,167 @@ export default {
...
+      if (this.allMovesAreDisabled) {
+        scuttleDisabledExplanation = this.disabledText;
       }
-      const scuttleMove = {
+      return {
         displayName: 'Scuttle',
         eventName: 'scuttle',
         moveDescription: 'Scrap a lower point card',
-        disabled: scuttleDisabled || allMovesAreDisabled,
+        disabled: scuttleDisabled || this.allMovesAreDisabled,
```
itsalaidbacklife, line 108: With the above change this line can just use the existing variable without ||
```suggestion
        disabled: scuttleDisabled,
```

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -70,181 +70,167 @@ export default {
...
+          }
+        }
+      }
+      return {
+        displayName: 'One-Off',
+        eventName: 'targetedOneOff',
+        moveDescription: this.selectedCard.ruleText,
+        disabled: oneOffDisabled || this.allMovesAreDisabled,
+        disabledExplanation: oneOffDisabledExplanation,
+      };
```
itsalaidbacklife, line 168: Same idea as above. It's cleaner to have one variable for `oneOffDisabled` and for `oneOffDisabledExplanation` which get used in the return value down below, instead of having the variable and also comparing to the global stuff in the return statement. Basically it's funky in the current setup that we're referring to `this.isPlayersTurn` here but also implicitly down below when we use `this.allMovesAreDisabled` because that also reads from `this.isPlayersTurn` in the same way.

So instead, let's rejigger this to try to more efficiently leverage those existing variables and to reduce the number of times we need to check whether all moves are disabled. I'm writing this in the web editor in GH so I'm not 100% I got the format right, but see if you can rework this like so.
```suggestion
      let oneOffDisabled = this.allMovesAreDisabled;
      let oneOffDisabledExplanation = this.disabledText;
      if (!this.allMovesAreDisabled) {
        if (this.opponentQueenCount >= 2) {
          oneOffDisabled = true;
          oneOffDisabledExplanation = `You can't play a ${this.selectedCard.rank} while your opponent has two or more queens`;
        } else {
          let validTargetExists;
          // Twos
          if (this.selectedCard.rank === 2) {
            const numOpFaceCards = this.$store.getters.opponent.faceCards.length;
            const numOpJacks = this.$store.getters.opponent.points.reduce(
              (jackCount, pointCard) => {
                return jackCount + pointCard.attachments.length;
              },
              0
            );
            const numTotalTargets = numOpFaceCards + numOpJacks;
            validTargetExists = numTotalTargets >= 1;
            if (!validTargetExists) {
              oneOffDisabled = true;
              oneOffDisabledExplanation = 'There are no Royals to target';
            }
            // Nines
          } else {
            const numValidTargets =
              this.$store.getters.opponent.points.length +
              this.$store.getters.opponent.faceCards.length;
           if (numValidTargets === 0) {
              oneOffDisabled = true;
              oneOffDisabledExplanation = 'There are no point cards or Royals to target';
            }
          }
        }
      }
      return {
        displayName: 'One-Off',
        eventName: 'targetedOneOff',
        moveDescription: this.selectedCard.ruleText,
        disabled: oneOffDisabled || this.allMovesAreDisabled,
        disabledExplanation: oneOffDisabledExplanation,
      };
```

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -70,181 +70,167 @@ export default {
...
+    },
+    jackMove() {
+      let ableToJack = false;
+      let disabledExplanation = '';
+      if (this.isPlayersTurn && !this.allMovesAreDisabled) {
```
itsalaidbacklife, line 173: `allMovesAreDisabled` already checks if it's your turn so we can just
```suggestion
      if (!this.allMovesAreDisabled) {
```

itsalaidbacklife (approved):
Looks good. I'm asking for a bit more refactoring of a couple of the computed props to reduce the number of times we check the same expressions

.github/workflows/pr-actions.yml
```diff
@@ -58,8 +58,8 @@ jobs:
     name: "Test: End to End"
     runs-on: ubuntu-latest
     needs: test-unit
-    # stop the job if it runs over 20 minutes to prevent a hanging process from eating minutes
-    timeout-minutes: 20
+    # stop the job if it runs over 60 minutes to prevent a hanging process from eating minutes
+    timeout-minutes: 60
```
itsalaidbacklife, line 62: We updated this in main as well in a separate PR, so just merge main into your branch and take the 'main' version of this file instead

TommySchmid: Looks like there's a bit of discrepancy in there now - https://github.com/cuttle-cards/cuttle/blob/b0f096887de357b0f29ffc179eb49c7595f113f3/.github/workflows/pr-actions.yml#L61-L62

I can make any changes in this pr if needed.

.github/workflows/pr-actions.yml
```diff
@@ -60,6 +60,7 @@ jobs:
     needs: test-unit
     # stop the job if it runs over 20 minutes to prevent a hanging process from eating minutes
     timeout-minutes: 45
+ main
```
itsalaidbacklife, line 63: This is a type leftover from the merge conflict
```suggestion
```

TommySchmid: Weird, after merging main into my branch via github gui, I don't see those changes reflected in my ide. Should I?

Nice catch, sorry about that. I'll merge again through github gui.

client/js/components/GameView/MoveChoiceOverlay.vue
TommySchmid: This change is causing several tests to fail from basicMoves, oneOffs, sevens. Here's an example from basicMoves:

`    // Player scuttles 6 of diamonds with 7 of clubs
    cy.get('[data-player-hand-card=7-0]').click(); // 7 of clubs
    cy.get('[data-move-choice=scuttle]').click();
    cy.get('[data-opponent-point-card=6-2]').click(); // 6 of hearts
    assertGameState(0, {
      p0Hand: [Card.ACE_OF_CLUBS, Card.ACE_OF_SPADES],
      p0Points: [Card.TWO_OF_CLUBS, Card.TEN_OF_HEARTS],
      p0FaceCards: [Card.KING_OF_SPADES],
      p1Hand: [Card.TEN_OF_CLUBS, Card.TEN_OF_SPADES],
      p1Points: [Card.ACE_OF_DIAMONDS],
      p1FaceCards: [Card.QUEEN_OF_HEARTS],
      scrap: [Card.SEVEN_OF_CLUBS, Card.SIX_OF_HEARTS],
    });
    cy.log('Successfully scuttled as p0');`

The scuttle move choice on the 7 was disabled with the 'You can only scuttle smaller point cards' disabled text. The move should be legal of course. Im not sure what the logic here is.

TommySchmid: The scuttle move choice on the 7 was disabled with the 'You can only scuttle smaller point cards' disabled text. The move should be legal of course. Im not sure what the logic here is.

client/js/components/GameView/MoveChoiceOverlay.vue
TommySchmid: This change also causes a few test failures, one being `Disables playing a frozen number card until the following turn` from oneOffs. The frozen disabled text is missing from the scuttle move choice with this change in place. Reverting it causes the test to pass.

client/js/components/GameView/MoveChoiceOverlay.vue
itsalaidbacklife: Oops I included a ! That shouldn't be there

itsalaidbacklife: it should be
```suggestion
      const scuttleDisabled = this.allMovesAreDisabled;
```

client/js/components/GameView/MoveChoiceOverlay.vue
itsalaidbacklife: I think it's because of my typo above. See the revised suggestion

client/js/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -70,181 +70,163 @@ export default {
...
+              oneOffDisabledExplanation = 'There are no point cards or Royals to target';
+            }
+          }
+        }
+      }
+      return {
+        displayName: 'One-Off',
+        eventName: 'targetedOneOff',
+        moveDescription: this.selectedCard.ruleText,
+        disabled: oneOffDisabled || this.allMovesAreDisabled,
```
itsalaidbacklife, line 162: Because you assign `oneOffDisabled` in the beginning, and only re-assign it if you're setting it to true, you can just
```suggestion
        disabled: oneOffDisabled,
```

PR #122, opened 2022-10-06
[Feature]: Lobby should show whether a game is ranked, resolves #106

### LINKED ISSUE #106: [Feature]: Lobby should show whether a game is ranked
### Feature Summary

When sitting in a game lobby, I'd like to be able to see if the game I am in will be ranked

### Detailed Description

The header with the game name in the Lobby could be updated to show whether the current game is ranked so players can tell if the current game will be counted

seriouslysean:
## Please check the following

- For New Features:

image

### DIFF
--- a/client/js/store/modules/game.js
+++ b/client/js/store/modules/game.js
@@ -22,6 +22,7 @@ function resetState() {
     oneOffTarget: null,
     waitingForOpponentToCounter: false,
     myTurnToCounter: false,
+    isRanked: false,
     // Threes
     waitingForOpponentToPickFromScrap: false,
     pickingFromScrap: false,
@@ -131,9 +132,8 @@ export default {
     hasGlassesEight(state, getters) {
       return getters.player.faceCards.filter((card) => card.rank === 8).length > 0;
     },
-    isRanked(state, getters, rootState) {
-      const game = rootState['gameList'].games.find(({ id }) => id === state.id);
-      return game ? game.isRanked : false;
+    isRanked(state) {
+      return state.isRanked;
     },
   },
   mutations: {
@@ -187,6 +187,8 @@ export default {
       if (Object.hasOwnProperty.call(newGame, 'oneOffTarget'))
         state.oneOffTarget = _.cloneDeep(newGame.oneOffTarget);
       else state.oneOffTarget = null;
+
+      if (Object.hasOwnProperty.call(newGame, 'isRanked')) state.isRanked = newGame.isRanked;
     },
     setMyPNum(state, val) {
       state.myPNum = val;
--- a/client/js/store/modules/game.js
+++ b/client/js/store/modules/game.js
@@ -131,6 +131,10 @@ export default {
     hasGlassesEight(state, getters) {
       return getters.player.faceCards.filter((card) => card.rank === 8).length > 0;
     },
+    isRanked(state, getters, rootState) {
+      const game = rootState['gameList'].games.find(({ id }) => id === state.id);
+      return game ? game.isRanked : false;
+    },
   },
   mutations: {
     setGameId(state, val) {
--- a/client/js/store/modules/game.js
+++ b/client/js/store/modules/game.js
@@ -132,9 +132,6 @@ export default {
     hasGlassesEight(state, getters) {
       return getters.player.faceCards.filter((card) => card.rank === 8).length > 0;
     },
-    isRanked(state) {
-      return state.isRanked;
-    },
   },
   mutations: {
     setGameId(state, val) {
--- a/client/js/views/Lobby.vue
+++ b/client/js/views/Lobby.vue
@@ -5,7 +5,12 @@
         <img id="logo" alt="Cuttle logo" src="../img/logo.png" />
       </v-col>
       <v-col md="8" class="my-auto">
-        <h1>Lobby for {{ gameName }}</h1>
+        <h1>
+          Lobby for {{ gameName }}
+          <small v-if="isRanked" class="lobby-ranked-text">
+            (Ranked <v-icon v-if="isRanked" class="ml-1" medium>mdi-trophy</v-icon>)
+          </small>
+        </h1>
       </v-col>
     </v-row>
     <!-- Usernames -->
@@ -36,6 +41,7 @@
       <v-col cols="3">
         <v-btn :loading="readying" contained color="primary" data-cy="ready-button" @click="ready">
           {{ readyButtonText }}
+          <v-icon v-if="isRanked" class="ml-1" small data-cy="ready-button-ranked-icon">mdi-trophy</v-icon>
         </v-btn>
       </v-col>
       <v-spacer />
@@ -43,8 +49,9 @@
   </v-container>
 </template>
 <script>
-import LobbyPlayerIndicator from '../components/LobbyPlayerIndicator';
 import { mapGetters } from 'vuex';
+
+import LobbyPlayerIndicator from '../components/LobbyPlayerIndicator';
 
 export default {
   name: 'Lobby',
@@ -57,7 +64,7 @@ export default {
     };
   },
   computed: {
-    ...mapGetters(['opponentIsReady', 'opponentUsername']),
+    ...mapGetters(['opponentIsReady', 'opponentUsername', 'isRanked']),
     gameId() {
       return this.$store.state.game.id;
     },
@@ -101,5 +108,8 @@ export default {
   height: 10vh;
   min-height: 64px;
   margin: 0 auto;
+}
+.lobby-ranked-text {
+  color: var(--v-neutral-darken2);
 }
 </style>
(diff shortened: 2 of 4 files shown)


client/js/store/modules/game.js
```diff
@@ -131,6 +131,10 @@ export default {
     hasGlassesEight(state, getters) {
       return getters.player.faceCards.filter((card) => card.rank === 8).length > 0;
     },
+    isRanked(state, getters, rootState) {
+      const game = rootState['gameList'].games.find(({ id }) => id === state.id);
+      return game ? game.isRanked : false;
+    },
```
seriouslysean, line 137: Now a getter 😎

client/js/views/Lobby.vue
```diff
@@ -102,4 +111,7 @@ export default {
   min-height: 64px;
   margin: 0 auto;
 }
+.lobby-ranked-text {
+  color: var(--v-neutral-darken2);
+}
```
seriouslysean, line 116: Just chose a standard grey.

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -5,11 +5,14 @@ import {
...
 
-function setup() {
+function setup(isRanked = false) {
   cy.wipeDatabase();
   cy.visit('/');
   cy.signupPlayer(validUsername, validPassword);
-  cy.createGamePlayer({ gameName: 'Test Game', isRanked: false }).then((gameSummary) => {
+  // This is required to show the game list (which will ensure the `addGameToList` action dispatches)
+  // Otherwise, tanked game creation does not register properly
+  cy.visit('/');
```
seriouslysean, line 14: Really weird edgecase bug that was preventing me from making a ranked game for testing as `addGametoList` wasn't firing since the "user" wasn't on the proper gamelist page. You have to revisit it to force the reflow, probably from us using the store directly.

client/js/store/modules/game.js
itsalaidbacklife: It makes more sense as state than a getter imo. Why do a linear lookup of the games and compare by id when this module already has the 'full' data for a single game (which could include it's ranked state)

tests/e2e/specs/out-of-game/lobby.spec.js
itsalaidbacklife: This smells. The reason that we 'need' the current game to appear in the `gameList` store module is because the `isRanked` getter is reading the ranked-ness from across store modules, right? Instead, it would be cleaner to have `isRanked` be a piece of state in _this_ module, describing whether the current game is ranked, and then we wouldn't need our game to appear in the `gameList` in order to correctly register it as ranked (and we could skip the extra `cy.visit('/');`.

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -56,6 +59,21 @@ describe('Lobby - Page Content', () => {
   });
 });
 
+describe.only('Lobby - Page Content (Ranked)', () => {
```
itsalaidbacklife, line 62:
```suggestion
describe('Lobby - Page Content (Ranked)', () => {
```

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -56,6 +59,21 @@ describe('Lobby - Page Content', () => {
...
+  beforeEach(() => {
+    setup(true);
+  });
+
+  it('Displays ranked header', () => {
+    cy.contains('h1 small', 'Ranked');
+    cy.log(cy.window().its('cuttle.app.$store.state.gameList.games'));
```
itsalaidbacklife, line 69: Is this log useful? Wondering if it's stale from tinkering stages or we want it. I think it's funky to log the list of games rather than the current one, but it's my same point about moving the `isRanked` check into the `game` module

client/js/views/Lobby.vue
```diff
@@ -5,7 +5,12 @@
         <img id="logo" alt="Cuttle logo" src="../img/logo.png" />
       </v-col>
       <v-col md="8" class="my-auto">
-        <h1>Lobby for {{ gameName }}</h1>
+        <h1>
+          Lobby for {{ gameName }}
+          <small v-if="isRanked" class="lobby-ranked-text">
+            (Ranked <v-icon v-if="isRanked" class="ml-1" medium>mdi-trophy</v-icon>)
```
itsalaidbacklife, line 11: Not gonna die on this hill but I pref multiples of 8px (1 unit in vuetify's spacing classes = 4px)
```suggestion
            (Ranked <v-icon v-if="isRanked" class="ml-2" medium>mdi-trophy</v-icon>)
```

itsalaidbacklife (changes requested):
Nice. UX is right; requesting refactoring the store so a getter becomes state

client/js/store/modules/game.js
seriouslysean: I didn't see those anywhere in the state and I wasn't sure if this feature should be hoisting those up at this point. I do agree it would be better to have it in there. Any reason why we didn't add it in the first place since it's directly related to the game object?

client/js/views/Lobby.vue
seriouslysean: I thought it looked uglier with 8px, but I don't care enough to argue, hahaha.

tests/e2e/specs/out-of-game/lobby.spec.js
seriouslysean: This is definitely a symptom of the way I implemented the change and not the direct cause. Once the `isRanked` property is pulled in to the game object and referenced directly I don't think we'll have this issue and it can be removed.

tests/e2e/specs/out-of-game/lobby.spec.js
seriouslysean: Story of my life.

tests/e2e/specs/out-of-game/lobby.spec.js
seriouslysean: Nope, just missed it while I was debugging the issue to try and find where the isRanked property was from. This is why I like code review, this 'works' but is not ideal and the 2nd/3rd iterations will be much better for the effort 👍🏼

seriouslysean, line 69:
```suggestion
```

client/js/views/Lobby.vue
seriouslysean: I tried it again and I refuse, it looks like absolute shit with the larger space.
image

I ended up just taking it away altogether, haha.
image

client/js/store/modules/game.js
```diff
@@ -183,6 +187,8 @@ export default {
       if (Object.hasOwnProperty.call(newGame, 'oneOffTarget'))
         state.oneOffTarget = _.cloneDeep(newGame.oneOffTarget);
       else state.oneOffTarget = null;
+
+      if (Object.hasOwnProperty.call(newGame, 'isRanked')) state.isRanked = newGame.isRanked;
```
seriouslysean, line 191: Boy do I hate the way we do this a lot. I know why now and I don't blame you for it at all, but it's highly confusing the have the same function fire with both a full game object and a limited game object. I would love to align on a singular schema for the game object that comes through and make sure it has all the properties each time so we don't need to be this defensive.

itsalaidbacklife: Yes I think that's completely doable. Worth an issue

itsalaidbacklife (approved):
Looks good. The `isRanked` getter smells but I dimly remember some weird vuex limitation that may be forcing our hand here. Remind me why that's there and we're good to go

itsalaidbacklife (approved):
Looks good. The `isRanked` getter smells but I dimly remember some weird vuex limitation that may be forcing our hand here. Remind me why that's there and we're good to go

client/js/views/Lobby.vue
```diff
@@ -57,7 +66,7 @@ export default {
     };
   },
   computed: {
-    ...mapGetters(['opponentIsReady', 'opponentUsername']),
+    ...mapGetters(['opponentIsReady', 'opponentUsername', 'isRanked']),
```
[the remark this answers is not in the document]
seriouslysean: No, and maybe this is me being naive but I always tend to gravitate to getters for the caching. That said we’re not actually mutating the value or calculating anything so I can switch it back since it’s not really doing anything here.

client/js/views/Lobby.vue
seriouslysean: It might be worth asking Sarankit to take a pass on the lobby screen and just redesign it. The groundwork here is good but the design is limping along with the bare minimum :)

client/js/store/modules/game.js
seriouslysean: https://github.com/cuttle-cards/cuttle/issues/126

PR #123, opened 2022-10-06
[Feature]: Track my preference for ranked vs unranked games, resolves…

seriouslysean:
… #109

## Please check the following

- For New Features:

### DIFF
--- a/client/js/components/CreateGameDialog.vue
+++ b/client/js/components/CreateGameDialog.vue
@@ -53,6 +53,8 @@
 </template>
 
 <script>
+const LS_PREFERS_RANKED_NAME = 'prefersRanked';
+
 export default {
   name: 'CreateGameDialog',
   data() {
@@ -62,6 +64,14 @@ export default {
       loading: false,
       isRanked: false,
     };
+  },
+  watch: {
+    isRanked(isRanked) {
+      this.setRankedPreference(isRanked);
+    },
+  },
+  mounted() {
+    this.isRanked = this.getRankedPreference();
   },
   methods: {
     submitNewGame() {
@@ -86,6 +96,17 @@ export default {
     handleError(err) {
       this.$emit('error', err);
       this.cancelCreateGame();
+    },
+    setRankedPreference(prefersRanked) {
+      try {
+        localStorage.setItem(LS_PREFERS_RANKED_NAME, prefersRanked === true);
+      } catch (err) {
+        // Local storage is full or something else happened so don't set the value
+      }
+      this.isRanked = prefersRanked;
+    },
+    getRankedPreference() {
+      return localStorage.getItem(LS_PREFERS_RANKED_NAME) === 'true';
     },
   },
 };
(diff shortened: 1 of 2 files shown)


client/js/components/CreateGameDialog.vue
```diff
@@ -87,6 +97,17 @@ export default {
...
+      try {
+        localStorage.setItem(LS_PREFERS_RANKED_NAME, prefersRanked === true);
+      } catch (err) {
+        // Local storage is full or something else happened so don't set the value
+      }
+      this.isRanked = prefersRanked;
+    },
+    getRankedPreference() {
+      return localStorage.getItem(LS_PREFERS_RANKED_NAME) === 'true';
+    },
```
itsalaidbacklife, line 110: Woo venturing into new lands here. It begs questions around future features and when to store data in the db vs local storage etc but I don't think we need to establish a comprehensive plan for that now.

client/js/components/CreateGameDialog.vue
```diff
@@ -63,6 +65,14 @@ export default {
...
     };
   },
+  watch: {
+    isRanked(isRanked) {
+      this.setRankedPreference(isRanked);
+    },
+  },
+  mounted() {
+    this.isRanked = this.getRankedPreference();
+  },
```
itsalaidbacklife, line 75: Any reason to prefer `mounted()` over `created()`?
```suggestion
  created() {
    this.isRanked = this.getRankedPreference();
  },
```

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -150,6 +150,37 @@ describe('Home - Game List', () => {
 
 describe('Home - Create Game', () => {
   beforeEach(setup);
+
+  it.only('Saves ranked setting between sessions', () => {
```
itsalaidbacklife, line 154:
```suggestion
  it('Saves ranked setting between sessions', () => {
```

itsalaidbacklife (approved):
Clean implementation. The second watcher in the app and the first use of localStorage! Exciting stuff

client/js/components/CreateGameDialog.vue
seriouslysean: I don't think we should do this. `mounted()` is just as good and while we're not a SSR app, created has the window only in the browser (which would require an `if (window)` check) or something with `process.env.VUE_ENV === 'client'` if we ever looked in to it. We get the same benefit with mounted and the knowledge that window is always available within it.

client/js/components/CreateGameDialog.vue
seriouslysean: We should create `localStorage` utilities once we start using this in more places, but ideally we'll add them to a user profile object. Maybe LS as a fallback for guests if we ever go that route.

tests/e2e/specs/out-of-game/home.spec.js
seriouslysean: Whoops, thank you. Missed that 😓 -- really need a precommit hook to catch those, haha.

PR #117, opened 2022-09-30
[Feature]: Upgrade Cypress to Version 11

seriouslysean:
## Please check the following

- [x] Do the tests still pass? (see [Run the Tests](https://github.com/cuttle-cards/cuttle#run-the-tests))
- [x] Is the code formatted properly? (see [Linting (Formatting)](https://github.com/cuttle-cards/cuttle#linting-formatting))
- For New Features:
  - [x] Have tests been added to cover any new features or fixes?
  - [x] Has the documentation been updated accordingly?

PR #124, opened 2022-10-06
[Bug]: User session is invalid when logging in consecutively, resolves #116

### LINKED ISSUE #116: [Bug]: User session is invalid when logging in consecutively
### Bug Summary

The `setupGameAsP0()` helper occasionally throws a Type Error due to a 403/forbidden response

### Detailed Description

Cypress appears to be attempting to create a new property on a 403 'error' response when the backend is only sending the string `forbidden`

```
     TypeError: Cannot create property 'stack' on string 'Forbidden'        
                                                                            
Because this error occurred during a `before each` hook we are skipping the 
remaining tests in the current suite: `Countering One-Offs`
```
This happens unpredictably across different tests in the `beforeEach()` hook when using the `setupGameAsP0()` helper function. The issue is likely that one of the commands in that helper function is firing before the appropriate setup has completed, causing the backend to throw a `forbiden()` response

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Steps to Reproduce

1.
2.
3.

seriouslysean:
## Please check the following

- For New Features:

This is an incredibly annoying issue and has to do with using websockets to refresh the user's session. We need to actually use XHR requests to make sure the headers come back with the proper cookies so they can be set by the server for the browser to action on. For example, as it stands currently this will happen:

1. initial load I get `s%3AyWGFsNo7JX0ROG1xG2lJdBd0DsKHsnPt.TGqw%2FLu5dL0Havuk6enxTnbKPBNSmfRfaUIFcgbOTl8` (initial value)
2. On login, I get `s%3AyWGFsNo7JX0ROG1xG2lJdBd0DsKHsnPt.TGqw%2FLu5dL0Havuk6enxTnbKPBNSmfRfaUIFcgbOTl8` (same value)
3. On logout, I get `s%3AyWGFsNo7JX0ROG1xG2lJdBd0DsKHsnPt.TGqw%2FLu5dL0Havuk6enxTnbKPBNSmfRfaUIFcgbOTl8` (same value, this needs to be deleted 😎)
4. On relogin, I get `s%3AyWGFsNo7JX0ROG1xG2lJdBd0DsKHsnPt.TGqw%2FLu5dL0Havuk6enxTnbKPBNSmfRfaUIFcgbOTl8`
5. On reload, I get `s%3AEVw6BGnlugAXdl5DJXYvLTa_do8D1rKj.yYgV4FnF7XNT1yvvd8cyBeg7g5RJ84%2Fm31HCbFwZMhk` (different value)

Ideally after each session refresh (sign in, sign up, login, logout), that cookie should be changing. With this PR it does but it comes with some downsides for the time being until I can work in socket reconnects or socket session refreshes.

This is also the reason for the intermittent Cypress failures we've been seeing in all of our PRs, and will directly unblock https://github.com/cuttle-cards/cuttle/pull/117.

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -16,8 +16,8 @@ module.exports = {
   signup: async function (req, res) {
     try {
       const { username, password } = req.body;
-      const foundUser = await User.find({ username: username });
-      if (foundUser.length > 0) {
+      const foundUser = await User.findOne({ username: username });
+      if (foundUser) {
         throw {
           message: 'That username is already registered to another user; try logging in!',
         };
@@ -36,7 +36,7 @@ module.exports = {
   login: async function (req, res) {
     try {
       const { username, password } = req.body;
-      const user = await userAPI.findUserByUsername(username);
+      const user = await User.findOne({ username: username });
       if (!user) {
         throw {
           message: 'Could not find that user with that username. Try signing up!',
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -14,14 +14,13 @@ const passwordAPI = sails.hooks['custompasswordhook'];
 
 module.exports = {
   signup: async function (req, res) {
-    const { username, password } = req.body;
-    if (!username || !password) {
-      return res.badRequest({ message: 'Username and password are required' });
-    }
     try {
+      const { username, password } = req.body;
       const foundUser = await User.find({ username: username });
       if (foundUser.length > 0) {
-        throw 'That username is already registered';
+        throw {
+          message: 'That username is already registered to another user; try logging in!',
+        };
       }
       // Encrypt pw and create new user
       const encryptedPassword = await passwordAPI.encryptPass(password);
@@ -31,26 +30,24 @@ module.exports = {
       req.session.usr = user.id;
       return res.ok(user.id);
     } catch (err) {
-      return res.badRequest({
-        message: 'Unable to register user',
-      });
+      return res.badRequest(err);
     }
   },
   login: async function (req, res) {
-    const { username, password } = req.body;
-    if (!username || !password) {
-      return res.badRequest({ message: 'Username and password are required' });
-    }
     try {
+      const { username, password } = req.body;
       const user = await userAPI.findUserByUsername(username);
+      if (!user) {
+        throw {
+          message: 'Could not find that user with that username. Try signing up!',
+        };
+      }
       await passwordAPI.checkPass(password, user.encryptedPassword);
       req.session.loggedIn = true;
       req.session.usr = user.id;
       return res.ok(user.id);
     } catch (err) {
-      return res.badRequest({
-        message: 'Unable to login',
-      });
+      return res.badRequest(err);
     }
   },
   reLogin: function (req, res) {
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -14,17 +14,14 @@ const passwordAPI = sails.hooks['custompasswordhook'];
 
 module.exports = {
   signup: async function (req, res) {
-    // Request was missing data
-    if (!req.body.password && !req.body.username) {
-      return res.badRequest('You did not submit a username or password');
+    const { username, password } = req.body;
+    if (!username || !password) {
+      return res.badRequest({ message: 'Username and password are required' });
     }
     try {
-      const { username, password } = req.body;
-      const users = await User.find({ username: username });
-      if (users.length > 0) {
-        return res.badRequest({
-          message: 'That username is already registered to another user; try logging in!',
-        });
+      const foundUser = await User.find({ username: username });
+      if (foundUser.length > 0) {
+        throw 'That username is already registered';
       }
       // Encrypt pw and create new user
       const encryptedPassword = await passwordAPI.encryptPass(password);
@@ -34,33 +31,26 @@ module.exports = {
       req.session.usr = user.id;
       return res.ok(user.id);
     } catch (err) {
-      return res.badRequest(err);
+      return res.badRequest({
+        message: 'Unable to register user',
+      });
     }
   },
-  login: function (req, res) {
-    const { username } = req.body;
-    if (username) {
-      userAPI
-        .findUserByUsername(username)
-        .then((user) => {
-          return passwordAPI
-            .checkPass(req.body.password, user.encryptedPassword)
-            .then(() => {
-              req.session.loggedIn = true;
-              req.session.usr = user.id;
-              return res.ok();
-            })
-            .catch((reason) => {
-              return res.badRequest(reason);
-            });
-        })
-        .catch(() => {
-          return res.badRequest({
-            message: 'Could not find that user with that username. Try signing up!',
-          });
-        });
-    } else {
-      return res.badRequest({ message: 'A username must be provided' });
+  login: async function (req, res) {
+    const { username, password } = req.body;
+    if (!username || !password) {
+      return res.badRequest({ message: 'Username and password are required' });
+    }
+    try {
+      const user = await userAPI.findUserByUsername(username);
+      await passwordAPI.checkPass(password, user.encryptedPassword);
+      req.session.loggedIn = true;
+      req.session.usr = user.id;
+      return res.ok(user.id);
+    } catch (err) {
+      return res.badRequest({
+        message: 'Unable to login',
+      });
     }
   },
   reLogin: function (req, res) {
--- a/client/js/plugins/sails.js
+++ b/client/js/plugins/sails.js
@@ -6,6 +6,11 @@ const { isProd } = require('../../../utils/config-utils');
 
 export const io = require('sails.io.js')(require('socket.io-client'));
 
+export const reconnectSockets = () => {
+  io.socket.disconnect();
+  io.socket.reconnect();
+};
+
 if (!isProd) {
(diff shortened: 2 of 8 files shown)


client/js/plugins/sails.js
```diff
@@ -6,6 +6,11 @@ const { isProd } = require('../../../utils/config-utils');
 
 export const io = require('sails.io.js')(require('socket.io-client'));
 
+export const reconnectSockets = () => {
+  io.socket.disconnect();
+  io.socket.reconnect();
+};
```
seriouslysean, line 12: We need this to refresh any session connections in progress to make sure they get the new session id when the user logs in or out and the cookie changes.

client/js/store/modules/auth.js
```diff
@@ -1,8 +1,36 @@
-import { io } from '../../plugins/sails.js';
+import { io, reconnectSockets } from '../../plugins/sails.js';
 import { ROUTE_NAME_LOBBY, ROUTE_NAME_GAME } from '@/router';
 
 import { getPlayerPnumByUsername } from '_/utils/game-utils.js';
 
+async function handleLogin(context, username, password, signup = false) {
```
seriouslysean, line 6: This was the biggest lift, I had to rewrite all of the sign in/sign up logic to use fetch instead of the sockets, much like we do for the status check due to the way that browsers get and set cookies from the server.

api/controllers/UserController.js
```diff
@@ -34,33 +31,26 @@ module.exports = {
...
-            });
-        })
-        .catch(() => {
-          return res.badRequest({
-            message: 'Could not find that user with that username. Try signing up!',
-          });
-        });
-    } else {
-      return res.badRequest({ message: 'A username must be provided' });
+  login: async function (req, res) {
```
seriouslysean, line 39: I also took the opportunity to rewrite the sails methods while I was in here to consolidate them and get them updated to async await.

client/js/store/modules/auth.js
```diff
@@ -1,8 +1,36 @@
...
+      credentials: 'include',
+    });
+    const data = await response.json();
+    if (response.status !== 200) {
+      throw data.message;
+    }
+    // If the response was successful, the user is logged in
+    context.commit('authSuccess', username);
+    reconnectSockets();
+    return;
```
itsalaidbacklife, line 27: Okay now I see. Socket would keep the old cookie value from its initial connection prior to auth

client/js/store/modules/auth.js
```diff
@@ -24,64 +52,22 @@ export default {
...
-            } else {
-              message = new Error('Unknown error signing up');
-            }
-            context.commit('clearAuth');
-            return reject(message);
-          }
-        );
-      });
+    async requestSignup(context, { username, password }) {
+      await handleLogin(context, username, password, true);
```
itsalaidbacklife, line 60: Same idea
```suggestion
    requestSignup(context, { username, password }) {
      return handleLogin(context, username, password, true);
```

client/js/store/modules/auth.js
```diff
@@ -1,8 +1,36 @@
-import { io } from '../../plugins/sails.js';
+import { io, reconnectSockets } from '../../plugins/sails.js';
 import { ROUTE_NAME_LOBBY, ROUTE_NAME_GAME } from '@/router';
 
 import { getPlayerPnumByUsername } from '_/utils/game-utils.js';
 
+async function handleLogin(context, username, password, signup = false) {
+  const type = signup ? 'signup' : 'login';
+  try {
+    const response = await fetch(`/user/${type}`, {
```
itsalaidbacklife, line 9: nit: I think this would be more clear as `authType`
```suggestion
  const authType = signup ? 'signup' : 'login';
  try {
    const response = await fetch(`/user/${authType}`, {
```

client/js/store/modules/auth.js
```diff
@@ -24,64 +52,22 @@ export default {
...
-              context.commit('authSuccess', data.username);
-              return resolve();
-            }
-            context.commit('clearAuth');
-            return reject(jwres.body.message);
-          }
-        );
-      });
+    async requestLogin(context, { username, password }) {
+      await handleLogin(context, username, password);
```
itsalaidbacklife, line 56: Maybe this should return so that the dispatched action can await the resulting promise. That way callers can manage loading states. I don't think this function needs to be async or to await at that point
```suggestion
    requestLogin(context, { username, password }) {
      return handleLogin(context, username, password);
```

client/js/store/modules/auth.js
```diff
@@ -24,64 +52,22 @@ export default {
...
-        io.socket.get('/user/logout', {}, function handleResponse(resData, jwres) {
-          if (jwres.statusCode === 200) {
-            context.commit('clearAuth');
-            return resolve();
-          }
-          return reject(new Error('Error logging out :('));
+    async requestLogout(context) {
+      try {
+        await fetch('/user/logout', {
+          credentials: 'include',
```
itsalaidbacklife, line 66: I think the same applies for log out
```suggestion
        return await fetch('/user/logout', {
          credentials: 'include',
```

client/js/store/modules/auth.js
```diff
@@ -24,64 +52,22 @@ export default {
...
         });
-      });
+      } catch (err) {
+        context.commit('clearAuth');
```
itsalaidbacklife, line 69: if so we'd throw here
```suggestion
        context.commit('clearAuth');
        throw new Error(err);
```

client/js/views/LoginSignup.vue
```diff
@@ -126,24 +126,17 @@ export default {
...
-          })
-          .then(this.handleLogin)
-          .catch(this.handleError);
+      const action = this.isLoggingIn ? 'requestLogin' : 'requestSignup';
+      try {
+        await this.$store.dispatch(action, {
+          username: this.username,
+          password: this.pw,
+        });
+        this.handleLogin();
```
itsalaidbacklife, line 137: I think there's fuckery here if the store actions don't return their respective promises, which is why I suggested the above changes. I would expect `this.handleLogin()` to fire without awaiting the store dispatch b/c of the missing return. I'm less familiar with `async/await` than with `.then().catch()` but it looks like we need that return in the store actions to me

itsalaidbacklife (approved):
Very cool. An instructive fix worth reviewing with the team I think

client/js/views/LoginSignup.vue
peterancourt: @itsalaidbacklife from my understanding, `this.handleLogin` will not fire until the dispatch above, and all `await`ed child calls _within_ the dispatch, wrap up execution.

As long as we are `await`ing all async calls within the dispatch:
```
// LoginSignup.vue
await this.$store.dispatch(requestLogin, {...}) 

// store/modules/auth.js (requestSignup())
await handleLogin(context, username, password, true);
...

// store/modules/auth.js (handleLogin())
const response = await fetch(`/user/${authType}`, {...})
...
return
```

then no returns should be necessary here with async/await structure! Since all async child calls from the dispatch are properly awaited, there should not be an issue with `this.handleLogin` firing prematurely

client/js/store/modules/auth.js
seriouslysean: Yes, that's exactly right. By disconnecting and reconnecting we force the socket to refresh the authentication via handshake. There's probably a better way but I'm not aware of it. I know that sockets and REST calls share middleware in sails, but the reality is that the sockets don't set cookies on the browser which was causing problems.

client/js/views/LoginSignup.vue
seriouslysean: Yep, this all tracks and I appreciate you all digging through this. We don't need these awaits as long as we're returning `this.handleLogin` since it's a promise, we're just passing it down the lane for the next bit of logic to handle. Good suggestions, thank you!

tests/e2e/specs/out-of-game/authentication.spec.js
```diff
@@ -71,7 +71,7 @@ describe('Logging In', () => {
   /**
    * Rejected logins
    */
-  it('Rejects login before signup', () => {
+  it.only('Rejects login before signup', () => {
```
itsalaidbacklife, line 74:
```suggestion
  it('Rejects login before signup', () => {
```

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -34,7 +34,7 @@ describe('Home - Page Content', () => {
     cy.get('[data-cy=rules-link]').click();
     cy.hash().should('eq', '#/rules');
   });
-  it('Logs user out', () => {
+  it.only('Logs user out', () => {
```
itsalaidbacklife, line 37:
```suggestion
  it('Logs user out', () => {
```

client/js/views/LoginSignup.vue
itsalaidbacklife: I think you're right. My comment is now handled by [this commit: ](https://github.com/cuttle-cards/cuttle/pull/124/commits/7cc7ee9d92f8d95261c1583f7fe203f724dbd707)

This line didn't itself need to change, but we just had to ensure that the actions themselves returned correctly in order to correctly `await` them here

tests/e2e/specs/out-of-game/authentication.spec.js
seriouslysean: God damn it

PR #128, opened 2022-10-07
GameView - background image update

peterancourt:
## Please check the following

- For New Features:

* The game's background should display as below and look good across screen sizes

Screen Shot 2022-10-07 at 1 48 06 PM

Screen Shot 2022-10-07 at 1 51 02 PM

Screen Shot 2022-10-07 at 1 51 34 PM


itsalaidbacklife (approved):
Yesssssssssssss

client/js/views/GameView.vue
```diff
@@ -1108,7 +1108,9 @@ export default {
   color: #fff;
   width: 100vw;
   height: 100%;
-  background: linear-gradient(180deg, #6202ee 14.61%, #fd6222 100%), #c4c4c4;
```
seriouslysean, line 1111: I feel like these should be css variables

client/js/views/GameView.vue
```diff
@@ -1108,7 +1108,9 @@ export default {
...
+  background-image: url('../img/game_images/board_background.svg');
```
seriouslysean, line 1111: Do we have aliases set up to use instead of these relative paths?

seriouslysean (approved):
2 questions, neither of them a blocker. Nice work!

client/js/views/GameView.vue
itsalaidbacklife: Yes but also they're being removed so screw it?

client/js/views/GameView.vue
itsalaidbacklife: My webpack chops are weak, but if anything it would be 


```suggestion
  background-image: url('@/js/img/game_images/board_background.svg');
```
right? @peterancourt wanna give that a try?

client/js/views/GameView.vue
seriouslysean: @peterancourt we may just want to do a full overhaul of raw hex to css variables in the future. In which case just leave these for now and file a follow up.

client/js/views/GameView.vue
seriouslysean: You can leave it for now, I redid a lot of this stuff in the Vue3 branch which we'll have to change anyway 🤷🏼

peterancourt: It looks like it would be `@/img/game_images/board_background.svg`, but with that the CSS is trying to load `./@/img/game_images/board_background.svg` - does the webpack naming convention only work for JS?

client/js/views/GameView.vue
peterancourt: If we use something like that in the future then I definitely agree

client/js/views/GameView.vue
peterancourt: Ah it was `~@/img/game_images/board_background.svg`, the more you know

PR #131, opened 2022-10-07
Revert "Create computed properties for each move type. Create computed proper…"

seriouslysean:
Reverts cuttle-cards/cuttle#120


seriouslysean:
@TommySchmid this doesn't appear to be the culprit after all— false alarm 😬

PR #139, opened 2022-10-08
Enhancement: move logout button to the bottom for mobiles (closes #135)

### LINKED ISSUE #135: [Feature]: Pin logout link to the bottom of the navigation menu
### Feature Summary

The logout link is too easy to click by accident so we should move it

### Detailed Description

F15D03CA-3D5D-4600-80D2-6C31B1F0963B

We have a collapse link that is already there on desktop that can be used as a guideline to make this work, but it might require some trial and error.

See https://github.com/cuttle-cards/cuttle/blob/main/client/js/components/NavigationDrawer.vue#L30.

There is a computer prop to help make sure this is mobile only, `isSmallDevice` that uses the vuetify breakpoints helper.

chikaldirick:
## Please check the following

This is a tiny UI improvement that moves logout link from the very top to the bottom for mobile screens. (closes #135)
P.S. I decided to make things more simple and straightforward in terms of `pageLinks` generation depending on authentication status and screen size. If you want to keep thing as short as it possible we may use a rest spread operator approach of `...(cond ? {a: 1} : {}),` but I'm not sure it'll be readable then.

### DIFF
--- a/client/js/components/NavigationDrawer.vue
+++ b/client/js/components/NavigationDrawer.vue
@@ -25,12 +25,7 @@
     <!-- Bottom section (logout, expand/collapse) -->
     <template #append>
       <v-list v-if="isSmallDevice">
-        <v-list-item
-          link
-          exact
-          :to="{ name: 'Logout' }"
-          data-nav="Logout"
-        >
+        <v-list-item link exact :to="{ name: 'Logout' }" data-nav="Logout">
           <v-icon class="mr-2">mdi-logout</v-icon>
           <template v-if="!isCollapsed">Logout</template>
         </v-list-item>
@@ -69,13 +64,13 @@ export default {
 
       if (!this.authenticated) {
         return [
-            {
-              text: 'Login',
-              icon: 'login',
-              page: { name: 'Login' },
-            },
-            ...rules,
-          ];
+          {
+            text: 'Login',
+            icon: 'login',
+            page: { name: 'Login' },
+          },
+          ...rules,
+        ];
       }
 
       return this.isSmallDevice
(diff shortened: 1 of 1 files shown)


itsalaidbacklife:
Looks like we're failing lint. @chikaldirick can you run `npm run lint:fix` and commit the changes?

chikaldirick:
> Looks like we're failing lint. @chikaldirick can you run `npm run lint:fix` and commit the changes?

Sure, will do now.

chikaldirick:
@itsalaidbacklife, thanks for your comment.
I have resolved prettier errors locally (most of them were related to the extra indentation level and one suggested to put all `<v-list-item>` attributes to a single line.

seriouslysean (changes requested):
Looking at this as a whole, I think we might need to update the desktop version, too. @itsalaidbacklife would love some feedback here. @chikaldirick did great on the mobile version, but I can't help but think we might want something like this for desktop, too? Align the logout being on the bottom and either having the collapse link be below it, or move it out to something like this on desktop as a single control that we can remove on mobile?

image

seriouslysean (changes requested):
@chikaldirick whoops, approved too quickly. I double checked all of this locally and the logout link is showing whether you're logged in or not and twice on mobile devices.

| desktop | desktop | mobile | mobile |
| -------- | -------- | ------- | ------- |
| image | image | image | image |

client/js/components/NavigationDrawer.vue
```diff
@@ -22,9 +22,15 @@
         {{ text }}
       </v-list-item>
     </v-list>
-    <!-- Expand/Collapse -->
+    <!-- Bottom section (logout, expand/collapse) -->
     <template #append>
-      <v-list v-if="!isSmallDevice">
+      <v-list v-if="isSmallDevice">
```
seriouslysean, line 27: This should only show if isSmallDevice and  not authenticated.

Issue #133, opened 2022-10-08
[Feature]: Improve mobile view for game list


seriouslysean:
### Feature Summary

Various improvements to improve layout on mobile

### Detailed Description

* fix title
* Full width game view
* Move buttons below, 3 across maybe? Might make more sense stacked 1 on top of the other
* potentially move logo to the top

![D2C5D6D9-99EF-41A2-8318-D3585EDCED55](https://user-images.githubusercontent.com/414475/194678386-150381e9-37a6-403e-b93e-aa88c57f28bd.jpeg)

vilasrhegde:
@seriouslysean can I work on this issue?

Naveen-Nair:
@seriouslysean Hey! I would like to work on this issue

PR #148, opened 2022-10-10
[Feature]: Improve mobile view for game list 

jerryh-jr:
## Please check the following

- For New Features:

Resolves number #133


client/js/components/CreateGameDialog.vue
```diff
@@ -9,7 +9,7 @@
         elevation="8"
         v-on="on"
       >
-        + Create Game
```
jerryh-jr, line 12: Remove the + from the Create Game button.

client/js/components/GameListItem.vue
```diff
@@ -1,15 +1,15 @@
 <template>
   <div>
     <v-row class="list-item" data-cy="game-list-item">
-      <v-col sm="2" lg="3">
+      <v-col sm="2" lg="3" v-if="!$vuetify.breakpoint.mdAndDown">
```
jerryh-jr, line 4: Added breakpoint so the image only shows on desktop version of screen.

client/js/views/Home.vue
```diff
@@ -11,6 +11,9 @@
           <v-col :cols="$vuetify.breakpoint.mdAndDown ? 12 : 9">
             <div id="card-content-header" class="mb-4">
               <h1 id="home-card-title">Games</h1>
+              <v-row  class="create-game--btn"> 
+               <create-game-dialog @error="handleError" />
+              </v-row>
```
jerryh-jr, line 16: Moved create game button to top of page from bottom.

client/js/views/Home.vue
```diff
@@ -61,9 +64,6 @@
             </v-btn>
           </v-col>
         </v-row>
-        <v-row class="d-flex justify-end mt-8">
-          <create-game-dialog @error="handleError" />
-        </v-row>
```
jerryh-jr, line 63: Removed create game bottom from bottom of page.

client/js/views/Home.vue
```diff
@@ -159,6 +149,25 @@ h1 {
   -webkit-text-fill-color: transparent;
```
jerryh-jr, line 149: Moved css around for media query to min-width vs Max-width so that css for the larger screens are inside the media query. 

image

image

client/js/components/GameListItem.vue
```diff
@@ -102,29 +102,50 @@ export default {
 </script>
```
jerryh-jr, line 102: Changed media query to be set for larger screens with a min width of at least 980px. Added css to resolve issues with the mobile view.

client/js/components/GameListItem.vue
peterancourt: Would it make sense to use `$vuetify.breakpoint.lgAndUp` here instead?

client/js/components/GameListItem.vue
```diff
@@ -102,29 +102,50 @@ export default {
...
+      margin: 3px auto;
+      
   }
 }
+
+.list-item--inner-text {
+  display: flex;
+  padding-bottom: 1rem;
+  padding-top: .25rem;
+}
```
peterancourt, line 126: We can use some SCSS and move this into the `list-item` class above as:
```
&--inner-text {
  display: flex;
  padding-bottom: 1rem;
  padding-top: .25rem;
}

Comment (peterancourt) [reply client/js/components/GameListItem.vue:126]: Since `inner-text` is more of an element than a modifier, we also may want to consider using `__` instead of `--` here, if we're going with a BEM naming scheme.

File: client/js/components/GameListItem.vue [hunk 8]
```diff
@@ -102,29 +102,50 @@ export default {
...
+
 .list-item-button {
   display: flex;
   align-items: center;
   justify-content: center;
+  margin-top: 0;
+  padding-top: .5rem;
 }
```
peterancourt, line 134: I know you didn't add this class, but this can also probably go in to `.list-item` as:
```
&__button {
  display: flex;
  align-items: center;
  justify-content: center;
  margin-top: 0;
  padding-top: .5rem;
}
```

client/js/components/GameListItem.vue
```diff
@@ -102,29 +102,50 @@ export default {
...
+  flex-direction: row;
+  padding: 0;
+  & .game-name {
+    font-size: 1rem;
+    margin-bottom: 1rem;
+  }
+}
+  .list-item--inner-text {
+    display: block;
+    padding: 0;
```
peterancourt, line 148: Same change here

client/js/views/Home.vue
```diff
@@ -1,14 +1,20 @@
 <template>
-  <div class="home">
-    <div class="container">
+  <div class="home pa-4" :class="{
+    'home--large': !$vuetify.breakpoint.mdAndDown,
```
peterancourt, line 3: We may also be able to make use of `$vuetify.breakpoint.lgAndUp` here

client/js/views/Home.vue
```diff
@@ -1,14 +1,20 @@
...
-          <v-col cols="9">
-            <v-row id="card-content-header" class="mb-4">
-              <v-col cols="4">
-                <h1 id="home-card-title">Games</h1>
-              </v-col>
-            </v-row>
+          <v-col :cols="$vuetify.breakpoint.mdAndDown ? 12 : 9">
+            <div id="card-content-header" class="mb-4">
+              <h1 id="home-card-title">Games</h1>
+              <v-row  class="create-game--btn">
```
peterancourt, line 14: Nitpicky, but it looks like we have an extra space here between `v-row` and `class`

client/js/views/Home.vue
```diff
@@ -24,8 +30,8 @@
               </div>
             </div>
           </v-col>
-          <v-col id="side-nav" cols="3">
-            <img id="logo" alt="Vue logo" src="../img/logo.png" />
+          <v-col id="side-nav" :cols="$vuetify.breakpoint.mdAndDown ? 12 : 3">
+            <img v-if="!$vuetify.breakpoint.mdAndDown" id="logo" alt="Vue logo" src="../img/logo.png" />
```
peterancourt, line 34: Same here with `lgAndUp`

client/js/views/Home.vue
```diff
@@ -154,25 +149,66 @@ h1 {
...
+  margin: 0 auto;
+  .home--large & {
+    height: auto;
+    width: 80%;
+    margin: 20px auto;
+  }
 }
 
-#card-content-header {
+.create-game--btn {
```
peterancourt, line 181: I'm a bit hesitant to use the `--` if we don't have a parent `create-game` class. Maybe we can name this `create-game-btn`?

client/js/views/Home.vue
```diff
@@ -154,25 +149,66 @@ h1 {
...
-#game-list-card {
-  border-radius: 15px;
-  margin-top: 8px;
+#logo {
+  height: 20vh;
+  margin: 0 auto;
+  .home--large & {
+    height: auto;
+    width: 80%;
+    margin: 20px auto;
```
peterancourt, line 177: I see we're using `--large` here on all screen sizes, do we just want to make this `.home` instead?

peterancourt (commented):
Looking good so far! A couple notes on class names/vuetify breakpoints

client/js/components/GameListItem.vue
jerryh-jr: I agree, and was thinking about making this change as well.

client/js/views/Home.vue
jerryh-jr: So if the first one is changed. We would have to change the number of cols or the image explodes in size. The second in this image where we are checking if it is not mdAndDown I have changed.

client/js/components/GameListItem.vue
```diff
@@ -1,21 +1,21 @@
 <template>
   <div>
     <v-row class="list-item" data-cy="game-list-item">
-      <v-col sm="2" lg="3" v-if="!$vuetify.breakpoint.mdAndDown">
+      <v-col sm="2" lg="3" v-if="$vuetify.breakpoint.lgAndUp">
```
jerryh-jr, line 4: Change breakpoint from !mdAndDown to just lgAndUp

client/js/components/GameListItem.vue
```diff
@@ -102,29 +102,47 @@ export default {
...
+  flex-direction: row;
+  padding: 0;
+  & .game-name {
+    font-size: 1rem;
+    margin-bottom: 1rem;
+  }
+  &__inner-text {
+    display: block;
+    padding: 0;
+  }
```
peterancourt, line 145: Let's bump up the indentation on these

client/js/views/Home.vue
```diff
@@ -24,8 +30,8 @@
               </div>
             </div>
           </v-col>
-          <v-col id="side-nav" cols="3">
-            <img id="logo" alt="Vue logo" src="../img/logo.png" />
+          <v-col id="side-nav" :cols="$vuetify.breakpoint.mdAndDown? 12 : 3">
```
peterancourt, line 33: Looks like we need a space here between `$vuetify.breakpoint.mdAndDown` and `?`

peterancourt (approved):

Issue #151, opened 2022-10-12
[Feature]: Add .npmrc file to suppress some output from npm


seriouslysean:
### Feature Summary

Clean up message output in npm

### Detailed Description

https://twitter.com/wesbos/status/1579895071272755200?s=46&t=pYEfeWIxVs98nek7vXcooA

> take the noise out of your npm install by putting a .npmrc file in your project with these settings
> You should still periodically run `npm audit` and `npm fund`, but this cleans it up for those of us who re-install often

```
fund=false
audit=false
legacy-peer-deps=false
loglevel=error
```

For the settings above:
- `fund`, https://docs.npmjs.com/cli/v8/using-npm/config#fund
- `audit`, https://docs.npmjs.com/cli/v8/using-npm/config#audit
- `legacy-peer-deps`, https://docs.npmjs.com/cli/v8/using-npm/config#legacy-peer-deps
    - This will be required when we upgrade npm to 8 as peer deps are strict and will prevent installs for mismatches. We can remove this once our package.json is fully compliant. cc @itsalaidbacklife 
- `loglevel`, https://docs.npmjs.com/cli/v8/using-npm/config#loglevel

Test the output here and compare it to what we see today to make sure it's actually having an impact. Might be worth looking to see if there are other settings that might be useful to have, too.

- https://docs.npmjs.com/cli/v8/using-npm/config
- https://docs.npmjs.com/cli/v8/configuring-npm/npmrc 
- https://stackabuse.com/the-ultimate-guide-to-configuring-npm/
- https://nodesource.com/blog/configuring-your-npmrc-for-an-optimal-node-js-environment/

lbalmaceda:
I've never done this before. Happy to explore the available options

seriouslysean:
@lbalmaceda would love to have you! New to me, too, so we can learn together, haha.

lbalmaceda:
Seems like `audit=false` would just stop submitting the dependencies to npm for a scan. That makes that last message disappear, but how often do you run package install to consider this message noise? Might be worth keeping around.
<img width="434" alt="Screenshot 2022-10-12 at 15 52 53" src="https://user-images.githubusercontent.com/3900123/195365032-f7eb8bbf-8540-4d27-9217-2ee79d85d116.png">


`fund=false` does remove the following message "xxx packages looking for funding" but does not stop the other messages from appearing.
[...]

PR #153, opened 2022-10-12
Reduce noise from npm output

lbalmaceda:
## Please check the following

- For New Features:

Attempt to reduce the console output from `npm install`. This could be tuned with the `loglevel` setting https://docs.npmjs.com/cli/v8/using-npm/config#loglevel.

- `fund=false` removes the message that N packages are accepting funding.
- `audit=false` stops npm from submitting a summary of the project dependencies for a scan. Might imply a security risk if we ignore these.

The readme change relates to installing dependencies. And is that some packages are using post-install scripts to run advertisement. We could in theory define a new script like "npm run deps" that would run "npm ci --ignore-scripts", however, I think devs would be more used to `npm i` and `npm ci`.

Relates to https://github.com/cuttle-cards/cuttle/issues/151


.npmrc
```diff
@@ -0,0 +1,8 @@
+# removes notice from projects looking for funding
+fund=true
+# stops auto-checking for dependencies' issues
+audit=true
+# ignore peerDependencies contract - in preparation of migration to npm 8
+legacy-peer-deps=true
```
itsalaidbacklife, line 6: Am I correct that these don't actually do anything because they are the default values, however the `loglevel=warn` actually accomplishes the desired suppression? Maybe we should remove these then?

lbalmaceda: `loglevel=warn` is already suppressing the output of `fund` and `audit`, so if we choose to keep `loglevel=warn` we can remove the other two. I'll rebase the branch and remove the first commit of the PR so they go away.

README.md
```diff
@@ -36,7 +36,7 @@ git clone https://github.com/cuttle-cards/cuttle
 #### Install Dependencies
 Open your shell/terminal of choice, `cd` into the root folder of this repo and run
 ```
-npm ci
+npm ci --ignore-scripts
```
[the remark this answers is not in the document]
seriouslysean: I think we're ok to just leave this out and not worry about it for now. Some output is ok, interesting the outcome of this. Never really dug in to the `.npmrc` side of things, nor what the settings do. Really appreciate the deep dive here!

.npmrc
```diff
@@ -0,0 +1,4 @@
...
+# reduces noise from console
+loglevel=warn
```
seriouslysean, line 4: Need a trailing return in this file, otherwise good to go
```suggestion
loglevel=warn

```

seriouslysean (changes requested):
Couple small comments, otherwise looks good to me!

seriouslysean (approved):
@lbalmaceda still missing the empty return at the end of the npmrc, but it’s good to go otherwise. 👍🏼 

Amazing discussion and research in this ticket, thank you so much!

itsalaidbacklife (approved):
Great work and an informative discussion

PR #157, opened 2022-10-12
Feature: Card Appearance Update

peterancourt:
## Please check the following

- For New Features:

* Build the app and run any test
* All cards should be updated and should display at the proper size

Screen Shot 2022-10-14 at 2 39 00 PM

**Kings, Queens, and A-4s**
Screen Shot 2022-10-07 at 2 19 04 PM

**5-7s, Glasses 8s, 9s and 10s**
Screen Shot 2022-10-12 at 4 51 48 PM

**8s and Jacks**
Screen Shot 2022-10-12 at 5 03 37 PM


client/js/components/GameView/Card.vue
```diff
@@ -35,11 +35,11 @@ export default {
   name: 'Card',
   props: {
     suit: {
-      type: Number,
+      type: [Number, String],
       required: true,
     },
     rank: {
-      type: Number,
+      type: [Number, String],
```
itsalaidbacklife, line 42: Why can these be strings?

itsalaidbacklife: Okay I see it's for the 'back' case. Is that necessary? The current implementation doesn't use the `Card` component for card backs. Curious what you and @seriouslysean both think about this. It's more generic this way, which is perhaps more flexible. 

Maybe it makes more sense to have a boolean `back` prop and to create a computed prop to return the path to the image? Then the question becomes what values if any to give for the `suit` and `rank` props. We could either
1. Pass the suit and rank anyway even though they are ignored when rendering the component b/c `back = true`. This arguably is less secure because the user isn't supposed to know which cards are in place given that these cards are displayed as card backs, but if you're already able to read the non-rendered values of props in the component hierarchy then you can already see that stuff in the store, so I think this might be the best/cleanest
2. Make those props optional and/or accept null values and set them to null for card back images

What do y'all think?

peterancourt: Agreed that 'back' feels a bit awkward here, but I do like the flexibility and consistency in styling with visible cards that adding this to `Card.vue` provides.

'1' Definitely seems like the most clean approach, although as the game grows I fear that people would realize over that the rank and suit are visible on face-down cards. Null definitely seems like a potential option as well! 

What if we used a bogus numerical value like `99`?

client/js/components/GameView/Card.vue
```diff
@@ -35,11 +35,11 @@ export default {
   name: 'Card',
   props: {
     suit: {
-      type: Number,
+      type: [Number, String],
```
seriouslysean, line 38: This definitely has a smell to it, feels like it should be one type. If suit is no longer always required, this should not be a required prop. Instead something like this:
```
suit: {
      type: Number,
```
which will default to `undefined` I think?

client/js/img/cards/card_1_9.svg
```diff
@@ -0,0 +1 @@
...,.94-4.32c0-.16,0-.43-.05-.82S29.14,30.53,29.1,30Zm-7-12.37a17,17,0,0,0,1.44,6.86c1,2.2,1.92,3.29,2.89,3.29a2,2,0,0,0,1.7-.81,4.36,4.36,0,0,0,.56-2.43,17.18,17.18,0,0,0-1.42-6.92q-1.43-3.29-3-3.28a1.8,1.8,0,0,0-1.6.84A4.46,4.46,0,0,0,22.06,17.65Z"/><use width="115.78" height="121.49" transform="translate(48.18 34.54) scale(0.43)" xlink:href="#Red_2"/><use width="115.78" height="121.49" transform="translate(116.02 34.54) scale(0.43)" xlink:href="#Red_2"/><use width="115.78" height="121.49" transform="translate(48.18 85.13) scale(0.43)" xlink:href="#Red_2"/><use width="115.78" height="121.49" transform="translate(116.02 85.13) scale(0.43)" xlink:href="#Red_2"/><use width="115.78" height="121.49" transform="translate(82.17 105.44) scale(0.43)" xlink:href="#Red_2"/><use width="115.78" height="121.49" transform="translate(48.18 133.72) scale(0.43)" xlink:href="#Red_2"/><use width="115.78" height="121.49" transform="translate(116.02 133.72) scale(0.43)" xlink:href="#Red_2"/><use width="115.78" height="121.49" transform="translate(48.18 183.31) scale(0.43)" xlink:href="#Red_2"/><use width="115.78" height="121.49" transform="translate(116.02 183.31) scale(0.43)" xlink:href="#Red_2"/></svg>
```
seriouslysean, line 1: Not this car specifically but these all need to be optimized. They have `id`s and all sorts of other attributes we probably don't want. Not for this PR, but maybe after Vue 3 (and Vite) we look in to an SVG loader or a tool we can run to optimize them.

client/js/components/GameView/Card.vue
```diff
@@ -78,6 +78,8 @@ export default {
           return 'Hearts';
         case 3:
           return 'Spades';
+        case 'back':
```
seriouslysean, line 81: Back also isn't a suit name so this doesn't belong here. This logic should be forked off and a decision made elsewhere before it get here. If you make the above changes, you could fast fail here with a number check and return `'Invalid Suit Error'` early or just remove the back entry here.

You'd probably want another computed prop for `isBack` that could use a check like `!suit || !rank` or whatever to return `true`, and use that instead.

client/js/components/GameView/Card.vue
```diff
@@ -35,11 +35,11 @@ export default {
...
     suit: {
-      type: Number,
+      type: [Number, String],
       required: true,
     },
     rank: {
-      type: Number,
+      type: [Number, String],
       required: true,
     },
```
seriouslysean, line 44: Same here, get rid of the `required` prop and switch type to just `Number`.

client/js/views/GameView.vue
```diff
@@ -103,15 +103,14 @@
...
+                      suit="back"
+                      rank="back"
+                      data-opponent-hand-card
                       class="transition-all opponent-card-back-wrapper opponent-hand-card mx-2"
-                    >
-                      <v-card class="opponent-card-back" data-opponent-hand-card>
-                        <v-img :src="require('../img/logo_head.svg')" contain />
-                      </v-card>
-                    </div>
+                    />
```
seriouslysean, line 113: With the removal of required for the `rank` and `suit` props, this could be simplified to just 
```
                    <card
                      v-for="(card, index) in opponent.hand"
                      :key="index"
                      data-opponent-hand-card
                      class="transition-all opponent-card-back-wrapper opponent-hand-card mx-2"
                    />
```

seriouslysean (changes requested):

peterancourt:
@seriouslysean awesome suggestions, that simplifies things a lot

itsalaidbacklife (changes requested):
Looking good! Last change request: remove the green border from around the deck while it's your turn. It look ugly now that the deck is a stack of cards, and I think the other affordances are sufficient anyway

seriouslysean (approved):
With the back changes, this looks much better. No other blockers from my side! Nice work, @peterancourt! 👍

PR #160, opened 2022-10-13
Updated rules page, tweaking the grammar and clarifying some things

byarmis:
## Please check the following

- For New Features:
  - ~~[ ] Have tests been added to cover any new features or `fixes?~~
  - ~~[ ] Has the documentation been updated accordingly?~~

* Removed "Preview" from the gif buttons on the rules page
* Capitalized Royals
* Removed unnecessary apostrophes
* `delete` -> `dealer`
* Sentences end with a period
* Replaced one `==` with a `===`
* Removed deprecated `frameborder`
* Added an FAQ about what happens if a 7 reveals two Jacks and your opponent has no points in play
* Added an FAQ about what happens if you 7 with one card left in the deck.
* Added note to the Draw option about passing.

### DIFF
--- a/api/services/cardService.js
+++ b/api/services/cardService.js
@@ -16,9 +16,8 @@ module.exports = {
             }
             return resolve(card);
           });
-      } else {
-        return Promise.reject({ message: 'Invalid arguments for findCard' });
       }
+      return Promise.reject({ message: 'Invalid arguments for findCard' });
     });
   },
 
--- a/client/js/views/Rules.vue
+++ b/client/js/views/Rules.vue
@@ -286,7 +286,7 @@ export default {
             title: 'Points',
             icon: 'numeric',
             description:
-              'Play a number card from your hand.  Worth its rank in points and lasts until scrapped or scuttled.',
+              'Play a number card from your hand. Worth its rank in points and lasts until scrapped.',
             staticImg: 'cuttle_points.png',
             animatedImg:
               'https://github.com/cuttle-cards/cuttle-assets/blob/main/assets/points.gif?raw=true',
--- a/client/js/views/Rules.vue
+++ b/client/js/views/Rules.vue
@@ -382,7 +382,7 @@ export default {
             title: 'Two - Effect 1',
             icon: 'delete',
             description:
-              "Two's have two alternative one-off effects: Counter target One-Off Effect.  This happens during your opponent's turn.",
+              "Two's have two alternative one-off effects: Counter target One-Off Effect. May be played during your opponent's turn.",
             staticImg: 'cuttle_counter.png',
             animatedImg:
               'https://github.com/cuttle-cards/cuttle-assets/blob/main/assets/counter.gif?raw=true',
--- a/client/js/views/Rules.vue
+++ b/client/js/views/Rules.vue
@@ -8,9 +8,9 @@
       <div>
         <h1 class="gradient-text">Rules of Cuttle</h1>
         <p class="d-block">
-          Cuttle is a 2 player battle card game played with a standard 52-card deck of cards. It has
-          the strategic nuance of trading card games like Magic, with the elegant balance of a
-          standard deck--and you can play it for free! Test your mettle in the deepest cardgame
+          Cuttle is a two player battle card game played with a standard 52-card deck of cards. It
+          has the strategic nuance of trading card games like Magic, with the elegant balance of a
+          standard deck--and you can play it for free! Test your mettle in the deepest card game
           under the sea!
         </p>
         <div class="d-flex justify-center">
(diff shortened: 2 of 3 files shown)


client/js/views/Rules.vue
```diff
@@ -275,7 +286,7 @@ export default {
             title: 'Points',
             icon: 'numeric',
             description:
-              'Play a number card from your hand. Worth its rank in points, lasts until scrapped',
+              'Play a number card from your hand.  Worth its rank in points and lasts until scrapped or scuttled.',
```
itsalaidbacklife, line 289: Extra space
```suggestion
              'Play a number card from your hand. Worth its rank in points and lasts until scrapped or scuttled.',
```

client/js/views/Rules.vue
```diff
@@ -286,7 +297,8 @@ export default {
           {
             title: 'Scuttle',
             icon: 'skull-crossbones',
-            description: 'Scrap an opponent’s point card with a bigger one from your hand',
+            description:
+              'Scrap an opponent’s point card with a bigger one from your hand.  Ties are broken by the suit: ♣️ (weakest) < ♦️ < ♥️ < ♠️ (strongest).',
```
itsalaidbacklife, line 301: emojies fts

client/js/views/Rules.vue
```diff
@@ -361,15 +373,16 @@ export default {
...
             staticImg: 'cuttle_one_off_ace.png',
             animatedImg:
               'https://github.com/cuttle-cards/cuttle-assets/blob/main/assets/ace.gif?raw=true',
           },
           {
             title: 'Two - Effect 1',
             icon: 'delete',
-            description: 'Twos have two alternative one-off effects: Counter target One-Off Effect',
+            description:
+              "Two's have two alternative one-off effects: Counter target One-Off Effect.  This happens during your opponent's turn.",
```
itsalaidbacklife, line 385: Got an extra space. Not all counters happen on your opponent's turn b/c if you counter their counter then it's still on your turn. Maybe a 'may' effect?
```suggestion
              "Two's have two alternative one-off effects: Counter target One-Off Effect. May be played during your opponent's turn.",
```

itsalaidbacklife (approved):
Looks good. Picking a few nits

api/services/cardService.js
```diff
@@ -16,9 +16,8 @@ module.exports = {
             }
             return resolve(card);
           });
-      } else {
-        return Promise.reject({ message: 'Invalid arguments for findCard' });
       }
+      return Promise.reject({ message: 'Invalid arguments for findCard' });
```
seriouslysean, line 20: This is going to conflict with the changes in https://github.com/cuttle-cards/cuttle/pull/161/files#diff-90d500bee1581144835b0ef79d6747a0359efdb86c692d69f6ccc1919c387563R20

client/js/views/Rules.vue
seriouslysean, line 301:
```suggestion
              'Scrap an opponent’s point card with a bigger one from your hand.  Ties are broken by the suit: &clubs; (weakest) &lt; &diams; &lt; &hearts; &lt; &spades; (strongest).',
```
I actually think these should be the html entities instead. I don't care enough to have it be a blocker since it's arguably easier to read this way, but since this _is_ html, the encoding makes more sense to me. Same with the less than symbols.

It's arguably more valid and looks the same
| html |
|-------|
| image |

seriouslysean (approved):
Nothing major, one callout about encoded html entities though.

client/js/views/Rules.vue
seriouslysean: At the very least we should encode the less than symbols, emojis I'm fine with either considering the ease of readibility with the images. 
image

client/js/views/Rules.vue
```diff
@@ -266,7 +276,8 @@ export default {
           {
             title: 'Draw',
             icon: 'cards-playing-spade-multiple',
-            description: 'Take one card from the Deck. You may not draw past the 8-card hand limit',
+            description:
+              'Take one card from the deck. You may not draw past the 8-card hand limit.  If there are no cards left in the deck, you may pass.  Three consecutive passes ends the game in a stalemate',
```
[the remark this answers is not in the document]
byarmis: image

Yeah, I'd say it looks reasonable (page is 1244px wide, I think)

api/services/cardService.js
byarmis: I don't think it will since they're both making the same (linted) change.  If it does though, I'll fix it in the second MR before that's merged in

client/js/views/Rules.vue
itsalaidbacklife: @seriouslysean it looks like this is native vue behavior:
https://codepen.io/aleph_one/pen/gOzyYrb

itsalaidbacklife (approved):
Nice work

client/js/views/Rules.vue
seriouslysean: Makes sense since we’re not using `v-html` I suppose. Thanks for the info!

Issue #156, opened 2022-10-12
[Feature]: Create version endpoint


byarmis:
### Feature Summary

Create an API endpoint that returns the current version number

### Detailed Description

This could probably live in a `/healthcheck` or `/api/healthcheck` route, but debugging the deploy with @itsalaidbacklife currently and an endpoint that returns the current version of the deployed code would be handy for testing things out.

Down the line, this endpoint could return if the backend can connect to the database correctly and things like that, but for starters, just the version would be helpful

lbalmaceda:
@byarmis By "version of the deployed code" you mean the version that lives in the package.json file here or where? 
https://github.com/cuttle-cards/cuttle/blob/main/package.json#L3

itsalaidbacklife:
@lbalmaceda yes I think the package.json version would be most appropriate. I think we should use `/status` as the endpoint url and we could put the endpoint in a new file `api/controllers/StatusController.js`

It would be ideal if it could return a json payload that includes both `version` taken from package json as you suggest and `status` indicating whether the server is able to query the database. It could make a simple query like `Game.find({limit: 1})` and set the `status` to "up" if there are no errors and one record is found, and to "down" otherwise

itsalaidbacklife:
Awesome thank you! While we are at it do you think it'd be feasible to include the SHA of the latest commit? I'm not sure how we'd approach it exactly but if we could have all three pieces of info (package version, last commit sha, connectivity status) that would be extremely helpful

lbalmaceda:
Sure! I'll have to understand how the server/api is being built as a deliverable/package. There might be a way to use the git CLI to evaluate and get the latest commit SHA during the "building" step.

lbalmaceda:
I've pushed the changes for the original request. The commit SHA is more challenging as it would mean the server would need to clone this repository and have it available locally for the git CLI to pick it up on runtime. I'll work on a separate PR that branches off this one to try to add a static file during the build process that we can reference to from the introduced controller. I hope that's OK with you!

itsalaidbacklife:
> I've pushed the changes for the original request. The commit SHA is more challenging as it would mean the server would need to clone this repository and have it available locally for the git CLI to pick it up on runtime. I'll work on a separate PR that branches off this one to try to add a static file during the build process that we can reference to from the introduced controller. I hope that's OK with you!

Yeah that's great. We can definitely take this as an initial pass and figure out the git stuff downstream

lbalmaceda:
[...]

PR #161, opened 2022-10-13
Add API status endpoint

lbalmaceda:
## Please check the following

- For New Features:

Adds a new endpoint `GET /api/status` that returns wether the DB is available or not, and the version form the `package.json` file.

Tests for these changes be run with the following command:
`npm run e2e:server -- --spec tests/e2e/specs/out-of-game/api-status.spec.js`

Relates to https://github.com/cuttle-cards/cuttle/issues/156


api/hooks/apiStatusHook/index.js
```diff
@@ -0,0 +1,17 @@
...
+  return {
+    getStatus: function () {
+      return new Promise(function (resolve, reject) {
+        Game.count().exec(function (error, game) {
+          if (error || game <= 0) {
+            return resolve(false);
+          }
+          return resolve(true);
+        });
+      });
```
itsalaidbacklife, line 14: This is just the logic I had in mind, but can we move it into helpers, rather than a hook? There isn't much practical difference except both hooks and services are deprecated at this point. Here's the docs [on helpers](https://sailsjs.com/documentation/concepts/helpers). Same basic idea but you don't need a wrapping directory and the function guts get tucked into the `fn` attribute of a larger object that also has metadata like `friendlyName` and `description`. 

You can check `api/helpers/get-seasons-without-rankings` for a simple example.

Also would you mind rejiggering this to leverage `async/await`? The `exec()` method predates native promises in JS and `async/await` haha. So we haven't cleaned up the entire codebase in that regard (since I started on it before ES6 was a thing. So going forward, generally `async/await` are preferable imo (though we can still combine them with `Promise.all()` when doing multiple queries)

api/hooks/apiStatusHook/index.js
```diff
@@ -0,0 +1,17 @@
+module.exports = function apiStatusHook() {
+  ////////////////
+  // API Status //
+  ////////////////
+  return {
+    getStatus: function () {
+      return new Promise(function (resolve, reject) {
+        Game.count().exec(function (error, game) {
```
seriouslysean, line 8: Not that I think anyone would do this, but as it stands it's possible to ddos the server by pinging this endpoint repeatedly. I think it might be worth caching the value for like 3-5 minutes at a time or something similar. I'm not sure the best way to do that since we don't really have a cache wrapper. But maybe in pro scenarios have it use the memcached server or something? cc @itsalaidbacklife

api/controllers/StatusController.js
```diff
@@ -0,0 +1,22 @@
+const statusAPI = sails.hooks['apistatushook'];
+
+const isDatabaseAvailable = () => {
+  return statusAPI.getStatus();
+};
+
+const getPackageVersion = () => {
+  // 'npm_package_version' is only available when the server is started via an npm script
+  return process.env.npm_package_version || 'unknown';
```
seriouslysean, line 9: Since we have access to internal files, I'm curious if you could just import the package.json and pluck off only the version key so that it's always available. Something like this:
```
const { version } = require('../../../package.json');
```
That path might not be right but the spirit is there.

api/services/cardService.js
```diff
@@ -16,9 +16,8 @@ module.exports = {
             }
             return resolve(card);
           });
-      } else {
-        return Promise.reject({ message: 'Invalid arguments for findCard' });
       }
+      return Promise.reject({ message: 'Invalid arguments for findCard' });
```
seriouslysean, line 20: I don't think this is a blocker by any means, but we should file a follow up to refactor this function to use async await and the modern commands for game data.

config/routes.js
```diff
@@ -52,6 +52,7 @@ module.exports.routes = {
   '/user/logout': 'userController.logout',
   '/user/status': 'userController.status',
 
+  'GET /api/status': 'StatusController.getStatus',
```
seriouslysean, line 55: I love that you jumped right in to sails and just started hooking things up-- really nice work!

tests/e2e/specs/out-of-game/api-status.spec.js
```diff
@@ -0,0 +1,32 @@
+const packageVersion = require('../../../../package.json').version;
```
seriouslysean, line 1: This can be ES6 syntax as far as I know:
```
const { version } = require('../../../../package.json');
```
Make sure it goes under the import below-- imports first, then conts.

tests/e2e/specs/out-of-game/api-status.spec.js
```diff
@@ -0,0 +1,32 @@
...
+
+import { setupGameAsP0 } from '../../support/helpers';
+
+describe.only('API Status Page', () => {
```
seriouslysean, line 5: Love that you did tests for this, awesome! 👏🏼

seriouslysean (changes requested):
Couple adjustments, nothing major-- this is great work! One thing I will call out is that I'm using `/user/status` for the initial call to get the game data. That said, I feel like this call should probably be something like `/health` instead that would return something along these lines:
```
{
    version: '3.1.x',
    sha: 'sha1' // once we figure this out
    alive: true|false,
}
```

api/hooks/apiStatusHook/index.js
itsalaidbacklife: interesting point. That said, is this any more vulnerable than any other endpoint?

tests/e2e/specs/out-of-game/api-status.spec.js
itsalaidbacklife: Bump for test love. Let's remove the `.only()` so that if we wanted to add an additional describe block here it could run
```suggestion
describe('API Status Page', () => {
```

api/hooks/apiStatusHook/index.js
seriouslysean: Not really but it seems a little easier to automate a hit against. I'm not really bothered either way, just something I see as a potential problem.

lbalmaceda: I agree with the DDOS but can share that is not just this endpoint, it would be all of them. I think you could require authentication for them so at least you could track who is doing this and limit or ban them, for example. That assuming you are monitoring the service. But there will be always endpoints that are meant to be public, like the sign up endpoint. 

I was reading that Sails supports [using Express middlewares](https://sailsjs.com/documentation/concepts/middleware#?http-middleware) and these might be a quick solution. Check this [blogpost](https://dev.to/brunohgv/limiting-node-js-api-calls-with-express-rate-limit-11kl) for guidance. Could be tackled in a separate PR.

seriouslysean: Yeah by no means a blocker, excellent discussion on everything— really appreciate the level of detail and research :)

api/hooks/apiStatusHook/index.js
lbalmaceda: I hope I did it right this time. I always get confused with promises and async/await 😂

lbalmaceda:
Pushed my changes! 

> One thing I will call out is that I'm using /user/status for the initial call to get the game data. 

@seriouslysean I'm not sure I followed you here. I kept the use of `Game.count()` as it seemed a simple way to check if a game was ongoing. That said, users would be definitely available before a game is created. Can you clarify what you meant with your comment?

itsalaidbacklife:
> @seriouslysean I'm not sure I followed you here. I kept the use of Game.count() as it seemed a simple way to check if a game was ongoing. That said, users would be definitely available before a game is created. Can you clarify what you meant with your comment?

He's not actually talking about the function logic, but rather the request mapping. There is an existing endpoint that's mapped to `/user/status` (used to check authentication status), which @seriouslysean is saying is confusingly similar to `/api/status`.

I think that's a fair point and perhaps we should this endpoint to `/health`? Sorry to move the requirements out from under you, but these discussions have been informative and I think this change would make things more clear

api/controllers/HealthController.js
```diff
@@ -0,0 +1,22 @@
+const { version: pkgVersion } = require('../../package.json');
+
+const isDatabaseAvailable = async () => {
+  return await sails.helpers.getApiHealth();
+};
```
itsalaidbacklife, line 5: I'm open to keeping this, but penny for your thoughts--why create the `isDatabaseAvailable` function rather than calling `await sails.helpers.getApiHealth();` directly in the endpoint?

lbalmaceda: It's a one liner so we can definitely merge it and remove the additional function. What I originally thought is, the `isDatabaseAvailable` function in this controller will hold the logic to get that value. However that's not true as before I was delegating that to a hook and now to a helper (one liners). I'll make the change :zap:

lbalmaceda:
> I think that's a fair point and perhaps we should this endpoint to /health?

@itsalaidbacklife All good then! That's done in this commit https://github.com/cuttle-cards/cuttle/pull/161/commits/7ff928af49ae9d3a1a63a1f23ff9dc2a0e558dfd#diff-33a52123e3f345cf172fd8514455ae4d0b14c69d1139198bb3bf73f00516ca81R55

I just pushed another one removing the extra function wrapper.

seriouslysean (approved):
@lbalmaceda this looks great, thank you! I’m going to pull it down and run it locally to double check but if that works we’re good to merge.

it was really great working with you, the back and forth combined with your knowledge was really refreshing. Would love to have you join us on Discord if you have time— we chat a bunch in the code channel but also play on Wednesday nights and Thursday afternoons. https://discord.gg/Q95en6Et

Thanks again!

api/helpers/get-api-health.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
+  friendlyName: 'Get API health',
+
+  description:
+    'Get the health of the API service. This includes database status and service version.',
+
+  fn: async (_, exits) => {
+    try {
+      const games = await Game.count();
```
seriouslysean, line 9: @lbalmaceda we need a different way to query to `alive` status, if there are no games this returns false. Tested locally http://localhost:1337/health
```
{
  "alive": false,
  "version": "3.1.0"
}
```
which sort of defeats the purpose in my mind. Maybe we can query the port of the db or something similar? cc @itsalaidbacklife

lbalmaceda: An easy fix would be counting with the query to throw an error if there's an issue with the DB adapter (or just any error honestly).
image

Their docs actually [suggest the offline scenario](https://sailsjs.com/documentation/concepts/models-and-orm/errors#?adapter-errors):
image


So I could reorder the try/catch into something like this:

```js
try {
  Game.count(); // ignore result
} catch (err) {
  return false;
}
return true;
```
image


If you have a scenario I can reproduce to bring the db down and check for the "db disconnected" status I'm happy to test this approach. I've spent more than an hour trying to stub/mock the database model so I can force it to throw an exception to test it but I couldn't make it work. It's like "Game" is not available from the test context.. And I've tried [lifting sails](https://sailsjs.com/documentation/concepts/testing#?lifecycletestjs) as explained there but that threw an "invalid left assignment" error. I'm blocked at this point. Any clue how I can access this instance to stub it? I've pushed this change with a "skip" for that test that fails. 

----- 

Otherwise, I've been reading the sails docs for [Datastores](https://sailsjs.com/documentation/reference/waterline-orm/datastores). These are like generic adapters for databases, which you can customize if you are using a different db engine i.e. mongodb. However, the datastore API doesn't expose any method to access the connection status. You can always access the "raw driver" to check that but then you'd be tied to the API of the specific provider i.e. mongodb client.

I'm thinking another approach could be using the "config/bootstrap.js" file to define a one-off [script that runs during the app startup](https://sailsjs.com/documentation/reference/configuration/sails-config-bootstrap). In there we could seed/provision the db with an entry. Say we create a new model "DbStatus" with no attributes (will have the defaults). Then we can create a new entry for that model (say we check if none existed before), which we could query from this `helper/get-api-health.js` file.  What do you think?

itsalaidbacklife: Thanks for the deep dive. So far as I can tell, there's no adapter-agnostic way to interact with the connection directly which will prevent us from testing this properly. That said, because it's fundamentally a dev experience issue rather than user experience, I'm comfortable skipping the tests for the failure case, since they will likely be a pain to configure to work across environments because we use postgresql in production but a mocked in-memory database (sails-disk) during local development.

So I say we take your suggestion above regarding the error handling (catching errors and returning false) and test only the positive case since that's what we have access to. If we figure out a consistent way to tap more directly into the db connection in a cross-environment manner, awesome, and if not then testing this can wait until a larger scale rework of models and dependencies. Great work @lbalmaceda!

seriouslysean: Yeah! This is my take on the issue, too. Thanks for all the information! LGTM.

package.json
```diff
@@ -79,6 +79,7 @@
     "prettier": "2.7.1",
     "sass": "1.32",
     "sass-loader": "8.0.0",
+    "sinon": "^14.0.1",
```
itsalaidbacklife, line 82: This looks like it's stale from the previous exploration of mocking 
```suggestion
```

tests/e2e/specs/out-of-game/health.spec.js
```diff
@@ -0,0 +1,38 @@
...
+    });
+  });
+
+  it.skip('is not available when db throws any error', () => {
+    stub = sinon.stub(Game, "count").throws();
+    cy.request('/health').then((res) => {
+      expect(res).property('status').to.equal(200);
+      expect(res.body).property('alive').to.eq(false);
+    });
+  });
```
itsalaidbacklife, line 30: This didn't work right? I'd expect an issue because Game won't be globally available outside of backend code. Let's remove it and Sinon altogether since we don't currently have a working use case
```suggestion
```

tests/e2e/specs/out-of-game/health.spec.js
```diff
@@ -0,0 +1,38 @@
+const { version: pkgVersion } = require('../../../../package.json');
+const sinon = require("sinon")
+const Game = require('../../../../api/models/Game')
```
itsalaidbacklife, line 3: Looks like the stubbing was a bust for now so let's remove it
```suggestion
```

tests/e2e/specs/out-of-game/health.spec.js
```diff
@@ -0,0 +1,38 @@
...
+let stub;
+describe('Health Page', () => {
+  beforeEach(() => {
+    cy.visit('/');
+  });
+
+  afterEach(() => {
+    if (stub) {
+      stub.restore();
+    }
```
itsalaidbacklife, line 14:
```suggestion
```

itsalaidbacklife (approved):
Since the mocking stuff didn't work let's remove it--and then for real we're good to merge

tests/e2e/specs/out-of-game/health.spec.js
lbalmaceda: Removed all, kept the try/catch update.

Issue #155, opened 2022-10-12
[Feature]: Leaderboard should break ties in rank based on number of wins


itsalaidbacklife:
### Feature Summary

This is a revision to the Stats page leaderboard. When two players have the same number of points in a given season, the player with more total wins for that season should be given the highest rank. Only if the totals of points and wins are equal should the players be tied for the same rank.

### Detailed Description

In the example below, seriouslysean should be given 8th place because they have fewer total wins (3) compared with ben@yarm.is
![image](https://user-images.githubusercontent.com/6520704/195384919-f7dc4463-b4b4-42b8-aeae-4f6dbf2f9304.png)

This change would be made in the `StatsLeaderboard.vue` component and its `rank()` function to account for the win totals in addition to the point totals.

billy-le:
I'm interested in working on this if no one else is assigned. Thanks!

itsalaidbacklife:
Awesome! Let me know if there's anything I can do to help you set up or to approach this issue. Of course feel free to ask questions in this issue thread, and if you'd like more real-time help/chat, you can use the #code channel of our [discord](https://discord.gg/6vQCFa6nvP) to talk to me or the other devs any time (I'm @aleph_one there)

billy-le:
hi @itsalaidbacklife , is there a way to seed the db with some data?

itsalaidbacklife:
Hi @billy-le, thanks again for checking this out.

Yes its definitely possible to seed data during localhost development. While booted in `development` mode (the default), you can make requests to `/test/loadSeasonFixture` and `/test/loadMatchFixtures` to seed records into those tables.

The easiest (and in my opinion best) way to do it is using test code. If you look at `tests/e2e/specs/out-of-game/stats.spec.js` you'll find the tests for the stats page, which are written using [cypress](https://docs.cypress. io). These tests already import fixture data from `tests/e2e/fixtures/statsFixtures.js` and use a custom cypress command to request those endpoints and seed that data into the database beforeEach test.

I think the best way to go will be to update that `statsFixtures` file's match data so that it it includes a tie in points that's broken by wins ie one player wins week 1 and another wins week 2 but have one or the other player win their week with more match wins than the other.

You can run the tests by booting the server and client and then running `npm run e2e:gui` in a separate terminal. That will open the cypress ui and then you can click the Stats test file to run its tests, which will seed the season and match data then make assertions about the stats page UI

billy-le:
[...]

PR #164, opened 2022-10-15
feature/tie-breaker

billy-le:
This is a fix for Issue #155 

added tie breaker when two players scores are equal

## Please check the following

For New Features:

test have been updated to reflect matches fixture changes


client/js/components/StatsLeaderboard.vue
```diff
@@ -289,9 +292,25 @@ export default {
...
+     * all scores sorted - [5, 5, 4]
+     * all wins sorted - [5, 4, 2]
+     *
+     * player 2 will be ranked 1st, player 1 will be ranked 2nd, player 3 will be ranked 3rd
+     */
+    rank({ totalScore, totalWins }) {
+      const scoreIndex = this.totalScoresSorted.indexOf(totalScore);
+      const winsIndex = this.totalWinsSorted.indexOf(totalWins);
+      const ranking = Math.max(scoreIndex, winsIndex) + 1;
+      return ranking;
```
itsalaidbacklife, line 313: So I think this addresses the change requested (breaking ties) but if I understand it correctly it makes other logical changes we want to avoid. Here are the intended requirements:

- Rankings should be determined first by points then by wins (total wins only matter to break a tie in points). Your change weighs both equally, but if one player gets 100 wins on week 1 and then never wins again, that should put them behind someone who took second place in that first week and then got first place every week after, even if the total wins are < 100 for that second player.

- If two players have the same number of total points, the tie should then be broken by the player with more total wins. If both players have the same total points and total wins, then they should have the same rank. So if two players tie for 4th place, they should both have a rank of 4 and the next player should have a rank of 6. This implementation would (I believe) give one of the players a rank of 5 somewhat randomly based on initial sort order

itsalaidbacklife (changes requested):
Right idea but we need to change to ranking logic further. See my comment for the requirements

client/js/components/StatsLeaderboard.vue
billy-le: Assuming that players scores will always be calculated correctly based on their winnings, this should theoretically work fine if a player's winning is 100 the first, that would mean their total points would also be very high as well.

I think I can make some improvements on how the data is mapped over, however. Let me provide another iteration on this. Thanks again for your input.

itsalaidbacklife: I think there is a miscommunication about how scoring works which is driving the misalignment. Score is determined each week according to who gets the most wins that week:

1. 5 points for the player who gets the most wins that week
2. 4 points for the player who gets the 2nd-most wins that week
3. 3 points for the player who gets the 3rd-most wins that week
4. 1 point to all players who got at least one win but didn't place 1st-3rd

The total number of points (which determines rankings) is the sum of the points you've earned each week. That means that if a player wins 100 matches in the first week, they earn a maximum of 5 points for that week, whereas someone who wins 3 matches every week could also get 5 points per week, if they had the most wins on that week (even if it was less than 100)

itsalaidbacklife: So the top-ranked player is the player who had the highest score from each week, not the one with the most total wins

billy-le: Okay cool, cool. Thanks for clarifying that. It makes much more sense now. I'll add that part to the README.md if it's not already there. Maybe we can break up the "How to Play" in a separate doc?

itsalaidbacklife: > Okay cool, cool. Thanks for clarifying that. It makes much more sense now. I'll add that part to the README.md if it's not already there. Maybe we can break up the "How to Play" in a separate doc?

As in separate from the rest of the README? I'm open to that

billy-le: Yes, I'll add an issue for that.

billy-le:
Yes, I can. I've just added it.

itsalaidbacklife (approved):
Perfect. Thanks so much for your help! This is actually just in time for our 3rd season championship, so it's timely and helpful for clarifying the tournament seeds. Really appreciate your engagement and our back and forth about how this should work, along with the README improvements.

If you'd like to hang out and play, please feel free to join [our discord](https://discord.gg/zZujC7Tp5j) (I'm aleph_one there), and if you're interested in contributing further, I would be more than happy to help you pick up another issue at any level of complexity or technical discipline that catches your eye.

PR #169, opened 2022-10-16
Add rate-limit middleware

lbalmaceda:
## Please check the following

- For New Features:

We briefly talked about rate-limits for the API in another issue that could impact the db, and the solution seemed relatively plug and play. I'm doing that here. I realized if we override the "order" array we need to keep the default values in place (these are pointed out in the existing comments of the file).

I'm checking the `isProd` variable to tell the limit of requests per minute, for the test env is 1000 and for prod is 30. I'm not sure how many request a single IP makes per minute but this can be tweaked. I originally put 200 for test but then realized locally it would return 429 errors fast.

I've not added tests but I'm thinking we could somehow set the env variable to be "prod" and burst more than 30 requests and expect a 429 at some point. Have you done a test like this in the past, where you change the prod variable before lifting the sails networking stack?

### DIFF
--- a/config/http.js
+++ b/config/http.js
@@ -20,6 +20,25 @@ module.exports.http = {
    ****************************************************************************/
 
   middleware: {
+    order: [
+      'cookieParser',
+      'session',
+      'bodyParser',
+      'compress',
+      'poweredBy',
+      'rateLimit',
+      'router',
+      'www',
+      'favicon',
+    ],
+    rateLimit: (function () {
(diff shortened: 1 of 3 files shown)


seriouslysean (changes requested):
While I like this in spirit, I don't think it's something we need to worry about in the short term. I will say I very much appreciate you putting in the work looking in to it and seeing it if was even worth doing. Thank you so much! 🙌🏼

Issue #163, opened 2022-10-14
[Feature]: Filter future seasons out from the data sent to the stats page


itsalaidbacklife:
### Feature Summary

Currently the stats page displays all the available Seasons that are listed in the database. We should the backend endpoint that services the stats page to filter out the future seasons (start time is after current time).

### Detailed Description

Currently if seasons are added to the database (e.g. Spades 2023) for a future start time, they will be displayed in the Stats page, and the latest-starting one will be pre-selected in the dropdown at the top of the page like so:
![image](https://user-images.githubusercontent.com/6520704/195924595-fcd8db58-58de-432e-b6e1-812dd3532537.png)

This is unideal from a maintenance standpoint because it means we need to wait until the end of the previous season to add the new one to the database. It would be much more efficient to have the backend endpoint that generates the stats payload to filter out the seasons that haven't started yet. That way we could add future seasons in one big batch without worrying about waiting until just before they start.

To do this we should update `api/helpers/get-seasons-without-rankings.js` which is a helper used to fetch the seasons. In that function, the query for all the seasons should receive an additional `where` clause to filter down to the seasons that are only have a start time that is less than the current time. See line 7 of that function:
```
const seasons = await Season.find({ sort: 'startTime DESC' }).populateAll(); // this should update to set the `where` inside the .find()
```
The where clause can be set according to the [sails docs on querying](https://sailsjs.com/documentation/concepts/models-and-orm/query-language) and the current time should be calculated using `dayjs().valueOf()` ([docs](https://day.js.org/docs/en/display/unix-timestamp-milliseconds)) which should compute the current time in milliseconds

P-DR0ZD:
Can I work on this @itsalaidbacklife

itsalaidbacklife:
@P-DR0ZD Yes please do! Please feel free to ask any questions you have here or in discord

PR #209, opened 2022-11-11
Added where condition to get-seasons-without-rankings.js so seasons that haven't started yet aren't shown to Fix issue #163

P-DR0ZD:
## Please check the following

- For New Features:

Added where condition so the user can't see seasons that haven't started yet from issue #163


itsalaidbacklife:
Looks good! Let's update the tests to check that this works as well. Can you update tests/e2e/fixtures/statsFixture.js to add an additional season that has a start time in the future (using dayjs to dynamically set the start time to the future)

P-DR0ZD:
I added the test you were suggesting I also added a test inside of stats.spec.js to check if the future date isn't being shown and changed one of the tests in stats.spec.js because it was falling because it's start date was a date in the future.

itsalaidbacklife:
@P-DR0ZD Looks like there is a failing test case in the stats file: `1) Selects different seasons to show their results`. This is likely due the changes, so it will warrant fixing. We can discuss here or on discord if it gives you trouble. Are you able to replicate the failure locally?

I'll add that there were additional tests failing in `reconnecting.spec.js` but that they are unrelated to your PR. I am fixing the reconnecting tests in #211 so updating from main should resolve that problem, though we still need to sort out the stats page test.

P-DR0ZD:
Yea I should be able to fix that later today

P-DR0ZD:
I think this fixed the issue. At least it did on my end

itsalaidbacklife:
Resolves #163

itsalaidbacklife (approved):
Looks good! Thanks for the contribution. This will save me time

PR #213, opened 2022-11-13
Add class to fix card width for resolving seven

### LINKED ISSUE #179: Jacks and Deck Cards (during 7) are small in Firefox
### Bug Summary

We recently switched the card images to .svg’s which need an explicit size set. It is likely that the jacks and the deck cards during 7 resolution are missing that explicit size

### Detailed Description

Cared images of jacks and of the cards in the deck display very small or even I fail to display on Firefox 

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Steps to Reproduce

1. Open the app in Firefox 
2. Play a jack or a 7
3. See the cards fail to load or load too small

thomasjvu:
## Please check the following

- For New Features:

Adds class to fix card width when resolving seven. Fixes #179

### DIFF
--- a/client/js/views/GameView.vue
+++ b/client/js/views/GameView.vue
@@ -150,7 +150,7 @@
                   :rank="topCard.rank"
                   :data-top-card="`${topCard.rank}-${topCard.suit}`"
                   :is-selected="topCardIsSelected"
-                  class="mb-4"
+                  class="mb-4 resolving-seven-size"
                   @click="selectTopCard"
                 />
                 <card
@@ -159,7 +159,7 @@
                   :rank="secondCard.rank"
                   :data-second-card="`${secondCard.rank}-${secondCard.suit}`"
                   :is-selected="secondCardIsSelected"
-                  class="mb-4"
+                  class="mb-4 resolving-seven-size"
                   @click="selectSecondCard"
                 />
               </div>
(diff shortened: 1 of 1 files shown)


thomasjvu:
Seven resolution now looks good on Firefox with card images appearing at a normal size. Screenshot of result below.

image

client/js/views/GameView.vue
```diff
@@ -1191,6 +1191,10 @@ export default {
   grid-area: decks;
 }
 
+.resolving-seven-size {
+    width: 9.5rem;
+}
```
itsalaidbacklife, line 1196: This looks good--let's move it down inside the block for `.reveal-top-two {}` block and add a & in front of the selector like 
```suggestion
& .resolving-seven-card {
    width: 9.5rem;
}
```
The & is an alias for the nesting, so that means it'd apply to something with the `.resolving-seven-card` class that's inside the `.reveal-top-two` element.

While we are at it, let's rename the class to resolving-seven-card instead of -size so that we can use the same selector in case we need to apply other styles to these cards besides size

itsalaidbacklife (changes requested):
Looks good. Let's change the class name and move the css 8 lines down into the block describing its parent

thomasjvu:
Hi, I just pushed a new commit onto my fork's main branch that accounts for the changes of:

- renaming the class to .resolving-seven-card
- moving the css into the parent .reveal-top-two while also adding the & nesting alias

I double checked to make sure that everything still looks okay, which it does. Screenshot below. 

image

itsalaidbacklife (approved):
Looks great! Thanks for the fix

PR #227, opened 2022-11-23
Feat/scuttle animation

itsalaidbacklife:
## Please check the following

- For New Features:

### DIFF
--- a/api/controllers/game/scuttle.js
+++ b/api/controllers/game/scuttle.js
@@ -80,6 +80,7 @@ module.exports = function (req, res) {
           victory,
           playedCardId: req.body.cardId,
           targetCardId: req.body.targetId,
+          playedBy: req.session.pNum,
         },
       });
       return res.ok();
--- a/api/controllers/game/scuttle.js
+++ b/api/controllers/game/scuttle.js
@@ -78,6 +78,8 @@ module.exports = function (req, res) {
           change: 'scuttle',
           game: fullGame,
           victory,
+          playedCardId: req.body.cardId,
+          targetCardId: req.body.targetId,
         },
       });
       return res.ok();
--- a/client/js/components/GameView/Card.vue
+++ b/client/js/components/GameView/Card.vue
@@ -21,6 +21,11 @@
       class="valid-move target-overlay"
       opacity=".8"
     />
+    <transition name="slide-above">
+      <v-overlay v-if="scuttledBy" absolute :value="true" class="scuttled-by-overlay">
+        <card :suit="scuttledBy.suit" :rank="scuttledBy.rank"></card>
+      </v-overlay>
+    </transition>
     <img
       v-if="isGlasses"
       :src="require(`../../img/cards/Glasses_${suitName}.png`)"
@@ -64,6 +69,17 @@ export default {
     isFrozen: {
       type: Boolean,
       default: false,
+    },
+    scuttledBy: {
+      type: Object,
+      default: null,
+      validator: (card) => {
+        return (
+          Object.hasOwnProperty.call(card, 'suit') &&
+          Object.hasOwnProperty.call(card, 'rank') &&
+          Object.hasOwnProperty.call(card, 'id')
+        );
+      },
     },
   },
   computed: {
@@ -196,6 +212,29 @@ export default {
   &:hover:after {
     opacity: 0;
   }
+}
+
+.scuttled-by-overlay {
+  height: 80%;
+  top: -32px;
+  transition: all 1s ease;
+}
+.slide-below-leave-active,
+.slide-above-leave-active,
+.in-below-out-left-leave-active {
+  position: absolute;
+}
+// slide-below (enter and leave below)
+.slide-below-enter,
+.slide-below-leave-to {
+  opacity: 0;
+  transform: translateY(32px);
+}
+// slide-above (enter and leave above)
+.slide-above-enter,
+.slide-above-leave-to {
+  opacity: 0;
+  transform: translateY(-32px);
 }
 
 @media (max-width: 600px) {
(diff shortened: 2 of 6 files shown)


api/controllers/game/scuttle.js
```diff
@@ -78,6 +78,9 @@ module.exports = function (req, res) {
           change: 'scuttle',
           game: fullGame,
           victory,
+          playedCardId: req.body.cardId,
+          targetCardId: req.body.targetId,
+          playedBy: req.session.pNum,
```
itsalaidbacklife, line 83: Need to send this data to know which cards to move

client/js/components/GameView/Card.vue
```diff
@@ -65,6 +70,17 @@ export default {
...
+      type: Object,
+      default: null,
+      validator: (card) => {
+        return (
+          Object.hasOwnProperty.call(card, 'suit') &&
+          Object.hasOwnProperty.call(card, 'rank') &&
+          Object.hasOwnProperty.call(card, 'id')
+        );
+      },
+    },
```
itsalaidbacklife, line 83: Probably silly to include this validator here and nowhere es\

client/js/components/GameView/Card.vue
```diff
@@ -198,6 +214,29 @@ export default {
...
+.slide-below-leave-to {
+  opacity: 0;
+  transform: translateY(32px);
+}
+// slide-above (enter and leave above)
+.slide-above-enter,
+.slide-above-leave-to {
+  opacity: 0;
+  transform: translateY(-32px);
+}
```
itsalaidbacklife, line 238: These are copied from GameView. It would be better I think to pull these into scss mixins we can include in both places. Okay for the moment imo

client/js/components/GameView/Card.vue
```diff
@@ -21,6 +21,11 @@
       class="valid-move target-overlay"
       opacity=".8"
     />
+    <transition name="slide-above">
```
seriouslysean, line 24: Do you know if the transition needs to be v-ifed or if if automatically goes away if there's nothing to render? I don't remember and want to make sure we're following convention here. Also seems weird this is using card but the other card below is using an img.

client/js/components/GameView/Card.vue
seriouslysean: File a follow up :), I agree we should dry this up. If it's easy to include might be worth doing here and not rushing this feature out the door. Do we have anywhere we store global mixins that are shared?

client/js/store/modules/game.js
```diff
@@ -269,6 +269,41 @@ export default {
...
+      }
+      // Remove played card from opponent hand and temporarily add to its targets attachments
+      const playedCardIndex = context.getters.opponent.hand.findIndex(
+        (card) => card.id === playedCardId
+      );
+      const targetCardIndex = context.getters.player.points.findIndex(
+        (card) => card.id === targetCardId
+      );
+
+      // Update game in one-step if mvoed cards are not found
```
seriouslysean, line 292:
```suggestion
      // Update game in one-step if moved cards are not found
```

client/js/store/modules/game.js
```diff
@@ -269,6 +269,41 @@ export default {
...
+      }
+
+      const [playedCard] = context.getters.opponent.hand.splice(playedCardIndex, 1);
+      const targetCard = context.getters.player.points[targetCardIndex];
+      targetCard.scuttledBy = playedCard;
+
+      // Finish complete update of the game state after 1s
+      setTimeout(() => {
+        context.dispatch('updateGameThenResetPNumIfNull', game);
+      }, 1000);
```
seriouslysean, line 305: Maybe it's just me but this whole function feels complicated. Are we not able to just track those new ids and use coordinates to animate a single absolute positioned element across the gameboard or does it have to be attached to use the transitions?

seriouslysean (approved):
Approved with quite a few questions.

client/js/components/GameView/Card.vue
itsalaidbacklife: Currently we have no scss mixins or reused css. The closes atm is global styles that are applied in the unscoped App.vue style block. This would likely be the first reused scss stuff so I'm thinking new issue

client/js/store/modules/game.js
itsalaidbacklife: Agreed it feels complicated and generally smells. That said imo a coordinate-based system would likely be an overall improvement but a more complicated one. I'm definitely in favor of re-examining the whole transition system accordingly, so I'm thinking separate issue.

That said the vast majority of the complexity here is due to the 2-stage nature of the intended animation itself:
1) the opponent's point card fades out downwards while the scuttling card simultaneously fades in from above on top of your point card, THEN
2) your point card with the scuttling card fades out to the left (towards scrap)

I'm open to suggestions for how to simplify. Would love to iscuss. My thought is it warrants a holistic second pass after this PR

client/js/components/GameView/Card.vue
itsalaidbacklife: An interesting point and I don't know. I'll do some digging

client/js/components/GameView/Card.vue
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/232

client/js/store/modules/game.js
itsalaidbacklife: Wrote up https://github.com/cuttle-cards/cuttle/issues/233 to track the issue of replacing the transition/animation system with an absolute coordinate one. I think doing so will be complicated and probably require a senior engineer, but it would be a significant improvement

client/js/components/GameView/Card.vue
itsalaidbacklife: Yeah the docs say to use transitions wrapped around elements which themselves have v-if: https://vuejs.org/guide/built-ins/transition.html#the-transition-component

PR #246, opened 2022-12-05
Feat/176 game history css update

jerryh-jr:
## Please check the following

- For New Features:


client/js/views/GameView.vue
```diff
@@ -1253,30 +1249,68 @@ export default {
...
     width: 80%;
     max-width: 80%;
     height: 100%;
+    border-radius: 20px;
+    box-shadow: 0px 0px 10px #111111;
   }
   #history {
-    background-color: rgba(0, 0, 0, 0.46);
+    background-color: rgba(241, 200, 160, 0.65);
+    color: #111111;
```
peterancourt, line 1269: @itsalaidbacklife A note for a future ticket, we may want to start looking into storing some of these hex values as color variables in scss files to get a true design system going.

client/js/views/GameView.vue
```diff
@@ -1253,30 +1249,68 @@ export default {
...
     & #history-logs {
-      height: 80%;
       overflow: auto;
+      height: 85%;
+      font-size: 0.75em;
+      letter-spacing: 0.25px;
+      font-family: 'Libre Baskerville', serif;
```
peterancourt, line 1275: @itsalaidbacklife same with the fonts

client/js/views/GameView.vue
```diff
@@ -1301,7 +1301,7 @@ export default {
         width: 100%;
         margin: 0 auto;
         overflow: auto;
-        font-size: 1.15em;
+        font-size: 1em;
```
jerryh-jr, line 1304: Lowered font to better fit the log. 
image

client/js/views/GameView.vue
```diff
@@ -1269,6 +1269,7 @@ export default {
     color: #111111;
     & #history-logs {
       overflow: auto;
+      overflow-wrap: anywhere;
```
jerryh-jr, line 1272: Resolved issue with overflowing text. 
image
image
image
image

jerryh-jr: @peterancourt what do you think?

client/js/views/GameView.vue
```diff
@@ -1253,30 +1249,69 @@ export default {
...
-    flex-direction: column;
-    justify-content: center;
-    align-items: center;
-    border: 1px solid #fff;
+    padding: 0.5rem;
     width: 80%;
     max-width: 80%;
     height: 100%;
+    border-radius: 20px;
+    box-shadow: 0px 0px 10px #111111;
```
itsalaidbacklife, line 1265: It would probably be more consistent to give this an elevation class rather than a custom box shadow

itsalaidbacklife: Check the Vuetify docs: https://vuetifyjs.com/en/styles/elevation/

Elevation is a material design concept of using specific box shadows to suggest depth and clarify what's on top of what. Since were generally using material design it'd make sense to use an elevation shadow that'd be more consistent with the rest of the app. Vuetify provides styles that we can apply using classes like `class="elevation-10"` to make this easy

client/js/views/GameView.vue
itsalaidbacklife: Yeah the fonts are actually more troublesome than the colors as far as where to put them because we're not being very semantically consistent overall with them. I'll write up a font reorganization ticket

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cuttle",
-  "version": "3.3.1",
+  "version": "3.3.2",
```
itsalaidbacklife, line 3: Let's bump this to 3.3.5
```suggestion
  "version": "3.3.5",
```

itsalaidbacklife (approved):
This is looking great! I'm requesting a couple refactoring changes but the bones all look right

client/js/App.vue
```diff
@@ -18,7 +18,7 @@ export default {
 </script>
 
 <style lang="scss">
-@import url('https://fonts.googleapis.com/css2?family=PT+Serif:wght@400;700&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Cormorant+Infant:wght@400;600;700&family=Libre+Baskerville:wght@400;700&family=PT+Serif:wght@400;700&display=swap');
```
seriouslysean, line 21: Good job combining these, would love to see if we need all the weights for each of these fonts. A future improvement could be to only load them when we need them (gameview), but that's grasping at straws for our usages and a higher level of complexity for a minor improvement.

client/js/views/GameView.vue
```diff
@@ -268,13 +268,13 @@
       <!-- History -->
       <div v-if="$vuetify.breakpoint.smAndUp" class="history-container">
         <div id="field-right">
-          <div id="history" class="rounded d-flex flex-column justify-start">
-            <h3>History</h3>
```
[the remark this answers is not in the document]
seriouslysean: Would it be better to make a new scoped class for this? Something like `rounded-20` as a global class so you can use it anywhere? I'm not saying we should, but since we're using Vuetify, it feels more Vuetify-like. I'm fine leaving it though. Would love to move these sorts of global things out to CSS variables at some point so we can keep a manifest of logical defaults higher in the stack.

client/js/views/GameView.vue
seriouslysean: I agree with this callout. ☝🏼

client/js/views/GameView.vue
seriouslysean: @peterancourt spot on, I would love to get a CSS config file of all of our base cuttle values in a `:root` rule block. I'm torn on Vuetify, I love it but would love to move away from it for some cuttle specific stuff since we're talking about pulling back on our global usages.

client/js/views/GameView.vue
seriouslysean: Why not `break-word` instead of `anywhere`? `anywhere` is buck wild and will break a string literally anywhere to make it fit.

client/js/views/GameView.vue
seriouslysean: We need logical fallbacks for all these fonts, something like a css variable in a config:
```
--font-family-primary: 'Libre Baskerville', Georgia,Times,Times New Roman,serif;
```

and then later on, you can use it like this:
```
font-family: var(--font-family-primary);
```

seriouslysean (approved):
Overall looking good, few callouts though.

client/js/views/GameView.vue
```diff
@@ -268,13 +268,13 @@
...
+          <div id="history" class="d-flex flex-column justify-start align-center">
+            <h3 class="history-title">History</h3>
```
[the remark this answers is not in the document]
jerryh-jr: @itsalaidbacklife @seriouslysean added elevation class and removed border-shadow.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cuttle",
-  "version": "3.3.2",
+  "version": "3.3.5",
```
jerryh-jr, line 3: @itsalaidbacklife bumped package version

client/js/views/GameView.vue
```diff
@@ -1273,15 +1272,15 @@ export default {
       height: 85%;
       font-size: 0.75em;
       letter-spacing: 0.25px;
-      font-family: 'Libre Baskerville', serif;
+      font-family: 'Libre Baskerville', Century Gothic, CenturyGothic, AppleGothic, sans-serif;
```
jerryh-jr, line 1275: @seriouslysean added font fall backs.

client/js/views/GameView.vue
jerryh-jr: I added an elevation-10 and removed border-shadow from the element.

package.json
jerryh-jr: This change was made.

client/js/views/GameView.vue
jerryh-jr: Break-word isn't working in the flex container. From what I have read this is due to break-word not shrinking the min-content intrinsic size where as anywhere does shrink it. With break word it still overflows.

client/js/App.vue
jerryh-jr: Good call out, currently we are only using one weight for each font. Not sure if that would change in the future.

client/js/views/GameView.vue
jerryh-jr: Added font fall backs to these in the latest commit.

client/js/views/GameView.vue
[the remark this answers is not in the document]
jerryh-jr: Yeah I agree and think it would be a great candidate for a variable.

client/js/views/GameView.vue
jerryh-jr: @seriouslysean from what I can gather. break-word will calculate the min content size for width as if no words have been broken. Where Anywhere will calculate the min-content size with the soft breaks included.

From what I have been gathering it looks like on the flex container we could set min-width: 0; overriding the default auto and that should allow us to use break-word and have it work.

itsalaidbacklife: @seriouslysean if the main issue is long usernames then won't break word not affect them?

seriouslysean: I didn't consider that as I hadn't tested it locally. I don't think we should be breaking usernames but I guess we kind of have to.

Maybe we eventually move to "Player/Opponent labels and some other way to show which user?

itsalaidbacklife: @seriouslysean I agree. Necessary evil for the time being and some kind of tokenizing for player/opponent would probably be best in the long run

client/js/views/GameView.vue
itsalaidbacklife: Don't we need to import these other fonts? Also what's the difference between Century Gothic and CenturyGothic? If we need the version with spaces we should put it in quotes like `'Century Gothic'`

jerryh-jr: @itsalaidbacklife these should all be Web Safe system fonts I got from here https://www.cssfontstack.com/Century-Gothic.

Since these are system fonts I do not believe we need to import them if we do please let me know and I will look into importing them. 

I can update the font to 'Century Gothic'.

itsalaidbacklife (approved):
Looks great! What a nice look and feel boost

PR #250, opened 2022-12-07
Created a Basic Design for a My Profile Page

P-DR0ZD:
## Please check the following

- For New Features:

I added a basic design for a My Profile page. It has a text field where you can input an email and a success and error snack bar based on if the email is Valid.

## Whats Left

- I want some extra input on the design Side of things

- I also want some help with the backend Things. I was able to update the User.js model with an email Field but I need help creating an API to access this field


client/js/router/index.js
```diff
@@ -70,6 +71,12 @@ const routes = [
     component: Stats,
     beforeEnter: mustBeAuthenticated,
   },
+  {
+    path: '/myprofile',
```
itsalaidbacklife, line 75:
```suggestion
    path: '/profile',
```

client/js/views/MyProfile.vue
```diff
@@ -0,0 +1,86 @@
...
+    };
+  },
+  computed: {},
+  mounted() {},
+  methods: {
+    async submitEmail() {
+      try {
+        //Email Testing Regex
+        if (/^\w+([.-]?\w+)*@\w+([.-]?\w+)*(\.\w{2,3})+$/.test(this.email)) {
+          //ToDo: Figure how to save email to Database
```
itsalaidbacklife, line 46: Great start. We'll need this to work full stack before we merge&deploy so we don't ship a broken feature.

The gist is we need a new backend endpoint, defined in `api/controllers/UserController.js` which we could call submitEmail() that would expect that the user sends an email in the request body and if so it would call User.updateOne() to update the correct user and add their email. We can discuss the details outside GitHub but the other key pieces will be 

adding a backend route in config/routes.js to map the new endpoint to

adding a new store action in the auth store module that fires a request to this new endpoint and returns the result as a promise. That'll let this page dispatch the store action to make the request and handle success/error based on that request

client/js/views/MyProfile.vue
```diff
@@ -0,0 +1,86 @@
...
+      colorValue: '',
+    };
+  },
+  computed: {},
+  mounted() {},
+  methods: {
+    async submitEmail() {
+      try {
+        //Email Testing Regex
+        if (/^\w+([.-]?\w+)*@\w+([.-]?\w+)*(\.\w{2,3})+$/.test(this.email)) {
```
itsalaidbacklife, line 45: Let's rewrite the validity check into a computed property on the stored input value and then disable submission with an error message on the input when the email is invalid

client/js/views/MyProfile.vue
```diff
@@ -0,0 +1,86 @@
...
+    <!-- <h3>Current Email: test</h3> -->
+  </v-container>
+</template>
+<script>
+export default {
+  name: 'MyProfile',
+  data() {
+    return {
+      user: this.$store.state.auth.username,
+      email: '',
```
itsalaidbacklife, line 33: Once this is hooked up in the full stack we'll want to initialize this value from the user's existing email in case it's already submitted

itsalaidbacklife (changes requested):
Off to a good start. The biggest lift left will be hooking up the backend. We can discuss here further or on discord as necessary

api/controllers/UserController.js
```diff
@@ -90,6 +90,32 @@ module.exports = {
     return res.ok();
   },
 
+  submitEmail: async function (req, res) {
+    try {
+      const { username, email } = req.body;
+      const updatedUser = await User.updateOne({ username: username }).set({ email: email });
```
itsalaidbacklife, line 96: This makes sense except that we should also verify that the requesting user is the one specified by the username. Otherwise anyone could request to change anyone else's email with an ajax request. This boils down to looking up the user based on `req.session.usr` and then verifying that the user you get back has the same username as the requesting one
```suggestion
      const requestingUser = await User.findOne({ id: req.session.usr });
      if (requestingUser.username !== username) {
        return res.forbidden({message: 'You can only change your own email address'});
      }
      const updatedUser = await User.updateOne({ username: username }).set({ email: email });
```
A couple notes on how this works
1. sails exposes `req.session` as a readable and writeable object that is unique to the browser window making the request
2. When you log in, the application sets the 'usr' attribute of the session to your user id. This is how we know who you are.
3. We should additionally use policies to ensure that anyone making this request is logged in and has provided a valid email address

itsalaidbacklife: Here's the sails [documentation on policies](https://sailsjs.com/documentation/concepts/policies#policies). The basic idea is that a policy is a middleware function that runs before a controller action does, usually to validate whether it's allowed or not. They have essentially 2 parts:

1. We'll define a new policy in `api/policies/hasValidEmail.js` which could use the same regex you're using on the frontend to validate that `req.body` has an `email` attribute and that the email and return a `res.badRequest()` if its missing or invalid
2. Then update `config/policies.js` to map this new endpoint to two policies: `'isLoggedIn'` and the new `'hasValidEmail'`
That way without complicating the controller logic itself we can ensure that this action only runs if the user is logged in and if they've provided a valid email

client/js/store/modules/auth.js
```diff
@@ -59,7 +63,49 @@ export default {
     async requestSignup(context, { username, password }) {
       return handleLogin(context, username, password, true);
     },
-
+    async submitEmail(context, { username, email }) {
+      try {
+        const response = await fetch(`/user/submitEmail`, {
```
itsalaidbacklife, line 68: pref single quotes over backticks unless you're using variables
```suggestion
        const response = await fetch('/user/submitEmail', {
```

config/policies.js
```diff
@@ -30,6 +30,8 @@ module.exports.policies = {
     reLogin: ['hasValidUsername'],
     logout: true,
     status: true,
+    submitEmail: ['hasValidUsername'],
```
itsalaidbacklife, line 33: Let's create a different `hasValidEmail` policy (because emails have different validation than usernames) and use both that and the existing `isLoggedIn`
```suggestion
    submitEmail: ['isLoggedIn', 'hasValidEmail'],
```

config/policies.js
```diff
@@ -30,6 +30,8 @@ module.exports.policies = {
...
+    findEmail: ['hasValidUsername'],
```
itsalaidbacklife, line 34: I think we can get rid of this endpoint altogether

client/js/store/modules/auth.js
```diff
@@ -59,7 +63,49 @@ export default {
...
+        });
+        const data = await response.json();
+        if (response.status !== 200) {
+          throw data.message;
+        }
+        context.commit('setEmail', data);
+      } catch (err) {
+        throw new Error(err);
+      }
+    },
```
itsalaidbacklife, line 108: Is this used to get the user email when loading the page? Isn't the `email` available on the User object itself in the `auth` module?

itsalaidbacklife: I think we can directly grab the `email` off of the `User` object when logging in, refreshing auth status etc and committing it to the store

itsalaidbacklife (commented):
I'm having difficulty getting it to accept an email submission both in terms of accepting the regex and then processing the request which gives a not found error I think. Let's discuss

P-DR0ZD:
> I'm having difficulty getting it to accept an email submission both in terms of accepting the regex and then processing the request which gives a not found error I think. Let's discuss

It's working from my end that's weird anything that i should look at

itsalaidbacklife:
@P-DR0ZD I'm closing this for now because there's more thinking to do about avatars and the function of this page, but it has all the bones of what we'll need when we re-open to create this page. Great work!

PR #115, opened 2022-09-27
Feature: Vue 3 (with Vite)

seriouslysean:
# 🎉 THIS IS REALLY BIG, I'M SORRY 🎉 

###### _(I literally can't break it up anymore)_

I did get it partially working, so much to do still though.

image

PR #259, opened 2022-12-29
chore: install and enable eslint-plugin-vuetify

KaelWD:
For #115, should help with updating vuetify components

Rules used: https://github.com/vuetifyjs/eslint-plugin-vuetify#deprecations


itsalaidbacklife (approved):
Thanks @KaelWD ! We really appreciate your help. You know as well as anyone what an undertaking a vue3 migration can be. We're just getting to the point where things *mostly* work enough to use our tests to assess % completion as we crank through the rest of it. This is timely and helpful

PR #277, opened 2023-01-04
[DevEx]: Enable Vuetify eslint errors and fix them across the app, fixes #275

### LINKED ISSUE #275: [DevEx]: Enable Vuetify eslint errors and fix them across the app
### Improvement Summary

The vuetify team ha an eslint plugin that throws errors for use of deprecated props. We should enable it and fix all the resulting errors as a part of our vue3 migration

### Detailed Description

The plugin is already installed; we just need to enable it and fix the resulting errors. To enable it, uncomment this line in `.eslintrc.js`
image

Then run `npm run lint` to see the resulting errors. At that point we can work through the errors file by file to get our vuetify components up to date

seriouslysean:
## Please check the following

- For New Features:

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -23,9 +23,10 @@ module.exports = {
     'max-len': [
       'warn',
       {
-        code: 110,
+        code: 120,
         ignoreStrings: true,
         ignoreTemplateLiterals: true,
+        ignoreUrls: true,
       },
     ],
     'vue/html-indent': ['error'],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,7 +15,7 @@ module.exports = {
   extends: [
     'eslint:recommended',
     'plugin:vue/base',
-    // 'plugin:vuetify/base', // TODO: remove after migration
+    'plugin:vuetify/base',
   ],
   plugins: ['cypress', 'jest', 'prettier'],
   ignorePatterns: ['/node_modules/*', '/assets/*'],
@@ -68,6 +68,9 @@ module.exports = {
       globals: {
         _: true,
         sails: true,
+        gameService: true,
+        cardService: true,
+        //
         Card: true,
         Season: true,
         Match: true,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -68,8 +68,9 @@ module.exports = {
       globals: {
         _: true,
         sails: true,
-        gameService: true,
         cardService: true,
+        gameService: true,
+        userService: true,
         //
         Card: true,
         Season: true,
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,14 @@
+{
+	"version": "2.0.0",
+	"tasks": [
+		{
+			"type": "npm",
+			"script": "lint",
+			"problemMatcher": [
+				"$eslint-stylish"
+			],
+			"label": "npm: lint",
+			"detail": "eslint --ext .js,.vue --ignore-path .gitignore ."
+		}
+	]
+}
--- a/src/components/GameListItem.vue
+++ b/src/components/GameListItem.vue
@@ -2,7 +2,12 @@
   <div>
     <v-row class="list-item" data-cy="game-list-item">
       <v-col v-if="$vuetify.display.lgAndUp" sm="2" lg="3">
-        <v-img src="/img/logo-head.svg" class="my-1" cover :height="$vuetify.display.smAndUp ? 62 : 32" />
+        <v-img
+          src="/img/logo-head.svg"
+          class="my-1"
+          cover
+          :width="62"
+        />
       </v-col>
       <v-col sm="7" lg="6" class="list-item__inner-text">
         <p class="game-name" data-cy="game-list-item-name">
@@ -18,10 +23,9 @@
           min-width="200"
           :disabled="!status"
           :loading="joiningGame"
-          :small="!$vuetify.display.lg"
           @click="subscribeToGame"
         >
-          <v-icon v-if="isRanked" class="mr-4" size="medium">mdi-trophy</v-icon>
+          <v-icon v-if="isRanked" class="mr-4" size="medium" icon="mdi-trophy" />
           {{ buttonText }}
         </v-btn>
       </v-col>
--- a/src/components/GameListItem.vue
+++ b/src/components/GameListItem.vue
@@ -2,7 +2,7 @@
   <div>
     <v-row class="list-item" data-cy="game-list-item">
       <v-col v-if="$vuetify.display.lgAndUp" sm="2" lg="3">
-        <v-img src="/img/logo-head.svg" class="my-1" contain :height="$vuetify.display.smAndUp ? 62 : 32" />
+        <v-img src="/img/logo-head.svg" class="my-1" cover :height="$vuetify.display.smAndUp ? 62 : 32" />
       </v-col>
       <v-col sm="7" lg="6" class="list-item__inner-text">
         <p class="game-name" data-cy="game-list-item-name">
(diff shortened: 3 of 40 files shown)


src/components/StatsLeaderboard.vue
```diff
@@ -19,11 +19,10 @@
         multiple
       />
     </div>
-    <!-- https://vuetifyjs.com/en/api/v-data-table/ -->
-    <v-data-table
+    <!-- https://next.vuetifyjs.com/en/api/v-table/ -->
+    <v-table
```
seriouslysean, line 23: Literally none of these props will work, there's a whole different setup to do a normal table vs. data table. This is just temporary to fix the linting error.

seriouslysean (commented):
All changes were made to fix Vuetify without any regard to style or functionality. We'll need a full pass on the visuals to fix those issues as we get to a more working state.

src/views/GameView.vue
```diff
@@ -1443,8 +1441,8 @@ export default {
       'opp-score opp-score opp-score opp-score opp-score opp-score opp-score opp-score'
       'field field field field field field field field'
       'decks decks decks decks decks decks decks decks'
-      'player-score player-score player-score player-score player-score player-score player-score player-score'
-      'player-hand player-hand player-hand player-hand player-hand player-hand player-hand player-hand';
+      'score score score score score score score score'
+      'hand hand hand hand hand hand hand hand';
```
seriouslysean, line 1445: Renaming to meet `max-len` reqs.

src/views/HomeView.vue
```diff
@@ -39,29 +39,41 @@
               variant="outlined"
               color="primary"
               class="mt-4"
-              :small="$vuetify.display.mdAndDown ? true : false"
               to="/rules"
               data-cy="rules-link"
+              v-bind="{
```
seriouslysean, line 44: These were a bit trickier, we need to bind the prop since they're no longer booleans.

src/views/LobbyView.vue
```diff
@@ -55,7 +61,7 @@ import { mapGetters, mapState } from 'vuex';
 import LobbyPlayerIndicator from '@/components/LobbyPlayerIndicator.vue';
 
 export default {
-  name: 'Lobby',
+  name: 'LobbyView',
```
seriouslysean, line 64: Aligned all component names with the component naming guidelines. 2 words, and for Cuttle specifically we can use `${name}View` for single word components, and to denote a top level view component.

src/views/LoginView.vue
```diff
@@ -59,7 +59,7 @@
     </v-row>
 
     <v-row class="mt-0">
-      <v-col md="6" sm="12" class="d-flex justify-start flex-column mt-4">
+      <v-col :cols="$vuetify.display.mdAndUp ? 6 : 12" class="d-flex justify-start flex-column mt-4">
```
seriouslysean, line 62: cols instead of by size now.

src/components/GameView/Card.vue
```diff
@@ -12,11 +12,10 @@
     <v-icon v-if="isFrozen" class="player-card-icon mr-1 mt-1" color="#00a5ff" icon="mdi-snowflake" />
     <v-overlay
       v-ripple
-      :value="isValidTarget"
+      :model-value="isValidTarget"
       absolute
-      color="accent"
+      scrim="accent"
       class="valid-move target-overlay"
-      opacity=".8"
```
seriouslysean, line 19: Opacity is no longer supported, we'll need a design pass before we release.

.eslintrc.js
```diff
@@ -68,6 +69,10 @@ module.exports = {
       globals: {
         _: true,
         sails: true,
+        cardService: true,
+        gameService: true,
+        userService: true,
```
seriouslysean, line 74: More globals to prevent so many `eslint` errors.

.vscode/tasks.json
```diff
@@ -0,0 +1,14 @@
...
+	"tasks": [
+		{
+			"type": "npm",
+			"script": "lint",
+			"problemMatcher": [
+				"$eslint-stylish"
+			],
+			"label": "npm: lint",
+			"detail": "eslint --ext .js,.vue --ignore-path .gitignore ."
+		}
```
seriouslysean, line 12: This will let us do `> Run lint` from VSCode and show errors from all files in the problems pane which is really nice.

src/components/GameListItem.vue
```diff
@@ -2,7 +2,12 @@
   <div>
     <v-row class="list-item" data-cy="game-list-item">
       <v-col v-if="$vuetify.display.lgAndUp" sm="2" lg="3">
-        <v-img src="/img/logo-head.svg" class="my-1" contain :height="$vuetify.display.smAndUp ? 62 : 32" />
+        <v-img
+          src="/img/logo-head.svg"
+          class="my-1"
+          cover
+          :width="62"
```
seriouslysean, line 9: `height` is no longer supported, so moved these sorts of things to width. Eventually we should move these to CSS imo.

src/components/GameListItem.vue
```diff
@@ -14,14 +19,13 @@
...
-          outlined
+          variant="outlined"
           min-width="200"
           :disabled="!status"
           :loading="joiningGame"
-          :small="!$vuetify.display.lg"
           @click="subscribeToGame"
         >
-          <v-icon v-if="isRanked" class="mr-4" medium>mdi-trophy</v-icon>
+          <v-icon v-if="isRanked" class="mr-4" size="medium" icon="mdi-trophy" />
```
seriouslysean, line 28: Moved a few more icons inline.

seriouslysean (commented):
Few callouts, nothing too major.

src/components/GameListItem.vue
itsalaidbacklife: agreed

src/views/About.vue
```diff
@@ -1,5 +0,0 @@
-<template>
-  <div class="about">
-    <h1>This is an about page</h1>
-  </div>
-</template>
```
itsalaidbacklife, line 5: nooooo the about page

src/views/GameView.vue
itsalaidbacklife: ngl feels like the linter is in the way here. This is fine for now but given the way grid css works it's annoying to be constrained by the max-len here

src/views/HomeView.vue
itsalaidbacklife: Yeah funky one. What do you think of swapping the ternary into the value instead of the whole object?

src/views/HomeView.vue
```diff
@@ -39,29 +39,41 @@
...
+                ...($vuetify.display.mdAndDown ? {
+                  size: 'small',
+                } : {})
+              }"
```
itsalaidbacklife, line 48: Like
```suggestion
              v-bind="{
                  size: $vuetify.display.mdAndDown ? 'small' : 'medium',
                } "
```

itsalaidbacklife: This is less flexible when it comes to potentially configuring multiple props based on the same size, but we're only doing one here and I think this might be more legible. Thoughts?

src/views/HomeView.vue
```diff
@@ -39,29 +39,41 @@
...
               class="mt-4"
-              :small="$vuetify.display.mdAndDown ? true : false"
               href="https://human-ai-interaction.github.io/cuttle-bot/"
               target="_blank"
               data-cy="ai-link"
+              v-bind="{
+                ...($vuetify.display.mdAndDown ? {
+                  size: 'small',
+                } : {})
+              }"
```
itsalaidbacklife, line 63:
```suggestion
              v-bind="{
                  size: $vuetify.display.mdAndDown ? 'small' : 'medium',
                } "
```

src/views/HomeView.vue
```diff
@@ -39,29 +39,41 @@
...
               variant="outlined"
               class="mt-4"
-              :small="$vuetify.display.mdAndDown"
               href="https://discord.gg/9vrAZ8xGyh"
               target="_blank"
+              v-bind="{
+                ...($vuetify.display.mdAndDown ? {
+                  size: 'small',
+                } : {})
+              }"
```
itsalaidbacklife, line 76: Seeing this 3 times in this file makes me think we want a computed prop for the v-bind object

itsalaidbacklife (approved):
Looks good. Got one suggestion regarding the `size` prop for buttons: putting it into a computed prop that uses the ternary for the object value rather than the whole object (since we only configure the `size` prop

src/views/GameView.vue
seriouslysean: I’m in agreement, but we also use opp-* for the other stuff. Maybe worth turning off line lengths in the css parts of SFCs.

PR #292, opened 2023-01-17
Feat: Fix home spec tests, resolves #268

### LINKED ISSUE #268: [DevEx]: Fix failing home.spec tests
### Improvement Summary

We should fix all failing automated tests

### Detailed Description

Vue3 updates have broken home.spec tests which needs to be examined and updated.

Failing test are marked `.skip`.

seriouslysean:
## Please check the following

- For New Features:

### DIFF
--- a/src/components/AwardCard.vue
+++ b/src/components/AwardCard.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-card class="d-flex align-center" :color="theme.neutral.lighten2" :data-tournament="`${place}${suffix}`">
+  <v-card class="d-flex align-center" :color="theme.neutral-lighten2" :data-tournament="`${place}${suffix}`">
     <div class="medal-icon pa-8 d-flex align-center">
       <v-icon size="48" :color="medalColor" icon="mdi-medal" />
     </div>
--- a/src/components/CreateGameDialog.vue
+++ b/src/components/CreateGameDialog.vue
@@ -12,13 +12,7 @@
       <v-card-text>
         <div class="d-flex align-center">
           <v-switch v-model="isRanked" label="Ranked" data-cy="create-game-ranked-switch" />
-          <stats-scoring-dialog :show-button-text="false" data-cy="ranked-info-button">
-            <template v-slot:button>
-              <v-btn color="primary" class="mb-2" variant="plain" data-cy="ranked-info-button">
-                <v-icon class="ml-1" icon="mdi-information-outline" />
-              </v-btn>
-            </template>
-          </stats-scoring-dialog>
+          <stats-scoring-dialog :show-button-text="false" />
         </div>
         <v-text-field
           v-model="gameName"
--- a/src/components/CreateGameDialog.vue
+++ b/src/components/CreateGameDialog.vue
@@ -12,7 +12,13 @@
       <v-card-text>
         <div class="d-flex align-center">
           <v-switch v-model="isRanked" label="Ranked" data-cy="create-game-ranked-switch" />
-          <stats-scoring-dialog :show-button-text="false" data-cy="ranked-info-button" />
+          <stats-scoring-dialog :show-button-text="false" data-cy="ranked-info-button">
+            <template v-slot:button>
+              <v-btn color="primary" class="mb-2" variant="plain" data-cy="ranked-info-button">
+                <v-icon class="ml-1" icon="mdi-information-outline" />
+              </v-btn>
+            </template>
+          </stats-scoring-dialog>
         </div>
         <v-text-field
           v-model="gameName"
--- a/src/components/StatsScoringDialog.vue
+++ b/src/components/StatsScoringDialog.vue
@@ -1,20 +1,13 @@
 <template>
   <v-dialog v-model="show">
     <!-- Activator -->
-    <template #activator="{ attrs, props }">
-      <v-btn
-        color="primary"
-        class="mb-2"
-        v-bind="{
-          ...$props,
-          ...props,
-          ...$attrs,
-          ...attrs,
-        }"
-        variant="plain">
-        <span v-if="showButtonText">How are ranks determined?</span>
-        <v-icon class="ml-1" icon="mdi-information-outline" />
-      </v-btn>
+    <template #activator="{ props }">
+      <slot name="button">
+        <v-btn color="primary" class="mb-2" v-bind="props" variant="plain">
+          <span v-if="showButtonText">How are ranks determined?</span>
+          <v-icon class="ml-1" icon="mdi-information-outline" />
+        </v-btn>
+      </slot>
     </template>
     <!-- Dialog -->
     <v-card>
--- a/src/components/StatsScoringDialog.vue
+++ b/src/components/StatsScoringDialog.vue
@@ -1,8 +1,17 @@
 <template>
   <v-dialog v-model="show">
     <!-- Activator -->
-    <template #activator="{ props }">
-      <v-btn color="primary" class="mb-2" v-bind="props" variant="plain">
+    <template #activator="{ attrs, props }">
+      <v-btn
+        color="primary"
+        class="mb-2"
+        v-bind="{
+          ...$props,
+          ...props,
+          ...$attrs,
+          ...attrs,
+        }"
+        variant="plain">
         <span v-if="showButtonText">How are ranks determined?</span>
         <v-icon class="ml-1" icon="mdi-information-outline" />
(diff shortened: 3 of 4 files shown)


src/components/StatsScoringDialog.vue
```diff
@@ -1,8 +1,17 @@
...
-      <v-btn color="primary" class="mb-2" v-bind="props" variant="plain">
+    <template #activator="{ attrs, props }">
+      <v-btn
+        color="primary"
+        class="mb-2"
+        v-bind="{
+          ...$props,
+          ...props,
+          ...$attrs,
+          ...attrs,
```
itsalaidbacklife, line 12: what's going on here? Where do `$props` and `$attrs` come from?

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -12,12 +12,22 @@ function setup() {
...
   cy.vueRoute('/');
 }
+
 function assertSuccessfulJoin(gameState) {
   expect(gameState.id).to.not.eq(null);
   cy.url().should('include', '/lobby/');
   cy.contains('h1', `Lobby for ${gameState.name}`);
 }
 
+function assertParentInputToggle(selector, checked = false) {
```
itsalaidbacklife, line 22: Nice; this DRY's things up a bit. Why `Parent`? Maybe just `toggleInput()`?

src/components/AwardCard.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <v-card class="d-flex align-center" :color="theme.neutral.lighten2" :data-tournament="`${place}${suffix}`">
+  <v-card class="d-flex align-center" :color="theme.neutral-lighten2" :data-tournament="`${place}${suffix}`">
```
seriouslysean, line 2: This isn't related to these tests but it was referencing a color that didn't exist and bombing the modal.

src/components/StatsScoringDialog.vue
```diff
@@ -66,7 +75,7 @@
 import AwardCard from '@/components/AwardCard.vue';
 
 export default {
-  name: 'RulesDialog',
+  name: 'StatsScoringDialog',
```
seriouslysean, line 78: We were using the wrong name :)

tests/e2e/specs/out-of-game/home.spec.js
seriouslysean: Because it doesn't just toggle an input. It's specifically a wrapper around toggling a parent (like vuetify provides) when we need to check the inner input. Check below, we add `${selector} input` to the selector string. Additionally we aren't just toggling it, we're asserting the state AND toggling it.

src/components/StatsScoringDialog.vue
seriouslysean: I know this is super weird, and eventually I think we should rework how we do this activator, but let me explain. This entirely has to do with how we're applying the data attribute to the component calling context.

In `src/components/CreateGameDialog.vue` we call `<stats-scoring-dialog :show-button-text="false" data-cy="ranked-info-button" />` which has the data attr we need.

Then in `src/components/StatsScoringDialog.vue` we have an activator defined (the output of that component's dialog activator) which is what needs that data attr. The activator in this component looks like this.
```html
<!-- Activator -->
<template #activator="{ attrs, props }">
  <v-btn
    color="primary"
    class="mb-2"
    v-bind="{
      ...$props,
      ...props,
      ...$attrs,
      ...attrs,
    }"
    variant="plain">
    <span v-if="showButtonText">How are ranks determined?</span>
    <v-icon class="ml-1" icon="mdi-information-outline" />
  </v-btn>
</template>
```

The issue is that the `v-bind` was only applying the props from a defined activator but we aren't defining one. We have two options, we can duplicate the button and do something like this.
```html
<v-btn color="primary" class="mb-2" variant="plain" data-cy="ranked-info-button">
  <v-icon class="ml-1" icon="mdi-information-outline" />
</v-btn>
```

...or we can proxy the attributes and props down from the parent on to the default activator via `$props` and `$attrs`, which are the parent values (that you saw on the `stats-scoring-dialog` component above). This would allow any calling context to apply attrs and props to the default activator element by binding them along with the activator props as far as I understand.

That said, I pushed up the other example that is a bit less DRY, but easier to understand.

See the following for more information, https://vuejs.org/api/component-instance.html#attrs.

tests/e2e/specs/out-of-game/home.spec.js
itsalaidbacklife: But we don't have any inputs that aren't wrapped, so from our consumption's standpoint the component sortof just is a sparkling input. Like it doesn't make sense for the 'switch' or checkbox or w/e to be checked if it's inner input isn't b/c the component's state is effectively computed from the input itself, right?

As for asserting, I could go either way, but it seems reasonable to me to say that the helper function serves the purpose of toggling an input and because this is a testing context we assert it's state before and after every time we do that

src/components/CreateGameDialog.vue
```diff
@@ -12,7 +12,13 @@
       <v-card-text>
         <div class="d-flex align-center">
           <v-switch v-model="isRanked" label="Ranked" data-cy="create-game-ranked-switch" />
-          <stats-scoring-dialog :show-button-text="false" data-cy="ranked-info-button" />
+          <stats-scoring-dialog :show-button-text="false" data-cy="ranked-info-button">
```
itsalaidbacklife, line 15: Should we remove the outer layer data-cy selector now that it's passed via slot?
```suggestion
          <stats-scoring-dialog :show-button-text="false">
```

src/components/CreateGameDialog.vue
```diff
@@ -12,7 +12,13 @@
...
+            <template v-slot:button>
```
itsalaidbacklife, line 16: Pref shorthand, since we're generally using it elsewhere. Could probably rummage up a linting rule for it
```suggestion
            <template #button>
```

src/components/StatsScoringDialog.vue
```diff
@@ -2,10 +2,12 @@
   <v-dialog v-model="show">
     <!-- Activator -->
     <template #activator="{ props }">
-      <v-btn color="primary" class="mb-2" v-bind="props" variant="plain">
-        <span v-if="showButtonText">How are ranks determined?</span>
-        <v-icon class="ml-1" icon="mdi-information-outline" />
-      </v-btn>
+      <slot name="button">
+        <v-btn color="primary" class="mb-2" v-bind="props" variant="plain">
+          <span v-if="showButtonText">How are ranks determined?</span>
```
itsalaidbacklife, line 7: The use of this prop feels redundant with the slot approach; I think we want one or the other, since both are about what goes into the button used to open the same dialog. I'd be to go either way, but I think if we should either nix the prop and make this text appear unless you override the slot, or make it entirely prop-driven and nix the slot.

Funnily enough, the main driver for the use of the slot was getting the `data-cy` attribute to drill down properly, right? The goofy thing is that there isn't really a need (as I see it) for the consumers of this component to have different data-cy selectors! We could just put the data-cy selector into this component itself, directly on the button, and then there's no need for the parent + child prop combining.

The only thing that changes functionally is that all consumers (2) of the component will have the same data-cy attribute on its activator, but I think that's actually fine. It's one component, just slightly customized so we can select it the same way. What do you think?

itsalaidbacklife (approved):
Nice work. Learned stuff from your use of `$attrs` and `$props`

src/components/CreateGameDialog.vue
seriouslysean: Yep, missed that. I was in POC mode 😬

src/components/CreateGameDialog.vue
seriouslysean: Forgot all about this, despite using it multiple times, hahaha.

src/components/StatsScoringDialog.vue
seriouslysean: Good idea, I'll switch to just using the prop and move the cy selector down. I probably should have spent more time looking in to the why, but figured that's better for post "everything working" so we have time to rethink architecture. Mostly in "get it working" mode at the moment.

This is a simple change that I'm in favor of though, since it'll reduce complexity of the implementation. Good stuff! 👍🏼

seriouslysean (commented):
Great feedback, thanks @itsalaidbacklife.

itsalaidbacklife (approved):
Looks great

PR #321, opened 2023-02-05
Convert snackbar to base component

seriouslysean:
## Please check the following

- For New Features:

Little bit of both with this PR, fixed the mobile positioning and started the consolidation of the base components so we can reduce (and better use) the vuetify components.

### DIFF
--- a/src/components/Global/BaseSnackbar.vue
+++ b/src/components/Global/BaseSnackbar.vue
@@ -8,11 +8,11 @@
     v-bind="$attrs"
   >
     <slot />
-    <slot name="actions">
+    <template #actions>
       <v-btn data-cy="close-snackbar" icon variant="text" @click="clear">
         <v-icon icon="mdi-close" />
       </v-btn>
-    </slot>
+    </template>
   </v-snackbar>
 </template>
 
--- a/src/components/Global/BaseSnackbar.vue
+++ b/src/components/Global/BaseSnackbar.vue
@@ -1,0 +1,31 @@
+<template>
+  <v-snackbar
+    v-model="showSnackBar"
+    :color="color"
+    class="base-snackbar"
+    position="absolute"
+    location="bottom"
+    data-cy="auth-snackbar"
+    v-bind="$attrs"
+  >
+    <slot />
+    <slot name="actions">
+      <v-btn data-cy="close-snackbar" icon variant="text" @click="clear">
+        <v-icon icon="mdi-close" />
+      </v-btn>
+    </slot>
+  </v-snackbar>
+</template>
+
+<script>
+export default {
+  name: 'BaseSnackbar',
+  emits: ['clear'],
+  props: ['showSnackBar', 'color'],
+  methods: {
+    clear() {
+      this.$emit('clear');
+    },
+  }
+}
+</script>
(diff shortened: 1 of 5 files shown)


src/components/Global/BaseSnackbar.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <v-snackbar
+    v-model="showSnackBar"
+    :color="color"
+    class="base-snackbar"
+    position="absolute"
```
seriouslysean, line 6: This fixes the position issue combined with the location prop. If we need more props here to be proxied, we'll have to add them manually to this component. As-is this cleans up a lot though.

src/components/Global/BaseSnackbar.vue
```diff
@@ -0,0 +1,31 @@
...
+    location="bottom"
+    data-cy="auth-snackbar"
+    v-bind="$attrs"
+  >
+    <slot />
+    <slot name="actions">
+      <v-btn data-cy="close-snackbar" icon variant="text" @click="clear">
+        <v-icon icon="mdi-close" />
+      </v-btn>
+    </slot>
```
seriouslysean, line 16: There's still a bug on mobile where this shows up stacked wrong, but not part of this initial fix imo.

seriouslysean: 246E215E-6474-4797-B4CF-E714C6078172

itsalaidbacklife: Nice. I think the fix to the remaining style bug is to wrap the a lot in a template that fills the v-snackbar slot with the same name (actions).

This defines an actions slot in the base component, but it never fills the actions slot of the v-snackbar

itsalaidbacklife, line 16: I think this will fix the alignment bug
```suggestion
    <template #actions>
      <slot name="actions">
        <v-btn data-cy="close-snackbar" icon variant="text" @click="clear">
          <v-icon icon="mdi-close" />
        </v-btn>
      </slot>
    </template>
```

itsalaidbacklife: Actually simpler yet. There's no need to customize the close button, really. So until we hit a use-case where we need to do that, all we have to change is swapping the slot for a template

itsalaidbacklife: Actually simpler yet. There's no need to customize the close button, really. So until we hit a use-case where we need to do that, all we have to change is swapping the slot for a template
```suggestion
    <template #actions>
      <v-btn data-cy="close-snackbar" icon variant="text" @click="clear">
        <v-icon icon="mdi-close" />
      </v-btn>
    </template>
```

itsalaidbacklife (approved):
Looks good. Solid use case for the base component wrapper

src/components/Global/BaseSnackbar.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <v-snackbar
+    v-model="showSnackBar"
+    :color="color"
+    class="base-snackbar"
+    position="absolute"
+    location="bottom"
+    data-cy="auth-snackbar"
```
itsalaidbacklife, line 8: Doesn't this need to read from the prop? I think we want a `data-cy` prop that this proxies down

seriouslysean, line 8:
```suggestion
```

itsalaidbacklife (approved):

PR #330, opened 2023-02-08
Feat/devtools -- resolves #316

### LINKED ISSUE #316: [DevEx]: Add the vue devtools so they can be used with cypress
### Improvement Summary

We removed the vue-devtools dev dependency during the Vue 3 update. The most useful thing about this dep was the ability to use the dev tools in cypress tests for debugging.

### Detailed Description

We should configure the repo so that it's possible to use the devtools in cypress. That could potentially be accomplished using the [cypress version of the devtools](https://github.com/cypress-io/vue-devtools)

Or using the standalone [npm module for the devtools](https://www.npmjs.com/package/@vue/devtools). Either way, we want it to be possible to use the devtools to view the variables and store status from inside cypress while debugging.

itsalaidbacklife:
## Please check the following

- For New Features:

Adds the vue devtools during local development, with new `npm run start:devtools` command, and updates docs accordingly. No added tests, but devtools are only added in non-production frontend environments.

This resolves #316


cypress.config.js
```diff
@@ -1,18 +1,23 @@
...
+const isDev = process.env.NODE_ENV === ENV_DEV;
+const isProd = process.env.NODE_ENV === ENV_PROD;
 
 module.exports = defineConfig({
   // https://docs.cypress.io/guides/references/configuration#Global
   env: {
-    NODE_ENV: isProd ? 'production' : 'dev',
+    NODE_ENV: process.env.NODE_ENV || ENV_DEV,
+    DEV: isDev,
+    PROD: isProd,
```
itsalaidbacklife, line 13: What are these used for?

cypress.config.js
```diff
@@ -1,18 +1,23 @@
...
   },
   // https://docs.cypress.io/guides/references/configuration#e2e
   e2e: {
     baseUrl: process.env.VITE_API_URL || 'http://localhost:8080',
     specPattern: [
       ...(isProd ? ['tests/e2e/specs/**/*.spec.prod.js'] : ['tests/e2e/specs/**/*.spec.dev.js']),
-      'tests/e2e/specs/**/*.spec.js',
+      // 'tests/e2e/specs/**/*.spec.js',
```
itsalaidbacklife, line 20: Should we delete this?

cypress.config.js
```diff
@@ -1,18 +1,23 @@
 const { defineConfig } = require('cypress');
 
-const { isProd } = require('./utils/config-utils');
+const ENV_DEV = 'development';
+const ENV_PROD = 'production';
+const isDev = process.env.NODE_ENV === ENV_DEV;
+const isProd = process.env.NODE_ENV === ENV_PROD;
```
seriouslysean, line 6: Moved all instances of the prod check here since only Cypress cares about it now.

cypress.config.js
seriouslysean, line 13: Keep track of the original `NODE_ENV` and the new `PROD` and `DEV` values (from Vite) in case we need them.

src/main.js
```diff
@@ -22,15 +21,5 @@ app.use(store);
...
-if (!import.meta.env.PROD || window.Cypress) {
-  // Expose app for debugging/testing
-  window.cuttle = {
-    app,
-  };
-  // Connect the devtools -- non-prod only
-  devtools.connect(null, 8098);
-}
+// Add Cuttle window object
+initCuttleGlobals(app);
```
seriouslysean, line 25: Moved all globals logic to a singular function. This also means we could mock it in Vittest if we really want extra coverage eventually.

tests/e2e/specs/global/tools.spec.dev.js
```diff
@@ -0,0 +1,21 @@
+function setup() {
+  cy.wipeDatabase();
+  cy.visit('/', {
+    onBeforeLoad(win) {
+      cy.stub(win.console, 'log').as('consoleLog')
+      cy.stub(win.console, 'error').as('consoleError')
+    },
```
seriouslysean, line 7: We stub the console now so we can listen for the console log from the devtools connection instead of watching for the element. 😎

tests/e2e/specs/global/tools.spec.dev.js
```diff
@@ -0,0 +1,21 @@
...
+  });
+}
+
+describe('Devtools', () => {
+  beforeEach(setup);
+
+  it('Should include cuttle.app in development builds', () => {
+    cy.window().its('cuttle.app').should('exist');
+  });
```
seriouslysean, line 16: Added additional coverage for the `cuttle.app` property.

tests/e2e/specs/global/tools.spec.prod.js
```diff
@@ -1,12 +1,21 @@
 function setup() {
   cy.wipeDatabase();
-  cy.visit('/');
+  cy.visit('/?mock-prod=true', {
```
seriouslysean, line 3: Added a query param to allow us to mock a "true prod" environment. I know this is basically a shim but it's at least better coverage than none.

utils/config-utils.js
```diff
@@ -1,10 +1,30 @@
...
-  isProd,
-};
+import devtools from '@vue/devtools'
+
+import { version } from '_/package.json';
+
+export function initCuttleGlobals(app) {
+  if (!window) {
+    console.error('No window object found');
+  }
```
seriouslysean, line 8: Defensive in case the window isn't available. This may not be required since we aren't SSR yet, I'd be ok removing it.

utils/config-utils.js
```diff
@@ -1,10 +1,30 @@
...
+
+  // This is required to mock a prod-like scenario in Cypress to check the existence (or lack
+  // there-of) of the devtools instance
+  const isMockProd = new URLSearchParams(window.location.search).get('mock-prod') === 'true';
+  const includeDevtools = (!import.meta.env.PROD || window.Cypress) && !isMockProd;
+
+  const test = window.Cypress != null;
+  const cuttle = {
+    version,
```
seriouslysean, line 17: Added version so we can easily verify in prod.

utils/config-utils.js
```diff
@@ -1,10 +1,30 @@
...
+    // Expose app for debugging/testing in lower envs
+    app: includeDevtools ? app : null,
```
seriouslysean, line 19: Conditionally added app as we need it.

utils/config-utils.js
```diff
@@ -1,10 +1,30 @@
...
+    test,
```
seriouslysean, line 20: We may not need this any longer, but I thought it was useful as a boolean instead of checking `window.Cuttle` (which is an object).

cypress.config.js
seriouslysean: We may not need them at all anymore but cypress adds them to the process env which lets us use cy.ENV I think. I’m good to remove them if these still pass.

cypress.config.js
seriouslysean: Already did :)

itsalaidbacklife:
@seriouslysean we're failing the devtool test w/ a type error
image


Maybe we should just remove the test and generally try to simplify the config so that the tests are the same regardless of environment

seriouslysean:
I don't really agree with that, considering it's different per environment. I can take a look at this later to see what's up.

itsalaidbacklife:
@seriouslysean per our convo let's use the `test` NODE_ENV for the CI case where the frontend is in a near-production state (built but without the devtools) and the backend is mocking the db. Additionally, let's wrap the devtools connect call in a try/catch to swallow the error into a console warning

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "4.0.3",
+  "version": "4.0.4",
```
itsalaidbacklife, line 4:
```suggestion
  "version": "4.0.5",
```

cypress.config.js
itsalaidbacklife: Yeah this is a nice trick. Definitely worth revisiting

src/main.js
itsalaidbacklife: NIce cleanup

.env
```diff
@@ -1,3 +1,2 @@
-CUTTLE_ENV=dev
```
itsalaidbacklife, line 1: @seriouslysean per our convo, nixing some of the environment-specific config while we sort out how we want this to work more precisely

utils/config-utils.js
```diff
@@ -1,10 +1,30 @@
...
+  window.cuttle = cuttle;
+
+  // Connect the devtools -- non-prod only
+  if (!import.meta.env.PROD) {
+    try {
+      devtools.connect(null, 8098);
+    } catch(err) {
+      console.warn('Failed to connect vue devtools. Try running npm run start:devtools');
+      return;
+    }
```
itsalaidbacklife, line 27: Could remove this try/catch b/c it doesn't seem to catch anything

itsalaidbacklife: @seriouslysean think it's better to leave it defensively? Overkill if it doesn't address the key issue when the devtools aren't open

utils/config-utils.js
```diff
@@ -1,10 +1,30 @@
...
+    version,
+    // Expose app for debugging/testing in lower envs
+    app: window.Cypress ? app : null,
+    test,
+  };
+
+  window.cuttle = cuttle;
+
+  // Connect the devtools -- non-prod only
+  if (!import.meta.env.PROD) {
```
itsalaidbacklife, line 21: If I'm the weirdo for wanting these open, we could use another env var for whether to connect devtools that I can just leave on for myself. Think the default should be off?

utils/config-utils.js
seriouslysean: I don't see any issue leaving this here, don't think you need to return though.

itsalaidbacklife: The idea of the return was so that we could assume success if we didn't hit the catch block. That's why the log says connected rather than connecting. It didn't really work though because connect doesn't actually error

seriouslysean: I just moved the return up to the try. Can probably take another look at this eventually, not sure the best way to be safe with calling the devtools script.

seriouslysean (approved):

PR #331, opened 2023-02-09
Reset state at initialize for each game - resolves #318

### LINKED ISSUE #318: [Bug]: Top two cards revealed at start of game
### Bug Summary

Resolving seven oneOff persists through concede to the next game

### Detailed Description

Players that end a game via stalemate or concede while a seven oneOff is being resolved will see the top two cards still revealed at the start of the next game.

We should clear out the necessary dialog and statuses (like resolvingSeven) when we first start a game.

TommySchmid:
## Please check the following

- For New Features:


src/components/GameView/GameOverlays.vue
```diff
@@ -108,7 +108,6 @@ export default {
       'opponent',
       'hasGlassesEight',
       'player',
-      'resolvingSeven',
```
TommySchmid, line 111: Unused getter

seriouslysean: Technically we should do this in a separate PR to keep the changesets siloed but I’ll leave it up to @itsalaidbacklife.

itsalaidbacklife: I'd rather just do it now --save bookkeeping. It's more effort to document that we ought to do this than to just do it

TommySchmid:
@seriouslysean Removing the getter doesn't solve a particular issue. I noticed it was unused in the rest of this .vue doc and removed it for organization's sake.

seriouslysean:
@TommySchmid all good! We’ve been trying to chunk updates out a bit better and improve our versioning so I just wanted to call it out to @itsalaidbacklife — appreciate the cleanup!

itsalaidbacklife:
> @TommySchmid does this solve a particular issue? If so which one? I’d also like to see testing steps so I can reproduce and verify locally.

@seriouslysean this should fix https://github.com/cuttle-cards/cuttle/issues/318

TommySchmid:
@seriouslysean To reproduce this issue, you'll want to remove the `store.commit('resetState');` line from the Initialize switch case in the `sails.js` file. 

Then run the 'Opponent concedes while seven oneOff is being resolved' test within gameOver.spec

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -401,6 +401,50 @@ describe('Stalemates', () => {
...
+      p1Points: [],
+      p1FaceCards: [],
+      topCard: Card.FOUR_OF_CLUBS,
+      secondCard: Card.SIX_OF_DIAMONDS,
+    });
+
+    cy.get('[data-top-card=4-0]').should('not.exist');
+    cy.get('[data-second-card=6-1]').should('not.exist');
+  });
+});
```
itsalaidbacklife, line 446: Nice test case. Clear and a solid replication. I'm wondering if we could add two additional test cases -- one for your opponent conceding while you resolve a four, and another for the player (not opponent) conceding during a three

itsalaidbacklife: I suspect all of these cases would fail without the implemented fix as they share the same root cause of not resetting state before starting a new game

itsalaidbacklife (approved):
Nice work; looks good

PR #336, opened 2023-02-15
Feat/177 point counter

jerryh-jr:
## Please check the following

- For New Features:


jerryh-jr:
@seriouslysean @itsalaidbacklife @peterancourt  Hey all, please let me know where you see issues and any updates I can do. I am curious on the CSS part for the components and looking for more guidance on any changes.

seriouslysean (changes requested):
This is shaping up really nicely and is almost there. You have conflicts in the `GameView.vue` file and some due diligence to do w/r/t your css implementation and variable usage in the template.

client/js/components/GameView/PointCounter.vue
```diff
@@ -0,0 +1,118 @@
...
+  },
+};
+</script>
+
+<style scoped>
+
+.point-counter__wrapper {
+  background-color: rgba(241, 200, 160, 0.65);
+  border: 2px solid rgb(241, 200, 160);
+  box-shadow: 0 4px 4px -4px black;
```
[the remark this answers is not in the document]
jerryh-jr: I have updated this to be a RGBA value and changed it to a drop-shadow instead of box.

client/js/components/GameView/PointCounter.vue
```diff
@@ -0,0 +1,118 @@
...
+  max-width: 64px;
+}
+
+.point-counter {
+  font-weight: bold;
+  color: #f3f3f3;
+  font-family: 'Cormorant Infant', Century Gothic, CenturyGothic, AppleGothic, sans-serif;
+  padding: 0 0.25rem;
+  text-shadow: 1px 1px 5px #111111;
+}
```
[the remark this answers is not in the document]
jerryh-jr: I changed the CSS to use justify-center on the element and it still looks off. I am looking into other options for this. 
image

client/js/components/GameView/PointCounter.vue
```diff
@@ -0,0 +1,118 @@
...
+    clipPath() {
+      return this.isPlayer
+        ? 'polygon(0 0, 100% 0, 100% 90%, 50% 95%, 0 90%)'
+        : 'polygon(0% 100%, 100% 100%, 100% 10%, 50% 5%, 0 10%)';
+    },
+    clipPathTransform() {
+      return this.isPlayer
+        ? 'rotate(90deg)'
+        : 'rotate(270deg)'
+    },
```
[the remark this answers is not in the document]
jerryh-jr: These have been removed as a computed prop and moved to CSS.

jerryh-jr:
Screen
<img width=
shot 2023-03-05 at 8 12 29 PM" src="https://user-images.githubusercontent.com/47369872/222998555-24e6e2cb-7e68-4b9e-a8a8-701a3bd9515c.png">
Screenshot 2023-03-05 at 8 15 16 PM

Here are some images as the component currently stands. The design has changed to be more consistent with the history, using the same colors and fonts.

src/components/GameView/PointCounter.vue
```diff
@@ -0,0 +1,219 @@
...
+              <v-icon color="black" icon="mdi-information" v-bind="iconSize"/>
+            </v-btn>
+          </template>
+          <v-list class="score-goal-explanation">
+            <v-list-item :class="{ 'current-goal': kingCount === 0 }"> 0 Kings: 21pts </v-list-item>
+            <v-list-item :class="{ 'current-goal': kingCount === 1 }"> 1 King: 14pts </v-list-item>
+            <v-list-item :class="{ 'current-goal': kingCount === 2 }"> 2 Kings: 10pts </v-list-item>
+            <v-list-item :class="{ 'current-goal': kingCount === 3 }"> 3 Kings: 5pts </v-list-item>
+            <v-list-item :class="{ 'current-goal': kingCount === 4 }"> 4 Kings: 0pts </v-list-item>
+          </v-list>
```
seriouslysean, line 30: This should become a computed prop since this is a net new component. I'd like to see something like this as a computed prop:
```
kingPointsToWin() {
    return [0, 5, 10, 14, 21];
},
```

And then something like this in the template:
```suggestion
          <v-list class="score-goal-explanation">
            <v-list-item v-for="(points, idx) in kingPointsToWin" :class="{ 'current-goal': kingCount === idx }"> {{ idx }} Kings: {{ points }}pts </v-list-item>
          </v-list>
```

seriouslysean (changes requested):
```
 1) Game Basic Moves - P0 Perspective
       Plays Points:
     AssertionError: Timed out retrying after 4000ms: expected '<h3#player-score>' to contain 'POINTS: 11'
```
Your changes are breaking the tests so you'll need to run those locally and either have the AC changed and talk to @itsalaidbacklife, or get designs to better match the expectations.

I know that the designs were agreed upon but I have a major problem with the color scheme. The contrast is great! But it looks like an absolute eyesore in the middle of the screen.

image

The information icon is far too large, it should be smaller but still maintain the larger trigger area on mobile for accessibility reasons (44px is the recommended size).

The white border between the scores doesn't match the aesthetic and while the font color works for the history, I'm not sure it works here. This is less about the work you did and more about the spirit of the feature, so I'd like @itsalaidbacklife to weigh in.

src/components/GameView/PointCounter.vue
```diff
@@ -0,0 +1,221 @@
...
+          class="point-counter__current-points"
+          :class="[ isPlayer ? 'point-counter__current-player' : 'point-counter__current-opponent' ]"
+        >
+          {{ currentPoints }}
+        </div>
+        <div class="point-counter__total-points">{{ pointsToWin }}</div>
+        <v-menu :location="isPlayer ? 'top' : 'bottom'">
+          <template #activator="{ props }">
+            <v-btn class="point-counter__btn" icon v-bind="props" variant="plain">
+              <v-icon color="black" icon="mdi-information" :size="[$vuetify.display.smAndDown ? 'small' : 'large  ']" />
```
seriouslysean, line 21:
```suggestion
              <v-icon color="black" icon="mdi-information" :size="[$vuetify.display.smAndDown ? 'small' : 'large']" />
```

seriouslysean:
@jerryh-jr there is still a failing test here, is this almost ready for re-review? cc @itsalaidbacklife

src/components/GameView/PointCounter.vue
```diff
@@ -0,0 +1,231 @@
...
+    background-color: rgba(241, 200, 160, 0.65);
+    top: 1px;
+    left: 1px;
+    width: 23px;
+    height: 98px;
+    color: rgba(27, 27, 27);
+  }
+
+  &__btn {
+    margin-top: -0.8em;
```
itsalaidbacklife, line 106: I don't think we want to do margins or spacing in em because we don't want them to scale with your system font size. @seriouslysean does that sound right to you? Maybe just px instead?

src/components/GameView/PointCounter.vue
```diff
@@ -0,0 +1,231 @@
...
+  &__opponent {
+    clip-path: polygon(0% 100%, 100% 100%, 100% 10%, 50% 5%, 0 10%);
+  }
+
+  &__current-opponent {
+    margin-top: 0.8em;
+  }
+
+  &__current-player {
+    margin-top: 0.55em;
```
itsalaidbacklife, line 131: Same here -- I think we want to convert these to px units. @seriouslysean sound right?

src/components/GameView/PointCounter.vue
```diff
@@ -0,0 +1,231 @@
...
+    &__current-points {
+      font-size: 3em;
+    }
+
+    &__total-points {
+      font-size: 2.5em;
+    }
+
+    &__current-player {
+      margin-top: 0.5em;
```
itsalaidbacklife, line 187: And here

src/components/GameView/PointCounter.vue
```diff
@@ -0,0 +1,231 @@
...
+    &__btn {
+      margin-top: 0.1em;
+    }
+
+    &__current-opponent {
+      margin-top: 0.8em;
+    }
+
+    &__current-player {
+      margin-top: 0.5em;
```
itsalaidbacklife, line 226: And here

src/components/GameView/PointCounter.vue
```diff
@@ -0,0 +1,231 @@
...
+      font-size: 3.5em;
+      border-bottom: 2px solid rgba(27, 27, 27);
+    }
+
+    &__total-points {
+      font-size: 2.75em;
+    }
+
+    &__btn {
+      margin-top: 0.1em;
```
itsalaidbacklife, line 218: And here

src/components/GameView/PointCounter.vue
```diff
@@ -0,0 +1,231 @@
...
+      margin-top: 0.8em;
+    }
+
+    &__current-player {
+      margin-top: 0.5em;
+    }
+  }
+
+}
+</style>
```
itsalaidbacklife, line 231:
```suggestion
</style>

```

src/views/GameView.vue
```diff
@@ -1177,6 +1182,47 @@ export default {
...
+}
+@media screen and (min-width: 1024px) {
+  #point-counter-field {
+    margin-left: 0.5rem;
+    margin-top: 0.75rem;
+  }
+
+
+  #player-point-counter {
+    margin-top: 2.5em;
```
itsalaidbacklife, line 1222: Same for these margins

tests/e2e/support/helpers.js
```diff
@@ -198,12 +198,14 @@ function assertDomMatchesFixture(pNum, fixture) {
...
-    .should('contain', `POINTS: ${expectedP1Points}`)
-    .should('contain', `GOAL: ${expectedP1PointsToWin}`);
+  cy.get(`#${p0Role}-point-counter  [data-cy=point-counter__current]`)
+    .should('contain', `${expectedP0Points}`)
+  cy.get(`#${p0Role}-point-counter [data-cy=point-counter__total]`)
+    .should('contain', `${expectedP0PointsToWin}`)
+  cy.get(`#${p1Role}-point-counter [data-cy=point-counter__current]`)
+    .should('contain', `${expectedP1Points}`)
+  cy.get(`#${p1Role}-point-counter [data-cy=point-counter__total]`)
+    .should('contain', `${expectedP1PointsToWin}`);
```
itsalaidbacklife, line 208: This is great. My only critique is I don't love the words "total" and "current" for the number of points you have and the amount you need to win. I how about we use "points" and "goal" instead. Also for the data-cy attributes we've been doing straight up kebab-case with one dash between words and no double underscore __.

So data-cy="point-counter-points" and data-cy="point-counter-goal"

itsalaidbacklife (approved):
This looks great! I will run it on my machine today or tomorrow to check the UX, but the code looks clean. Only changes I'm requesting is to to rename the data-cy attributes and to swap the em units for px ones for the margins

itsalaidbacklife (changes requested):
This is getting really close! I've got one bug to report and some issues with the grid to work out.

**Bug**: the `current-goal` doesn't highlight correctly in the .score-goal-explanation
Expected:
image

Actual:
image

Responsiveness: Generally as the screen gets narrower, the point counters shrink and move upward. I think we want to keep them vertically aligned with the deck and scrap in all viewport sizes except mobile portrait, and for that to have the opponent point counter above and the player point counter below. Here's the current state
Medium desktop (want point counters to align with deck and scrap)
image

Small size desktop (Want counters to still align with deck and scrap)
image

Mobile Portrait: (let's move the player point counter down)
image

src/components/GameView/PointCounter.vue
seriouslysean: I would follow the standard we have elsewhere. Ideally we would move to em eventually and standardize but we aren't there yet.

src/components/GameView/PointCounter.vue
```diff
@@ -0,0 +1,214 @@
...
+    margin-top: 12px;
+  }
+
+  &__current-player {
+    margin-top: 9px;
+  }
+
+  &__current-opponent {
+    margin-top: 16px;
+  }
```
seriouslysean, line 139: I really don't like these margins, the values are all over the place and go against the larger use of vuetify imo. Is there a reason we have so many values?

src/components/GameView/PointCounter.vue
```diff
@@ -0,0 +1,214 @@
...
+      width: 30px;
+      height: 125px;
+    }
+
+    &__inner-container {
+      top: 2px;
+      left: 2px;
+      width: 26px;
+      height: 121px;
+    }
```
seriouslysean, line 159: Same for these, the values are all over the place. I know why you're doing it, to get the visuals "just right", but it will make this a bear to update and maintain in the future. I don't think there's a particular fix for it right now without deep diving on it, and I don't think it's a blocker, but it is something we should align on longterm.

We probably need a longer term design overhaul imo, at least in terms of standards. Some sort of a Cuttle design system with logical defaults and global values.

src/components/GameView/PointCounter.vue
```diff
@@ -0,0 +1,214 @@
...
+      margin-top: 2px;
+    }
+
+    &__current-opponent {
+      margin-top: 28px;
+    }
+
+    &__current-player {
+      margin-top: 16px;
+    }
```
seriouslysean, line 210: Same with these, but I already made that point above.

src/views/GameView.vue
seriouslysean: Agreed here, your values are all over the place using px, rem and em. Pick one and stick with it. In this case lets continue using px for now until we do a larger overhaul.

seriouslysean (changes requested):
Screenshot 2023-03-22 at 9 25 23 AM

I think that the desktop view should be more consistent. Opponent totals pinned to top, player totals pinned to the bottom.

Screenshot 2023-03-22 at 9 26 54 AM

Portrait looks pretty gross and feels a bit worse than the current version, but I don't think this is a blocker. I think we need to just re-evaluate this whole view.

Screenshot 2023-03-22 at 9 27 47 AM

Mobile landscape looks good, but we should be consistent with my first point. Pin to top and bottom respectively.

Screenshot 2023-03-22 at 9 27 30 AM

I also saw a Vue warning in the console while running this locally so you'll need to look in to that, too.

My change request still stands for now with these new points.

peterancourt:
Getting the alignment right on desktop and making things look good on mobile are both tricky here - a couple of thoughts:
1. Since we pretty much always want the point counters aligned with the decks, I wonder if we should consider moving the point counter into the same div/grid area as the decks (`deck-container`)? I have to imagine it would be easier to make sure they are aligned that way.

2. re: portrait mobile, I don't see the new point counter design working for mobile after the deck shifts to the bottom without feeling out of place. I feel like we may need to use the old/current version of the point counter there, or try and rethink the point counter design on mobile so that hangs out horizontally below the opponents cards, and above the player's cards. We can always think on that a bit. Point is we might want to make this change for desktop/landscape mobile for now, and come back to the mobile stuff.

itsalaidbacklife:
> Getting the alignment right on desktop and making things look good on mobile are both tricky here - a couple of thoughts:
> 
> 1. Since we pretty much always want the point counters aligned with the decks, I wonder if we should consider moving the point counter into the same div/grid area as the decks (`deck-container`)? I have to imagine it would be easier to make sure they are aligned that way.
> 2. re: portrait mobile, I don't see the new point counter design working for mobile after the deck shifts to the bottom without feeling out of place. I feel like we may need to use the old/current version of the point counter there, or try and rethink the point counter design on mobile so that hangs out horizontally below the opponents cards, and above the player's cards. We can always think on that a bit. Point is we might want to make this change for desktop/landscape mobile for now, and come back to the mobile stuff.

This makes sense. I agree with points 1 and 2. Let's try putting the point counters in the same grid area with the deck and scrap and let's use the old text-style point counters for portrait mobile.

src/views/GameView.vue
```diff
@@ -113,61 +113,68 @@
...
+                </div>
+              </template>
+            </v-card>
+            <point-counter
+              id="opponent-point-counter"
+              class="point-container point-counter"
+              :king-count="opponentKingCount"
+              :points-to-win="opponentPointsToWin"
+              :is-player="false"
+            />
```
peterancourt, line 156: @jerryh-jr Here's the layout change - since we want the point counters aligned with the deck/scrap piles, I moved the point counter components to the `deck-container` grid area. Beyond that, I added each point counter to a div with it's respective deck element, with a class called `deck-outer` that handles the alignment

peterancourt (commented):
@jerryh-jr sorry for the delay, finally fixed those github issues on my personal machine. I pushed my layout commit that should hopefully steer us in the right direction for wrapping this up, and left a few comments explaining what changed. Also, a couple comments on some changes to get this looking good on mobile until we figure out how we want to lay out the new point counters on small screens. 

Otherwise, we might need to bump up the size of the point counters and this should be ready to go. Great job on this!

Screen Shot 2023-05-18 at 11 22 32 AM

itsalaidbacklife:
Closing due to inactivity. This can be picked up later

PR #337, opened 2023-02-16
test: implement failing test case for recovering missing socket event

itsalaidbacklife:
## Please check the following

- For New Features:

Fixes bug where random socket disconnects caused missing updates

### DIFF
--- a/src/plugins/sails.js
+++ b/src/plugins/sails.js
@@ -3,6 +3,7 @@ import socketIoClient from 'socket.io-client';
 import { cloneDeep } from 'lodash';
 import store from '@/store/store.js';
 import router from '@/router.js';
+import { ROUTE_NAME_GAME } from '@/router';
 
 export const io = sails(socketIoClient);
 
@@ -188,5 +189,16 @@ io.socket.on('leftGame', function (evData) {
     store.commit('opponentLeft');
   } else {
     store.commit('otherLeftGame', evData.id);
+  }
+});
+
+//////////////////
+// Connectivity //
+//////////////////
+io.socket.on('connect', () => {
+  const { username } = store.state.auth;
+  const inGame = router.currentRoute.value.name === ROUTE_NAME_GAME;
+  if (inGame) {
+    return store.dispatch('requestReauthenticate', { username });
   }
 });
--- a/src/plugins/sails.js
+++ b/src/plugins/sails.js
@@ -199,7 +199,8 @@ io.socket.on('connect', () => {
   // Request latest game state if socket reconnects during game
   const { username } = store.state.auth;
   const inGame = router.currentRoute.value.name === ROUTE_NAME_GAME;
-  if (inGame) {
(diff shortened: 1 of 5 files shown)


src/plugins/sails.js
```diff
@@ -190,3 +191,15 @@ io.socket.on('leftGame', function (evData) {
...
 });
+
+//////////////////
+// Connectivity //
+//////////////////
+io.socket.on('connect', () => {
+  // Request latest game state if socket reconnects during game
+  const { username } = store.state.auth;
+  const inGame = router.currentRoute.value.name === ROUTE_NAME_GAME;
+  if (inGame) {
```
seriouslysean, line 202: I’d flip this to return early, otherwise do this logic after the if check.

utils/config-utils.js
```diff
@@ -11,6 +11,7 @@ export function initCuttleGlobals(app) {
     version,
     // Expose app for debugging/testing in lower envs
     app: test ? app : null,
```
seriouslysean, line 13: Convert this to a spread operator with an object with both values.

seriouslysean (changes requested):
I’ll look in to the failure, but try disconnecting via an action instead of pulling the whole of io in to the window object.

utils/config-utils.js
itsalaidbacklife: Removed the socket from the window altogether so that vitest can import this file without exploding

src/plugins/sails.js
itsalaidbacklife: Are you suggesting this?
```
  if (!inGame) {
    return;
  }
  return store.dispatch('requestReauthenticate', { username });
```

seriouslysean: Yeah exactly, I always order a fast fail approach.

src/store/modules/auth.js
```diff
@@ -155,5 +155,11 @@ export default {
         context.commit('clearAuth');
       }
     },
+    disconnectSocket() {
+      io.socket.disconnect();
```
seriouslysean, line 159: Despite feeling like an anti pattern, this does collocate everything and prevent s as more complicated test solve— glad it worked!

seriouslysean (approved):
Really nice work dude, this looks great!

src/plugins/sails.js
seriouslysean: For anyone else reading, my personal reasoning is that I like to see the return cases clearly defined, tends to make more sense to me when working through the happy path as the final return. There’s probably a case to be made for it being more performant in some cases too, depends on the code.

itsalaidbacklife: in the fullness of time, I think we'll actually wind up beefing this up to do different logic depending on the page i.e. get the gameList from the `HomeView` vs rehydrating the game data when you're in a game. So if and when we get there we may wind up shifting to a switch and/or something that's more case driven vs a sortof on/off logic like we have now

PR #339, opened 2023-02-22
Feat/Game over rework - resolves #105

### LINKED ISSUE #105: [Feature]: Show status of match in GameOverDialog
### Feature Summary

When a game is finished, if it is a ranked game, we should show players the status of their current match

### Detailed Description

As a player competing in a ranked match, I would like to see my score against my current opponent in our current match for this week after our ranked game ends.

We should redesign the `GameOverDialog` to display this information.

It should show you how many games you've played during this match, in order, with whether each was a win, a loss, or a stalemate, and it should say whether your match is in progress, or won/lost like

1. You won your match against <opponent_username>
2. You lost your match against <opponent_username>
3. Your match against <opponent_username> is in progress

lihaobhsfer:
## Please check the following

- For New Features:

### DIFF
--- a/api/controllers/game/concede.js
+++ b/api/controllers/game/concede.js
@@ -5,16 +5,13 @@ module.exports = async function (req, res) {
     await Game.updateOne(req.session.game).set({ result: winner });
     const game = await gameService.populateGame({ gameId: req.session.game });
     await gameService.clearGame({ userId: req.session.usr });
-    let currentMatch;
-    if (game.isRanked) {
-      currentMatch = await sails.helpers.addGameToMatch(game);
-    }
+    const currentMatch = game.isRanked ? await sails.helpers.addGameToMatch(game) : null;
     // Send socket message
     const victory = {
       gameOver: true,
       winner,
       conceded: true,
-      currentMatch: currentMatch,
+      currentMatch,
     };
     Game.publish([game.id], {
       verb: 'updated',
--- a/api/controllers/game/concede.js
+++ b/api/controllers/game/concede.js
@@ -5,14 +5,16 @@ module.exports = async function (req, res) {
     await Game.updateOne(req.session.game).set({ result: winner });
     const game = await gameService.populateGame({ gameId: req.session.game });
     await gameService.clearGame({ userId: req.session.usr });
+    let currentMatch;
     if (game.isRanked) {
-      await sails.helpers.addGameToMatch(game);
+      currentMatch = await sails.helpers.addGameToMatch(game);
     }
     // Send socket message
     const victory = {
       gameOver: true,
       winner,
       conceded: true,
+      currentMatch: currentMatch,
     };
     Game.publish([game.id], {
       verb: 'updated',
--- a/api/controllers/game/pass.js
+++ b/api/controllers/game/pass.js
@@ -57,8 +57,7 @@ module.exports = function (req, res) {
         };
         await Game.updateOne({ id: game.id }).set(gameUpdates);
         if (game.isRanked) {
-          const currentMatch = await sails.helpers.addGameToMatch(game);
-          victory.currentMatch = currentMatch;
+          victory.currentMatch = await sails.helpers.addGameToMatch(game);
         }
         await gameService.clearGame({ userId: req.session.usr });
       }
--- a/api/controllers/game/pass.js
+++ b/api/controllers/game/pass.js
@@ -45,6 +45,7 @@ module.exports = function (req, res) {
       const victory = {
         gameOver: false,
         winner: null,
+        currentMatch: null,
       };
       // Game ends in stalemate if 3 passes are made consecutively
       if (game.passes > 2) {
@@ -56,7 +57,8 @@ module.exports = function (req, res) {
         };
         await Game.updateOne({ id: game.id }).set(gameUpdates);
         if (game.isRanked) {
-          await sails.helpers.addGameToMatch(game);
+          const currentMatch = await sails.helpers.addGameToMatch(game);
+          victory.currentMatch = currentMatch;
         }
         await gameService.clearGame({ userId: req.session.usr });
       }
--- a/api/controllers/game/stalemate.js
+++ b/api/controllers/game/stalemate.js
@@ -39,8 +39,7 @@ module.exports = async function (req, res) {
     await Promise.all(updatePromises);
 
     if (victory.gameOver && gameUpdates.result === gameService.GameResult.STALEMATE && game.isRanked) {
-      const currentMatch = await sails.helpers.addGameToMatch(game);
-      victory.currentMatch = currentMatch;
+      victory.currentMatch = await sails.helpers.addGameToMatch(game);
     }
 
     Game.publish([game.id], {
--- a/api/controllers/game/stalemate.js
+++ b/api/controllers/game/stalemate.js
@@ -20,6 +20,7 @@ module.exports = async function (req, res) {
       gameOver: false,
       winner: null,
       conceded: false,
+      currentMatch: null,
     };
 
     // End in stalemate if both players requested stalemate this turn
@@ -32,13 +33,15 @@ module.exports = async function (req, res) {
         result: gameService.GameResult.STALEMATE,
       };
       updatePromises.push(gameService.clearGame({ userId }));
-      if (game.isRanked) {
-        updatePromises.push(sails.helpers.addGameToMatch(game));
-      }
     }
 
     updatePromises.push(Game.updateOne({ id: gameId }).set(gameUpdates));
     await Promise.all(updatePromises);
+
+    if (victory.gameOver && gameUpdates.result === gameService.GameResult.STALEMATE && game.isRanked) {
+      const currentMatch = await sails.helpers.addGameToMatch(game);
+      victory.currentMatch = currentMatch;
+    }
 
     Game.publish([game.id], {
       verb: 'updated',
--- a/api/helpers/add-game-to-match.js
+++ b/api/helpers/add-game-to-match.js
@@ -64,6 +64,7 @@ module.exports = {
           winner: relevantMatch.player2,
         });
       }
+      relevantMatch = await sails.helpers.findOrCreateCurrentMatch(player1.id, player2.id);
 
       return exits.success(relevantMatch);
     } catch (err) {
--- a/api/services/gameService.js
+++ b/api/services/gameService.js
@@ -117,7 +117,7 @@ module.exports = {
               })
               // then find points
               .then(function findPoints(values) {
-                const [ game, p0, p1 ] = values;
+                const [game, p0, p1] = values;
                 const p0Points = cardService.findPoints({ userId: p0.id });
                 const p1Points = cardService.findPoints({ userId: p1.id });
                 return Promise.all([
@@ -130,7 +130,7 @@ module.exports = {
               })
               // then format results & resolve
               .then(function finish(values) {
-                const [ game, p0, p1, p0Points, p1Points ] = values;
+                const [game, p0, p1, p0Points, p1Points] = values;
                 const populatedP0 = new tempUser(p0, p0Points);
                 const populatedP1 = new tempUser(p1, p1Points);
                 const result = new tempGame(game, populatedP0, populatedP1);
@@ -157,6 +157,7 @@ module.exports = {
       gameOver: false,
       winner: null,
       conceded: false,
+      currentMatch: null,
     };
     let { game } = options;
     const p0Wins = userService.checkWin({ user: game.players[0] });
@@ -177,7 +178,8 @@ module.exports = {
         ...gameUpdates,
       };
       if (game.isRanked) {
-        await sails.helpers.addGameToMatch(game);
+        const currentMatch = await sails.helpers.addGameToMatch(game);
+        res.currentMatch = currentMatch;
       }
     }
     return res;
--- a/api/services/gameService.js
+++ b/api/services/gameService.js
@@ -178,8 +178,7 @@ module.exports = {
         ...gameUpdates,
       };
       if (game.isRanked) {
-        const currentMatch = await sails.helpers.addGameToMatch(game);
-        res.currentMatch = currentMatch;
+        res.currentMatch = await sails.helpers.addGameToMatch(game);
       }
     }
     return res;
--- a/src/components/GameView/GameOverDialog.vue
+++ b/src/components/GameView/GameOverDialog.vue
@@ -1,24 +1,26 @@
 <template>
   <v-dialog v-model="show" persistent max-width="650">
     <v-card id="game-over-dialog">
-      <v-card-title :data-cy="headingDataAttr">
-        <h1>{{ heading }}</h1>
-      </v-card-title>
-      <v-card-text class="d-flex justify-center">
-        <v-img
-          :src="logoSrc"
-          :data-cy="logoDataAttr"
-          class="logo-image"
-        />
-      </v-card-text>
+      <div class="d-flex">
+        <v-card-title :data-cy="headingDataAttr" class="mt-8">
+          <h1>{{ heading }}</h1>
+        </v-card-title>
+        <v-card-text class="d-flex justify-end mr-4 mt-4">
+          <v-img
+            :src="logoSrc"
+            :data-cy="logoDataAttr"
+            class="logo-image"
+          />
+        </v-card-text>
+      </div>
       <v-card-text v-if="currentMatch" data-cy="match-result-section">
         Match against {{ opponent.username }}
+        <span>: {{currentMatchWinner ? "Finished" : "In Progress"}}</span>
       </v-card-text>
       <v-card-text v-if="currentMatchWinner" data-cy="match-winner-message">
         You {{ playerWinsMatch ? "won" :"lost" }} your game against {{opponent.username}}
       </v-card-text>
       <v-card-text data-cy="match-result-games">
-        current match games: {{matchGameStats.length}}
         <div class="d-flex">
           <div
             class="d-flex flex-column align-center"
@@ -175,12 +177,9 @@ export default {
 .logo-image {
   height: auto;
   max-width: 180px;
+}
 
-  // TODO: replace with an actual css variable for breakpoint width
-  // https://github.com/vuetifyjs/vuetify/blob/v3.1.3/packages/vuetify/src/styles/settings/_variables.scss#L95
-  @media(min-width: 601px) {
-    max-width: 300px;
-  }
-
+h1 {
+  font-family: 'PT Serif', serif !important;
 }
 </style>
--- a/src/components/GameView/GameOverDialog.vue
+++ b/src/components/GameView/GameOverDialog.vue
@@ -3,49 +3,40 @@
     <v-card id="game-over-dialog">
       <div class="d-flex">
         <v-card-title :data-cy="headingDataAttr" class="mt-8">
-          <h1>{{ heading }}</h1>
+          <h1 class="dialog-header">{{ heading }}</h1>
         </v-card-title>
         <v-card-text class="d-flex justify-end mr-4 mt-4">
-          <v-img
-            :src="logoSrc"
-            :data-cy="logoDataAttr"
-            class="logo-image"
-          />
+          <v-img :src="logoSrc" :data-cy="logoDataAttr" class="logo-image" />
         </v-card-text>
       </div>
-      <v-card-text v-if="currentMatch" data-cy="match-result-section">
+      <v-card-text class="dialog-text" v-if="currentMatch" data-cy="match-result-section">
         Match against {{ opponent.username }}
-        <span>: {{currentMatchWinner ? "Finished" : "In Progress"}}</span>
+        <span>: {{ currentMatchWinner ? 'Finished' : 'In Progress' }}</span>
       </v-card-text>
-      <v-card-text v-if="currentMatchWinner" data-cy="match-winner-message">
-        You {{ playerWinsMatch ? "won" :"lost" }} your game against {{opponent.username}}
+      <v-card-text class="dialog-text" v-if="currentMatchWinner" data-cy="match-winner-message">
+        You {{ playerWinsMatch ? 'won' : 'lost' }} your game against {{ opponent.username }}
       </v-card-text>
-      <v-card-text data-cy="match-result-games">
+      <v-card-text data-cy="match-result-games" class="dialog-text">
         <div class="d-flex">
           <div
-            class="d-flex flex-column align-center"
+            class="d-flex flex-column mr-4 align-center"
             v-for="(gameStatus, i) in matchGameStats"
-            :key="i"
+            :key="`${gameStatus}-${i}`"
           >
             <v-icon size="x-large" color="black" :icon="iconFromGameStatus(gameStatus)" />
-            <span>
-              {{ gameStatus }}
-            </span>
-          </div> 
+            {{ gameStatus }}
+          </div>
         </div>
       </v-card-text>
       <v-card-actions class="d-flex justify-end">
-        <v-btn color="primary" variant="flat" data-cy="gameover-go-home" @click="goHome">
-          Go Home
-        </v-btn>
+        <v-btn color="primary" variant="flat" data-cy="gameover-go-home" @click="goHome"> Go Home </v-btn>
       </v-card-actions>
     </v-card>
   </v-dialog>
 </template>
 
 <script>
-import { mapGetters } from 'vuex';
-import { mapState } from 'vuex';
+import { mapGetters, mapState } from 'vuex';
 
 export default {
   name: 'GameOverDialog',
@@ -76,10 +67,7 @@ export default {
       playingFromDeck: ({ game }) => game.playingFromDeck,
       game: ({ game }) => game,
     }),
-    ...mapGetters([
-      'player',
-      'opponent',
-    ]),
+    ...mapGetters(['player', 'opponent']),
     heading() {
       if (this.stalemate) {
         return 'Stalemate';
@@ -124,27 +112,30 @@ export default {
 
       return 'loss-img';
     },
-    currentMatch(){
-      return this.game.currentMatch
+    currentMatch() {
+      return this.game.currentMatch;
     },
-    matchGameStats(){
-      if(!this.game.currentMatch?.games) return []
-      const result= this.game.currentMatch.games.map(game=>{
-        if (game.result === -1) return 'I'
-        if(game.result === 0 || game.result === 1){
-          if(game.p0 === this.player.id) return 'W'
-          return 'L'
+    matchGameStats() {
+      const currentMatchGames = this.game.currentMatch?.games ?? []
+      return currentMatchGames.map((game) => {
+        switch(game.result) {
+          case 0: // p0 won game
+            return game.p0 === this.player.id ? 'W' : 'L';
+          case 1: // p1 won game
+            return game.p1 === this.player.id ? 'W' : 'L';
+          case 2: // draw
+            return 'D';
+          default: // incomplete
+            return 'I';
         }
-        return 'D'
-      })
-      return result
+      });
     },
     currentMatchWinner() {
-      return this.game.currentMatch?.winner
+      return this.game.currentMatch?.winner;
     },
-    playerWinsMatch(){
-      return this.game.currentMatch?.winner === this.player.id
-    }
+    playerWinsMatch() {
+      return this.game.currentMatch?.winner === this.player.id;
+    },
   },
   methods: {
     goHome() {
@@ -157,18 +148,18 @@ export default {
         });
       });
     },
-    iconFromGameStatus(gameStatus){
-      switch (gameStatus){
+    iconFromGameStatus(gameStatus) {
+      switch (gameStatus) {
         case 'W':
-          return 'mdi-trophy'
+          return 'mdi-thumb-up-outline';
         case 'L':
-          return 'mdi-close-thick'
+          return 'mdi-thumb-down-outline';
         case 'I':
-          return 'mdi-account-clock'
+          return 'mdi-account-clock-outline';
         case 'D':
-          return 'mdi-handshake-outline'
+          return 'mdi-handshake-outline';
       }
-    }
+    },
   },
 };
 </script>
@@ -176,10 +167,10 @@ export default {
 <style scoped lang="scss">
 .logo-image {
   height: auto;
-  max-width: 180px;
+  max-width: 90px;
 }
 
-h1 {
-  font-family: 'PT Serif', serif !important;
+.dialog-header, .dialog-text {
+  font-family: 'PT Serif', serif;
 }
 </style>
--- a/src/components/GameView/GameOverDialog.vue
+++ b/src/components/GameView/GameOverDialog.vue
@@ -21,9 +21,9 @@
       </div>
       <v-card-text class="dialog-text" v-if="currentMatch" data-cy="match-result-section">
         Match against {{ opponent.username }}
-        <span>: {{ currentMatchWinner ? 'Finished' : 'In Progress' }}</span>
+        <span>: {{ matchIsOver ? 'Finished' : 'In Progress' }}</span>
       </v-card-text>
-      <v-card-text class="dialog-text" v-if="currentMatchWinner" data-cy="match-winner-message">
+      <v-card-text class="dialog-text" v-if="matchIsOver" data-cy="match-winner-message">
         You {{ playerWinsMatch ? 'won' : 'lost' }} your game against {{ opponent.username }}
       </v-card-text>
       <v-card-text data-cy="match-result-games" class="dialog-text">
@@ -52,7 +52,7 @@ export default {
       type: Boolean,
       required: true,
     },
-    playerWins: {
+    playerWinsGame: {
       type: Boolean,
       required: true,
     },
@@ -76,12 +76,12 @@ export default {
     }),
     ...mapGetters(['player', 'opponent']),
     heading() {
-      if (this.currentMatchWinner) {
+      if (this.matchIsOver) {
         return this.playerWinsMatch ? 'You Win the Match' : 'You Lose the Match';
       }
       const currentMatchGames = this.game.currentMatch?.games ?? [];
       const gameNumberPrefix = currentMatchGames.length > 0 ? `Game ${currentMatchGames.length}: ` : '';
-      const winnerMessage = this.stalemate ? 'Draw' : this.playerWins ? 'You Win' : 'You Lose';
+      const winnerMessage = this.stalemate ? 'Draw' : this.playerWinsGame ? 'You Win' : 'You Lose';
 
       return `${gameNumberPrefix}${winnerMessage}`;
     },
@@ -90,7 +90,7 @@ export default {
         return 'stalemate-heading';
       }
 
-      if (this.playerWins) {
+      if (this.playerWinsGame) {
         return 'victory-heading';
       }
 
@@ -101,7 +101,7 @@ export default {
         return '/img/logo-stalemate.svg';
       }
 
-      if (this.playerWins) {
+      if (this.playerWinsGame) {
         return '/img/logo-body-no-text.svg';
       }
 
@@ -112,7 +112,7 @@ export default {
         return 'stalemate-img';
       }
 
-      if (this.playerWins) {
+      if (this.playerWinsGame) {
         return 'victory-img';
       }
 
@@ -136,7 +136,7 @@ export default {
         }
       });
     },
-    currentMatchWinner() {
+    matchIsOver() {
       return this.game.currentMatch?.winner;
     },
     playerWinsMatch() {
--- a/src/components/GameView/GameOverDialog.vue
+++ b/src/components/GameView/GameOverDialog.vue
@@ -11,6 +11,27 @@
           class="logo-image"
         />
       </v-card-text>
+      <v-card-text v-if="currentMatch" data-cy="match-result-section">
+        Match against {{ opponent.username }}
+      </v-card-text>
+      <v-card-text v-if="currentMatchWinner" data-cy="match-winner-message">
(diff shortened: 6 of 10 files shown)


lihaobhsfer:
some work in progress

In progress match

image


Finished match
image

Unranked
image

api/controllers/game/concede.js
```diff
@@ -5,14 +5,16 @@ module.exports = async function (req, res) {
     await Game.updateOne(req.session.game).set({ result: winner });
     const game = await gameService.populateGame({ gameId: req.session.game });
     await gameService.clearGame({ userId: req.session.usr });
+    let currentMatch;
     if (game.isRanked) {
-      await sails.helpers.addGameToMatch(game);
+      currentMatch = await sails.helpers.addGameToMatch(game);
     }
```
seriouslysean, line 11: Something like this might work better:
```
const currentMatch = game.isRanked ? await sails.helpers.addGameToMatch(game) : null;
```
That way the key always comes through but is set according to the context.

api/controllers/game/concede.js
```diff
@@ -5,14 +5,16 @@ module.exports = async function (req, res) {
...
     // Send socket message
     const victory = {
       gameOver: true,
       winner,
       conceded: true,
+      currentMatch: currentMatch,
```
seriouslysean, line 17:
```suggestion
      currentMatch,
```
Shorthand will work here.

api/controllers/game/pass.js
```diff
@@ -45,6 +45,7 @@ module.exports = function (req, res) {
       const victory = {
         gameOver: false,
         winner: null,
+        currentMatch: null,
```
seriouslysean, line 48: Doing null above also matches the shape we're going for here so we don't get different schemas based on request.

api/controllers/game/pass.js
```diff
@@ -56,7 +57,8 @@ module.exports = function (req, res) {
         };
         await Game.updateOne({ id: game.id }).set(gameUpdates);
         if (game.isRanked) {
-          await sails.helpers.addGameToMatch(game);
+          const currentMatch = await sails.helpers.addGameToMatch(game);
+          victory.currentMatch = currentMatch;
```
seriouslysean, line 61:
```suggestion
          victory.currentMatch = await sails.helpers.addGameToMatch(game);
```
Unless this gets referenced below, no need to assign it to a const.

api/controllers/game/stalemate.js
```diff
@@ -32,14 +33,16 @@ module.exports = async function (req, res) {
...
-        updatePromises.push(sails.helpers.addGameToMatch(game));
-      }
     }
 
     updatePromises.push(Game.updateOne({ id: gameId }).set(gameUpdates));
     await Promise.all(updatePromises);
 
+    if (victory.gameOver && gameUpdates.result === gameService.GameResult.STALEMATE && game.isRanked) {
+      const currentMatch = await sails.helpers.addGameToMatch(game);
+      victory.currentMatch = currentMatch;
```
seriouslysean, line 43:
```suggestion
      victory.currentMatch = await sails.helpers.addGameToMatch(game);
```
Same here.

api/helpers/add-game-to-match.js
```diff
@@ -64,6 +64,7 @@ module.exports = {
           winner: relevantMatch.player2,
         });
       }
+      relevantMatch = await sails.helpers.findOrCreateCurrentMatch(player1.id, player2.id);
```
seriouslysean, line 67: This feels like it should be in an else at first, but I'm not entirely sure. It's weird to set `relevantMatch` conditionally based on player wins above, and then reset it again immediately at the end anyway, overwriting what we did before.

If this is intentional, a comment explaining why this is necessary would be really useful.

api/services/gameService.js
```diff
@@ -177,7 +178,8 @@ module.exports = {
         ...gameUpdates,
       };
       if (game.isRanked) {
-        await sails.helpers.addGameToMatch(game);
+        const currentMatch = await sails.helpers.addGameToMatch(game);
+        res.currentMatch = currentMatch;
```
seriouslysean, line 182:
```suggestion
        res.currentMatch = await sails.helpers.addGameToMatch(game);
```
Doesn't seem to be reused below.

src/components/GameView/GameOverDialog.vue
```diff
@@ -1,15 +1,38 @@
...
+      </v-card-text>
+      <v-card-text v-if="currentMatchWinner" data-cy="match-winner-message">
+        You {{ playerWinsMatch ? "won" :"lost" }} your game against {{opponent.username}}
+      </v-card-text>
+      <v-card-text data-cy="match-result-games">
+        <div class="d-flex">
+          <div
+            class="d-flex flex-column align-center"
+            v-for="(gameStatus, i) in matchGameStats"
+            :key="i"
```
seriouslysean, line 28: Key should be more unique, something like this to prevent collisions.
```suggestion
            :key="`${gameStatus}-${i}`"
```

src/components/GameView/GameOverDialog.vue
```diff
@@ -1,15 +1,38 @@
...
+          >
+            <v-icon size="x-large" color="black" :icon="iconFromGameStatus(gameStatus)" />
+            <span>
+              {{ gameStatus }}
+            </span>
```
seriouslysean, line 33: Empty span feels weird here, does this need a class? Would a `<p>` tag suffice or can it just be added inline?

src/components/GameView/GameOverDialog.vue
```diff
@@ -21,6 +44,9 @@
 </template>
 
 <script>
+import { mapGetters } from 'vuex';
+import { mapState } from 'vuex';
```
seriouslysean, line 48:
```suggestion
import { mapGetters, mapState } from 'vuex';
```
These imports are from the same package and can be combined.

src/components/GameView/GameOverDialog.vue
```diff
@@ -46,6 +72,14 @@ export default {
         // do nothing - parent controls whether dialog is open
       },
     },
+    ...mapState({
+      playingFromDeck: ({ game }) => game.playingFromDeck,
+      game: ({ game }) => game,
```
seriouslysean, line 77: Mapping the game this way feels weird but I get why you're doing it. It might be better to eventually convert these to getters, nothing that will hold up this PR though. I think we currently just use a computed prop in each component that references the game state as-is which is already weird, too. cc @itsalaidbacklife

src/components/GameView/GameOverDialog.vue
```diff
@@ -90,6 +124,27 @@ export default {
...
+      const result= this.game.currentMatch.games.map(game=>{
+        if (game.result === -1) return 'I'
+        if(game.result === 0 || game.result === 1){
+          if(game.p0 === this.player.id) return 'W'
+          return 'L'
+        }
+        return 'D'
+      })
+      return result
+    },
```
seriouslysean, line 141: Stay consistent with your coding style, spaces after the if statements, spaced after the equals for const declarations.

src/components/GameView/GameOverDialog.vue
```diff
@@ -90,6 +124,27 @@ export default {
 
       return 'loss-img';
     },
+    currentMatch(){
+      return this.game.currentMatch
+    },
+    matchGameStats(){
+      if(!this.game.currentMatch?.games) return []
+      const result= this.game.currentMatch.games.map(game=>{
```
seriouslysean, line 132: You are optionally chaining the currentMatch variable, do we expect it to be empty? If so, I'd declare it as a variable above, optionally chain it and then use that in the rest of this function. You can even set a default value if we don't need to return early to prevent an extra fork.
```suggestion
      const currentMatchGames = this.game.currentMatch?.games ?? []; 
      const result = currentMatchGames.map((game) => {
```

src/components/GameView/GameOverDialog.vue
```diff
@@ -90,6 +124,27 @@ export default {
...
+        if (game.result === -1) return 'I'
+        if(game.result === 0 || game.result === 1){
+          if(game.p0 === this.player.id) return 'W'
+          return 'L'
+        }
+        return 'D'
+      })
+      return result
```
seriouslysean, line 140: You're declaring a variable just to return it, you could directly return the result without needing to assign it at all.
```
return this.game.currentMatch?.games?.map?.((game) => { ... });
```

src/components/GameView/GameOverDialog.vue
```diff
@@ -90,6 +124,27 @@ export default {
...
+    },
+    matchGameStats(){
+      if(!this.game.currentMatch?.games) return []
+      const result= this.game.currentMatch.games.map(game=>{
+        if (game.result === -1) return 'I'
+        if(game.result === 0 || game.result === 1){
+          if(game.p0 === this.player.id) return 'W'
+          return 'L'
+        }
+        return 'D'
```
seriouslysean, line 138: This nest of ifs is really confusing, it would be preferable to block them out a bit more legibly. Combing returns if you can would look good. Ideally, some sort of map or case statement (and @itsalaidbacklife knows I hate those) might even be better.
```suggestion
if (game.result < 0) {
  return 'I';
}

if ([0, 1].includes(game.result)) {
  return game.p0 === this.player.id ?
    'W' :
    'L';
}

return 'D';
```

src/components/GameView/GameOverDialog.vue
```diff
@@ -90,6 +124,27 @@ export default {
...
+      })
+      return result
+    },
+    currentMatchWinner() {
+      return this.game.currentMatch?.winner
+    },
+    playerWinsMatch(){
+      return this.game.currentMatch?.winner === this.player.id
```
seriouslysean, line 146: Make sure to always include a semicolons to end a statement. @itsalaidbacklife we should do a pass on our linting rules and get some precommit linting in place to keep a consistent coding style as a follow up.

src/components/GameView/GameOverDialog.vue
```diff
@@ -110,12 +177,9 @@ export default {
...
   max-width: 180px;
+}
 
-  // TODO: replace with an actual css variable for breakpoint width
-  // https://github.com/vuetifyjs/vuetify/blob/v3.1.3/packages/vuetify/src/styles/settings/_variables.scss#L95
-  @media(min-width: 601px) {
-    max-width: 300px;
-  }
-
+h1 {
```
seriouslysean, line 182: It's preferable to class something you're styling unless it's a global element. In this case, I'd prefer some sort or class like `dialog-heading` or similar. I'm trying to get @itsalaidbacklife to get with BEM for standardizing our naming practices, but we'll see, haha.

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -509,14 +539,21 @@ describe('Creating And Updating Ranked Matches', () => {
     cy.loginPlayer(playerOne.username, playerOne.password);
     cy.setupGameAsP0(true, true);
   });
-  it('Creates a match when two players play a ranked game for the first time this week', function () {
+  it.only('Creates a match when two players play a ranked game for the first time this week', function () {
```
seriouslysean, line 542: Remove this .only here.

seriouslysean (changes requested):
This is looking really great so far! Lots of feedback on patterns and ways to clean up what's there, but the bones are solid! Nice work.

api/controllers/game/concede.js
itsalaidbacklife: I like this as well. Keeps a consistent payload shape

api/controllers/game/stalemate.js
```diff
@@ -32,14 +33,16 @@ module.exports = async function (req, res) {
...
       updatePromises.push(gameService.clearGame({ userId }));
-      if (game.isRanked) {
-        updatePromises.push(sails.helpers.addGameToMatch(game));
-      }
     }
 
     updatePromises.push(Game.updateOne({ id: gameId }).set(gameUpdates));
     await Promise.all(updatePromises);
 
+    if (victory.gameOver && gameUpdates.result === gameService.GameResult.STALEMATE && game.isRanked) {
```
itsalaidbacklife, line 41: Why flatten here? Previously this was inside the above conditional i.e. in the case where we are already ending the game, which seems fairly straightfoward to me. Is it about decreasing the total levels of nesting? That's a reasonable concern, idk what to make of it. I think if we do keep this here that we might not need the stalemate check at this moment because it should be the case that the only way the game can end from this endpoint is because of a stalemate
```suggestion
    if (victory.gameOver && game.isRanked) {
```

api/helpers/add-game-to-match.js
itsalaidbacklife: I believe the fundamental issue here is that the database updating operations on line 53, 56, and 62 don't ensure that the `relevantMatch()` variable is up to date with the state of the database (as generally database-updating methods don't always return the resulting object, they just update it for the next time the db is read).

@lihaobhsfer please feel free to chime in here, but I think the possibility for de-sync is threefold depending on the conditionasls on lines 56 and 61:

@lihaobhsfer am I correct that the issue you encountered was consistently that the most recently played game was not appearing in `relevantMatch()` because  `Match.addToCollection()` does not update this variable in-memory?


The thing that is most awkward about line 67 is that it re-queries the database to generate the match object that we _pretty much_ already have but not completely (because it's missing the current game).

What might be better would be to simply update `relevantMatch` in memory to reflect the expected state in the database i.e. add the current game to it. We could move the logic that adds the game to the relevant match down below the logic that updates the relevant match in case it's over (someone got 2 wins). What about something like this?

```
      // End the match if this game clinches it
      if (numPlayer1Wins >= 2) {
        relevantMatch = await Match.updateOne(relevantMatch.id).set({
          endTime: dayjs().valueOf(),
          winner: relevantMatch.player1,
        });
      } else if (numPlayer2Wins >= 2) {
        relevantMatch = await Match.updateOne(relevantMatch.id).set({
          endTime: dayjs().valueOf(),
          winner: relevantMatch.player2,
        });
      }

      // Add game to current match
     await Match.addToCollection(relevantMatch.id, 'games').members([game.id]);
     relevantMatch.games.push(game); 
```

That way we group up the database write operation with the part where we update the `relevantMatch` variable in-memory, and we avoid re-querying for data that should already be up to date. @seriouslysean and @lihaobhsfer does that make sense?

src/components/GameView/GameOverDialog.vue
itsalaidbacklife: huh, yeah. Likely worth revisiting this holistically but it'll change when we move to pinia so let's wait until then to dig in. Imo this is fine but agreed it's slightly wonky as is having a separate computed prop

src/components/GameView/GameOverDialog.vue
itsalaidbacklife: Yeah unranked games won't have matches so it could be empty. This seems a reasonable simplification

src/components/GameView/GameOverDialog.vue
itsalaidbacklife: @seriouslysean I think the includes block is wrong because the `game.result` value is the pNum of whichever player won the game so we need to differentiate between whether the player is p0 or p1 differently based on whether it's 0 or 1. I think this is a good use case for switch: one separate return statement per possible value of a specified variable

```suggestion
        switch(game.result) {
          case 0: // p0 won game
            return game.p0 === this.player.id ? 'W' : 'L';
          case 1: // p1 won game
            return game.p1 === this.player.id ? 'W' : 'L';;
          case 2: // draw
            return 'D';
          default: // incomplete
            return 'I';
        }
```

src/components/GameView/GameOverDialog.vue
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/340

src/components/GameView/GameOverDialog.vue
```diff
@@ -102,6 +157,18 @@ export default {
...
+      switch (gameStatus){
+        case 'W':
+          return 'mdi-trophy'
+        case 'L':
+          return 'mdi-close-thick'
+        case 'I':
+          return 'mdi-account-clock'
+        case 'D':
+          return 'mdi-handshake-outline'
+      }
```
itsalaidbacklife, line 170: Let's discuss the iconography. This is a reasonable first pass, but I think we can sort it out even clearer. I like the handshake for the draw. I think we want win and loss to look like obvious opposites, which I don't think trophy does well because what's the opposite of a trophy? 

So maybe we lean on the handshake and say it's thumbs up, thumbs down, and handshake?
image


image

There's also this possible option for stalemate/draw
image


@seriouslysean @lihaobhsfer what do y'all think? I'd also pondered faces like smily, frown/dead for W/L. Penny for your thoughts

src/components/GameView/GameOverDialog.vue
itsalaidbacklife: Would also pref to skip `!important` if we can use a more specific selector

src/store/modules/game.js
```diff
@@ -237,10 +238,11 @@ export default {
       state.waitingForOpponentToPlayFromDeck = val;
     },
     // Game Over
-    setGameOver(state, { gameOver, conceded, winner }) {
+    setGameOver(state, { gameOver, conceded, winner, currentMatch }) {
       state.gameIsOver = gameOver;
       state.conceded = conceded;
       state.winnerPNum = winner;
+      state.currentMatch = currentMatch;
```
itsalaidbacklife, line 245: I think we also need to set this in `resetState()` so it initializes correctly (do we want `null`?

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -7,18 +7,48 @@ const dayjs = require('dayjs');
...
   cy.get('#game-over-dialog').should('be.visible').get('[data-cy=stalemate-heading]').should('be.visible');
   cy.get('[data-cy=stalemate-img]').should('be.visible');
+  cy.window()
+    .its('cuttle.app.config.globalProperties.$store.state.game')
+    .then((game) => {
+      if (game.isRanked)
+        cy.get('#game-over-dialog')
+          .should('be.visible')
+          .get('[data-cy=match-result-section]')
+          .should('be.visible');
```
itsalaidbacklife, line 49: Pref braces; they're easier to read
```suggestion
      if (game.isRanked) {
        cy.get('#game-over-dialog')
          .should('be.visible')
          .get('[data-cy=match-result-section]')
          .should('be.visible');
      }
```

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -7,18 +7,48 @@ const dayjs = require('dayjs');
...
+      else cy.get('#game-over-dialog').should('be.visible').should('not.contain', 'Match against');
```
itsalaidbacklife, line 50:
```suggestion
      if (game.isRanked) {
        cy.get('#game-over-dialog')
          .should('be.visible')
          .get('[data-cy=match-result-section]')
          .should('be.visible');
      } else cy.get('#game-over-dialog').should('be.visible').should('not.contain', 'Match against');
```

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -7,18 +7,48 @@ const dayjs = require('dayjs');
...
   cy.get('[data-cy=loss-img]').should('be.visible');
+  cy.window()
+    .its('cuttle.app.config.globalProperties.$store.state.game')
+    .then((game) => {
+      if (game.isRanked)
+        cy.get('#game-over-dialog')
+          .should('be.visible')
+          .get('[data-cy=match-result-section]')
+          .should('be.visible');
+      else cy.get('#game-over-dialog').should('be.visible').should('not.contain', 'Match against');
```
itsalaidbacklife, line 34:
```suggestion
      if (game.isRanked) {
        cy.get('#game-over-dialog')
          .should('be.visible')
          .get('[data-cy=match-result-section]')
          .should('be.visible');
      } else cy.get('#game-over-dialog').should('be.visible').should('not.contain', 'Match against');
```

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -7,18 +7,48 @@ const dayjs = require('dayjs');
...
   cy.get('#game-over-dialog').should('be.visible').get('[data-cy=victory-heading]').should('be.visible');
+  cy.window()
+    .its('cuttle.app.config.globalProperties.$store.state.game')
+    .then((game) => {
+      if (game.isRanked)
+        cy.get('#game-over-dialog')
+          .should('be.visible')
+          .get('[data-cy=match-result-section]')
+          .should('be.visible');
+      else cy.get('#game-over-dialog').should('be.visible').should('not.contain', 'Match against');
```
itsalaidbacklife, line 18:
```suggestion
      if (game.isRanked) {
        cy.get('#game-over-dialog')
          .should('be.visible')
          .get('[data-cy=match-result-section]')
          .should('be.visible');
      } else cy.get('#game-over-dialog').should('be.visible').should('not.contain', 'Match against');
```

itsalaidbacklife (changes requested):
Great start; this looks like it functions and hits all the key points. Time for refactoring and some UX cleanup. The code stuff is already in the comments, so I'll focus on the look of the dialog itself.

1. We should shrink down the image in the righthand corner so that it has a height to match the header text. That way we don't get a bunch of empty vertical space in the dialog. 
2. Let's discuss and rework the iconography. I have a thread going on the computed prop where you return the icon names
3. The non-ranked games should look the same as they do now i.e. rather than using the upper-righthand corner, let's keep the image in the center in the same way it is now. I think it looks better that way and works in the middle when we don't have any other info to display
4. This wasn't originally in the issue, but @seriouslysean and @lihaobhsfer what do you think of updating the header to say which game you won or whether the match is over e.g. "You Win Game 2" or "You Lose the Match"? The idea is to help clarify which info pertains to the game vs the match itself. This might also warrant updating the text in the body of the dialog. Here's a [discord thread](https://discord.com/channels/833694928684187669/1072564154927431703/1077960339100008468) with an example: 

image

itsalaidbacklife:
We should also norm on whether to call a tie a stalemate vs a draw. I think it's inconsistent to use D for draw but to show 'Stalemate' as the dialog header -- what do y'all think?

itsalaidbacklife:
@lihaobhsfer and @seriouslysean I added the minor-version label to this issue; I think it's a big enough UX change to warrant bumping the package version a minor #

seriouslysean:
@itsalaidbacklife I'd like consts with whatever letter we need so it's more clear inline. Something like `return GAME_TYPE_STALEMATE;`.

src/components/GameView/GameOverDialog.vue
lihaobhsfer: For all linting related comments, I'm surprised the linting tools didn't catch all these. I'll need to rely on those since I won't be able to catch all these. 

I'll look into my setup on why these are not working even when I ran `npm run lint:fix`. However we should think about getting these to work on any machine out of the box.

lihaobhsfer: And this will save us time once we agree to a set of linting rules to respect, and have the code linted consistently

src/components/GameView/GameOverDialog.vue
lihaobhsfer: BEM would be hard to keep track of if the project grows. The names can get quite long in some cases and eventually it's going to be hard to manage in my opinion. It'll be better to have our css tools to do this automatically for us. Thinking css modules in React.

I thought the styles are scoped to the local component, but I could be wrong.

tests/e2e/specs/in-game/gameOver.spec.js
lihaobhsfer: thanks for catching this! I always forget to remove.only.

src/components/GameView/GameOverDialog.vue
lihaobhsfer: I don't have much opinion here. As long as the size of the icons are consistent.

tests/e2e/specs/in-game/gameOver.spec.js
lihaobhsfer: should we add a lint rule to always use curly brackets even if there's only one statement for each condition :) ?

api/helpers/add-game-to-match.js
seriouslysean: Thanks for chiming in, this makes sense to me. I’m not as well versed on the sails side so I was relying on your breakdown, haha.

src/components/GameView/GameOverDialog.vue
seriouslysean: Yep, that works. Just calling it out so we don’t lose it in the shuffle.

src/components/GameView/GameOverDialog.vue
seriouslysean: Linting is something I plan on improving as part of my arch adjustments. Pinia, linting, typescript. Some sort of rough order there, haha.

src/components/GameView/GameOverDialog.vue
seriouslysean: I wasn’t entirely sure and tired to match it as best I could. This is one of those cases a switch actually makes sense so I’m in to it.

src/components/GameView/GameOverDialog.vue
seriouslysean: I have some thought around this to lint only the staged files since it doesn’t take long at all but I’ll proof it out first to make sure we don’t cause too much of a disruption. We definitely need it for consistency though.

src/components/GameView/GameOverDialog.vue
seriouslysean: I have a lot of thoughts around this and I just don’t agree with you at all. I maintain a very large codebase for Urban Outfitters that drives 3 brands  from one codebase.

I use BEM extensively and if you’re getting in to a situation there they’re too long, something is wrong. There’s definitely a case for scoped styles, but all are solving similar problems.

ideally if your class names are too long you break the code out in to a new component to reduce the complexity of the nesting.

It also adds clarity around the component and its children which is easier to understand at a glance.

It’s  certainly not the only way but I’m quite fond of it.

RE: BEM for Cuttle specifically, I’ll put together some sort of POC first so we can all see how it would work first and if it’s not right for the codebase we can drop it.

tests/e2e/specs/in-game/gameOver.spec.js
seriouslysean: I also always forget them, lol. Got @itsalaidbacklife working overtime catching them. Nice to switch it up and catch someone else for a change.

src/components/GameView/GameOverDialog.vue
seriouslysean: I like the thumbs instead of the other ones.

tests/e2e/specs/in-game/gameOver.spec.js
seriouslysean: Yeah we can do a linting pass when I add some more integrated rules.

tests/e2e/specs/in-game/gameOver.spec.js
itsalaidbacklife: Yeah I would love to work out a linting rule for this. No `.only(` in .spec files

api/controllers/game/stalemate.js
lihaobhsfer: it's the database operation. `await sails.helpers.addGameToMatch(game);` needs to happen after `Promise.all()` otherwise the game state would be `incomplete` even if it is actually `stalemate`

src/components/GameView/GameOverDialog.vue
lihaobhsfer: I think previously I was console logging the result to see if it was the desired shape. Declaring a variable made it easier

api/controllers/game/stalemate.js
itsalaidbacklife: Which game state? The one on `game` itself? Is the reason that this is bad because the resulting value of `victory.currentMatch()` is out of date if the game `result` hasn't yet been updated?

src/components/GameView/GameOverDialog.vue
itsalaidbacklife: I think we can sidebar this for now. @seriouslysean I'm interested to see the BEM POC, though my current thinking is that what would be optimal would be a combination of:

1. Scoped styles in each component (currently in use)
2. Better configured global style options for things like typography
3. Established style code conventions to help standardize our approach to create more standardized css

api/controllers/game/stalemate.js
lihaobhsfer: `victory.currentMatch.games.map(game => game.result)`. `result` would be `incomplete` if we do it the orginal way.

src/components/GameView/GameOverDialog.vue
```diff
@@ -90,6 +112,30 @@ export default {
...
+    matchGameStats() {
+      const currentMatchGames = this.game.currentMatch?.games ?? []
+      return currentMatchGames.map((game) => {
+        switch(game.result) {
+          case 0: // p0 won game
+            return game.p0 === this.player.id ? 'W' : 'L';
+          case 1: // p1 won game
+            return game.p1 === this.player.id ? 'W' : 'L';
+          case 2: // draw
+            return 'D';
```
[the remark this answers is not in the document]
lihaobhsfer: I like the idea of keeping the terms consistent.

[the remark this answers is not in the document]
seriouslysean: I'd prefer draw even though stalemate sounds cooler just for the ease of use.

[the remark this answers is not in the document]
itsalaidbacklife: Alright let's go with Draw. @lihaobhsfer can you update the heading text as well?

api/controllers/game/stalemate.js
itsalaidbacklife: That's specifically for the current game that we're adding to the match right now, correct? This is reasonable; my only wondering is whether it makes sense to try to skip a database read by updating the game that gets added to the match in memory without re-querying for it. But this may be best as is. More consistent with the db to just re-query

api/helpers/add-game-to-match.js
itsalaidbacklife: @lihaobhsfer do you think this is reasonable and that it would solve the desync issue while avoiding an extra query?

src/components/GameView/GameOverDialog.vue
```diff
@@ -102,20 +148,29 @@ export default {
...
 .logo-image {
   height: auto;
-  max-width: 180px;
-
-  // TODO: replace with an actual css variable for breakpoint width
-  // https://github.com/vuetifyjs/vuetify/blob/v3.1.3/packages/vuetify/src/styles/settings/_variables.scss#L95
-  @media(min-width: 601px) {
-    max-width: 300px;
-  }
+  max-width: 90px;
```
itsalaidbacklife, line 170: let's add max-height as well so the stalemate image isn't super tall
```suggestion
  max-width: 90px;
  max-height: 90px;
```

src/components/GameView/GameOverDialog.vue
```diff
@@ -1,26 +1,43 @@
...
-      </v-card-title>
-      <v-card-text class="d-flex justify-center">
-        <v-img
-          :src="logoSrc"
-          :data-cy="logoDataAttr"
-          class="logo-image"
-        />
+      <div class="d-flex">
+        <v-card-title :data-cy="headingDataAttr" class="mt-8">
+          <h1 class="dialog-header">{{ heading }}</h1>
```
itsalaidbacklife, line 6: In addition to calling ties 'Draw' instead of stalemate, can we update the heading to specify which game this was or whether you won the match? I'm envisioning the headings reading like:

1. You Win Game 1
2. You Lose Game 2
3. Game 3: Draw
4. You Win the Match

Notice the 4th won doesn't say which game it is, just that you won the match.  I hope this can help players keep track of which game is which and when the match is over (I know we have the paragraph text already but I hope this makes it even more clear).

@lihaobhsfer and @seriouslysean what do you think about this from a UX perspective?

src/components/GameView/GameOverDialog.vue
```diff
@@ -1,26 +1,43 @@
...
     <v-card id="game-over-dialog">
-      <v-card-title :data-cy="headingDataAttr">
-        <h1>{{ heading }}</h1>
-      </v-card-title>
-      <v-card-text class="d-flex justify-center">
-        <v-img
-          :src="logoSrc"
-          :data-cy="logoDataAttr"
-          class="logo-image"
-        />
```
itsalaidbacklife, line 12: Let's make the non-ranked dialog look exactly as it does now, without changes (big img in the center) since the match stuff doesn't matter in that context and we may as well continue to use the same real estate. If it's easier to manage, we could split this into two components to handle the ranked and unranked cases and then include both in a higher order component that chooses which dialog to show based on ranked/unranked

src/components/GameView/GameOverDialog.vue
lihaobhsfer: How does
```
Game 1: You Win 
Game 2: You Lose
Game 3: Draw
You Win the Match
```
look like? Fewer conditionals to format the message

itsalaidbacklife: Sounds good

api/controllers/game/concede.js
```diff
@@ -5,14 +5,13 @@ module.exports = async function (req, res) {
     await Game.updateOne(req.session.game).set({ result: winner });
     const game = await gameService.populateGame({ gameId: req.session.game });
     await gameService.clearGame({ userId: req.session.usr });
-    if (game.isRanked) {
-      await sails.helpers.addGameToMatch(game);
-    }
+    const currentMatch = game.isRanked ? await sails.helpers.addGameToMatch(game) : null;
```
seriouslysean, line 8: 👍🏼

api/controllers/game/pass.js
```diff
@@ -56,7 +57,7 @@ module.exports = function (req, res) {
         };
         await Game.updateOne({ id: game.id }).set(gameUpdates);
         if (game.isRanked) {
-          await sails.helpers.addGameToMatch(game);
+          victory.currentMatch = await sails.helpers.addGameToMatch(game);
```
seriouslysean, line 60: We have a pattern of doing this elsewhere (read: above) but are we ok adding this directly to the victory object?

api/helpers/add-game-to-match.js
seriouslysean: I'm curious to the answer above, but otherwise I'm fine with what you both decide. It might be worth adding a follow up to look at this whole chain in detail and rearchitect it, but that's not the point of this PR as long as it's working and tests well.

src/components/GameView/GameOverDialog.vue
seriouslysean: I agree with this. and if you need to split this I think it should be a follow up as long as you can reasonably finish this without making too many forks in the template.

src/components/GameView/GameOverDialog.vue
```diff
@@ -46,16 +70,20 @@ export default {
...
+      playingFromDeck: ({ game }) => game.playingFromDeck,
+      game: ({ game }) => game,
+    }),
+    ...mapGetters(['player', 'opponent']),
     heading() {
-      if (this.stalemate) {
-        return 'Stalemate';
+      if (this.currentMatchWinner) {
+        return this.playerWinsMatch ? 'You Win the Match' : 'You Lose the Match';
       }
```
seriouslysean, line 81: It seems like this is a little unclear to me. We're checking if there's a winner at all and then switching based on that first? Otherwise we then check... sort of the same stuff below? Maybe I'm missing something but I'd prefer this to be much more straightforward with the logic and perhaps use a map of conditions, or a switch instead (I hate you for using them @itsalaidbacklife, lol, but it makes sense here).

src/components/GameView/GameOverDialog.vue
```diff
@@ -90,6 +118,30 @@ export default {
 
       return 'loss-img';
     },
+    currentMatch() {
+      return this.game.currentMatch;
+    },
+    matchGameStats() {
+      const currentMatchGames = this.game.currentMatch?.games ?? [];
+      return currentMatchGames.map((game) => {
+        switch (game.result) {
```
seriouslysean, line 127: This is still pretty confusing, and it seems like a lookup map might do better, but as long as it works I'm ok with it for now. It feels like a refactor would be well warranted with better naming of some of the conditionals.

seriouslysean (approved):
Overall, I'm good with this but haven't fully regressed this entire feature. Leaving final approval to @itsalaidbacklife and @peterancourt. Good job putting this together, and thanks for working through the feedback @lihaobhsfer! 👍🏼

src/components/GameView/GameOverDialog.vue
itsalaidbacklife: It's tricky largely because the stored data is a bit funky to work with. It stores which user id was p0 and which was p1, and it stores whether p1 or p0 won, so then given the fact that p0 won (result = 0) you need to check if the current player was p0 or p1. I think it may make sense to revise the data architecture to store the user id of who won rather than the pNum of them (or maybe both)

itsalaidbacklife: I think this function is reasonable given what's currently stored

src/components/GameView/GameOverDialog.vue
itsalaidbacklife: The tricky bit here is that we want to do different stuff based on whether the match is over and also whether you won or not. That means there's a few different variables which makes it difficult to use a switch since we have different things to compare.

I think this might be easier to understand if we rename `currentMatchWinner` to `matchIsOver` and `playerWins` to `playerWinsGame`. Idea being to differentiate the stuff about the match from the stuff about the latest game, both of which factor into the dialog header

itsalaidbacklife (approved):
Great work. Last two things and we're good to merge.

1. Please rename `currentMatchWinner` to `matchIsOver` and `playerWins` to `playerWinsGame`
2. Please bump the version in package.json a minor-version to 4.3.0

Very nicely done

itsalaidbacklife (approved):
Excellent work

Issue #328, opened 2023-02-06
[Bug]: GameOverlays headings are too large on portrait mobile


itsalaidbacklife:
### Bug Summary

All the headings in the GameOverlays (except in MoveChoiceOverlay) are too large on portrait mobile. They should be responsively shrunk for those cases

### Detailed Description

See image
![image](https://user-images.githubusercontent.com/6520704/217015644-7d7bed1a-4df1-4758-a025-9deecd0660ed.png)

We'd like this to instead fit in one line

### Steps to Reproduce

1.
2.
3.

PR #341, opened 2023-02-24
Game overlays heading mobile #328

TommySchmid:
## Please check the following

- For New Features:


src/components/GameView/GameOverlays.vue
```diff
@@ -148,4 +148,11 @@ export default {
     font-family: 'PT Serif', serif !important;
   }
 }
+
+@media (max-width: 600px) {
+  .text-h3 {
+      text-align: center;
+      font-size: 30px !important;
```
seriouslysean, line 155: We should not be using important here.

seriouslysean (changes requested):

TommySchmid:
Here are a few example screenshots of what these changes look like
Screen Shot 2023-03-01 at 2 28 28 PM
Screen Shot 2023-03-01 at 2 28 52 PM
Screen Shot 2023-03-01 at 4 15 34 PM
Screen Shot 2023-03-01 at 4 15 49 PM

src/components/GameView/GameOverlays.vue
```diff
@@ -129,6 +141,9 @@ export default {
     showWaitingForOpponentToPlayFromDeck() {
       return this.waitingForOpponentToPlayFromDeck && !this.showWaitingForOpponentToDiscardJackFromDeck;
     },
+    useMobileStyling() {
+      return this.$vuetify.display.xs;
```
seriouslysean, line 145: Just online this in the template, no need to obfuscate it.

src/components/GameView/GameOverlays.vue
```diff
@@ -148,4 +163,9 @@ export default {
     font-family: 'PT Serif', serif !important;
   }
 }
```
seriouslysean, line 165: I think this is already a vuetify text style

src/components/GameView/GameOverlays.vue
```diff
@@ -5,42 +5,54 @@
...
     </v-overlay>
     <v-overlay
       id="waiting-for-opponent-stalemate-scrim"
       v-model="waitingForOpponentToStalemate"
       class="game-overlay"
     >
-      <h1 class="text-h3">Opponent Considering Stalemate Request</h1>
+      <h1 :class="[this.$vuetify.display.xs === true ? 'text-h5' : 'text-h3']">
+        <div>Opponent Considering Stalemate Request</div>
+      </h1>
```
seriouslysean, line 55: These are all much better, however, I have a separate problem with these all being H1s. Accessibiity states that each page should only have ONE primary heading. These should all probably be h2s I'd guess. Leave that for another ticket though since they were h1s already.

The classes themselves look good! 👍🏼

src/components/GameView/GameOverlays.vue
```diff
@@ -143,9 +155,13 @@ export default {
   display: flex;
   justify-content: center;
   align-items: center;
+  text-align: center;
+  font-family: 'PT Serif', 'serif';
```
seriouslysean, line 159: I still don't think we should be doing this, this should be a build in text class in Vuetify somewhere and not a one off here. This also feels like a follow up improvement to me since it will likely involve modifying the build system to include or modify a sass config file from vuetify however. cc @itsalaidbacklife

vite.config.js
```diff
@@ -15,6 +16,10 @@ export default defineConfig({
     vue(),
     vuetify({
       autoImport: false,
+      styles: 'sass'
+    }),
+    new VuetifyPlugin({
+      styles: { configFile: 'src/sass/variables.scss'}
```
seriouslysean, line 22: This is not required, we're already doing this above with the `vuetify()` call. Look at the import targets:
- `import vuetify from 'vite-plugin-vuetify';`
- `import VuetifyPlugin from 'vite-plugin-vuetify';`

This is a blocker and this can't go in that way.

src/sass/variables.scss
```diff
@@ -0,0 +1,9 @@
+/**
+  * Vuetify Overrides
+ */
+ @use 'vuetify/settings' with (
+  $body-font-family: (
+    'PT Serif',
+    serif,
+  ),
+ );
```
seriouslysean, line 9: This feels better but will convert the whole app to use this font and I don't think that's out intention. It seems like we only want headers to use this so may need to hit the drawing board and come up with some logical defaults we want to use.

seriouslysean (changes requested):
Some more changes needed here and a config blocker.

src/components/GameView/GameOverlays.vue
TommySchmid: I've removed the `font-family` style as it's now unneeded in this file.

src/sass/variables.scss
seriouslysean: I’ll let @itsalaidbacklife check on it but I’m pretty sure it would since it’s a global variable, haven’t tested it though.

itsalaidbacklife: I'd expect this to set it for the entire app, which would likely be an improvement in that we could hopefully make things more consistent. We should see how it looks, because I do expect this would become a pretty sweeping change. @TommySchmid Could we get some screenshots of the login, home page, and the game view with and without an  overlay?

src/components/GameView/GameOverlays.vue
itsalaidbacklife: I'm not as strong on accessibility so @seriouslysean penny for your thoughts: does the page only rendering one of these (at most, exclusively) reduce the badness? The <template> code does have more than 1 `<h1>` but the rendered html has only 1 at a time

src/components/GameView/GameOverlays.vue
itsalaidbacklife: Because this is now globally configured with the scss var, right?

src/components/GameView/GameOverlays.vue
seriouslysean: Only if this is the only h1 on the page. I’m fine leaving it and eventually doing an accessibility pass.

src/components/GameView/GameOverlays.vue
seriouslysean: Yep, the app should be using that font family var.

src/sass/variables.scss
TommySchmid: Screen Shot 2023-03-17 at 9 32 20 AM
Screen Shot 2023-03-17 at 9 32 54 AM
Screen Shot 2023-03-17 at 9 34 50 AM
Screen Shot 2023-03-17 at 9 34 58 AM
Screen Shot 2023-03-17 at 9 35 11 AM

itsalaidbacklife: I think this looks good

src/components/GameView/GameOverlays.vue
itsalaidbacklife: @TommySchmid it looks like we're still setting the font family on line 159 here. We should be safe to remove this now that the Vuetify theme is setup

TommySchmid: Removed, in addition to `font-family` references in `App.vue` and `StatsView.vue`.

itsalaidbacklife (approved):
Nice work!

Issue #332, opened 2023-02-11
[Feature]: Revise scoring system to 5, 4, 3, 2, 1 for each week


itsalaidbacklife:
### Feature Summary

Currently players earn 1 point for the week if they get a match win without placing in the top 3. We want to recalculate scoring so that winning a match (without placing) for the week earns you 2 points (up from 1) and so that any players who play at least one match during the week without winning at all earn 1 point.

### Detailed Description

We already give 5 points to the week's winner, 4, points for 2nd place, and 3 for third place, so this will add a 2 point possibility, which will make the possible scores for a week a smooth 5, 4, 3, 2, 1. 

It will also solve the problem that currently when a user has no wins for the week (or all season), that they don't get a chip for the week (or the total score), which means we have nothing to click to bring up the menu showing who they've played against. Giving 1 point to players without wins will make it easier for them to see who they've already played against.

The chips for players with a 1 point score for the week should be the outlined variant and grey, or black in color

itsalaidbacklife:
Resolved by #342

PR #342, opened 2023-02-24
[Feature]: Revise scoring system to 5, 4, 3, 2, 1 for each week

peterancourt:
## Please check the following

- [/] Do the tests still pass? (see [Run the Tests](https://github.com/cuttle-cards/cuttle#run-the-tests))
- [/] Is the code formatted properly? (see [Linting (Formatting)](https://github.com/cuttle-cards/cuttle#linting-formatting))
- For New Features:

With this change, those who have one at least one match in a given week without placing will receive 2 points, and those with at least one match played but no wins will receive 1 point

**To Test:**
* In `tests/e2e/specs/out-of-game/stats.spec.js`, add `.only` to `it('Displays Headers, Cards, and Table')`
* Run the test
* _Player4_ should now have 1 point in week 2
* _Player5_ should now have 2 points in week 2

Screen Shot 2023-02-24 at 11 29 40 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "4.0.7",
+  "version": "4.1.7",
   "dependencies": {
     "connect-redis": "3.0.2",
     "dayjs": "1.11.3",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "4.1.7",
+  "version": "4.1.0",
   "dependencies": {
     "connect-redis": "3.0.2",
     "dayjs": "1.11.3",
--- a/src/components/StatsLeaderboard.vue
+++ b/src/components/StatsLeaderboard.vue
@@ -109,7 +109,7 @@ export default {
       }
 
       return this.season.rankings.map((playerStats, index) => {
-        const playerWins = this.playerWins[index];
+        const playerMatches = this.playerMatches[index];
         const playerScores = this.playerScores[index];
         // Count matches completed per week and in total
         const playerMatchesCount = Object.values(playerStats.matches).reduce(
@@ -122,34 +122,36 @@ export default {
         const res = {
           username: playerStats.username,
           week_total_count: playerMatchesCount,
-          week_total_wins: playerWins.total,
+          week_total_wins: playerMatches.totalWins,
           week_total_points: playerScores.total,
           week_total: playerScores.total,
-          rank: this.rank({ totalScore: playerScores.total, totalWins: playerWins.total }),
+          rank: this.rank({ totalScore: playerScores.total, totalWins: playerMatches.totalWins }),
         };
         for (const weekNum in playerStats.matches) {
           res[`week_${weekNum}`] = playerScores[weekNum];
           res[`week_${weekNum}_count`] = playerStats.matches[weekNum].length;
-          res[`week_${weekNum}_wins`] = playerWins[weekNum].wins;
+          res[`week_${weekNum}_wins`] = playerMatches[weekNum].winCount;
           res[`week_${weekNum}_points`] = playerScores[weekNum];
         }
         return res;
       }).sort((player1, player2) => player1.rank - player2.rank);
     },
-    playerWins() {
+    playerMatches() {
       if (this.noSeasonRankingsExist) {
         return [];
       }
       return this.season.rankings.map((playerStats) => {
-        const res = { total: 0 };
+        const res = { totalWins: 0 };
         const playerMatches = playerStats.matches;
         for (const weekNum of this.weeks.map((week) => week.value)) {
           const matchesThisWeek = playerMatches[weekNum] || [];
           const wins = matchesThisWeek.filter((match) => match.result === Result.WON);
-          res.total += wins.length;
+          const losses = matchesThisWeek.filter((match) => match.result === Result.LOST);
+          res.totalWins += wins.length;
           res[`${weekNum}`] = {
-            wins: wins.length,
-            matchesPlayed: matchesThisWeek.length,
+            winCount: wins.length,
+            lossCount: losses.length,
+            totalCount: matchesThisWeek.length,
           }
         }
         return res;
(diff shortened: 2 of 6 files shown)


src/components/StatsLeaderboard.vue
```diff
@@ -84,8 +84,11 @@ export default {
     };
   },
   computed: {
+    noSeasonRankingsExist() {
+      return !this.season || !this.season.rankings || this.season.rankings.length === 0
+    },
```
peterancourt, line 89: This snippet was used in a lot of different places, so I added a variable for it

src/components/StatsLeaderboard.vue
```diff
@@ -144,13 +147,16 @@ export default {
           const matchesThisWeek = playerMatches[weekNum] || [];
           const wins = matchesThisWeek.filter((match) => match.result === Result.WON);
           res.total += wins.length;
-          res[`${weekNum}`] = wins.length;
+          res[`${weekNum}`] = {
+            wins: wins.length,
+            matchesPlayed: matchesThisWeek.length,
+          }
```
peterancourt, line 153: To avoid excess repetition, I added the number of matches played by a user in a given week to `playerWins` instead of creating a new variable. It feels like it fits here

itsalaidbacklife: This makes sense in the control flow but it smells a bit; I think mostly because the existing setup is already a bit funky. I haven't fully baked my thoughts on the issue, but two points feel a bit weird:

1. It seems odd that the playerWins computed prop includes matches that you lost. This is mostly a naming issue--maybe the computed prop could be called playerMatches and its constituent objects could have `winCount` and `matchCount` attributes

2. It feels a bit funky that we have match count, wins, and points as these relevant variables for each player and two out of three are grouped together while one (points) is separated. I wonder if it makes sense to set things up so that points makes its way into the same object as the point and match counts. That might be weird in terms of control flow, given that we need the first two to calculate the third. So maybe this change doesn't make sense. What do y'all think @peterancourt and @seriouslysean ?

src/components/StatsLeaderboard.vue
```diff
@@ -144,13 +147,16 @@ export default {
           const matchesThisWeek = playerMatches[weekNum] || [];
           const wins = matchesThisWeek.filter((match) => match.result === Result.WON);
           res.total += wins.length;
-          res[`${weekNum}`] = wins.length;
+          res[`${weekNum}`] = {
+            wins: wins.length,
+            matchesPlayed: matchesThisWeek.length,
```
itsalaidbacklife, line 152: I believe that matchesThisWeek could have incomplete matches. We should double check this, but we might need to `.filter()` this based on the match result to only include complete matches

itsalaidbacklife (changes requested):
Great work. Requesting 4 changes:

1. Please bump us a minor version in package.json
2. How do the black chips look? Can you add a screenshot with the black color and one with say neutral-darken-3 for comparison? Which looks better?
3. I mentioned some potential refactoring in the comments. Let's discuss whether my suggestions make sense
4. Let's update at least one of the test cases to confirm that the 1-point chip shows up, and that you don't get a point for an incomplete match

src/components/StatsLeaderboard.vue
seriouslysean: I like `playerMatches` instead. Avoid the naming issue-- I think they belong together but agree naming was the issue. Re: organization issue, it feels like `wins` and `matchesPlayed` belong together, with the consideration we change to count variables like @itsalaidbacklife said.

Since you need those two the calculate the third, it should probably become a variable you calculate in vue as a computed prop. Does the server really care about it at all?

Alternatively, would it be better to just return `matches` and have it be a flatMap of wins losses that you can calculate in Vue? Not sure the server needs to know unless we're putting together a completely full payload. Otherwise we can call player matches and do something to filter the mapping for wins or losses, or count it for total matches.

seriouslysean: I keep looking at this and I think I'm starting to agree with @itsalaidbacklife more-- `winCount` and `matchCount` make sense. Can we just do something like this in here (with the assumption that server side processing is faster):
```
res[`${weekNum}`] = {
            wins: wins.length,
            losses: matchesThisWeek.length - wins.length,
            matches: matchesThisWeek.length,
};
```
What about stalemates? Do they count as a loss here?

itsalaidbacklife: There are no stalemate matches only incomplete matches (stalemate games don't count towards completing matches and a match doesn't end until one player gets two wins)

src/components/StatsLeaderboard.vue
itsalaidbacklife: I stand corrected; there are no incomplete matches in the DTO; this should work as written

src/components/StatsLeaderboard.vue
itsalaidbacklife: There are no stalemate matches. A match can be incomplete in the database, but the endpoint serving the stats data filters them out, so all the matches here should be complete, and your losses are all the matches you didn't win. @peterancourt let's rename the computed prop `playerMatches` -- I could go either way on whether to include `losses` explicitly.

For explicitness sake, how about we call the attributes `winCount`, `lossCount`, `totalCount`?

itsalaidbacklife (commented):
@peterancourt excellent work; almost there. Let's rename the one computed prop per the only open comment, and can you please also update the `StatsScoringDialog` to reflect the changes in scoring?

Change top 4 to top 8 in the opening paragraph:
> Competitive Cuttle is divided into 4 seasons per year, one per suit: Clubs, Diamonds, Hearts, and Spades. At the end of each season, the top **_4_** players play a double elimination championship tournament and the

Change the bit about winning a match to say you get two points:

```
          <v-list-item> Each other player who won at least one match gets 1 point for the week </v-list-item>
```
becomes

```
          <v-list-item> Each other player who won at least one match gets 2 points for the week </v-list-item>
          <v-list-item> Each other player who completed a match without winning gets 1 points for the week </v-list-item>
```

Also the bottom paragraph: top 4 => top 8

> You can view the statistics and rankings for each season on this page. At the end of each season,
          the top **_4_** players compete in a double elimination championship tournament. Do you have what it takes
          to become

src/components/StatsLeaderboard.vue
peterancourt: sounds good to me 👍

peterancourt:
@itsalaidbacklife here's some options for the color of the chip, both outlined and filled in. Let me know which you prefer!

**Neutral darken 3 (Outline only)**
Screen Shot 2023-03-08 at 11 48 30 AM

**Neutral darken 3**
Screen Shot 2023-03-08 at 11 54 12 AM

**Black**
Screen Shot 2023-03-08 at 11 56 58 AM

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "4.0.7",
+  "version": "4.1.7",
```
itsalaidbacklife, line 4: The patch version resets to 0 when we bump a minor version
```suggestion
  "version": "4.1.0",
```

peterancourt: Ah right, silly mistake

itsalaidbacklife (approved):
Excellent work. This resolves #332 and makes a significant improvement to the balance of the competitive system while also resolving a QoL issue around seeing who you've played when you've lost all your matches

PR #345, opened 2023-03-03
Feature/spectate games

itsalaidbacklife:
## Please check the following

- For New Features:

**Work in progress**
Creates a new spectate mode so users can watch other people play in real time. Adds tests accordlingly


itsalaidbacklife (commented):
Checkpoint. The fundamentals are in place and the feature works in its essential form. I've got some additional case handling to do and some cleanup -- @seriouslysean I'd like to go over some organizational and arch considerations

src/plugins/sails.js
```diff
@@ -169,7 +169,7 @@ io.socket.on('gameCreated', function (evData) {
 });
 
 io.socket.on('gameStarting', function (evData) {
-  store.commit('removeGame', evData);
+  store.commit('gameStarted', evData);
```
itsalaidbacklife, line 172: Changed the name of this mutation because it's no longer removing a game from the list, now it just moves the game over to a different list for the spectatable games

api/controllers/TestController.js
```diff
@@ -41,4 +44,34 @@ module.exports = {
...
+      }
+
+      const user = await User.findOne({username});
+
+      if (!user) {
+        return res.ok(false);
+      }
+
+      // Set session from user data in db
+      req.session.loggedIn = true;
```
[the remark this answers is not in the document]
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/347

tests/e2e/support/commands.js
```diff
@@ -240,6 +267,39 @@ Cypress.Commands.add('playPointsOpponent', (card) => {
     });
 });
 
+/**
+ * @param card {suit: number, rank: number}
+ */
+Cypress.Commands.add('playPointsSpectator', (card, pNum) => {
```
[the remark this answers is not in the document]
itsalaidbacklife: It also goes hand in hand with switching your session data. It might make sense to make this a higher order command that both switches your session data and executes the points move

itsalaidbacklife:
This updates the home page to separate the game list into the ones you can play vs spectate
image

image


image


The specatator is alsways presented with the view of player 0.
image


TODO: Reveal the opponent hand

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -61,98 +64,174 @@ describe('Home - Page Content', () => {
...
-     */
-    cy.createGamePlayer({ gameName: 'Test Game', isRanked: false }).then((gameData) => {
-      // Test that JOIN button starts enabled
-      cy.contains('button.v-btn', 'Play').should('not.be.disabled');
-      // Sign up 2 users and subscribe them to game
-      cy.signupOpponent('secondUser@aol.com', 'myNewPassword');
-      cy.subscribeOpponent(gameData.gameId);
-      cy.signupOpponent('thirdUser@facebook.com', 'anotherUserPw');
-      cy.subscribeOpponent(gameData.gameId);
+  describe('Spectating games', () => {
```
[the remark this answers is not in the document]
itsalaidbacklife: Todo: add test case for cant spectate a completed game

tests/e2e/specs/in-game/spectating.spec.js
```diff
@@ -0,0 +1,58 @@
...
+    }, true);
+  });
+
+
+  it('Continues spectating after page refresh', () => {});
+
+  it('Continues spectating after socket disconnect', () => {});
+
+  it('Prevents spectator from making moves', () => {});
+});
```
itsalaidbacklife, line 58:
```suggestion
});

```

itsalaidbacklife (commented):
@seriouslysean submitting a checkpoint comment review to submit comments. I think this feature is 75% atm

tests/e2e/specs/out-of-game/home.spec.js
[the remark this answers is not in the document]
itsalaidbacklife: Per our conversation, I've split these test into the in-game/specating.spec.js (about receiving updates when moves happen) and the ones here in home.spec.js which pertain to getting into the games

src/store/modules/game.js
```diff
@@ -23,6 +23,7 @@ function resetState() {
     waitingForOpponentToCounter: false,
     myTurnToCounter: false,
     isRanked: false,
+    isSpectating: false,
```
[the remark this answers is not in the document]
itsalaidbacklife: I'm leaning towards taking this out of the store and making it a computed prop in gameView based on the route. I don't think it's used outside the GameView component itself so that seems reasonable.

api/controllers/TestController.js
```diff
@@ -41,4 +44,34 @@ module.exports = {
     }
     return res.ok();
   },
+
+  // Switch to a specified user's session
+  recoverSession: async function (req, res) {
```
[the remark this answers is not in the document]
itsalaidbacklife: TODO: Can we just use `UserController.reLogin`?

src/store/modules/game.js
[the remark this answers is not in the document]
itsalaidbacklife: Moved into computed prop; I figure if it _can_ be computed, it's generally better that it be computed to prevent potential desync

api/controllers/TestController.js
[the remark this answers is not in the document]
itsalaidbacklife: TODO: Can we just use `UserController.reLogin`?

src/plugins/sails.js
```diff
@@ -168,8 +168,12 @@ io.socket.on('gameCreated', function (evData) {
   store.commit('addGameToList', newGame);
 });
 
-io.socket.on('gameStarting', function (evData) {
-  store.commit('removeGame', evData);
+io.socket.on('gameStarting', function ({ gameId }) {
```
itsalaidbacklife, line 171: Should normalize the tense here. Events make more sense in past tense imo so `gameStarted`

itsalaidbacklife (commented):
Adding comments with notes and questions

tests/e2e/support/commands.js
```diff
@@ -197,6 +247,18 @@ Cypress.Commands.add('leaveLobbyOpponent', (id) => {
...
+Cypress.Commands.add('recoverSessionOpponent', (username) => {
+  return new Cypress.Promise((resolve, reject) => {
+    io.socket.get('/test/recoverSession', { username }, function handleResponse(res, jwres) {
+      if (jwres.statusCode !== 200 || res === false) {
+        return reject(new Error('error recovering session'));
+      }
+      return resolve();
+    });
+  });
+});
```
itsalaidbacklife, line 260: This might work by hitting the existing reLogin endpoint, which would save the need to add a new testing-only endpoint. I'd also like to revise the command itself to input a player object with username and password so it can accept the user fixtures exactly as exported in the fixture file

tests/e2e/support/commands.js
```diff
@@ -240,6 +303,39 @@ Cypress.Commands.add('playPointsOpponent', (card) => {
     });
 });
 
+/**
+ * @param card {suit: number, rank: number}
+ */
+Cypress.Commands.add('playPointsSpectator', (card, pNum) => {
```
itsalaidbacklife, line 309: @seriouslysean We talked about wrapping this into a separate command -- atm since points is the only move the tests require I think it's okay to leave this as is and to revisit it as a part of the follow up issue where we add test cases for a wider range of moves

tests/e2e/support/helpers.js
```diff
@@ -245,7 +245,7 @@ function assertDomMatchesFixture(pNum, fixture) {
     } else {
       cy.get('[data-opponent-hand-card]').should('not.exist');
     }
-    if (playerHasGlasses) {
+    if (playerHasGlasses || spectating) {
```
itsalaidbacklife, line 248: If we're asserting during spectate mode we expect the opponent hand cards to be visible

tests/e2e/support/helpers.js
```diff
@@ -377,6 +377,7 @@ export const SnackBarError = {
   ILLEGAL_SCUTTLE:
     "You can only scuttle an opponent's point card with a higher rank point card, or the same rank with a higher suit. Suit order (low to high) is: Clubs < Diamonds < Hearts < Spades",
   FROZEN_CARD: 'That card is frozen! You must wait a turn to play it',
+  NOT_IN_GAME: "You can't do that because you're not a player in this game",
```
itsalaidbacklife, line 380: It might make sense to move this object into /fixtures. Same goes for the Card enum. @seriouslysean thoughts?

itsalaidbacklife (commented):
Few more comments with considerations for future issues

tests/e2e/specs/in-game/spectating.spec.js
```diff
@@ -0,0 +1,184 @@
...
+    cy.get('[data-move-choice=targetedOneOff]').click();
+    cy.get('[data-opponent-face-card=13-2]').click();
+    assertSnackbarError(SnackBarError.NOT_IN_GAME);
+    cy.log('Correctly prevented from targeted one-off');
+
+    // Can't resolve three
+    // Can't resolve four
+    // Can't resolve seven
+    // Can't counter
+    // Can't resolve
```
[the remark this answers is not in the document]
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/349

api/controllers/game/spectate.js
```diff
@@ -0,0 +1,19 @@
...
+    const game = await gameService.populateGame({ gameId });
+
+    if (game.status || game.players.length !== 2) {
+      return res.badRequest({message: 'You can only spectate an ongoing game with two players'});
+    }
+
+    // TODO: Can't spectate a game you're in
+
+    // Subscribe socket to game
+    Game.subscribe(req, [ gameId ]);
```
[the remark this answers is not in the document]
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/350

config/policies.js
```diff
@@ -36,6 +36,7 @@ module.exports.policies = {
   'game/create': ['isLoggedIn', 'hasGameName'],
   'game/get-list': 'isLoggedIn',
   'game/subscribe': 'isLoggedIn',
+  'game/spectate': ['isLoggedIn'],
```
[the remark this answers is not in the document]
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/351

src/views/GameView.vue
```diff
@@ -442,6 +443,12 @@ export default {
       'resolvingSeven',
       'hasGlassesEight',
     ]),
+    isSpectating() {
+      return this.$router.currentRoute.value.name === ROUTE_NAME_SPECTATE;
+    },
```
itsalaidbacklife, line 448: TODO: show which one-off's and counters are on the stack so that spectators can see which things have been played while waiting for an opponent to counter

itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/353

api/controllers/game/concede.js
```diff
@@ -22,6 +22,10 @@ module.exports = async function (req, res) {
         victory,
       },
     });
+
+    // Inform all clients this game is over
+    sails.sockets.blast('gameFinished', { gameId: game.id });
```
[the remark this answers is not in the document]
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/354

api/controllers/TestController.js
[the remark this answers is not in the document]
itsalaidbacklife: Yes we can. Updated to use UserController.reLogin and avoid the need for a new endpoint

api/controllers/UserController.js
```diff
@@ -50,33 +51,37 @@ module.exports = {
...
-        Game.publish([game.id], {
-          verb: 'updated',
-          data: {
-            ...game.lastEvent,
-            game,
-          },
-        });
-
+        // Set session data & send socket msg if in game
+        if (game) {
```
seriouslysean, line 74: This seems new, and now it's behind a check for game's existence. Do we ever expect this to not be set at this point?

api/controllers/game/concede.js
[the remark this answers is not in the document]
seriouslysean: As long as it's not majorly breaking anything (and it's already broken today), I think forking it off is a better idea since I much prefer more concise and pointed PRs.

api/controllers/game/get-list.js
```diff
@@ -4,11 +4,15 @@ module.exports = async function (req, res) {
   // HANDLE REQ.SESSION.GAME = GAME ID CASE
   sails.sockets.join(req, 'GameList');
   try {
-    const games = await gameAPI.findOpenGames();
+    const [ openGames, spectatableGames ] = await Promise.all([
```
seriouslysean, line 7:
```suggestion
    const [ openGames, spectatableGames ] = await Promise.all([
```
I really don't like `spectatableGames` as a name, but I don't have a better alternative.

api/controllers/game/ready.js
```diff
@@ -28,7 +28,7 @@ module.exports = function (req, res) {
         }
         if (bothReady) {
           // Inform all clients this game is starting
```
seriouslysean, line 30:
```suggestion
          // Inform all clients this game has started
```
Stick with your new case. I usually prefer the now tense, but I see why you've don it this way instead. As long as we're consistent I don't mind either way.

api/controllers/game/ready.js
```diff
@@ -28,7 +28,7 @@ module.exports = function (req, res) {
...
-          sails.sockets.blast('gameStarting', { gameId: game.id });
+          sails.sockets.blast('gameStarted', { gameId: game.id });
```
seriouslysean, line 31: Make sure anything using this identifier is aware of this new name. It think these ids should live in a constants file we share eventually so we can make sure it's always correct, something like

```
const SOCKET_EVENT_GAME_STARTED = 'gameStarted';
```

api/controllers/game/spectate.js
```diff
@@ -0,0 +1,17 @@
+module.exports = async function (req, res) {
+  try {
+    const { gameId } = req.body;
+    const game = await gameService.populateGame({ gameId });
+
+    if (game.status || game.players.length !== 2) {
```
seriouslysean, line 6:
```suggestion
    if (game.status || game.players.length < 2) {
```
It's probably much better to treat this like an actual number and use less then. Same effect but easier to understand imo. Unless you need it to be 2 exactly, in which case we can leave it. Wanted to pose the suggestion either way.

api/controllers/game/spectate.js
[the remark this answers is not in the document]
seriouslysean: I think this is a totally fair compromise to make for MVP. Get it out there and working to make sure there aren't any major bugs, then we can work on tracking and validation and stuff later.

api/helpers/ find-spectatable-games.js
```diff
@@ -0,0 +1,23 @@
+const dayjs = require('dayjs');
+
+module.exports = {
+  friendlyName: 'Find Spectatable Games',
+
+  description: 'Finds all the games that are available to spectate',
+
+  fn: async (_, exits) => {
+
+    const recentUpdateThreshhold = dayjs().subtract(10, 'minute').valueOf();
```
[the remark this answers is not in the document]
seriouslysean: I think 10 minutes feels too long, 5 might be better. If you think about it no one is sitting on a turn for more than a minute or two at a time, and at 10 minutes I'm afraid you're just going to constantly jump in to empty games (though we never really have that many).

No harm going full bore first and tightening this up as needed, I'll leave it to you since you play more often than I do.

api/helpers/ find-spectatable-games.js
```diff
@@ -0,0 +1,23 @@
...
+
+    try {
+      const games = await Game.find({
+        status: false, // not open to join
+        result: -1, // not finished yet
```
seriouslysean, line 15: This is a weird key name for this, but it's probably based on the DB so I don't think it matters as much. It's a little confusing to me though.

api/policies/isInGame.js
```diff
@@ -16,5 +16,5 @@ module.exports = function (req, res, next) {
     }
     return res.badRequest({ message: "Error with user's session: 'game' must be an integer" });
   }
-  return res.badRequest({ message: 'Session Error: You must join a game to perform that move' });
+  return res.badRequest({ message: "You can't do that because you're not a player in this game" });
```
seriouslysean, line 19:
```suggestion
  return res.badRequest({ message: 'You can not do that because you are not a player in this game' });
```
I think removing the contractions is also an ok middle ground, doesn't change much unless you want it more informal. Could also use backticks.

src/components/GameListItem.vue
```diff
@@ -16,17 +16,28 @@
...
       </v-col>
       <v-col cols="3" class="list-item__button">
+        <!-- Join Button -->
         <v-btn
-          color="primary"
-          rounded
-          variant="outlined"
-          min-width="200"
+          v-if="!isSpectatable"
+          v-bind="buttonAttrs"
```
seriouslysean, line 22: I know why you moved these since you're re-using them, but it makes it harder to figure out the button style at a glance. Don't hate it, but there's a slight cognitive impact now.

src/components/GameListItem.vue
```diff
@@ -16,17 +16,28 @@
         <p>{{ readyText }} players</p>
       </v-col>
       <v-col cols="3" class="list-item__button">
+        <!-- Join Button -->
```
seriouslysean, line 19: Did you test joining/leaving a game to make sure it's updating immediately? If a sync is missed somehow, do we need a poll or regular check in to make sure it's good to go?

src/plugins/sails.js
```diff
@@ -54,12 +54,21 @@ io.socket.on('game', function (evData) {
...
+
+          // Update state
           store.commit('resetState');
           store.dispatch('updateGameThenResetPNumIfNull', evData.data.game);
-          const gameRoute = `/game/${store.state.game.id}`;
-          const currentRoute = router.currentRoute.fullPath;
-          if (gameRoute !== currentRoute) {
-            router.push(gameRoute);
+          if (isSpectating) {
+            store.commit('setMyPNum', 0); // always spectate as p0
```
seriouslysean, line 64: Is there a follow up to look in to this? Or are we just using this as the baseline and changing based on player vs. observer to show cards, etc?

src/plugins/sails.js
```diff
@@ -54,12 +54,21 @@ io.socket.on('game', function (evData) {
...
+          }
+
+          // Validate current route & navigate if incorrect
+          const expectedRoutePrefix = isSpectating ? 'spectate' : 'game';
```
seriouslysean, line 68: These should use route constants if we have them (which I think we do). Might need to move them to a more local position so it can be shared though. If it's too much work or annoying you can make a ticket for it.

src/plugins/sails.js
```diff
@@ -198,9 +211,14 @@ io.socket.on('leftGame', function (evData) {
...
   // Request latest game state if socket reconnects during game
   const { username } = store.state.auth;
-  const inGame = router.currentRoute.value.name === ROUTE_NAME_GAME;
-  if (!inGame) {
-    return;
+  switch (router.currentRoute.value.name) {
+    case ROUTE_NAME_GAME:
+      return store.dispatch('requestReauthenticate', { username });
+    case ROUTE_NAME_SPECTATE: {
+      const gameId = Number(router.currentRoute.value.params.gameId);
```
seriouslysean, line 218: You might need to be defensive here. Just because you're coercing it doesn't mean it's valid.

src/store/modules/game.js
```diff
@@ -323,6 +324,26 @@ export default {
...
+        {
+          gameId,
+        },
+        function handleResponse(res, jwres) {
+          if (jwres.statusCode === 200) {
+            context.commit('updateGame', res);
+            context.commit('setMyPNum', 0);
+            return resolve();
+          }
+          return reject(new Error('Unable to spectate game'));
```
seriouslysean, line 341: You probably want the actual error message in here for debugging, unless this goes right to the snack bar.

src/store/modules/gameList.js
```diff
@@ -8,51 +8,75 @@ class GameSummary {
     this.numPlayers = Object.prototype.hasOwnProperty.call(obj, 'players') ? obj.players.length : 0;
     this.status = Object.prototype.hasOwnProperty.call(obj, 'status') ? obj.status : false;
     this.isRanked = Object.prototype.hasOwnProperty.call(obj, 'isRanked') ? obj.isRanked : false;
+    this.isOver = false;
```
seriouslysean, line 11: Should this be hardcoded here?

src/store/modules/gameList.js
```diff
@@ -8,51 +8,75 @@ class GameSummary {
...
   }
 }
 export default {
   state: {
-    games: [],
+    openGames: [],
+    spectateGames: [],
```
seriouslysean, line 17: Would it make more sense to keep games, and then have two computed getters that are filtered versions of the state for caching purposes (`spectatableGames` and `openGames`)?

src/store/modules/gameList.js
```diff
@@ -8,51 +8,75 @@ class GameSummary {
...
       return new Promise((resolve, reject) => {
         io.socket.get('/game/getList', function handleResponse(resData, jwres) {
           if (jwres.statusCode === 200) {
-            const games = cloneDeep(resData.games);
-            context.commit('refreshGames', games);
-            return resolve(resData.games);
+            const openGames = cloneDeep(resData.openGames);
+            const spectatableGames = cloneDeep(resData.spectatableGames);
+            context.commit('refreshGames', openGames);
+            context.commit('refreshSpectateGames', spectatableGames);
```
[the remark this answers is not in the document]
seriouslysean: I agree with this given we never, ever need to call them separately. A mutation is generally a single update so I don't have two of them here. You could make a n action called refresh games and have these done together but still separate.

src/views/GameView.vue
```diff
@@ -59,7 +59,7 @@
               <div id="opponent-hand-cards" class="d-flex justify-center align-start">
                 <transition name="slide-below" mode="out-in">
                   <transition-group
-                    v-if="hasGlassesEight"
+                    v-if="revealOpponentHand"
```
seriouslysean, line 62:
```suggestion
                    v-if="showOpponentHand"
```
Reveal is kinda weird, I like show for the simplicity but I'm open to whatever.

src/views/GameView.vue
```diff
@@ -702,6 +699,11 @@ export default {
     },
   },
   mounted() {
+    if (this.isSpectating && !this.$store.state.game.id) {
```
seriouslysean, line 702: Are we defensive enough if there's no game set in the store? If it's null or undefined I think this would bomb. Util we have a redirect in place for games that aren't valid or don't exist this might be an issue.

src/views/GameView.vue
```diff
@@ -702,6 +699,11 @@ export default {
...
+      const { gameId } = this.$router.currentRoute.value.params;
+      this.$store.dispatch('requestSpectate', Number(gameId));
```
seriouslysean, line 704: Same here, this may not be defensive enough since it will check any game id from the route params.

src/views/HomeView.vue
```diff
@@ -19,18 +19,49 @@
...
+                      :status="game.status"
+                      :num-players="game.numPlayers"
+                      :is-ranked="game.isRanked"
+                      :isSpectatable="true"
+                      :disableSpectate="game.isOver"
+                      @error="handleError"
+                    />
+                  </div>
+                </v-window-item>
+              </v-window>
```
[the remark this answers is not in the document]
seriouslysean: I think this feels fine for a UX improvement/refactor. Try to keep a separation of concerns for this particular implementation, it's already pretty complex.

src/views/HomeView.vue
```diff
@@ -104,13 +135,15 @@ export default {
   },
   data() {
     return {
+      tab: 'play',
```
seriouslysean, line 138: You should probably have consts in component for the valid tabs and use those instead.

src/views/HomeView.vue
```diff
@@ -104,13 +135,15 @@ export default {
...
       showSnackBar: false,
       snackBarMessage: '',
     };
   },
   computed: {
     ...mapState({
-      gameList: ({ gameList }) => gameList.games,
+      playableGameList: ({ gameList }) => gameList.openGames,
```
seriouslysean, line 145: Do you need a default here? 
```suggestion
      playableGameList: ({ gameList }) => gameList.openGames || [],
```
Ideally this should be handled in the whatever logic is happening in the store though.

seriouslysean (commented):
Initial review. I did a pretty decent pass but haven't tested it yet. Looks really great so far!

api/controllers/UserController.js
itsalaidbacklife: So I may need to roll back on this but the usecase is for switching sessions between "other" users during testing, which we do in the home.spec while both those "other" users are not already in the game so they can ready up etc.

This is a new change as a part of removing my previous implementation of session switching (which used a new testing-only endpoint) in favor of reusing this existing one.

That change broke a reconnection spec so it may need to roll back. That's the CI failure and I need to debug still

api/controllers/game/get-list.js
itsalaidbacklife: Yeah I also don't love it and also  haven't come up with a better alternative. Maybe spectateGames?

Also is this actually a suggestion? I can't actually see the difference 😅

api/controllers/game/ready.js
itsalaidbacklife: Yeah I think that's worth making it's own issue

api/controllers/game/spectate.js
itsalaidbacklife: I think that's reasonable

api/helpers/ find-spectatable-games.js
itsalaidbacklife: That smells; I'd love to pick your brain on how to clarify it. 

So the idea here is to find games that are being actively played. To do so we're querying for games with two criteria:

1. `status: false` -- `status` is a boolean for whether a game is open for players to join. It defaults to true when the game is made and is set to `false` when a 2nd player joins. So `false` means the game has at least two people in it. 

2. `result: -1` -- `result` is an enum describing the end state of a game. -1 is the initial value before a game starts and then when a game ends it's set to 0 if p0 wins, 1 if p1 wins, and 2 for a stalemate. 

In the short term at minimum this should be compared using an enum value (Result.INCOMPLETE) instead of the raw -1. More importantly, I believe there is actually a bug here where games that have two players (so status = false) but that haven't started yet (not everyone is ready) would be turned up by this, possibly letting you spectate a game that hasn't been dealt yet. I'll see if I can replicate in a test case; it should be fixable by checking p0Ready and p1Ready, but that is a bit of a kludge and speaks to the awkwardness of the data architecture here.

In the longer term, I think this hits on the smell of having these two separate fields (result and status) for what is logically more one state on a spectrum of values, ie a game is created, started, and  ends with one of three possible states (p0 win, p1 win, draw). These two columns (status and result) could likely be collapsed into one column that represents which of those states the game is in. It would be important to name it clearly and it would potentially have downsides in that some checks would need to compare this new column to a list of values instead of just one. That point could use further baking, penny for your thoughts. If we can land on a coherent vision for it, I'll write up an issue

api/policies/isInGame.js
itsalaidbacklife: Yeah that's reasonable. I kinda like the informality of the contractions but mixed quote types are funky

src/plugins/sails.js
itsalaidbacklife: Like whether you spectate as p0 or p1? Look into what?

seriouslysean: Meaning should there be a special player num for observers or does p0 do what you require? Feels kind of weird to share a pnum despite knowing it’s just a way to get the events/game moves.

src/store/modules/gameList.js
itsalaidbacklife: I think yes. GameSummaries are first instantiated clients side when we request the games from the server (open and spectate) and the server doesn't send finished games, so all of them are initially `isFinished = false` we only flip this to true when a game was not finished during initial page load and then finishes while we're on the home page (which emits a socket event)

src/store/modules/gameList.js
itsalaidbacklife: Yeah I considered the same and I'm still on the fence. I think a part of the weirdness is that the queries for these two lists of games are different so the backend already has them in different arrays and it seems natural to just keep them separate at that point rather than combining them again in order to separate them via getters.

That said maybe we could rework the query to return both types of games in a single query and then it might make sense to separate them via client side getters

src/views/GameView.vue
itsalaidbacklife: `game` is a store module, so that should generally exist right? It's possible that it will be empty state in which case the `id` will be `null`, but I don't see how this could type error

src/views/HomeView.vue
itsalaidbacklife: Yeah the store itself should always have a value for this which inits to an empty array, so I think this should be fine

src/store/modules/gameList.js
seriouslysean: Sounds like a follow up to me :)

src/views/GameView.vue
seriouslysean: I couldn’t remember if we defaulted to null or an empty object, so wanted to check.

src/views/GameView.vue
seriouslysean: Correct. Not expected, but maybe not impossible.

src/views/HomeView.vue
seriouslysean: Nice, thanks.

api/controllers/UserController.js
```diff
@@ -50,33 +51,37 @@ module.exports = {
...
       .findUserByUsername(req.body.username)
       .then(function gotUser(user) {
+        const { game: gameId, loggedIn } = req.session;
         const checkPass =
-          !req.session.loggedIn && req.body.password
+          !loggedIn && req.body.password
             ? passwordAPI.checkPass(req.body.password, user.encryptedPassword)
             : null;
-        const promiseGame = gameService.populateGame({ gameId: user.game });
+        const promiseGame = gameId ? gameService.findGame({ gameId }) : null;
```
itsalaidbacklife: Oops. Swapped `.populateGame()` for `.findGame()` which doesn't grab the points. This caused the CI failure

src/components/GameListItem.vue
itsalaidbacklife: Yeah the alternative as I see it is to make it just one button with computed props for the text, icon, data-cy attribute, and a higher order click handler that delegates to the correct method based on the prop for spectating. My hunch is that this is simpler; penny for your thoughts

src/components/GameListItem.vue
itsalaidbacklife: What do you mean by updating immediately? I sense you're onto a possible bug but I don't fully grasp the issue

seriouslysean: So imagine opening two tabs, I want to verify that it’s being reactive and that a blast isn’t being missed by either post and leading to stale UX.

src/plugins/sails.js
itsalaidbacklife: Okay yeah I hear you. The main reason for using a 'normal' 'valid' pNum is that it leaves all the existing getter logic around indentifying which stuff goes at the bottom of the screen i.e. the player's cards vs the top i.e. the opponent's cards. Making the spectator one or the other player (atm it's just always p0) means we don't need to make all those other getters and such pay attention to whether your pNum is something else that _isn't_ an index in the array of players, which feels simpler to me.

Instead we keep spectating as a getter based on the current route, which lets consumers who care about that difference use it however they need to while letting most of the presentational logic do it's thing in the usual way. Otherwise I think there will be a lot of places that need to handle the pNum being something other than 1 or 0

src/plugins/sails.js
itsalaidbacklife: Yeah it could also be done via route names instead of paths, which might be less brittle. I think we can do that in the same PR; shouldn't be too big a deal

src/plugins/sails.js
itsalaidbacklife: How should we handle an invalid gameId? You kinda need to go to the homepage at that point, right? So maybe a dialog that explains the problem with a button to go home? Otherwise could be an error snackbar and then going home

api/helpers/ find-spectatable-games.js
seriouslysean: I really like the idea of combining this in to a single call because at the end of the day they’re all just games. We only care about properties on them that say whether they’re observable, playable, etc.

src/store/modules/game.js
itsalaidbacklife: This does go right into the snackbar. The component `.catch()`'s the store's `dispatch()` and puts the error into a snackbar. Generally I like that paradigm as it keeps the errors human readable. I think troubleshooting the specific cause of the error is more of a development/debugging task that doesn't need to be user facing

api/policies/isInGame.js
seriouslysean: Personally I like the contractions but wanted to call out consistency around the strings. Bsckticks would be a good middle ground.

src/store/modules/gameList.js
itsalaidbacklife: And possibly one that rolls into the change regarding `status` and `result` which I think are confusing and could possibly be combined

src/views/GameView.vue
itsalaidbacklife: So the game store module itself is a populated object which defaults to having empty values for the specified keys. It's controlled by the `restState()` function in the `game.js module`:

```
function resetState() {
  return {
    id: null,
    chat: [],
    deck: [],
    log: [],
    name: null,
    p0Ready: false,
    p1Ready: false,
    passes: 0,
    players: [],
    scrap: [],
    turn: 0,
    twos: [],
    myPNum: null,
    topCard: null,
    secondCard: null,
    oneOff: null,
    oneOffTarget: null,
    waitingForOpponentToCounter: false,
    myTurnToCounter: false,
    isRanked: false,
    // Threes
    waitingForOpponentToPickFromScrap: false,
    pickingFromScrap: false,
    // Fours
    discarding: false,
    waitingForOpponentToDiscard: false,
    // Sevens
    playingFromDeck: false,
    waitingForOpponentToPlayFromDeck: false,
    // Last Event
    lastEventChange: null,
    lastEventOneOffRank: null,
    lastEventTargetType: null,
    // GameOver
    gameIsOver: false,
    winnerPNum: null,
    conceded: false,
    waitingForOpponentToStalemate: false,
    consideringOpponentStalemateRequest: false,
  };
}
```

so `game.id` will be `null` by default, but the reference is typesafe

src/views/GameView.vue
itsalaidbacklife: Thoughts on what the UX should be to handle this error? I think we need to navigate home if that happens. The same would apply in the GameView itself when playing

src/plugins/sails.js
seriouslysean: This makes a lot more sense now, thanks. It might make sense to change this long run for clarity but for now, for the MVP, I agree this is the best move for now.

src/plugins/sails.js
seriouslysean: Agreed.

src/plugins/sails.js
seriouslysean: I had a solution for this at one point. It should just be an auto direct to the homepage with a snack bar message (eventually) that it’s an invalid game id.

src/store/modules/game.js
seriouslysean: Makes sense, we can always log separately once you ever set me up that error logging account. Wink wink.

src/store/modules/gameList.js
seriouslysean: Man, I must have looked at this a dozen times and completely missed this was the class usage. Ignore me, haha. This is right, I completely missed this is the original instantiating of the variable.

api/controllers/UserController.js
itsalaidbacklife: Fixed by reworking UserController.reLogin: https://github.com/cuttle-cards/cuttle/pull/345/commits/91fa83a63ac42b99197bc8d029b29929c52c8508

src/views/HomeView.vue
itsalaidbacklife: done

src/store/modules/game.js
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/368

api/helpers/ find-spectatable-games.js
itsalaidbacklife: So another way this could go down would be to have two columns:
`status` which would be an enum and `winner` which would be the id of the user who won.

`status` could allow the values:
`UNSTARTED`, `STARTED`, `FINISHED`

So games that are `game.status === Status.FINISHED` that have `game.winner === null` would be stalemates, and otherwise, the `winner` field would tell you which player won. 

What this doesn't properly capture in and of itself is that the `status` variable currently represents whether the game is joinable. That may or may not be worth capturing separately. In principle you could omit this from the database and use the count of players to determine this i.e. you can't join unless there's < 2 players already in the game. The thing that this sorta misses is the ability to manually close games in the database by flipping their status to false, which I sometimes do to clean them up. There's a way to get all of these, I'm sure, but I don't have it fully mapped out yet.

@seriouslysean this speaks to a documentation question: given that we know this is something that we're interested to refactor but that we don't currently have a clear and specific vision for, should we write up an issue for it? How should we avoid losing the seed of this idea while acknowledging that it's underbaked for execution? Could be an issue with a needs-refinement label perhaps

src/components/GameListItem.vue
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/373

api/helpers/ find-spectatable-games.js
itsalaidbacklife: Test case and improvement is in place to fix the bug around joining unstarted games. And https://github.com/cuttle-cards/cuttle/issues/374 is documented to investigate a reorganization of these database columns

Issue #355, opened 2023-03-13
[Bug]: Fix StatsScoringDialog's reference to the championship tournaments


itsalaidbacklife:
### Bug Summary

The StatsScoringDialog has out of date text and could use some style updates

### Detailed Description

See this annotated image with requested changes. Involves:
1. Switching the button type of the cancel button
2. Adding vertical margin between the top text and the AwardCard's
3. Updating the text at the top and bottom to say 'top 8 players' instead of top 4
4. Adding a sentence about the World Championship

![image](https://user-images.githubusercontent.com/6520704/224726403-cd6d27ec-1581-45d3-94f3-5fb834a81620.png)

All of these changes could be implemented in the `src/components/StatsScoringDialog.vue`

### Steps to Reproduce

1. Navigate to the stats page at https://cuttle.cards/#/stats
2. Click the 'How are Ranks Determined' button
3. Dialog will appear and will have the problems outlined above

Haviles04:
I'd like to take this one!

PR #356, opened 2023-03-16
Bug#355

Haviles04:
## Please check the following

- For New Features:

I added vertical margin for the award card container, fixed the spelling on permanently, added text, changed button variant to text, added styling to make the award card look better on mobile.

### DIFF
--- a/src/components/StatsScoringDialog.vue
+++ b/src/components/StatsScoringDialog.vue
@@ -3,12 +3,7 @@
     <!-- Activator -->
     <template #activator="{ props }">
       <slot name="button">
-        <v-btn
-          v-bind="props"
-          color="primary"
-          class="mb-2"
-          variant="plain"
-          data-cy="ranked-info-button">
+        <v-btn v-bind="props" color="primary" class="mb-2" variant="text" data-cy="ranked-info-button">
           <span v-if="showButtonText">How are ranks determined?</span>
           <v-icon class="ml-1" icon="mdi-information-outline" />
         </v-btn>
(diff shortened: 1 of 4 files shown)


itsalaidbacklife (commented):
Resolves #355 
Looking good! Would you also please replace the trophy icons with `v-chip`'s to match the leaderboard? See `StatsLeaderBoardCell` for an example.

The first, second, and third place chips should have `variant="elevanted"`, and the colors
`this.$vuetify.theme.themes.cuttleTheme.colors.firstPlace`, `this.$vuetify.theme.themes.cuttleTheme.colors.secondPlace` and
`this.$vuetify.theme.themes.cuttleTheme.colors.thirdPlace` respectively

and the 2 point and 1 point chips should have `variant="outlined"` and colors `'primary'` and `'#000'` respectively.

Last, this is a patch-version update, so could you please bump the patch version in package.json?

src/components/StatsScoringDialog.vue
```diff
@@ -41,26 +36,55 @@
         </p>
         <v-list>
           <v-list-item>
-            <v-icon :color="theme.firstPlace" class="mr-2" icon="mdi-medal" />
+            <v-chip
+              variant="elevated"
+              class="mr-2 mb-1"
+              :color="this.$vuetify.theme.themes.cuttleTheme.colors.firstPlace"
```
itsalaidbacklife, line 42: For brevity's sake, let's define a computed property called `theme` that returns `theme()` that returns `this.$vuetify.theme.themes.cuttleTheme.colors` so that each of these colors can just be `theme.firstPlace`, `theme.secondPlace`, etc

itsalaidbacklife: Ha! I see we have a `theme` comptuted property already but it's out of date. The above snippet is what it should return: 
```
theme() {
  return this.$vuetify.theme.themes.cuttleTheme.colors;
}
```

itsalaidbacklife: Also there is a bit of a gotcha in Vue that the html/template code _doesn't_ need **this**, but the javascript _does._ So even without the computed property these colors should look like

```
:color="$vuetify.theme.themes.cuttleTheme.colors.firstPlace" // notice I removed 'this'
```
but this will be cleaner if we do fix the `theme` computed prop as written above and then this will just be

```
:color="theme.firstPlace"
```
which is easier to read and less repetitive given that we use the same expression 4x here

itsalaidbacklife (approved):
Looks great! Requesting one refactoring change about how the colors are written (will look the same in the UI). Nice work -- thanks for your help!

src/components/StatsScoringDialog.vue
Haviles04: Definitely reads much better this way, good call!

PR #365, opened 2023-03-20
fix: set allowNull: true on non-required model attributes

itsalaidbacklife:
## Please check the following

- For New Features:

This addresses a console warning when connecting to a live db: 
```
Warning: After transforming columnNames back to attribute names for model `season`,
 a record in the result has a value of `null` for property `bracketLink`.
Since the `bracketLink` attribute declares `type: 'string'`,
without ALSO declaring `allowNull: true`, this `null` value is unexpected.
(To resolve, either change this attribute to `allowNull: true` or update
existing records in the database accordingly.)
```

This doesn't pop up until connecting to a postgres database with pre-existing data, which is why it doesn't happen during local development. Basically in addition to the value not being required, it also needs to explicitly allow null to work properly with how postgres sets missing values as null


itsalaidbacklife:
This needs a migration to move the old `0`-valued match endTime's to `null`:

```
update "match" set endTime = null where endTime = 0;
```

PR #367, opened 2023-03-21
fix: set default timezone to 'America/New York' - Resolves #358

### LINKED ISSUE #358: [Bug]: Match creation and stats calculations are not accounting for Daylight Savings Time (or timezones)
### Bug Summary

Week 11 of Clubs 2023 should have started on Wednesday March 15th at 8:30pm EST, but it seems to have started ~an hour later. This indicates that the server is not accounting for Daylight Savings Time

### Detailed Description

Here's the leaderboard as displayed on production (where the machine runs in UTC time)
image

Here is the leaderboard on localhost while connected to the same production database (local machine running EST)
image

These pictures were taken back-to-back with the same data. Notice that week 11 has fewer matches played in production. This may be because these same matches are being evaluated relative to different timezones with and without daylight savings.

The relevant files are likely `api/controllers/StatsController.js` which parses the match data and groups by week, and `api/helpers/find-or-create-match.js` which identifies which match belongs to which game, and creates new matches, setting their timestamps

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Steps to Reproduce

_No response_

itsalaidbacklife:
## Please check the following

- For New Features:

Resolves #358 by setting default timezone to America/New York wherever dayjs is configured

### DIFF
--- a/api/controllers/StatsController.js
+++ b/api/controllers/StatsController.js
@@ -1,8 +1,13 @@
 const dayjs = require('dayjs');
 const isBetween = require('dayjs/plugin/isBetween');
+const utc = require('dayjs/plugin/utc');
+const timezone = require('dayjs/plugin/timezone');
 const Result = require('../../types/Result.es5');
-dayjs.extend(isBetween);
 
+dayjs.extend(isBetween);
+dayjs.extend(utc);
+dayjs.extend(timezone);
+dayjs.tz.setDefault('America/New_York');
 /////////////
 // Helpers //
 /////////////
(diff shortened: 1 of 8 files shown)


api/controllers/StatsController.js
```diff
@@ -1,8 +1,13 @@
 const dayjs = require('dayjs');
 const isBetween = require('dayjs/plugin/isBetween');
+const utc = require('dayjs/plugin/utc');
```
seriouslysean, line 3: You should pull these out in to a utility somehow or some sort of global. Your code is super wet.

seriouslysean (changes requested):

seriouslysean (approved):
Actually, since this is so hard to reproduce why don’t you push it, test it, and then if it works we can revert and fix it better.

itsalaidbacklife:
Closing because this didn't work when experimentally deployed

PR #371, opened 2023-03-22
Lobbysounds

DonPierreDon:
## Please check the following

- For New Features:


public/sounds/lobby/enter-lobby.mp3
itsalaidbacklife, line 1: Let's replace this with WaterDrop.mp3

public/sounds/lobby/leave-lobby.mp3
itsalaidbacklife, line 1: Let's replace this with BubblePop2.mp3

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "4.2.0",
+  "version": "4.2.1",
```
itsalaidbacklife, line 4: Since another pr was merged recently, we should bump this to 4.2.2
```suggestion
  "version": "4.2.2",
```

src/views/LobbyView.vue
```diff
@@ -70,6 +72,18 @@ export default {
...
+    opponentUsername: function(newVal, oldVal) {
+      // Play audio when player join lobby
+      if (oldVal === null && newVal !== null) {
+        this.$refs.enterLobbySound.play()
+      }
+      // Play audio when player leave lobby
+      else if (oldVal !== null && newVal === null) {
+        this.$refs.leaveLobbySound.play()
+      }
+    }
```
itsalaidbacklife, line 85: Nice, this works. I think we can slightly simplify by looking only at the new value since if there is an opponent after this update (i.e. `newVal` is truthy) then an opponent just joined. As a super minor syntax convention, we favor the shorthand `somefunctionName() {}` over `someFunctionName: function() {}` since it saves some characters and the parentheses imply that it's a function definition.
```suggestion
    opponentUsername(newVal) {
      // Opponent joined
      if (newVall) {
        this.$refs.enterLobbySound.play()
      // Opponent left
      } else {
        this.$refs.leaveLobbySound.play()
      }
    }
```

itsalaidbacklife (approved):
Nice work! I'm requesting the audio files be swapped, the package.json version be bumped again (there was another recent update) and a refactor of the watcher that plays the audio. 

Implementation works and is clear to understand. Well done. We can merge when the PR comments are addressed

itsalaidbacklife (approved):
Great work. This proved more complicated than expected, but I think it's come out nicely. Well done!

Issue #354, opened 2023-03-13
[Feature]: Blast 'gameFinished' socket event when games end due to offensive play, passing, or stalemate request


itsalaidbacklife:
### Feature Summary

When a game finishes, it is no longer available to spectate. This is prevented on the backend but we should also disable the 'Spectate' button on the game in the HomeView page.

### Detailed Description

There is already frontend setup to handle disabling a game when an event comes in to register that it is finished: the backend just needs to emit a 'gameFinished' event like so:

```
// see api/controllers/game/concede.js
sails.sockets.blast('gameFinished', { gameId: game.id });
```

As shown above, this already works when a player concedes. To apply this for the other ways that games can end, we should add the same line into the parts of the other backend actions that can end the game: passing, requesting a stalemate, or calculating that the game is over due to an offensive play:
1. `api/controllers/game/pass.js`
2. `api/controllers/game/stalemate.js`
3. `api/controllers/services/gameService.js` -- `checkWinGame()` function

When this works, it will result in the 'Spectate' button disabling for those games if you are on the HomeView before the game finishes. We should update the specatating `describe()` block in the `tests/e2e/specs/out-of-game/home.spec.js` tests to check these cases when we fix them.

Haviles04:
I'm working on this one!

PR #372, opened 2023-03-25
Feature #354

Haviles04:
## Please check the following

- For New Features:

Added socket blast, as well as test for all possible game ending scenarios


seriouslysean (changes requested):
Looks pretty good. Didn’t test it yet but have one callout.

api/controllers/game/pass.js
```diff
@@ -60,6 +60,8 @@ module.exports = function (req, res) {
           victory.currentMatch = await sails.helpers.addGameToMatch(game);
         }
         await gameService.clearGame({ userId: req.session.usr });
+        // Inform all clients this game is over
+        sails.sockets.blast('gameFinished', { gameId: game.id });
```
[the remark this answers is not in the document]
Haviles04: I moved this to get pushed into updatePromises inside gameService.cleargame to reduce it down to one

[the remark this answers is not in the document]
seriouslysean: Yeah that’s perfect! I do the same thing with my team at work when PRing. It usually takes like ~3 edits to get things perfect. This is way better.

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -29,6 +30,15 @@ function toggleInput(selector, checked = false) {
...
 }
 
+function recoverOpponentsInNewGame(gameId) {
+  cy.recoverSessionOpponent(playerOne);
+  cy.subscribeOpponent(gameId);
+  cy.readyOpponent(gameId);
+  cy.recoverSessionOpponent(playerTwo);
+  cy.subscribeOpponent(gameId);
+  cy.readyOpponent(gameId);
+}
```
itsalaidbacklife, line 40: This is a handy utility and it looks like it gets a fair bit of use here. It looks like it's only used immediately after creating a game, so it likely makes sense to combine this with the game creation itself. The only issue with doing that is how to get access to the `gameId` of the newly created game, since that's provided by the `createGameOpponent()`. Here we can use [Cypress aliases](https://docs.cypress.io/guides/core-concepts/variables-and-aliases) to capture the reference to the gameId.

```suggestion
function setupGameBetweenTwoUnseenPlayers(gameName) {
  cy.createGameOpponent(gameName).then(({ gameId }) => {
    cy.wrap(gameId).as(`${gameName}GameId`); // this creates an alias for the gameId
    cy.recoverSessionOpponent(playerOne);
    cy.subscribeOpponent(gameId);
    cy.readyOpponent(gameId);
    cy.recoverSessionOpponent(playerTwo);
    cy.subscribeOpponent(gameId);
    cy.readyOpponent(gameId);
    });
}
```

Then you can use it in tests like so:
```
setupGameBetweenTwoUnseenPlayers('conceded');
cy.concedeOpponent();
cy.get(`[data-cy-spectate-game=${this.concededGameId}]`).should('be.disabled');
```

One gotcha is that the reference to `this.concededGameId` only works when the test's callback uses the actual `function` keyword instead of an arrow function `() = >{}`. So we'd have to update the test like so:

```
it('Disables spectate button if on home view before game finishes', function () {
  // test code
});
```

Vs the current arrow function wouldn't allow the reference to `this.concededGameId`

```
it('Does not work without the function keyword', () => {
  // this won't work because arrow functions treat `this` differently
});
```

This might simplify the consumers of the function here since they all involve creating the game in addition to joining it as the other players

api/services/gameService.js
```diff
@@ -247,6 +247,8 @@ module.exports = {
                 Card.destroy({
                   or: deleteCardsCriteria,
                 }),
+                // Inform all clients this game is over
+                sails.sockets.blast('gameFinished', { gameId: game.id }),
```
itsalaidbacklife, line 251: This is great. Much cleaner than my previous implementation which cluttered all the individual endpoints with the need for this blast

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -306,6 +314,119 @@ describe('Home - Game List', () => {
...
+        p1Points: [Card.SEVEN_OF_DIAMONDS, Card.SEVEN_OF_CLUBS],
+        p1FaceCards: [],
+      });
+      cy.get('[data-player-hand-card]').should('have.length', 1);
+      cy.recoverSessionOpponent(playerOne);
+      cy.vueRoute('/');
+      cy.get('[data-cy-game-list-selector=spectate]').click();
+      cy.get(`[data-cy-spectate-game]`).should('be.not.disabled');
+      cy.playPointsSpectator(Card.EIGHT_OF_SPADES, 0);
+      cy.log('playedPoints');
```
itsalaidbacklife, line 406: Let's get rid of all the cy.log() calls that are here and instead add two cy.log calls per  test situation (before and after) the case worked successfully e.g. 

```
      cy.log('Setup game end by passing')
      cy.setupGameAsSpectator();
      cy.loadGameFixture({
        p0Hand: [Card.SEVEN_OF_CLUBS],
        p0Points: [Card.SEVEN_OF_DIAMONDS, Card.SEVEN_OF_HEARTS],
        p0FaceCards: [],
        p1Hand: [],
        p1Points: [],
        p1FaceCards: [],
      });
      cy.get('[data-player-hand-card]').should('have.length', 1);
      //Delete deck, draw last two card, both players pass until stalemate
      cy.deleteDeck();
      cy.get('#deck').should('contain', '(2)');
      cy.recoverSessionOpponent(playerOne);
      cy.drawCardOpponent();
      cy.get('#deck').should('contain', '(1)');
      cy.recoverSessionOpponent(playerTwo);
      cy.drawCardOpponent();
      cy.get('#deck').should('contain', '(0)');
      cy.log('Deck empty');
      cy.recoverSessionOpponent(playerOne);
      cy.passOpponent();
      cy.get('#history').should('contain', `${playerOne.username} passes`);
      cy.recoverSessionOpponent(playerTwo);
      cy.passOpponent();
      cy.get('#history').should('contain', `${playerTwo.username} passes`);
      cy.recoverSessionOpponent({ username: playerUsername, password: playerPassword });
      cy.vueRoute('/');
      cy.get('[data-cy-game-list-selector=spectate]').click();
      cy.get(`[data-cy-spectate-game]`).should('be.not.disabled');
      cy.recoverSessionOpponent(playerOne);
      cy.passOpponent();
      cy.get(`[data-cy-spectate-game]`).should('be.disabled');
      cy.log('Game ended by passing - successfully disabled spectate button');
```

This will clean up the log output and focus the logs on the outcome of a test case in broad strokes

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -306,6 +314,119 @@ describe('Home - Game List', () => {
         cy.get(`[data-cy-spectate-game=${gameId}]`).click();
       });
     });
+
+    it('Disables spectate button if on home view before game finishes', () => {
+      cy.signupOpponent(playerOne.username, playerOne.password);
+      cy.signupOpponent(playerTwo.username, playerTwo.password);
+
+      //Navigate to homepage and select spectate tab
```
itsalaidbacklife, line 322: Nitpick: put a space before the comment for all the comments
```suggestion
      // Navigate to homepage and select spectate tab
```

itsalaidbacklife, line 322: Nitpick: put a space before the comment for all the comments (there are more in this file, but I'll just comment it here)
```suggestion
      // Navigate to homepage and select spectate tab
```

itsalaidbacklife (approved):
Nice work; solid improvement. Please bump the package.json version to 4.3.1.

I've requested some refactoring of the test code and a bit of cleanup, but generally this is very nice work

PR #381, opened 2023-04-04
Fix/socket subscriptions

### LINKED ISSUE #380: [Bug]: Old socket subscriptions can cause you to suddenly switch games
### Bug Summary

If your socket is subscribed to an old game e.g. one you had previously spectated then left, you might still receive an socket event for that old game even though you're no longer 'in' the game. This can cause your UI to suddenly update to show you the game you were previously spectating, even in the middle of trying to play another game.

### Detailed Description

This can be a jarring and frustrating experience for players. Basically you spectate a game, leave, then start playing a game and suddenly when a move happens in the game you were spectating, you see the old spectated game instead of the new one. 

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Steps to Reproduce

1. Create and start a game as two players
2. Spectate the game as a 3rd user
3. Have the (spectating) 3rd user navigate back to the homepage directly via the url bar
4. Have the 3rd (no longer spectating) user start a new game with a 4th user
5. Have the active players from the first game make a move
6. The 3rd user will suddenly see the game they were previously spectating

itsalaidbacklife:
## Please check the following
Resolves #380 with additional test case.

Fixes bug where UI would jump you into a game you'd previously been spectating by making the socket logic pay attention to gameId.
- For New Features:

### DIFF
--- a/api/controllers/game/ready.js
+++ b/api/controllers/game/ready.js
@@ -103,6 +103,7 @@ module.exports = function (req, res) {
             change: 'ready',
             userId: user.id,
             pNum: user.pNum,
+            gameId: game.id,
           },
         });
         return Game.updateOne({ id: game.id }).set(gameUpdates);
--- a/src/plugins/sails.js
+++ b/src/plugins/sails.js
@@ -43,8 +43,9 @@ io.sails.reconnection = true;
 io.socket.on('game', function (evData) {
   switch (evData.verb) {
     case 'updated': {
+      const currentRoute = router.currentRoute.value;
       // No-op if the event's gameId doesn't match the url
-      const { gameId: urlGameId } = router.currentRoute.value.params;
+      const { gameId: urlGameId } = currentRoute.params;
       const eventGameId = evData.data?.game?.id ?? evData.data.gameId;
       if (urlGameId && Number(urlGameId) !== eventGameId) {
         return;
@@ -60,7 +61,6 @@ io.socket.on('game', function (evData) {
           store.commit('updateReady', evData.data.pNum);
           break;
         case 'Initialize': {
-          const currentRoute = router.currentRoute.value;
           const isSpectating = currentRoute.name === ROUTE_NAME_SPECTATE;
 
           // Update state
--- a/src/plugins/sails.js
+++ b/src/plugins/sails.js
@@ -42,7 +42,13 @@ io.sails.reconnection = true;
 // Handles socket updates of game data
 io.socket.on('game', function (evData) {
   switch (evData.verb) {
-    case 'updated':
+    case 'updated': {
+      // No-op if the event's gameId doesn't match the url
+      const { gameId: urlGameId } = router.currentRoute.value.params;
+      const eventGameId = evData.data?.game?.id ?? evData.data.gameId;
+      if (urlGameId && Number(urlGameId) !== eventGameId) {
+        return;
+      }
       // Handle GameOver
       if (evData.data.victory && evData.data.victory.gameOver) {
         setTimeout(() => {
@@ -173,7 +179,7 @@ io.socket.on('game', function (evData) {
           store.commit('setConsideringOpponentStalemateRequest', false);
           store.commit('setWaitingForOpponentToStalemate', false);
           break;
-      }
+      }}
   }
 });
 
--- a/tests/e2e/specs/in-game/spectating.spec.js
+++ b/tests/e2e/specs/in-game/spectating.spec.js
@@ -10,7 +10,7 @@ import {
   SnackBarError,
   getCardId,
 } from '../../support/helpers';
-// import { assertLoss } from './gameOver.spec';
+import { assertLoss } from './gameOver.spec';
 
 function setup() {
   cy.wipeDatabase();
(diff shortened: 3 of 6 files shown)


api/controllers/game/ready.js
```diff
@@ -103,6 +103,7 @@ module.exports = function (req, res) {
             change: 'ready',
             userId: user.id,
             pNum: user.pNum,
+            gameId: game.id,
```
itsalaidbacklife, line 106: Added this so we can filter out socket events for games we aren't viewing right now

src/plugins/sails.js
```diff
@@ -42,7 +42,14 @@ io.sails.reconnection = true;
...
   switch (evData.verb) {
-    case 'updated':
+    case 'updated': {
+      const currentRoute = router.currentRoute.value;
+      // No-op if the event's gameId doesn't match the url
+      const { gameId: urlGameId } = currentRoute.params;
+      const eventGameId = evData.data?.game?.id ?? evData.data.gameId;
+      if (urlGameId && Number(urlGameId) !== eventGameId) {
+        return;
+      }
```
itsalaidbacklife, line 52: This was the fix. Compare the gameId in the url to the one in the event payload and early-return if they don't match to prevent stale updates

tests/e2e/specs/in-game/spectating.spec.js
```diff
@@ -137,6 +140,53 @@ describe('Spectating Games', () => {
...
+      p1FaceCards: [],
+    });
+    cy.get('[data-player-hand-card]').should('have.length', 2);
+
+    cy.recoverSessionOpponent(playerOne);
+    cy.get('@aceOfSpades').then((aceOfSpadesId) => {
+      cy.playPointsById(aceOfSpadesId);
+    });
+
+    cy.wait(3000);
```
itsalaidbacklife, line 178: I hate this `cy.wait()` but I wasn't sure how to avoid it. The idea here is that after a move happens in the "other" game that we were _previously_ spectating, we want to make sure that our UI _doesn't_ update to reflect that change. The tricky thing there is that cypress assertions run until they pass, so our assertGameState() call will pass erroneously if it runs before the actual problem happens because things were right, but then they weren't.

Waiting long enough for the issue to occur and _then_ asserting the expected game state is my best solution so far for this

tests/e2e/support/commands.js
```diff
@@ -118,7 +118,7 @@ Cypress.Commands.add('setupGameAsSpectator', () => {
   cy.visit('/');
   cy.signupPlayer(username, validPassword);
   cy.vueRoute('/');
-  cy.createGamePlayer({ gameName: 'Test Game', isRanked: false }).then((gameData) => {
+  cy.createGamePlayer({ gameName: 'Spectator Game', isRanked: false }).then((gameData) => {
```
itsalaidbacklife, line 121: Updated this to make it easier to tell the difference between the spectator game and the one we're playing

tests/e2e/support/commands.js
```diff
@@ -342,6 +342,21 @@ Cypress.Commands.add('playPointsSpectator', (card, pNum) => {
...
+      cardId,
+    },
+    function handleResponse(_, jwres) {
+      if (jwres.statusCode !== 200) {
+        throw new Error(jwres.body.message);
+      }
+      return jwres;
+    },
+  );
+});
```
itsalaidbacklife, line 358: The need for this command is awkward and warrants further thought. Most of our moves input a Card object like 
```
{ suit: 1, rank: 12 }
```

and then look up the `id` of the card (needed for the api call) by looking in the `game` store module clientside for a card with the same suit and rank. This doesn't work for making moves in games where the game object isn't in the store, however. So my solution was to alias the card id that I want to play _while the visible user is spectating_ and the card's id _is_ in the store, then to use this new command after we leave the game (and join a new one) to make the move in the game we aren't watching using that stored card id. 

It works, but it smells and it feels extra weird that we have some commands for doing things as a spectator where you specify which pNum and then this where you give the cardId directly. Worth investigating a cleaner way to consolidate.

@seriouslysean penny for your thoughts

tests/e2e/specs/in-game/spectating.spec.js
seriouslysean: Could you try catch a state update by element and if it doesn’t happen assert true?

itsalaidbacklife: The tricky part is that the point of the test is that when an update happens in this no-longer-relevant-to-you game, there is **no** update in your app, either in the store or the UI. So I'm not sure what to check for to ensure that the update went through. Basically it should be when the request that the test command sends for `cy.playPointsById()` gets a response, but there can be a delay between then and the actual socket message, so I can't think of a surefire way to check that the update you're not supposed to see has happened

itsalaidbacklife: That's an interesting idea. Basically make a knowingly failing assertion. I don't think cypress gives us a way to catch failed assertions though. I'm not sure about the retryability either

itsalaidbacklife: I'm asking in the cypress discord because it's an interesting technical challenge

itsalaidbacklife: After an interesting convo, I think the manual `cy.wait()` is the best way to replicate the failure. Keeping it as-is

PR #382, opened 2023-04-05
Feat/show one offs in overlay

### LINKED ISSUE #353: [Feature]: Display the last played one-off or two in the #waiting-for-opponent-counter-scrim
### Feature Summary

When a player plays a one-off or counters their opponent's one-off they see an overlay while their opponent decides whether to counter. It would be helpful, especially for spectating, if this overlay showed which card the opponent was considering countering.

### Detailed Description

This overlay is defined in `src/components/GameView/gameDialogs.vue`:

image

We could display the one-off itself below the text to clarify what card might resolve. This would be especially useful for spectating:
image

This should also show the twos that have been played. Below we see the Ace of clubs as the root one-off and three counters (most recently the two of spades played by the visible player) 
image

itsalaidbacklife:
## Please check the following
Resolves #353 

Changes the visual style of in-game overlays and adds the one-offs on the stack to the #waiting-for-opponent-counter-scrim

Adds background color to overlays in addition to showing the one-off cards
image

- For New Features:

### DIFF
--- a/src/components/GameView/GameCard.vue
+++ b/src/components/GameView/GameCard.vue
@@ -75,6 +75,10 @@ export default {
       default: '',
       validator: (val) => ['', 'player', 'opponent'].includes(val),
     },
+    highElevation: {
+      type: Boolean,
+      default: false,
+    },
   },
   computed: {
     suitName() {
@@ -127,7 +131,13 @@ export default {
       return `${this.rankName} of ${this.suitName}`;
     },
     elevation() {
-      return this.isGlasses ? '0' : '1';
+      if (this.isGlasses) {
+        return '0';
+      }
+      if (this.highElevation) {
+        return '7';
+      }
+      return '1';
     },
     isBack() {
       return !this.suit && !this.rank;
--- a/src/components/GameView/GameOverlays.vue
+++ b/src/components/GameView/GameOverlays.vue
@@ -28,7 +28,7 @@
       <h1 :class="[this.$vuetify.display.xs === true ? 'text-h5' : 'text-h3', 'overlay-header']">
         {{ showWaitingForOpponetToCounterMessage }}
       </h1>
-      <div class="mt-4 d-flex justify-center">
+      <div id="counter-scrim-cards">
         <game-card
           v-if="oneOff"
           :rank="oneOff.rank"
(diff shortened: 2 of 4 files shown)


src/components/GameView/GameCard.vue
```diff
@@ -75,6 +75,10 @@ export default {
       default: '',
       validator: (val) => ['', 'player', 'opponent'].includes(val),
     },
+    highElevation: {
+      type: Boolean,
+      default: false,
+    },
```
itsalaidbacklife, line 81: This is used to set a higher elevation for cards in an overlay. I considered making it a numeric value that get explicitly set to a particular elevation (rather than a high/low boolean), but I think this will be easier for maintaining a consistent style if and when we need cards displayed in other overlays

src/components/GameView/GameOverlays.vue
```diff
@@ -18,60 +19,88 @@
...
+          class="overlay-card"
+        />
+        <div>
+          <game-card
+            v-for="(two, index) in twos"
+            :key="`overlay-two-${two.id}`"
+            :rank="two.rank"
+            :suit="two.suit"
+            :data-overlay-counter="`${two.rank}-${two.suit}`"
+            :class="`overlay-card overlay-two overlay-two-${index}`"
```
itsalaidbacklife, line 47: These classes are used to apply the different rotation transforms to each two in this list

tests/e2e/specs/in-game/countering.spec.js
```diff
@@ -283,12 +283,17 @@ describe('Countering One-Offs', () => {
     cy.get('#counter-dialog').should('be.visible').get('[data-cy=counter]').click();
     cy.get('#choose-two-dialog').should('be.visible').get('[data-counter-dialog-card=2-2]').click();
     cy.get('#waiting-for-opponent-counter-scrim').should('be.visible');
+    cy.get('#waiting-for-opponent-counter-scrim [data-overlay-one-off=1-0]').should('exist');
+    cy.get('#waiting-for-opponent-counter-scrim [data-overlay-counter=2-2]').should('exist');
```
itsalaidbacklife, line 287: Interestingly, these have to be `'exist'` assertions rather than `'be.visible'` because the cards cover each other which cypress considers to mean not visible

src/components/GameView/GameOverlays.vue
```diff
@@ -191,16 +224,45 @@ export default {
 </script>
 
 <style scoped lang="scss">
+:deep(.v-overlay__content) {
+  left: 0;
+}
```
itsalaidbacklife, line 229: I'm not sure why this is necessary but the overlay content was misaligning with the window without it

PR #384, opened 2023-04-06
Feature #350 display usernames of active spectators in game

### LINKED ISSUE #350: [Feature]: Display usernames of active spectators in game
### Feature Summary

Currently players have no indication of when other users are spectating their game. We should add a 'Spectating' area to the UI that indicates who is watching the game

### Detailed Description

Currently the app doesn't store any info in the database about who is spectating. We could track this, or instead, we could simply emit a socket event to all active players when someone joins as a spectator. This could be done by updating the backend api endpoint `api/controllers/game/spectate.js` to publish a socket event on the Game model (see how this is done in `api/controllers/game/concede.js` for an example), and then adding a socket handler to `src/plugins/saills.js` to handle the event. 

This could then commit an update to the game store module in `src/store/modules/game.js` which updates a new `spectators` variable that stores a list of the string username of the players spectating you.

Haviles04:
## Please check the following

- For New Features:

fixed #350


api/controllers/game/spectate.js
```diff
@@ -1,17 +1,44 @@
...
-      return res.badRequest({message: 'You can only spectate an ongoing game with two players'});
-    }
-
-    // Subscribe socket to game
-    Game.subscribe(req, [ gameId ]);
-    req.session.spectating = gameId;
-
-    return res.ok(game);
+    Promise.all([promiseGame, promisePlayer])
+      .then(function checkAndSubscribe(values) {
```
itsalaidbacklife, line 8: Earlier in this function we're using `async/await` rather than `Promise.then()`; these are alternatives to each other and it's best to use one or the other without mixing them. Generally `async/await` is the newer syntax and is our preference for this project when writing new code.

`async/await` syntax can still be used with Promise.all() and would be cleaner here:

```
const [ game, spectator ] = await Promise.all([ promiseGame, promiseSpectator ]);
// Now you can use `game` and `spectator`
```

api/controllers/game/spectate.js
```diff
@@ -1,17 +1,44 @@
 module.exports = async function (req, res) {
   try {
     const { gameId } = req.body;
-    const game = await gameService.populateGame({ gameId });
+    const promiseGame = await gameService.populateGame({ gameId });
+    const promisePlayer = await userService.findUser({ userId: req.session.usr });
```
itsalaidbacklife, line 5: Let's rename `promisePlayer` to `promiseSpectator` since they aren't playing in this game

api/controllers/game/spectate.js
```diff
@@ -1,17 +1,44 @@
...
-    return res.ok(game);
+    Promise.all([promiseGame, promisePlayer])
+      .then(function checkAndSubscribe(values) {
+        const [game, player] = values;
+        if (game.status || game.players.length < 2) {
+          return res.badRequest({ message: 'You can only spectate an ongoing game with two players' });
+        }
+        // Subscribe socket to game
+        const updatePromises = [Game.subscribe(req, [gameId]), (req.session.spectating = gameId)];
+        return Promise.all([game, player, ...updatePromises]);
```
itsalaidbacklife, line 15: I believe both of these are actually synchronous operations so we don't need the `Promise` wrapping here
```suggestion
        Game.subscribe(req, [gameId]);
        req.session.spectating = gameId
```

api/services/gameService.js
```diff
@@ -117,9 +118,15 @@ module.exports = {
                   message: "Can't populate game, because it does not have players collection",
                 });
               })
+              // find spectators
+              .then(function findSpectators(values) {
+                const [game, p0, p1] = values;
+                const { spectators } = game;
+                return [game, p0, p1, spectators];
+              })
```
itsalaidbacklife, line 126: We don't need a separate `.then()` handler for this because these are all synchronous operations to just read existing values
```suggestion
              const { spectators } = game;
```

api/controllers/game/spectate.js
```diff
@@ -1,17 +1,37 @@
 module.exports = async function (req, res) {
   try {
     const { gameId } = req.body;
-    const game = await gameService.populateGame({ gameId });
+    const promiseGame = await gameService.populateGame({ gameId });
+    const promiseSpectator = await userService.findUser({ userId: req.session.usr });
 
+    const [game, spectator] = await Promise.all([promiseGame, promiseSpectator]);
```
seriouslysean, line 7: Doing these in parallel is great 👍🏼 

Considering you're only using these variables once to reset them, you could also do something like this
```
    const promiseGame = ;
    const promiseSpectator = await userService.findUser({ userId: req.session.usr });

    const [game, spectator] = await Promise.all([
        gameService.populateGame({ gameId }),
        userService.findUser({ userId: req.session.usr }),
    ]);
``` 
I think those are promises by default so you likely don't need the await keyword for the individual use cases, but test it out to see.

api/controllers/game/spectate.js
```diff
@@ -1,17 +1,37 @@
...
     }
-
     // Subscribe socket to game
-    Game.subscribe(req, [ gameId ]);
+    Game.subscribe(req, [gameId]);
     req.session.spectating = gameId;
 
-    return res.ok(game);
+    return Promise.all([game, spectator])
+      .then(async function updateDataBase() {
```
seriouslysean, line 16:
```suggestion
      .then(async function updateDatabase() {
```
Casing here is off. It's also weird here to use chaining when async await would likely fit the bill. Not entirely sure, but something like this (high level, of course) makes more sense to me:
```
// game and spectator are already set, so you probably don't need to await them again imo
try {
    await updateDatabase(...);
    await populateGame(...);
    await publishAndRespond(...);
    // You'll need to set fullGame before this from one of those calls
    return res.ok(fullGame);
} catch {
   return res.badRequest(err);
}
```

api/services/gameService.js
```diff
@@ -109,7 +111,8 @@ module.exports = {
                   if (game.players.length > 1) {
                     const p0 = userService.findUser({ userId: game.players[0].id });
                     const p1 = userService.findUser({ userId: game.players[1].id });
-                    return Promise.all([Promise.resolve(game), p0, p1]);
+                    const spectatingUsers = game.spectatingUsers.map((user) => user.username);
```
seriouslysean, line 114: Will this always be an array?

src/plugins/sails.js
```diff
@@ -171,6 +171,9 @@ io.socket.on('game', function (evData) {
         case 'reLogin':
           store.dispatch('updateGameThenResetPNumIfNull', evData.data.game);
           break;
+        case 'spectatingUsers':
+          store.dispatch('updateGameThenResetPNumIfNull', evData.data.game);
+          break;
```
seriouslysean, line 176: These can likely be combined since they're the same.
```
        case 'reLogin':
        case 'spectatingUsers':
          break;
```

src/store/modules/game.js
```diff
@@ -100,6 +101,9 @@ export default {
       }
       return state.players[(state.myPNum + 1) % 2];
     },
+    spectatingUsers(state) {
+      return state.spectatingUsers;
```
seriouslysean, line 105: Do you need to be type safe here? 
```
return state.spectatingUsers || [];
```

src/store/modules/game.js
```diff
@@ -168,6 +172,8 @@ export default {
       if (Object.hasOwnProperty.call(newGame, 'p1Ready')) state.p1Ready = newGame.p1Ready;
       if (Object.hasOwnProperty.call(newGame, 'passes')) state.passes = newGame.passes;
       if (Object.hasOwnProperty.call(newGame, 'players')) state.players = cloneDeep(newGame.players);
+      if (Object.hasOwnProperty.call(newGame, 'spectatingUsers'))
```
seriouslysean, line 175: @itsalaidbacklife less about this PR and more higher order, do we always need has own property or would some other check here be better long term?

seriouslysean (commented):
Few things that I noticed, leaving final approval to @itsalaidbacklife since I haven't been as involved in this feature.

api/controllers/game/spectate.js
```diff
@@ -1,17 +1,37 @@
 module.exports = async function (req, res) {
   try {
     const { gameId } = req.body;
-    const game = await gameService.populateGame({ gameId });
+    const promiseGame = await gameService.populateGame({ gameId });
+    const promiseSpectator = await userService.findUser({ userId: req.session.usr });
```
itsalaidbacklife, line 5: To @seriouslysean 's point below, we actually _don't_ want `await` on these lines because we _instead_ want to await the `Promise.all()` of both of them.

```suggestion
    const promiseGame = gameService.populateGame({ gameId });
    const promiseSpectator = userService.findUser({ userId: req.session.usr });
```

itsalaidbacklife: That said, @seriouslysean's suggestion to inline them inside the Promise.all() call itself is reasonable imo. It gets rid of the variable declarations and shortens things up without being difficult to read because the actual calls themselves are pretty short. I say we go with that suggestion

api/controllers/game/spectate.js
itsalaidbacklife: I like this b/c it's shorter and the calls to the promise returning functions are already short and easy to read. I say we remove the `promiseGame` and `promiseSpectator` variables altogether and use 
```
    const [game, spectator] = await Promise.all([
        gameService.populateGame({ gameId }),
        userService.findUser({ userId: req.session.usr }),
    ]);
```
as suggested above

src/store/modules/game.js
itsalaidbacklife: I think typescript would save us the need altogether. It's basically a paranoid type safety check that we could skip if we had proper types on our client and server

api/controllers/game/spectate.js
Haviles04: This is a pro move

api/services/gameService.js
Haviles04: When there are no spectators this is an empty array in the database

src/store/modules/game.js
Haviles04: I think we are okay here, because spectatingUsers is always an array

api/services/gameService.js
```diff
@@ -38,9 +38,10 @@ function tempUser(usr, points) {
   this.faceCards.sort(comapreByRankThenSuit);
 }
 
-function tempGame(game, p0, p1) {
+function tempGame(game, p0, p1, spectatingUsers) {
```
itsalaidbacklife, line 41: since `spectatingUsers` is already on the `game` object when we populate that collection, let's remove it from here as an argument and instead assign the value of `this.spectators` to the `.map()` call that you're doing below;

```
this.spectatingUsers = game.spectatingUsers.map((user) => user.username);
```

src/store/modules/game.js
itsalaidbacklife: The other thing this does is allow for partial updates. Which granted we could do with a spread instead

src/store/modules/game.js
```diff
@@ -328,24 +334,24 @@ export default {
...
+            if (jwres.statusCode === 200) {
+              context.commit('updateGame', res);
+              context.commit('setMyPNum', 0);
+              return resolve();
+            }
+            return reject(new Error('Unable to spectate game'));
+          },
+        );
+      });
+    },
```
itsalaidbacklife, line 354: It looks like this is overly indented past where the other functions in this file are

src/views/GameView.vue
```diff
@@ -1480,5 +1482,9 @@ export default {
     grid-column-start: 1;
     grid-column-end: span 12;
   }
+
+  .btnWrapper {
+    flex-direction: column !important;
+  }
```
itsalaidbacklife, line 1488: `!important` generally smells. I think we can remove this and use a class above

src/views/GameView.vue
```diff
@@ -6,8 +6,11 @@
     </template>
     <!-- Authenticated View -->
     <template v-else>
-      <div id="game-menu-wrapper" class="d-flex flex-column align-center">
-        <game-menu />
+      <div id="game-menu-wrapper" class="d-flex flex-column justify-center align-center">
+        <div class="d-flex align-center justify-center btnWrapper">
```
itsalaidbacklife, line 10: I think this would achieve the same column affect ([docs](https://vuetifyjs.com/en/styles/flex/)). 
```suggestion
        <div class="d-flex align-center justify-center flex-column">
```

src/views/GameView.vue
itsalaidbacklife:
```suggestion
```

src/views/GameView.vue
```diff
@@ -6,8 +6,11 @@
...
+          <SpectatorListOverlay :spectatingUsers="spectatingUsers" />
```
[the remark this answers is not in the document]
Haviles04: I think we should maybe change it to spectator-list-menu since it's not longer an overlay

itsalaidbacklife:
Adding 'needs migration' label because this requires a database update to merge

src/components/GameView/SpectatorListMenu.vue
```diff
@@ -0,0 +1,37 @@
...
+        icon
+        variant="text"
+      >
+        <v-icon class="mr-1" icon="mdi-eye-outline" size="large" />
+        <span class="pr-2" v-if="!$vuetify.display.xs">{{ spectatingUsers.length }}</span>
+      </v-btn>
+    </template>
+    <!-- Menu -->
+    <v-list class="pl-2 pr-2" id="spectatorList" data-cy="spectate-list-menu">
+      <v-list-title>Spectators</v-list-title>
```
itsalaidbacklife, line 19: This component has been renamed to v-list-item-title
```suggestion
      <v-list-item-title>Spectators</v-list-item-title>
```

seriouslysean:
@Haviles04 are you having trouble fixing a merge conflict or something? You don’t need to close the branch to fix it. Let me know if you need help resolving it.

Haviles04:
> @Haviles04 are you having trouble fixing a merge conflict or something? You don’t need to close the branch to fix it. Let me know if you need help resolving it.

Yea, I had some problems with merge conflicts and my GH and local branch were out of sync, Aleph helped me fix my problem but thank you!

PR #388, opened 2023-04-14
Refactortests

DonPierreDon:
## Please check the following

- For New Features:


tests/e2e/support/commands.js
```diff
@@ -112,48 +102,13 @@ Cypress.Commands.add('setupGameAsP1', (alreadyAuthenticated = false, isRanked =
...
-    cy.get('[data-cy-spectate-game]').click();
-
-    cy.url().should('include', '/spectate/');
-    cy.window()
-    .its('cuttle.app.config.globalProperties.$store.state.game')
-    .then((gameState) => {
-      expect(gameState.id).to.not.eq(null);
-    });
-  });
-});
```
itsalaidbacklife, line 148: It looks like this command got taken out during the merge. That's what failed the test in CI

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -78,366 +71,358 @@ describe('Home - Page Content', () => {
...
-          .its('cuttle.app.config.globalProperties.$store.state.game')
-          .then((gameState) => {
-            expect(gameState.id).to.not.eq(null);
-          });
-      });
+      cy.leaveLobbyOpponent(gameData.gameId);
+      cy.contains('[data-cy-join-game]', 'Play').should('not.be.disabled');
     });
+  });
+});
```
itsalaidbacklife, line 167: These were previously wrapped in a `describe('Open Games')` block. Let's re-add that wrapping layer around these tests

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -78,366 +71,358 @@ describe('Home - Page Content', () => {
...
-          expect(gameState.id).to.eq(null);
-        });
-      cy.createGamePlayer({ gameName: 'Test Game', isRanked: false });
-      cy.get('[data-cy=game-list-item]').contains('[data-cy-join-game]', 'Play').click();
+  it('Displays a game for every open game on the server', () => {
+    cy.createGamePlayer({ gameName: '111', isRanked: false });
+    cy.createGamePlayer({ gameName: '33', isRanked: false });
+    cy.get('[data-cy=game-list-item]').should('have.length', 2);
+  });
+  it('Displays placeholder text when no games are available', () => {
```
itsalaidbacklife, line 79: Let's re-add the newline in between tests
```suggestion
  });

  it('Displays placeholder text when no games are available', () => {
```

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -1,5 +1,12 @@
 import { assertSnackbarError, Card } from '../../support/helpers';
-import { myUser, opponentOne, opponentTwo, playerOne, playerTwo } from '../../fixtures/userFixtures';
+import {
+  myUser,
+  opponentOne,
+  opponentTwo,
+  playerOne,
+  playerTwo,
+  playerThree,
```
itsalaidbacklife, line 8: Looks like this variable isn't used; let's remove the import. This is the cause of the linting error in CI
```suggestion
```

tests/e2e/support/commands.js
```diff
@@ -140,20 +129,20 @@ Cypress.Commands.add('setupGameAsSpectator', () => {
...
-    });
+      .its('cuttle.app.config.globalProperties.$store.state.game')
+      .then((gameState) => {
+        expect(gameState.id).to.not.eq(null);
+      });
   });
 });
 
-Cypress.Commands.add('signupOpponent', (username, password) => {
+Cypress.Commands.add('signupOpponent', (opponent) => {
```
itsalaidbacklife, line 139: One trick we can bring to bear here is to destructure ([docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions#function_parameters)) the `opponent` parameter here in order to clarify that it's expected to have both a username and password attribute. This would also let us use the shorthand syntax down below. This isn't necessary, but may be cleaner. What do you think?
```suggestion
Cypress.Commands.add('signupOpponent', ({ username, password }) => {
```

tests/e2e/support/commands.js
```diff
@@ -140,20 +129,20 @@ Cypress.Commands.add('setupGameAsSpectator', () => {
...
   return new Cypress.Promise((resolve, reject) => {
     io.socket.get(
       'localhost:1337/user/signup',
       {
-        username,
-        password,
+        username: opponent.username,
+        password: opponent.password,
```
itsalaidbacklife, line 145: Then `username` and `password` would already be defined and this part could stay as-is
```suggestion
        username,
        password,
```

tests/e2e/support/commands.js
```diff
@@ -164,19 +153,17 @@ Cypress.Commands.add('signupOpponent', (username, password) => {
     );
   });
 });
-
-Cypress.Commands.add('signupPlayer', (username, password) => {
+Cypress.Commands.add('signupPlayer', (player) => {
```
itsalaidbacklife, line 156: If we destructure above then we would do it here too
```suggestion
Cypress.Commands.add('signupPlayer', ({ username, password }) => {
```

tests/e2e/support/commands.js
```diff
@@ -164,19 +153,17 @@ Cypress.Commands.add('signupOpponent', (username, password) => {
...
 });
-
-Cypress.Commands.add('loginPlayer', (username, password) => {
+Cypress.Commands.add('loginPlayer', (player) => {
   cy.window()
     .its('cuttle.app.config.globalProperties.$store')
-    .invoke('dispatch', 'requestLogin', { username, password });
-  cy.log(`Logged in as player ${username}`);
+    .invoke('dispatch', 'requestLogin', { username: player.username, password: player.password });
+  cy.log(`Logged in as player ${player.username}`);
```
itsalaidbacklife, line 166: In which case we'd do the same move with the `.dispatch()` call arguments since `username` and `password` would already be defined

tests/e2e/support/commands.js
```diff
@@ -1179,13 +1166,13 @@ Cypress.Commands.add('rejectStalemateOpponent', () => {
...
-Cypress.Commands.add('reconnectOpponent', (username, password) => {
+Cypress.Commands.add('reconnectOpponent', (opponent) => {
   cy.log('Opponent Reconnects');
   io.socket.get(
     '/user/reLogin',
     {
-      username,
-      password,
+      username: opponent.username,
+      password: opponent.password,
```
itsalaidbacklife, line 1175: Same opportunity here

itsalaidbacklife (approved):
This looks great! There's one linter error about an unused variable (I commented on the line), and we should bump the `version` field in `package.json` a patch version. I also suggested a destructuring approach for the auth cypress commands. Well done

PR #389, opened 2023-04-16
Style/card colors

### LINKED ISSUE #385: [Feature]: Revise in-game dialogs and move-choice cards to use new colors
### Feature Summary

We have a revised design for the in-game dialogs and move choices. The basic idea is to use a dark surface color (~mahogony/chocolate) with slight transparency as the background color and the cream color (that the cards use) as the text color, occasionally inverting the two e.g. when hovering over a move choice.

### Detailed Description

Here is an example dialog:

image

And the move choice cards (note the colors are inverted on the left card because it is being hovered over):
image

Lastly we should change the text color on the `client/components/GameView/GameOverlays.vue` to be this mahogany color instead of black
image

To do this we should add the mahogany and cream colors from the new WIP [Cuttle design system](https://www.figma.com/file/OIi7SV6qAFsqAyZHaF2I4r/Cuttle-Design-System?node-id=13-2975&t=c55yOxuuDtH7gwcu-0) to the theme defined in `client/plugins/vuetify.js`, and then we should define the darker mahogany color as the `$card-background` in the `client/sass/variables.scss` file.

We should then define a new CuttleDialog.vue component as a higher order component that adds the transparency of .95 to the background and the cream-colored inner border, as shown in the first picture, and make all the instances of the in-game dialogs use this CuttleDialog.vue instead of the raw v-dialog so that they are all styled the same

itsalaidbacklife:
## Please check the following
Resolves #385 by applying styles to in-game dialogs and overlays. 

- For New Features:

Applies new styles to the in-game dialogs, overlays, and move choice cards using chocolate (#4A2416) and cream (#FFF4D7) aka surface-1 and surface-2 respectively.

### Move Choice Cards
Switch background to chocolate and text to cream and invert them on hover/click
image

### Overlays
Switch text color to chocolate from black
image

### Dialogs
Apply chocolate and cream colors, remove use of primary purple:
image


src/components/GameView/CannotCounterDialog.vue
```diff
@@ -1,47 +1,62 @@
...
-      <v-card-title>Cannot Counter</v-card-title>
-      <v-card-text>
-        <span v-if="!opponentLastTwo">
-          Your opponent has played the {{ oneOff.name }} as a one-off
-          <span v-if="target"> targeting your {{ target.name }}</span>
+  <base-dialog v-if="oneOff" v-model="show" title="Cannot Counter" id="cannot-counter-dialog">
+    <template #body>
+      <div v-if="!opponentLastTwo" class="my-2">
+        Your opponent has played the 
+        <game-card-name :card-name="oneOff.name" />
```
itsalaidbacklife, line 6: This new component basically serves to make the card names easier to read against the dark chocolate background (surface-1) by wrapping them in the lighter cream color (surface-1) with some padding 
image


Interestingly it adds a bit of a gotcha when testing because the assertions for text content need to account for nested elements with the `{includeShadowDom: true}` option

src/components/GameView/ChooseTwoDialog.vue
```diff
@@ -0,0 +1,90 @@
+<template>
```
itsalaidbacklife, line 1: I split the `CounterDialog` into this component (`ChooseTwoDialog`) and `ChooseWhetherToCounterDialog` to reduce the amount of mixing and matching. There was overlap between the two but it was messy and weird to keep straight so this way there's just a separate dialog for asking whether you'd like to counter (the other one) and this one for picking which two to use once you've decided to counter

src/components/GameView/ChooseWhetherToCounterDialog.vue
```diff
@@ -0,0 +1,123 @@
+<template>
```
itsalaidbacklife, line 1: Counterpart to the above `ChooseTwoDialog`. This is the one that first pops up to ask if you'd like to counter, then we switch to the one to which two to counter with if you do choose to counter.

src/components/GameView/CounterDialog.vue
```diff
@@ -1,72 +1,35 @@
...
+
+  <choose-two-dialog
+    v-if="oneOff"
+    :model-value="show && choseToCounter"
+    :one-off="oneOff"
+    :twos-in-hand="twosInHand"
+    :twos-played="twosPlayed"
+    @counter="counter($event)"
+    @resolve="resolve"
+  />
```
itsalaidbacklife, line 21: This is much simpler imo now that we have two components for the two flows

src/components/GameView/GameOverDialog.vue
```diff
@@ -1,61 +1,61 @@
...
-            <h1 class="dialog-header">{{ heading }}</h1>
-          </v-card-title>
-          <v-card-text class="d-flex justify-center">
-            <v-img :src="logoSrc" :data-cy="logoDataAttr" class="logo-image" />
-          </v-card-text>
+  <base-dialog v-model="show" id="game-over-dialog">
+    <template #title>
+      <h1 :data-cy="headingDataAttr" class="dialog-header">{{ heading }}</h1>
+      <v-img v-if="currentMatch && !isMobilePortrait" :src="logoSrc" :data-cy="logoDataAttr" class="logo-image-match" />
+    </template>
```
itsalaidbacklife, line 6: The base-dialog has a `title` prop for when the title is just text, and a `#title` slot for when the header should have additional content. Here we're using the slot (this is the use case that drove creating the slot in order to have the logo image appear in the upper-right corner

image

src/components/GameView/GameOverDialog.vue
```diff
@@ -1,61 +1,61 @@
...
+
+    <template #body>
+      <template v-if="isMobilePortrait || !currentMatch">
+        <div class="d-flex justify-center">
+          <v-img :src="logoSrc" :data-cy="logoDataAttr" class="logo-image" />
```
itsalaidbacklife, line 11: The centered image now appears on mobile portrait for ranked games in addition to showing up for unranked games on all device sizes.

Mobile portrait for ranked game
image

Midsized non-ranked
image

src/components/GameView/GameOverlays.vue
```diff
@@ -235,7 +236,8 @@ export default {
 }
 .overlay-header {
     font-weight: bold;
-    background-color: #FFF4D7;
+    background-color: rgba(var(--v-theme-surface-2));
+    color: rgba(var(--v-theme-surface-1));
```
itsalaidbacklife, line 240: Updated color reference to use theme and set the text to surface-1 (chocolate) from black

src/components/GameView/MoveChoiceCard.vue
```diff
@@ -85,8 +85,15 @@ export default {
   cursor: pointer;
 }
 .move-choice-card {
-  border: 2px solid #fff;
+  background-color: rgba(var(--v-theme-surface-1));
+  color: rgba(var(--v-theme-surface-2)) !important;
```
itsalaidbacklife, line 89: ` !important` smells here but it didn't work without it and I don't think it's a huge deal

src/components/GameView/OpponentRequestedStalemateDialog.vue
```diff
@@ -53,9 +58,9 @@ export default {
       get() {
         return this.modelValue;
       },
-      set() {
-        // do nothing - parent controls whether dialog is open
-      },  
+      set(val) {
+        this.$emit('update:modelValue', val);
+      },
```
itsalaidbacklife, line 63: This fixes a bug where the dialog wouldn't close when the request completed. Normally we handle this with socket updates but since requesting a stalemate should generally close the dialog it seemed reasonable to close it here unlike the moves you make through a dialog

src/components/Global/BaseDialog.vue
```diff
@@ -0,0 +1,70 @@
...
+      <v-card-title class="d-flex justify-space-between pt-4">
+        <h1 v-if="title">{{ title }}</h1>
+        <slot name="title" />
+      </v-card-title>
+      <v-card-text>
+        <slot name="body" />
+      </v-card-text>
+      <div class="dialog-actions d-flex justify-end">
+        <slot name="actions" />
+      </div>
```
itsalaidbacklife, line 13: I had this wrapped in a v-card-actions initially but it did weird stuff with the padding so I just used a div instead

src/components/Global/BaseDialog.vue
```diff
@@ -0,0 +1,70 @@
...
+      default: '',
+    },
+    id: {
+      type: String,
+      required: true,
+    },
+    scrollable: {
+      type: Boolean,
+      default: false,
+    },
```
itsalaidbacklife, line 38: We could have used a generic attribute binding to set any vuetify props from the parent (this one is needed for the Three and Scrap dialogs since they have long lists of cards), but I figured making it more explicit would be preferable. It has the downside of requiring an update to this component if we want a consumer to make use of an additional vuetify dialog prop but at least the list of options will be clear when looking at this component itself. Probably make it easier to refactor it out of vuetify if we ever do that as well

src/plugins/vuetify.js
```diff
@@ -34,6 +34,8 @@ const cuttleTheme = {
     'neutral-lighten3': '#F2F4F5',
     'neutral-lighten4': '#FAFAFA',
     'neutral-lighten5': '#FAFBFC',
+    'surface-1': '#4A2416',
```
itsalaidbacklife, line 37: i.e. chocolate

src/plugins/vuetify.js
```diff
@@ -34,6 +34,8 @@ const cuttleTheme = {
...
+    'surface-2': '#FFF4D7',
```
itsalaidbacklife, line 38: i.e. cream

src/plugins/sails.js
```diff
@@ -171,11 +171,12 @@ io.socket.on('game', function (evData) {
...
           store.dispatch('updateGameThenResetPNumIfNull', evData.data.game);
           break;
-        case 'requestStalemate':
-          if (evData.data.requestedByPNum !== store.state.game.myPNum && !evData.data.victory.gameOver) {
-            store.commit('setConsideringOpponentStalemateRequest', true);
-          }
+        case 'requestStalemate': {
+          const opponentConsideringStalemate = evData.data.requestedByPNum !== store.state.game.myPNum &&
+          !evData.data.victory.gameOver;
+          store.commit('setConsideringOpponentStalemateRequest', opponentConsideringStalemate);
```
itsalaidbacklife, line 177: This closes the dialog for asking your opponent to stalemate when they respond to your request. It's redundant to the set functionality inside the OpponentRequestsStalemate dialog and might be better to just handle here.

It felt reasonable to overkill it just to make sure the damn thing closed but maybe I should remove the setter logic in the dialog for consistency. 

Notably I'm using the ternary assignment on 175 here to prevent the need for a conditional. That said it's now a multi-line boolean expression which may not be better. @seriouslysean penny for your thoughts

src/plugins/vuetify.js
seriouslysean: The names here aren't very descriptive, but that's something we can mess with later. I think we should hit up some of the designer folks and get a naming structure for our colors. Maybe work up a panel and have something like:
- `background-primary`
- `background-secondary`
- `button-primary`
- `button-secondary`

There are more, maybe @peterancourt would want to weigh in. Adding these to Storybook seems like a no-brainer.

tests/e2e/specs/in-game/countering.spec.js
```diff
@@ -241,8 +241,8 @@ describe('Countering One-Offs', () => {
     cy.log('Player counters again (3rd counter');
     cy.get('#counter-dialog')
       .should('be.visible')
-      .should('contain', 'Your opponent has played 2♣️ to Counter your 2♥️.')
-      .get('[data-cy=counter]')
+      .contains('Your opponent has played 2♣️ to Counter your 2♥️', {includeShadowDom: true});
```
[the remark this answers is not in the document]
seriouslysean: Might be better to abstract this in to utility if it's going to be common and/or look in to other methods of asserting here.

src/components/GameView/CannotCounterDialog.vue
```diff
@@ -1,47 +1,62 @@
...
+    <template #actions>
+      <v-btn
+        data-cy="cannot-counter-resolve"
+        color="surface-1"
+        variant="flat" 
+        @click="$emit('resolve')"
+      >
+        Resolve
+      </v-btn>
+    </template>
```
seriouslysean, line 45: I don't hate passing buttons in this way but does lead to being able to pass in any elements which might not be desired. It might be better to eventually come up with a structure of data to pass to generate the buttons instead.

src/components/GameView/ChooseTwoDialog.vue
seriouslysean: This is a tough one but naming is always hard. Web components are recommended to have a 2 word name. PascalCase is also the recommended format for referencing them in the template so we may want a follow up ticket to go through and rename everything. 😬
- https://vuejs.org/guide/components/registration.html#local-registration

This is Vue 2, but still interesting and informative:
- https://v2.vuejs.org/v2/style-guide/?redirect=true#Multi-word-component-names-essential

src/components/GameView/ChooseTwoDialog.vue
```diff
@@ -0,0 +1,83 @@
...
+    show: {
+      get() {
+        return this.modelValue;
+      },
+      set() {
+        // do nothing - parent controls whether dialog is open
+      },
+    },
+    opponentLastTwo() {
+      return this.twosPlayed && this.twosPlayed.length > 0
```
seriouslysean, line 64: `this.twosPlayed && this.twosPlayed.length` is being used a couple times, might want to make that a computed prop though the duplication is hardly a breaker.

src/components/GameView/ChooseWhetherToCounterDialog.vue
```diff
@@ -0,0 +1,118 @@
...
+    opponentLastTwo() {
+      return this.twosPlayed && this.twosPlayed.length > 0
+        ? this.twosPlayed[this.twosPlayed.length - 1]
+        : null;
+    },
+    playerLastTwo() {
+      return this.twosPlayed && this.twosPlayed.length > 1
+        ? this.twosPlayed[this.twosPlayed.length - 2]
+        : null;
+    },
```
seriouslysean, line 98: There's a level of duplication between this and the other component that smells. It's fine for now but longterm this feels like a problem waiting to happen.

src/components/GameView/MoveChoiceCard.vue
seriouslysean: I'd urge you to reconsider adding this, but it's fine for now. Longer term I think you need to just increase the specificity. I'm also surprised since these component styles are scoped. Could also add an ID for the unique dialogs which would supercede it, but that also doesn't feel right.

It's possible you're using this to override the children when maybe the children rules need to be adjusted instead.

src/components/GameView/ReauthenticateDialog.vue
```diff
@@ -1,52 +1,56 @@
...
+      <template #body>
+        <p>You have disconnected due to inactivity. Log in again to resume your session</p>
+        <!-- Login Form -->
+        <v-text-field
+          v-model="username"
+          variant="outlined"
+          :dense="$vuetify.display.mdAndDown"
+          label="Username"
+          data-cy="username"
+          @keydown.enter="login"
```
seriouslysean, line 14: This seems weird, the text fields should be submitting on enter anyway based on a form action. Not the end of the world though.

src/components/Global/BaseDialog.vue
seriouslysean: I think requiring an update to modify it is fine, we have a specific way to use this component so it doesn't need to be totally generic.

src/plugins/sails.js
seriouslysean: I'm very in favor of doing this in one location. Remove it from the dialog and maybe leave a comment that it's handled here instead.

seriouslysean (approved):
No real blockers, few suggestions though.

src/components/GameView/ChooseTwoDialog.vue
itsalaidbacklife: I thought the recommendation was just that the name be multi-word not exactly 2. Don't these names conform to that rule?

I've created https://github.com/cuttle-cards/cuttle/issues/396 to swap to PascalCase template instances. Better to tackle that one in one fell swoop in a separate PR

src/components/GameView/CannotCounterDialog.vue
itsalaidbacklife: That's an interesting point. I think it's nice to err on the side of flexibility for now and keep an eye out for how to put this on rails in the future. Maybe a prop based setup for structuring the buttons would make sense. But atm this leaves it up to the consumer and avoids the need to bubble the events b/c you can bind the handlers directly on the buttons you put inside the slot, which I think is nice

src/components/GameView/MoveChoiceCard.vue
itsalaidbacklife: I'll do some further thinkering to see if I can get the specificity right. I also initially thought scoped styles would do it, but I think that because they ultimately just boil down to classes with hashed names that the scoped styles actually _don't_ increase the specificity, they just prevent style leaks into other files. I think your suggestion of an id selector makes sense; I'll try that first

src/components/GameView/ChooseWhetherToCounterDialog.vue
itsalaidbacklife: Since we have the HoC I think the move is to compute these in the parent and send them as a prop to keep the logic centralized

itsalaidbacklife: Actually that only applies if the computed props are used in both children. I'm silly both of these are only need in one of them. I'm nixing this from the other too. Silly me 😅

src/components/GameView/ChooseTwoDialog.vue
itsalaidbacklife: DRY'd this up by pulling that expression into a computed prop as you suggested

src/plugins/vuetify.js
itsalaidbacklife: Agreed both on adding to storybook and redefining the whole color system. Honestly I think it will likely make sense to get rid of the purple and orange in the theme altogether and make the chocolate and cream colors the primary and secondary, but we should get some design input before making the change

itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/397

src/components/GameView/OpponentRequestedStalemateDialog.vue
itsalaidbacklife: Removed in favor of handling the update in the store like the other dialogs

src/components/Global/BaseDialog.vue
```diff
@@ -0,0 +1,70 @@
...
+  },
+}
+</script>
+
+<style scoped>
+.dialog-card {
+  color: rgba(var(--v-theme-surface-2)) !important;
+  opacity: .92;
+  border: 4px solid rgba(var(--v-theme-surface-2));
+  border-radius: 12px !important;
```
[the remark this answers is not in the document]
itsalaidbacklife: @seriouslysean so interestingly the style that I'm bumping into for the color here is this:
image


It's injected by vuetify either for cards or for dialogs, I'm not sure atm. It's tricky to override because vuetify is using important and because we don't have an easy high specificity selector since we want this to apply to the whole dialog.

Interestingly we can make it work by changing the theme color name to `on-surface-1` which is the material design name for the color you put on top of your surfaces. So that could do it, but then it's not necessarily flexible for inverting the colors in some areas of the application. I'm still not sure what's a good way to handle it

src/components/Global/BaseDialog.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+  <v-dialog v-model="show" persistent max-width="650" :scrollable="scrollable" elevation scrim="surface-1">
+    <v-card :id="id" class="dialog-card" color="surface-1">
+      <v-card-title class="d-flex justify-space-between pt-4">
+        <h1 v-if="title">{{ title }}</h1>
+        <slot name="title" />
```
seriouslysean, line 6: This should probably be a v-if/else, so that you can use the prop for simple use-cases, or the slot for more complex use cases. If you're going to use both then they should be named different/uniquely.

src/components/Global/BaseDialog.vue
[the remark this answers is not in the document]
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/398

src/components/GameView/MoveChoiceCard.vue
itsalaidbacklife: Created https://github.com/cuttle-cards/cuttle/issues/398 to track fixing the `!important` usages

tests/e2e/specs/in-game/countering.spec.js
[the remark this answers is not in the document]
itsalaidbacklife: I'm not sure what's the right way of doing that but the best I can come up with is to define a custom child cypress command that we can chain off of `cy.get()` to make contains assertions with `includeShadowDom` set to `true`. I'm not sure that's actually more clear though since it basically trades a one-liner for another one-liner

src/components/GameView/ReauthenticateDialog.vue
itsalaidbacklife: This seems like a good compromise. I'll give it a shot

tests/e2e/specs/in-game/countering.spec.js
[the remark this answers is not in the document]
itsalaidbacklife: I think it's best to hold off on this and keep an eye out for patterns in when we need `includeShadowDom`. Gonna hold off on making an issue pending a clearer strategy for dealing with this

src/components/GameView/ReauthenticateDialog.vue
itsalaidbacklife: Okay I am up for this. One gotcha is that we still need `@keydown.enter` handlers on the text fields because the form won't submit-on-enter when the form doesn't have a submit button. I considered adding a hidden submit button, but I think this is more clear

itsalaidbacklife: Now it's this

```
        <form ref="form" @submit.prevent="login">
          <v-text-field
            v-model="username"
            variant="outlined"
            :dense="$vuetify.display.mdAndDown"
            label="Username"
            :loading="isLoggingIn"
            data-cy="username"
            @keydown.enter="submit"
          />
          <v-text-field
            v-model="password"
            variant="outlined"
            label="Password"
            :dense="$vuetify.display.mdAndDown"
            type="password"
            :loading="isLoggingIn"
            data-cy="password"
            @keydown.enter="submit"
          />
        </form>
      </template>

      <template #actions>
        <v-btn variant="text" color="primary" @click="leaveGame">
          Leave Game
        </v-btn>
        <v-btn
          color="primary"
          variant="flat"
          data-cy="login"
          :loading="isLoggingIn"
          @click="submit"
        >
          Log In
        </v-btn>
      </template>
```

Issue #374, opened 2023-03-26
[DevEx]: Reorganize `status` and `result` database columns


itsalaidbacklife:
### Improvement Summary

Currently, the `game` model and database table has two interrelated columns: `status` which represents whether a game can be joined by a new player, and `result`, which represents the outcome of the game. These could potentially be combined or reorganized for improved clarity

### Detailed Description

`status` is a boolean which is true if the game is available to join and false otherwise. This generally corresponds to whether the game has fewer than two players and has not yet started. So it’s possible that this column isn’t necessary and instead can be computed based on other data. The thing this fails to capture is that sometimes we manually set this value to false in order to soft delete games, but perhaps we could hard delete or use a separate isActive column for that purpose.

The `result` field is an enum that captures whether the game is incomplete (-1, the default value), stalemate (2), won by p0 (0) or won by p1 (1). This could potentially be improved to differentiate between games that have been started or not and also to make it easier to tell which specific user won (currently you need to cross reference who was which player against this value).

This ticket represents an investigation of possible database and api architectures that could better represent these states and their relationships.

itsalaidbacklife:
Proposal:

replace the `status` and `result` with the following two columns:

`status`: (int enum) represents the game's lifecycle. It can have three values:
1  - NEW (game has been created but has not started)
2 - STARTED (game has been started but has not yet finished)
3 - FINISHED (game has been completed)
4 - ARCHIVED (game has been removed from the list without being started)

`winner`: int fk(user) represents the id of the user who won the game

- It will be null unless the game has `status = FINISHED`. 
- It will be null if the game ended in a stalemate

Putting these together, we can better accommodate common use-cases:

- A game can only be joined by a player if it has `status = STARTED` AND < 2 players (this will fix #346)
- We can tell that a game ended in a stalemate if it has `status = FINISHED` and `winner = null` 
- We can more easily find the games that any player won by looking for the games where the `winner = <player_id>`

PR #403, opened 2023-05-08
Fix/prevent joining closed games

### LINKED ISSUE #346: [BUG]: Backend allows subscribing to closed games
### Bug Summary

When a game becomes full (has 2 players) the frontend disables the join button for that game, but the backend still accepts requests to subscribe to it

### Detailed Description

The /game/subscribe endpoint makes no check that the game is open for players to join. Currently users are prevented from subscribing to games client side, but we should also prevent this on the backend.

Basically this endpoint should query for the game being requested *before* subscribing you to it, and should return a badRequest response if that game has a `status === false` or if the number of players it has is > 1. 

The frontend should then show a snackbar when this error occurs 

### Current Behavior

If you remove the `disabled` prop from the join button in `GameListItem.Vue` and create a game, join it as two players then try to join it as a third player, it will bring the new player to the lobby of the game and add them to the list of players, causing weird effects in the game.

### Expected Behavior

When a user attempts to join a game that’s closed or that has two players in it already, the backend should throw an error that’s displayed in a snackbar to the user saying “Oops, that game is already full”

### Steps to Reproduce

1. Remove the `disabled` prop from `GameListItem.Vue`
2. Start the front and back ends, sign up three accounts in three separate browsers
3. Create a game, and join it as two players
4. Join it as the third player and you will be incorrectly brought to the lobby

itsalaidbacklife:
## Please check the following
Resolves #346 by throwing backend error when subscribing to game that's already full. Added appropriate assertions to the relevant test case.

- For New Features:

Throw an error when request to subscribe is made on a full game, disable the join button and add display a snackbar error
image

### DIFF
--- a/api/controllers/game/subscribe.js
+++ b/api/controllers/game/subscribe.js
@@ -17,7 +17,7 @@ module.exports = function (req, res) {
       const [game, user] = arr;
 
       // Fast fail if game is full
-      const gameIsFull = !game.status || game.players.length >= 2 || game.log.length > 0;
+      const gameIsFull = sails.helpers.isGameFull(game);
       if (gameIsFull) {
         throw { message: `Cannot join that game because it's already full` };
       }
--- a/api/controllers/game/subscribe.js
+++ b/api/controllers/game/subscribe.js
@@ -15,6 +15,10 @@ module.exports = function (req, res) {
     .then(async function success(arr) {
       // Catch promise values
       const [game, user] = arr;
+
+      if (!game.status) {
+        throw {message: `Cannot join that game because it's already full`};
+      }
       // Does the user already have a pnum for this game?
       let pNum = getPlayerPnumByUsername(game.players, user.username);
       if (!pNumIsValid(pNum) && game.players) {
--- a/api/controllers/game/subscribe.js
+++ b/api/controllers/game/subscribe.js
@@ -9,7 +9,7 @@ module.exports = function (req, res) {
   }
   Game.subscribe(req, [req.body.id]);
   const promiseClearOldGame = gameService.clearGame({ userId: req.session.usr });
-  const promiseGame = gameAPI.findGame(req.body.id).populate('players');
+  const promiseGame = gameAPI.findGame(req.body.id);
   const promiseUser = userAPI.findUser(req.session.usr);
   Promise.all([promiseGame, promiseUser, promiseClearOldGame])
     .then(async function success(arr) {
--- a/api/controllers/game/subscribe.js
+++ b/api/controllers/game/subscribe.js
@@ -9,15 +9,17 @@ module.exports = function (req, res) {
   }
   Game.subscribe(req, [req.body.id]);
   const promiseClearOldGame = gameService.clearGame({ userId: req.session.usr });
-  const promiseGame = gameAPI.findGame(req.body.id);
+  const promiseGame = gameAPI.findGame(req.body.id).populate('players');
   const promiseUser = userAPI.findUser(req.session.usr);
   Promise.all([promiseGame, promiseUser, promiseClearOldGame])
     .then(async function success(arr) {
       // Catch promise values
       const [game, user] = arr;
 
-      if (!game.status) {
-        throw {message: `Cannot join that game because it's already full`};
+      // Fast fail if game is full
+      const gameIsFull = !game.status || game.players.length >= 2 || game.log.length > 0;
+      if (gameIsFull) {
+        throw { message: `Cannot join that game because it's already full` };
       }
       // Does the user already have a pnum for this game?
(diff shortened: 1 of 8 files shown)


src/store/modules/game.js
```diff
@@ -326,7 +326,8 @@ export default {
               });
               return resolve();
             }
-            return reject(new Error('error subscribing'));
+            const message = res.message ?? 'error subscribing';
+            return reject(new Error(message));
```
itsalaidbacklife, line 330: Updated to show specific error message if it is available

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -148,6 +149,18 @@ describe('Home - Game List', () => {
 
       // Test that join button is now disabled
       cy.contains('[data-cy-join-game]', 'Play').should('be.disabled');
+
+      // Manually re-enable join button to confirm backend rejects request
+      cy.window().its('cuttle.app.config.globalProperties.$store')
+        .invoke('commit', 'updateGameStatus', {id: gameData.gameId, newStatus: true});
```
itsalaidbacklife, line 155: I spent a while trying to figure out how to test this issue. I tried having the test directly 
```
.invoke('dispatch', 'requestSubscribe', {id: gameData.gameId})
```

But I had trouble managing the promises to make the test pass if we got the expected error and fail otherwise. Plus that doesn't confirm that the snackbar shows up displaying the error message. I thought about disconnecting and reconnecting the socket so the player misses that the game is over, but that's a bit brittle because eventually we'd like the socket reconnect logic to rehydrate the list of which games are available.

Eventually I settled on having the test code manually switch a game to `status: true` by invoking the store mutation directly here. I'm really pleased with how this turned out because it lets the test control the game looking 'available' on the clientside without weird backend implications and then the actual interactions and assertions (clicking the button then asserting the button disables and the snackbar appears) happen organically through the app as usual

api/controllers/game/subscribe.js
```diff
@@ -15,6 +15,12 @@ module.exports = function (req, res) {
     .then(async function success(arr) {
       // Catch promise values
       const [game, user] = arr;
+
+      // Fast fail if game is full
+      const gameIsFull = !game.status || game.players.length >= 2 || game.log.length > 0;
```
seriouslysean, line 20: I feel like this should be moved to a helper function if you’ll need it in more than one place ever. It could take game as an argument.

api/controllers/game/subscribe.js
```diff
@@ -15,6 +15,12 @@ module.exports = function (req, res) {
...
+      if (gameIsFull) {
+        throw { message: `Cannot join that game because it's already full` };
+      }
```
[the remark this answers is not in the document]
seriouslysean: Is it worth walking through each way a game could be considered full? Or possibly updating the logic to better track when a game is full? Could even add a few second delay before folks are able to join a game but that seems kinda weird.

I’m not too familiar with how quickly everything updates though.

src/views/HomeView.vue
```diff
@@ -35,7 +35,7 @@
                       :status="game.status"
                       :num-players="game.numPlayers"
                       :is-ranked="game.isRanked"
-                      @error="handleError"
+                      @error="handleSubscribeError(game.id, $event)"
```
seriouslysean, line 38: You’re sending the event but then calling it message in the method? That seems kind of confusing to me.

seriouslysean (approved):
This seems fine to me, just a few questions.

api/controllers/game/subscribe.js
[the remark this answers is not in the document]
itsalaidbacklife: I definitely want to update the logic to better track when a game is full. That's essentially #374, which would hopefully make this more water tight. For now I think this is as specific as we can get with the data that's currently stored.

api/controllers/game/subscribe.js
itsalaidbacklife: Pulled into a new synchronous `isGameFull(game)` helper

src/views/HomeView.vue
itsalaidbacklife: So the event payload is the error message. It seemed more descriptive for the method argument to describe what the event is rather than just calling it `$event` inside of `handldeSubscribeError()`. Do you disagree? Do you think we should call it `$event` in both places?

We're stuck using $event in the event handler here in the html, so if we want them to be the same I would think we'd have to call both of them `$event`. But isn't it more clear to have the method argument say what the event _is_?

itsalaidbacklife: Is it the name `message` that's specifically confusing? Would it be better as `errMessage`?

PR #406, opened 2023-05-09
Feature #402 Enable stop spectating

### LINKED ISSUE #402: [Feature]: Enable stop spectating.
### Feature Summary

Enable stop spectating/ update game menu

### Detailed Description

As of now, there is no option to stop spectating a match. We should replace the 'concede' and 'stalemate' option in the game menu with a 'Stop spectating' option.

This will require a little bit of refactoring on the sails model, we will need to create a new model with a through association instead of the many to many that is currently in place, we should add an activelySpectating boolean value to each user that is spectating, and then filter out on the front end by activelySpectating.

We may have to manually query this information.
[Sails Through Associations](https://sailsjs.com/documentation/concepts/models-and-orm/associations/through-associations)

Haviles04:
## Please check the following

- For New Features:

Fixes #402 

image

- Added stop-spectating "Go Home" game menu option 
- Added corresponding spectate-leave.js and logic
- Hid "concede" and "request stalemate" when spectating
- Refactored spectate.js to better reflect who's actively spectating
- Added a new Sails.js model, and changed the way the models and spectators are associated
- Added corresponding test

Database Migration 
`` CREATE TABLE "userspectatinggame"(
"createdAt" INT8,
"updatedAt" INT8,
id SERIAL NOT NULL,
"activelySpectating" BOOLEAN,
"gameSpectated" INTEGER REFERENCES "game" ON DELETE CASCADE,
"spectator" INTEGER REFERENCES "user" ON DELETE CASCADE,
PRIMARY KEY(id));  ``

`` DROP TABLE "game_spectatingUsers__user_spectatedGames"; ``


api/services/gameService.js
```diff
@@ -111,7 +120,8 @@ module.exports = {
                   if (game.players.length > 1) {
                     const p0 = userService.findUser({ userId: game.players[0].id });
                     const p1 = userService.findUser({ userId: game.players[1].id });
-                    return Promise.all([Promise.resolve(game), p0, p1]);
+                    const spectators = fetchSpectatorUsernames(game.id);
```
Haviles04, line 123: I'm having an issue when this function is called, I am getting an empty array and it seems like it's just skipping this function completely.

api/controllers/game/spectate-leave.js
```diff
@@ -0,0 +1,20 @@
...
+    });
+
+    const fullGame = await gameService.populateGame({ gameId: gameId });
+    Game.publish([fullGame.id], {
+      change: 'spectatorLeft',
+      game: fullGame,
+    });
+    return res.ok(fullGame);
+  } catch (err) {
+    console.log(err);
```
itsalaidbacklife, line 17: We don't log errors to the server console b/c no one reads them. The right thing to do is to set up error logging, which we have an issue for but haven't completed. For now we can just nix this line
```suggestion
```

api/controllers/game/spectate.js
```diff
@@ -13,15 +13,19 @@ module.exports = async function (req, res) {
...
+    );
 
+    const fullGame = await gameService.populateGame({ gameId: game.id });
     Game.publish([fullGame.id], {
       change: 'spectatorJoined',
       game: fullGame,
     });
     return res.ok(fullGame);
   } catch (err) {
+    console.log(err);
```
itsalaidbacklife, line 28:
```suggestion
```

itsalaidbacklife, line 28:
```suggestion
```

api/models/UserSpectatingGame.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * UserSpectatingGame.js
+ *
+ * @description :: TODO: You might write a short summary of how this model works and what it represents here.
```
itsalaidbacklife, line 4:
```suggestion
 * @description :: Represents a user's relationship to a game they are spectating
```

src/components/GameView/GameMenu.vue
```diff
@@ -14,10 +14,15 @@
...
-        </v-list-item>
+        <v-item-group v-if="!isSpectating">
+          <v-list-item data-cy="concede-initiate" @click.stop="openConcedeDialog"> Concede</v-list-item>
+          <v-list-item data-cy="stalemate-initiate" @click.stop="openStalemateDialog">
+            Request Stalemate
+          </v-list-item>
+        </v-item-group>
+
+        <!-- Stop Spectating -->
+        <v-list-item v-else data-cy="stop-spectating" @click.stop="stopSpectate"> Go Home</v-list-item>
```
itsalaidbacklife, line 25: It's often easier to read if/else logic if you can avoid the need for !'s. So I'd prefer to put the spectating case above the non-specating stuff. Additionally we don't want a v-item-group because we don't want to create a functional difference in the UI, we just want to group these tags together for use with the v-if directive. So we should use a `<template>` as the wrapper instead (template is the generic wrapper in vue that doesn't get actually rendered -- it's the correct way to group things together to apply directives to the group)

```suggestion
        <!-- Stop Spectating -->
        <v-list-item v-if="isSpectating" data-cy="stop-spectating" @click.stop="stopSpectate"> Go Home</v-list-item>
        <template v-else="!isSpectating">
          <v-list-item data-cy="concede-initiate" @click.stop="openConcedeDialog"> Concede</v-list-item>
          <v-list-item data-cy="stalemate-initiate" @click.stop="openStalemateDialog">
            Request Stalemate
          </v-list-item>
        </template>
```

src/components/GameView/GameMenu.vue
```diff
@@ -132,6 +143,11 @@ export default {
       this.loading = false;
       this.showStalemateDialog = false;
     },
+    async stopSpectate() {
+      this.loading = true;
+      await this.$store.dispatch('requestSpectateLeave');
+      this.$router.push('/');
```
itsalaidbacklife, line 149: Just to be safe let's set the loading state back to false before going home
```suggestion
      this.loading = false;
      this.$router.push('/');
```

src/store/modules/game.js
```diff
@@ -350,7 +350,17 @@ export default {
...
     },
-
+    async requestSpectateLeave(context) {
+      return new Promise((resolve, reject) => {
+        io.socket.get('/game/spectateLeave', function handleResponse(res, jwres) {
+          if (jwres.statusCode === 200) {
+            context.commit('resetState');
+            return resolve();
+          }
+          return reject(new Error('Error leaving spectator'));
```
itsalaidbacklife, line 360: What should we do when we get this error? I think the menu should at least navigate us home still so we aren't trapped in a game we don't want to watch anymore. It's okay to throw this error here as long as it's handled above.
```suggestion
          return reject(new Error('Error leaving game as spectator'));
```

src/components/GameView/GameMenu.vue
itsalaidbacklife, line 149: We should handle the possible error from the store dispatch and set the loading state to false before navigating away. In this case we want to navigate home even if there's an error so I think we can use `finally` without a catch

```suggestion
      try {
        this.loading = true;
        await this.$store.dispatch('requestSpectateLeave');
      } finally {
        this.loading = false;
        this.$router.push('/');
      }
```

itsalaidbacklife: Also the loading state is currently only used in the dialog, but we don't show the dialog during this flow, so we should add a loading state into the menu itself (see my suggestion above)

itsalaidbacklife (approved):
Looking good! Most of my comments are in the GameMenu

api/controllers/game/spectate-leave.js
Haviles04: Oops! Yea haha this was just for myself. Forgot to delete them lol

src/components/GameView/GameMenu.vue
Haviles04: ***v-list-group

itsalaidbacklife:
Commenting with a formatted version of the migration script

```
CREATE TABLE "userspectatinggame"
(
	"createdAt" INT8, 
	"updatedAt" INT8,
	id SERIAL NOT NULL,
	"activelySpectating" BOOLEAN, 
	"gameSpectated" INTEGER REFERENCES "game" ON DELETE CASCADE,
	"spectator" INTEGER REFERENCES "user" ON DELETE CASCADE,
	PRIMARY KEY(id)
);
```

```
DROP TABLE "game_spectatingUsers__user_spectatedGames";
```

PR #408, opened 2023-05-12
Out of game dialogs style/color

DonPierreDon:
## Please check the following

- For New Features:

The home.spec test is not passing, i have less lines in the home.spec.js than in the official repo but when i try to pull it says everything is up to date.


src/components/GameView/GameMenu.vue
```diff
@@ -8,19 +8,28 @@
...
-          Request Stalemate
+        <v-list-item data-cy="concede-initiate">
+          <v-btn class="menu-button" color="surface-2" :size="buttonSize" @click.stop="openConcedeDialog">
+            Concede
+          </v-btn>
+        </v-list-item>
+        <v-list-item data-cy="stalemate-initiate">
+          <v-btn class="menu-button" color="surface-2" :size="buttonSize" @click.stop="openStalemateDialog">
+            Request Stalemate
+          </v-btn>
```
itsalaidbacklife, line 26: Instead of nesting `v-btn` elements inside the list items, we can just use the list item itself and give each list item a different`value` prop. That'll give it the hover and click-ripple effects you'd expect from a button. Check the code from the first example in the vuetify [docs on menus](https://vuetifyjs.com/en/components/menus/)

src/components/CreateGameDialog.vue
```diff
@@ -1,61 +1,62 @@
...
+      </div>
+      <v-text-field
+        v-model="gameName"
+        autofocus
+        :disabled="loading"
+        label="Game Name"
+        variant="outlined"
+        data-cy="game-name-input"
+        @keydown.enter="submitNewGame"
+      />
```
seriouslysean, line 26: This should be wrapped in an actual form, and then you won't need the `.enter` modifier. It's also more valid and accessible with the form wrapper.

src/components/CreateGameDialog.vue
```diff
@@ -1,61 +1,62 @@
...
+      </v-btn>
+      <v-btn
+        data-cy="submit-create-game"
+        :loading="loading"
+        color="surface-1"
+        variant="flat"
+        @click="submitNewGame"
+      >
+        Create Game
+      </v-btn>
```
seriouslysean, line 46: When made in to a form, this would need to be type submit most likely.

src/components/GameView/GameMenu.vue
```diff
@@ -8,19 +8,22 @@
         </v-btn>
       </template>
       <!-- Menu -->
-      <v-list id="game-menu">
-        <v-list-item data-cy="rules-open" @click.stop="showRulesDialog = true">
+      <v-list id="game-menu" class="bg-surface-2">
+        <v-list-item data-cy="rules-open" :size="buttonSize" @click.stop="openRulesDialog">
```
seriouslysean, line 12: Does this actually need the modifier `.stop`? Not sure if there are other elements that would fire a click action passed this point.

src/components/RulesDialog.vue
```diff
@@ -1,240 +1,224 @@
...
     },
   },
   methods: {
     close() {
-      this.internalShow = false;
       this.$emit('close');
     },
+    open() {
+      this.$emit('open');
+    },
```
seriouslysean, line 221: I feel like I'm missing something, how does this get called? Is it something Vuetify is doing under the hood?

src/components/CreateGameDialog.vue
DonPierreDon: like this ? :

```vue
      <v-form>
        <v-text-field
          v-model="gameName"
          autofocus
          :disabled="loading"
          label="Game Name"
          variant="outlined"
          data-cy="game-name-input"
          @keydown="submitNewGame"
        />
      </v-form>
```

src/components/GameView/GameMenu.vue
DonPierreDon: Yes you are right, i am removing it, thanks for the feedback !

src/components/RulesDialog.vue
DonPierreDon: I think it is something i forgot to remove, after removing it it still shows the appropriate behavior, thanks for the feedback !

src/components/CreateGameDialog.vue
DonPierreDon: /
```vue
      <v-form @submit.prevent="submitNewGame">
        <v-text-field
          v-model="gameName"
          autofocus
          :disabled="loading"
          label="Game Name"
          variant="outlined"
          data-cy="game-name-input"
        />
      </v-form>
``` 


```vue
        <v-btn
          type="submit"
          data-cy="submit-create-game"
          :loading="loading"
          color="surface-1"
          variant="flat"
          @click="submitNewGame"
        >
          Create Game
        </v-btn>
``` 
like this right ?

src/components/GameView/GameMenu.vue
DonPierreDon:
```vue
    <v-menu v-model="showGameMenu">
      <!-- Activator -->
        <v-btn class="ml-0" id="game-menu-activator" v-bind="props" icon variant="text">
          <v-icon color="neutral-lighten-2" icon="mdi-cog" />
        </v-btn>
      </template>
      <!-- Menu -->
      <v-list id="game-menu" class="bg-surface-2">
        <v-list-item data-cy="rules-open" @click="openRulesDialog"> Rules </v-list-item>
        <!-- Concede Dialog (Initiate + Confirm) -->
        <v-list-item data-cy="concede-initiate" @click="openConcedeDialog"> Concede </v-list-item>
        <v-list-item data-cy="stalemate-initiate" @click="openStalemateDialog">
          Request Stalemate
        </v-list-item>
      </v-list>
    </v-menu>
``` 
I did it like that, i am not sure its ok, but it looks fine to me, thanks for the review !

src/components/CreateGameDialog.vue
itsalaidbacklife: Close, but there's a gotcha because the #actions slot separates the main content from the buttons in a way that messes up putting both the form content + the buttons inside the form.

Check the `ReauthenticateDialog` for an example of how to do this. Basically we can put an empty, hidden `<button type="submit">` inside the form just to make it submittable, and then put the actual button (as a non-submit type button) in the actions slot as usual, giving it a click handler that uses a template ref to submit the form. Then the form itself can have an @submit.prevent handler to do the actual request to create the game

src/components/CreateGameDialog.vue
itsalaidbacklife: Yes, see this example: https://www.dev-diaries.com/social-posts/html-submit-button-outside-form/#:~:text=For%20a%20HTML%20form%20element,with%20the%20button's%20form%20property.

I actually did it incorrectly in the ReauthenticateDialog I think because I used a ref to submit the form but if we just use the name attribute in the form and the form attribute in the button and make the button a type="submit", that should do it

src/components/CreateGameDialog.vue
```diff
@@ -1,61 +1,62 @@
 <template>
-  <v-dialog v-model="show">
-    <template #activator="{ props }">
-      <v-btn v-bind="props" color="primary" data-cy="create-game-btn" rounded elevation="8">
+  <base-dialog v-model="show" id="create-game" title="Create Game">
```
itsalaidbacklife, line 2: If we use the id prop to set the data-cy attribute then we can name this id create-game-dialog and it should have the correct selector for the test
```suggestion
  <base-dialog v-model="show" id="create-game-dialog" title="Create Game">
```

itsalaidbacklife (commented):
Looking good. I have one suggestion for how to implement the form, per @seriouslysean 's suggestion, and a suggested fix for the failing test

src/components/CreateGameDialog.vue
```diff
@@ -1,61 +1,62 @@
...
-        <v-text-field
-          v-model="gameName"
-          autofocus
-          :disabled="loading"
-          label="Game Name"
-          variant="outlined"
-          data-cy="game-name-input"
-          @keydown.enter="submitNewGame"
+    <template #body>
+      <div class="d-flex align-center">
```
itsalaidbacklife, line 9: Let's make this a `<form name="create_game_form">` and then use a `<button type="submit" form="create_game_form">` in the `#actions` to do the submit. That way we've got more semantic html and the use of the `name` and `form` attributes can make the form submit work even though the buttons in the `#actions` slot which is outside the `<form>` itself

itsalaidbacklife:
Replaced by #470

PR #410, opened 2023-05-13
Feature #281 enable stats page link

### LINKED ISSUE #281: [Feature]: Stats Page - Enable creating a link to a specific season's leaderboard
### Feature Summary

Currently the Stats page uses internal data/variables to manage which season is currently selected. It would be better if the selection were managed via a url param, so that you could then copy the url and send it to someone, bringing them directly to the season you were looking at when you copied the url.

### Detailed Description

Notice how these two seasons both have the same url: https://cuttle.cards/#/stats
image

image

If you send someone the url https://cuttle.cards/#/stats, they will always be brought to the latest season. It would be better if you could specifically link someone to the Spades 2022 season with a url like https://www.cuttle.cards/#/stats?seasonIndex=3

This season index should be loaded from the url when the page is created, and it should be updated whenever the dropdown is used to select another season. All of the changes should be doable using only src/client/js/views/Stats.vue

Haviles04:
## Please check the following

- For New Features:

fixes #281 

Added test and logic to add url params to /stats page.  e.g /stats/4

### DIFF
--- a/src/views/StatsView.vue
+++ b/src/views/StatsView.vue
@@ -68,7 +68,6 @@
         </h2>
         <stats-leaderboard :loading="loadingData" :season="selectedSeason" />
       </template>
-      <h1 class="d-flex justify-center" v-if="!selectedSeason">Oops this season doesn't exist!</h1>
     </section>
   </div>
 </template>
@@ -92,14 +91,11 @@ export default {
       loadingData: false,
       selectedSeason: null,
       seasons: [],
-      seasonIndex: this.$route.query.seasonIndex,
     };
   },
   watch: {
     selectedSeason() {
-      if (this.selectedSeason !== this.seasonIndex) {
-        this.$router.push({ path: '/stats', query: { seasonIndex: `${this.selectedSeason.id}` } });
-      }
+      this.$router.push({ name: 'StatsBySeason', params: { seasonId: this.selectedSeason.id } });
     },
   },
   computed: {
@@ -131,8 +127,11 @@ export default {
     this.loadingData = true;
     io.socket.get('/stats', (res) => {
       this.seasons = res;
-      if (this.seasonIndex) {
-        this.selectedSeason = this.seasons.filter(({ id }) => id === parseInt(this.seasonIndex));
+      const seasonId = parseInt(this.$route.params.seasonId);
+      //check if route params matches any season ID
+      if (this.seasons.some(({ id }) => id === seasonId)) {
+        const [selectedSeason] = this.seasons.filter(({ id }) => id === seasonId);
+        this.selectedSeason = selectedSeason;
       } else {
         const [selectedSeason] = this.seasons;
         this.selectedSeason = selectedSeason;
--- a/src/views/StatsView.vue
+++ b/src/views/StatsView.vue
@@ -61,7 +61,7 @@
         </p>
       </div>
       <!-- Rankings Table -->
-      <template v-if="selectedSeason && selectedSeason.rankings && selectedSeason.rankings.length > 0">
+      <template v-if="weeklyRanking">
         <h2 class="text-h2 mb-4">
           Weekly Rankings
           <stats-scoring-dialog />
@@ -88,17 +88,22 @@ export default {
   },
   data() {
     return {
-      loadingData: false,
+      loadingData: true,
       selectedSeason: null,
       seasons: [],
     };
   },
   watch: {
     selectedSeason() {
-      this.$router.replace({ name: 'StatsBySeason', params: { seasonId: this.selectedSeason.id } });
+      this.selectedSeason
+        ? this.$router.replace({ name: 'StatsBySeason', params: { seasonId: this.selectedSeason.id } })
+        : null;
     },
   },
   computed: {
+    weeklyRanking() {
+      return this.selectedSeason && this.selectedSeason.rankings && this.selectedSeason.rankings.length > 0;
+    },
     seasonStartFormatted() {
       return !this.selectedSeason ? '' : dayjs(this.selectedSeason.startTime).format('YYYY/MM/DD');
     },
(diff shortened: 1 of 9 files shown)


src/views/StatsView.vue
```diff
@@ -61,9 +61,7 @@
         </p>
       </div>
       <!-- Rankings Table -->
-      <template
-        v-if="selectedSeason && selectedSeason.rankings && selectedSeason.rankings.length > 0"
-      >
+      <template v-if="selectedSeason && selectedSeason.rankings && selectedSeason.rankings.length > 0">
```
seriouslysean, line 64: This should be converted to a computed prop since you have so many conditionals in here.

src/views/StatsView.vue
```diff
@@ -95,6 +93,11 @@ export default {
       seasons: [],
     };
   },
+  watch: {
+    selectedSeason() {
+      this.$router.replace({ name: 'StatsBySeason', params: { seasonId: this.selectedSeason.id } });
```
seriouslysean, line 98: This should be null safe against a selectedSeason not being set.

src/views/StatsView.vue
```diff
@@ -115,19 +118,39 @@ export default {
...
-          value: season
-        }
+          value: season,
+        };
       });
-    }
+    },
+  },
+  methods: {
+    checkAndSelectSeason(seasonId) {
```
seriouslysean, line 127: Is there ever the potential to call this without a seasonId?

src/views/StatsView.vue
```diff
@@ -115,19 +118,39 @@ export default {
...
+      if (this.seasons.some(({ id }) => id === seasonId)) {
+        const [selectedSeason] = this.seasons.filter(({ id }) => id === seasonId);
+        this.selectedSeason = selectedSeason;
+      } else {
+        const [selectedSeason] = this.seasons;
+        this.selectedSeason = selectedSeason;
+      }
```
seriouslysean, line 134: Might be better to fast fail here so it's easier to understand, but @itsalaidbacklife and I argue about if/else blocks all the time so I'll let him pick.
```js
const hasSeason = this.seasons.some(({ id }) => id === seasonId);

if (hasSeason) {
  const [selectedSeason] = this.seasons.filter(({ id }) => id === seasonId);
  this.selectedSeason = selectedSeason;
  return;
}

const [selectedSeason] = this.seasons;
this.selectedSeason = selectedSeason;
```

There's also probably an opportunity to not need to both do some and filter via reducer, but I don't think we need to needle on performance here.

src/views/StatsView.vue
```diff
@@ -115,19 +118,39 @@ export default {
...
-      this.selectedSeason = selectedSeason;
-      this.loadingData = false;
+      if (res.length > 0) {
+        this.seasons = res;
+        const seasonId = parseInt(this.$route.params.seasonId);
+        this.checkAndSelectSeason(seasonId);
+      }
     });
+    this.loadingData = false;
+  },
```
seriouslysean, line 147:
```js
created() {
  this.loadingData = true;

  io.socket.get('/stats', (res) => {
    if (!res.length) {
      // fast fail
      return;
    }
    this.seasons = res;
    const seasonId = parseInt(this.$route.params.seasonId);
    this.checkAndSelectSeason(seasonId);
  });
  
  this.loadingData = false;
},
```
Fast fails are preferred imo.

Also a little weird to me to set loadingData to true initially here, would it be better to have the default in the component state set to true and then just set to false when done? You may also want to wrap in a try catch because if the socket call fails in an uncaught way your false declaration at the bottom may never get called properly.

src/views/StatsView.vue
```diff
@@ -115,19 +118,39 @@ export default {
...
+  beforeRouteUpdate(to, from, next) {
+    this.loadingData = true;
+    const seasonId = parseInt(to.params.seasonId);
+    this.checkAndSelectSeason(seasonId);
+    this.loadingData = false;
+    next();
```
seriouslysean, line 153: We very likely need some better way to track loading states since we're doing this all over the place @itsalaidbacklife. I'm thinking at this point some sort of mixin for providing `$isLoading` as a state value to any component that includes it. Doesn't need to be for this PR though.

seriouslysean (commented):
Leaving final comments and approvals to @itsalaidbacklife.

src/views/StatsView.vue
Haviles04: I think we can get rid of some here, and just filter the result, and then just check for length

src/views/StatsView.vue
Haviles04: I don't believe so

src/views/StatsView.vue
itsalaidbacklife: Yeah I think we could also explore the use of the `<Suspense>` tag as a piece of the puzzle ([docs](https://vuejs.org/guide/built-ins/suspense.html)). I think this would largely manifest as a composable, but the trick will be in establishing a uniform api given that different components have different requirements for when loading states should display

src/views/StatsView.vue
```diff
@@ -60,16 +60,20 @@
...
+      <template v-if="weeklyRanking">
         <h2 class="text-h2 mb-4">
           Weekly Rankings
           <stats-scoring-dialog />
         </h2>
         <stats-leaderboard :loading="loadingData" :season="selectedSeason" />
       </template>
+      <!-- Error display -->
+      <div v-if="error" class="d-flex flex-column align-center text-center">
+        <h2>Oops! There was a problem loading the leaderboard. Refresh the page to try again</h2>
```
itsalaidbacklife, line 74: Let's make this a <p>
```suggestion
        <p class="text-body-1">Oops! There was a problem loading the leaderboard. Refresh the page to try again</p>
```

src/views/StatsView.vue
```diff
@@ -115,20 +131,36 @@ export default {
...
+    },
   },
   created() {
-    this.loadingData = true;
     io.socket.get('/stats', (res) => {
-      this.seasons = res;
-      const [selectedSeason] = this.seasons;
-      this.selectedSeason = selectedSeason;
       this.loadingData = false;
+      if (!res.length) {
```
itsalaidbacklife, line 148:
```suggestion
      if (!res?.length) {
```

PR #418, opened 2023-05-24
Bug #405 blank background

### LINKED ISSUE #405: [Bug]: Navigating to the url of a completed game gives you a blank background image
### Bug Summary

Navigating to a game url for a game that's over currently causes an uncaught error and leaves the user stuck on a dead page. We should detect this error and instead open a dialog that invites the user to navigate home.

### Detailed Description

Navigating to a completed game url like /#/game/1 currently causes an error that leaves the user on a blank page like this:
image

We should fix this so that users instead get a dialog that explains that this game is over and that they should go home, with a button to bring them to the home screen

### Current Behavior

The blank screen appears

### Expected Behavior

A dialog appears with:

Title: Oops!
Body: This game is over, nothing to see here
Actions: Button that says Go Home which behaves the same as the GameOverDialog's Go Home button

### Steps to Reproduce

1. Sign up and start a game with two players
2. End the game (by playing it out, conceding etc)
3. Wait for the GameOver dialog to appear and refresh the page
4. You'll see this screen
5. 
image

Haviles04:
## Issue number

Relevant issue number: Fixes #405 

## Please check the following

- For New Features:

- Added a Game unavailable overlay that matches waiting for game to start overlay
- Moved new overlay and reauthenticate dialog into a new Game-unavailable-view component.
- Wrote test for new overlay.

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "cuttle",
-  "version": "4.3.22",
+  "version": "4.3.21",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "cuttle",
-  "version": "4.3.21",
+  "version": "4.3.22",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
+  "version": "4.3.22",
   "version": "4.3.24",
   "dependencies": {
     "connect-redis": "3.0.2",
--- a/package.json
+++ b/package.json
@@ -1,8 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "4.3.22",
-  "version": "4.3.24",
+  "version": "4.3.25",
   "dependencies": {
     "connect-redis": "3.0.2",
     "dayjs": "1.11.3",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "4.3.25",
+  "version": "4.3.26",
   "dependencies": {
     "connect-redis": "3.0.2",
     "dayjs": "1.11.3",
--- a/src/components/GameView/GameUnavailableOverlay.vue
+++ b/src/components/GameView/GameUnavailableOverlay.vue
@@ -2,7 +2,7 @@
   <v-overlay
     data-cy="unavailable-game-overlay"
     id="unavailable-game-scrim"
-    v-model="unavailableGame"
+    :model-value="show"
     class="game-overlay text-center d-flex justify-center align-center"
   >
     <div class="overlay-header">
@@ -27,7 +27,7 @@
 export default {
   name: 'GameUnavailable',
   props: {
-    unavailableGame: {
+    show: {
       type: Boolean,
       required: true,
     },
--- a/src/components/GameView/GameUnavailableOverlay.vue
+++ b/src/components/GameView/GameUnavailableOverlay.vue
@@ -1,0 +1,56 @@
+<template>
+  <v-overlay
+    data-cy="unavailable-game-overlay"
+    id="unavailable-game-scrim"
+    v-model="unavailableGame"
+    class="game-overlay text-center d-flex justify-center align-center"
+  >
+    <div class="overlay-header">
+      <h1 :class="[this.$vuetify.display.xs === true ? 'text-h5' : 'text-h3', ['font-weight-bold']]">
+        Oops!
+      </h1>
+      <p>Looks like this game isn't available</p>
+    </div>
+    <v-btn
+      color="secondary"
+      class="mt-4"
+      data-cy="leave-unavailable-game-button"
+      :loading="leavingGame"
+      @click="this.$router.push('/')"
+    >
+      Go Home
+    </v-btn>
+  </v-overlay>
+</template>
+
+<script>
+export default {
+  name: 'GameUnavailable',
+  props: {
+    unavailableGame: {
+      type: Boolean,
+      required: true,
+    },
+  },
+};
+</script>
+
+<style scoped lang="scss">
+:deep(.v-overlay__content) {
+  left: 0;
+}
+.game-overlay {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  text-align: center;
+}
+.overlay-header {
+  font-weight: bold;
+  background-color: rgba(var(--v-theme-surface-2));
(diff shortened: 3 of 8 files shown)


itsalaidbacklife (changes requested):
Looking good. Requesting to pull the game unavailable overlay into its own component and for us to take a look at the computed prop that renders it

src/components/GameView/GameUnavailableView.vue
```diff
@@ -0,0 +1,75 @@
...
+        <v-btn
+          color="secondary"
+          class="mt-4"
+          data-cy="leave-unavailable-game-button"
+          :loading="leavingGame"
+          @click="this.$router.push('/')"
+        >
+          Go Home
+        </v-btn>
+      </v-overlay>
```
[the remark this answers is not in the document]
seriouslysean: Split it for sure, separation of concerns imo.

src/components/GameView/GameUnavailableOverlay.vue
```diff
@@ -0,0 +1,56 @@
...
+      Go Home
+    </v-btn>
+  </v-overlay>
+</template>
+
+<script>
+export default {
+  name: 'GameUnavailable',
+  props: {
+    unavailableGame: {
```
itsalaidbacklife, line 30: Let's rename this `show`
```suggestion
    show: {
```

src/components/GameView/GameUnavailableOverlay.vue
```diff
@@ -0,0 +1,56 @@
+<template>
+  <v-overlay
+    data-cy="unavailable-game-overlay"
+    id="unavailable-game-scrim"
+    v-model="unavailableGame"
```
itsalaidbacklife, line 5: Instead of a 2-way v-model binding let's just use a one-way model-value binding:
```suggestion
    :model-value="show"
```

itsalaidbacklife: v-model is for when you want to set the child's model value from the parent but also sync up the parent's variable with the child when the child emits an update, which we don't need to do here. This is more straightforward top-down prop assignment

src/views/GameView.vue
```diff
@@ -1,9 +1,10 @@
 <template>
   <div id="game-view-wrapper">
-    <!-- Unauthenticated/Must re-log in -->
+    <!-- Unauthenticated/Must re-log in/ Unavailable game -->
     <template v-if="$store.state.game.myPNum === null">
```
itsalaidbacklife, line 4: Do we still need this v-if? it looks like it'd be unnecessary with the logic in the `<game-unavailable-view>` itself

src/components/GameView/GameUnavailableOverlay.vue
```diff
@@ -0,0 +1,56 @@
...
+  text-align: center;
+}
+.overlay-header {
+  font-weight: bold;
+  background-color: rgba(var(--v-theme-surface-2));
+  color: rgba(var(--v-theme-surface-1));
+  padding: 24px;
+  text-align: center;
+  width: 100vw;
+}
```
itsalaidbacklife, line 55: Now that we're using these styles in another place, it'd be good to define a `BaseOverlay` component that can accommodate this and the component here as well as the styles in the existing `GameOverlays`. That can reasonably be scoped into a separate issue, though

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cuttle",
-  "version": "4.3.20",
+  "version": "4.3.22",
```
itsalaidbacklife, line 3: Going forward, let's refrain from updating package-lock.json's version field manually. It updates to match `package.json` whenever we run `npm install`, which means that we can use the version here as a record of the last time that command was run. This is fine for now, but we can leave it off next time

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "4.3.21",
+  "version": "4.3.22",
```
itsalaidbacklife, line 4: Merge main into this and then bump to 4.3.25

src/components/GameView/GameUnavailableOverlay.vue
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/430

src/views/GameView.vue
Haviles04: without it, I get a blank game background.

src/components/GameView/GameUnavailableView.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+  <div>
+    <!-- Reauthenticate-->
+    <template v-if="mustReauthenticate">
+      <reauthenticate-dialog v-model="mustReauthenticate" />
+    </template>
+    <!-- Game doesn't exist-->
+    <template v-else>
+      <game-unavailable-overlay :unavailableGame="unavailableGame" />
+    </template>
```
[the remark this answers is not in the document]
Haviles04: I pulled up devtools after I switched them and it looks like there is no store,state,game.id until you're authenticated so the reauthenticate dialog no longer works

src/views/GameView.vue
Haviles04: So, I was able to get the game to load properly, but upon removing this v-if I'm getting a typeerror 
` kingCount(player) {
      return player.faceCards.reduce((kingCount, card) => kingCount + (card.rank === 13 ? 1 : 0), 0);
    }, `  TypeError: Cannot read properties of null (reading 'faceCards')

src/store/modules/game.js
Haviles04, line 1: I added resetting the state on game end, the store wasn't resetting at the end of the game, so if you had just finished and then went to the url of any game number, the screen would display the last setup of the last game you played.

itsalaidbacklife: Good catch

tests/e2e/specs/in-game/reconnecting.spec.js
```diff
@@ -664,3 +662,34 @@ describe('Reconnecting to a game', () => {
...
+
+  it('Shows unavailable game dialog, then return home', () => {
+    cy.loadGameFixture(0, {
+      p0Hand: [Card.SEVEN_OF_CLUBS],
+      p0Points: [Card.SEVEN_OF_DIAMONDS, Card.SEVEN_OF_HEARTS],
+      p0FaceCards: [],
+      p1Hand: [],
+      p1Points: [],
+      p1FaceCards: [],
+    });
```
itsalaidbacklife, line 679: Since the test doesn't care which cards are where we can skip loading a flame fixture 
```suggestion
```

itsalaidbacklife (approved):
Nice work. One last nit in the test

src/views/GameView.vue
itsalaidbacklife: I'm comfortable leaving it for now. We can revisit this if it proves valuable to do so

itsalaidbacklife (approved):
👍

Issue #419, opened 2023-05-25
[Bug]: Playing a 5 when you are at the 8 card hand limit duplicates a card in the deck


itsalaidbacklife:
### Bug Summary

The logic for resolving 5's has a specific case that fails to correctly clean up. In particular, `game.secondCard` is a separate variable from `game.deck` and when we move a card from the deck to the secondCard, we don't remove that card from the deck itself.

### Detailed Description

See lines 161-187 of api/controllers/game/resolve.js. The `else` block where the player has 7 cards in hand doesn't correctly remove the newly selected `secondCard` from the `deck`.

### Steps to Reproduce

1. Start a game
2. Draw until you have 8 cards
3. Play a 5
4. Opponent resolves the 5
5. There will now be 53 total cards in the deck

Issue #404, opened 2023-05-08
[DevEx]: Update PR template to expect an issue link


itsalaidbacklife:
### Improvement Summary

We should update the `.github/PULL_REQUEST_TEMPLATE.md` with an optional field for the Issue # that adds the issue number to the issue description

### Detailed Description

In order for github to automatically close issues when a PR is merged, the issue needs to be linked in the PR's description. Often new contributors don't know this or don't know how to format the issue number and either don't include it, or put it in the title rather than the description, etc. 

We should update the template so that the first line asks contributors to add the relevant issue number like:

## Issue number
Relevant issue number (e.g. #404):

This way contributors are encouraged to add the issue number to the PR description, which will hopefully help with automatically tracking and closing issues as they are completed.

PR #420, opened 2023-05-25
Fix: Prevent duplicating card when resolving 5 with 7 cards in hand

itsalaidbacklife:
## Issue number

Relevant issue number (e.g. #404): #419 

## Please check the following

- For New Features:


api/controllers/game/resolve.js
```diff
@@ -172,6 +172,7 @@ module.exports = function (req, res) {
                     newSecondCard = _.sample(game.deck).id;
                   }
                   gameUpdates.secondCard = newSecondCard;
+                  cardsToRemoveFromDeck.push(newSecondCard);
```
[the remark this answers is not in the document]
itsalaidbacklife: 1. You're definitely right about breaking this up. I think we likely want one file/function per card effect and for this one to just delegate to the appropriate helper.
2. Good catch on the while loop...this one is actually unnecessary! This was copied from the logic for drawing two cards, but it's simpler here because you only draw one. If you look on lines 137-150, you can get the idea of what this was emulating, but again it's unnecessary here because we're only drawing one card. I'll remove that loop altogether b/c it's garbage here.

But for the record the idea here was that when the 5 draws _two cards_ (again not the case for this flow), the topCard and secondCard are put into the player's hand. Then if there's at least one card left in the deck, we sample from the deck to pick a new topCard. Then we if there's still another card in the deck, we sample again to pick the new secondCard. For the secondCard we want to make sure that we don't accidentally pick the same card we chose for the newTopCard, so the while loop is intended to make sure the secondCard selection is a new card that hasn't been drawn.

That said the while loop is a dumb way of doing it. What would be better would be to sample by index rather than getting the object reference for the card in the deck _then remove the card from the deck in memory_ before sampling the second time, so that you can safely assume the second selection won't collide. Then there's no need for a while loop.

Think I should fix that while I'm in here or leave this update smaller? Imo it makes sense to rework this part of the logic for clarity's sake since it still won't be a huge delta

api/controllers/game/resolve.js
```diff
@@ -135,17 +135,15 @@ module.exports = function (req, res) {
 
                 //Replace top card, if there's a card in deck
                 if (game.deck.length >= 1) {
-                  const newTopCard = _.sample(game.deck).id;
+                  const newTopCardIndex = _.random(game.deck.length - 1);
+                  const newTopCard = game.deck.splice(newTopCardIndex, 1)[0].id;
```
itsalaidbacklife, line 139: Picking the card via a random index and then removing the selected card from `game.deck` in memory like this lets us select the second card with confidence it won't redundantly collide with the new topCard. This prevents the need for the `while` loop resampling

api/controllers/game/resolve.js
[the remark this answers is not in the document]
itsalaidbacklife: I refactored to remove the while loops from 5 resolution

tests/e2e/specs/in-game/oneOffs.spec.js
```diff
@@ -42,40 +42,93 @@ describe('Untargeted One-Offs', () => {
...
-      // Player is P0
-      p0Hand: [Card.ACE_OF_CLUBS, Card.FIVE_OF_HEARTS, Card.THREE_OF_CLUBS, Card.EIGHT_OF_HEARTS],
-      p0Points: [Card.TEN_OF_SPADES, Card.ACE_OF_SPADES],
-      p0FaceCards: [Card.KING_OF_SPADES],
-      // Opponent is P1
-      p1Hand: [Card.ACE_OF_HEARTS],
-      p1Points: [Card.TEN_OF_HEARTS, Card.ACE_OF_DIAMONDS],
-      p1FaceCards: [Card.KING_OF_HEARTS],
-      scrap: [Card.FIVE_OF_SPADES],
+    it('Plays a five to draw one card when at 8 cards in hand', () => {
```
itsalaidbacklife, line 81: We should probably also add test cases for:
1. Playing a 5 with <= 7 cards in hand when there's 2 cards in the deck
2. Playing a 5 with <= 7 cards in hand when there's 1 card in the deck
3. Playing a 5 with 8 cards in hand when there's 2 cards in the deck
4. Playing a 5 with 8 cards in hand when there's 1 card in the deck

api/controllers/game/resolve.js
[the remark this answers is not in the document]
seriouslysean: Awesome! Thanks for the explanation. I would fork the net new work to a new ticket though just to keep a separation of concerns.

All of that makes sense but I have a question about how you maintain randomness of the deck for a new game. Does sample work to pick a random card or are you actually shuffling the deck per game and tracking a net new dictionary of the card order in which case wouldn’t sample not be required?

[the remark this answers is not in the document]
itsalaidbacklife: So the deck is basically unordered. There are two cards topCard and secondCard that are explicitly specified and then the other cards are just "in the deck" until we need one to become the topCard or secondCard at which point we sample one randomly. 

I refactored to remove the while loops bc it's just a few lines and it's relevant to the current task even if it's separable. Does the new implementation make more sense?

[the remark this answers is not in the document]
itsalaidbacklife: I'll add that when I say the deck is unordered, that what I mean is that while the actual game.deck attribute in memory is an array, that the order of cards in game.deck is not guaranteed to be consistent across different times we query the database to construct the game. So it technically it has an order, it's just that we can't depend on the order for logic because it's not persistent

api/controllers/game/load-fixture.js
```diff
@@ -112,8 +112,16 @@ module.exports = function (req, res) {
 
       return Promise.all([game, ...updatePromises]);
     })
-    .then(function populateGame(values) {
+    .then(async function removeCardsFromDeck(values) {
       const [game] = values;
+      // If deck was specified, delete all other cards from the deck
+      const {deck} = req.body;
+      if (deck) {
+        await Game.replaceCollection(game.id, 'deck').members(deck);
```
itsalaidbacklife, line 120: So it kinda stinks to make this `async` in the middle of a `.then()` promise chain. I think ultimately the whole endpoint should be rewritten to use `async/await`, but I wanted to try to limit the scope of this pr.

The main reason that I wanted async here is that it works nicely for a conditional update like we have below. I only want to overwrite the `deck` if the fixture explicitly specifies it. imo that's a nice use case for `await`, since including it otherwise would require a `Promise.all()` and conditionally pushing an extra update promise to the returned result. I think this is better for now, though I don't love it

api/controllers/game/resolve.js
```diff
@@ -166,23 +164,22 @@ module.exports = function (req, res) {
...
-                    `The ${game.oneOff.name} one-off resolves; ${player.username} draws one card (last in deck) to reach the hand limit (8).`,
-                  ];
+
+                // Player draws last card in deck, to reach hand limit (only draws 1)
                 }
+              } else {
+                gameUpdates.log = [
+                  ...game.log,
+                  `The ${game.oneOff.name} one-off resolves; ${player.username} draws one card (last in deck) to reach the hand limit (8).`,
+                ];
```
itsalaidbacklife, line 182: This was previously in the wrong place and would never fire; a new test case made that clear.

tests/e2e/specs/in-game/oneOffs.spec.js
```diff
@@ -42,40 +42,268 @@ describe('Untargeted One-Offs', () => {
...
+      cy.loadGameFixture(0, {
+        p0Hand: [Card.FIVE_OF_CLUBS],
+        p0Points: [],
+        p0FaceCards: [],
+        p1Hand: [Card.TWO_OF_CLUBS],
+        p1Points: [],
+        p1FaceCards: [],
+        topCard: Card.THREE_OF_CLUBS,
+        secondCard: Card.EIGHT_OF_HEARTS,
+        deck: [Card.ACE_OF_DIAMONDS]
```
itsalaidbacklife, line 127: My initial implementation of 'delete the deck but leave some cards in it' was a revision to `cy.deleteDeck()` so that you'd first load the game fixture without this line then call
```
cy.deleteDeck([Card.ACE_OF_DIAMONDS]);
```

I think it's much more legible folding this into the `cy.loadGameFixture()` call like this instead, in part because it feels weird to have the `cy.deleteDeck()` call take an argument of things to _not_ delete, and in part because ultimately deleting the cards in the deck is still just setup for the test so it makes sense as test fixture.

Ultimately, this actually deprecates the entire need for the `cy.deleteDeck()` command and corresponding endpoint, as any tests that call it can instead just update their `loadFixture()` call to set the `deck: []`. To mitigate scope creep, I'll refrain from updating all those other tests in this update and write up an issue to remove the deleteDeck command + endpoint and replace its use with revised `loadFixture()` calls

itsalaidbacklife: #421

tests/e2e/support/commands.js
```diff
@@ -1242,13 +1242,18 @@ Cypress.Commands.add('playOneOffAndResolveAsPlayer', (card) => {
     });
 });
 
+/**
+ * Deletes all the cards in the deck EXCEPT topCard and secondCard
+ * @param cardsToLeaveInDeck: {suit: number, rank: number}[] list of cards to leave in the deck
+ */
```
itsalaidbacklife, line 1248: Todo: remove this comment and roll changes back to HEAD

tests/e2e/support/commands.js
```diff
@@ -1310,6 +1314,11 @@ Cypress.Commands.add('loadGameFixture', (pNum, fixture) => {
         reqBody.scrapCardIds = scrapCardIds;
       }
 
+      if (fixture.deck) {
```
itsalaidbacklife, line 1317: Todo: update comment above with deck attribute in fixture arg

itsalaidbacklife (commented):
Adding todos re: comment cleanup

api/controllers/game/load-fixture.js
seriouslysean: I don't see anything wrong with it, refactor the rest later.

seriouslysean (approved):

api/controllers/game/resolve.js
itsalaidbacklife: I considered adding an assertion about the log to the test, but skipped it. We're not testing the log in other places and it's somewhat iffy how important this distinction is here. If anything needs log testing it'd be the resolution of 3's and 4's where the log is the only way to see which cards were selected

PR #426, opened 2023-05-29
Dev ex#317 start dev command

### LINKED ISSUE #317: [DevEx]: Add a command to boot front & back end servers simultaneously
### Improvement Summary

Previously we used the dev dependency `concurrently` to define commands that ran multiple commands concurrently. In particular we had the `start:dev` command, which booted the front and back end servers simultaneously.

### Detailed Description

We should re-add the `start:dev` command so that this is possible again. This should simplify onboarding for new devs and generally streamline the development process. Here's the [npm page for concurrently](https://www.npmjs.com/package/concurrently).

We should also update the `docs/setup-development.md` docs when we talk about booting up to describe booting this way as the first way we describe, since it will be easier. This was the section of the docs we had before removing the command:

#### Start the App

You can simultaneously run the server on `localhost:1337` and the client on `localhost:8080` with one command:
```
npm run start:dev
```
This will also automatically open up [Vue Devtools]. Alternatively, you can run them independantly with separate commands.

TPBarker:
## Issue number

Relevant issue number: Fixes #317 

## Please check the following

- For New Features:

- All tests and linting tools passed locally.
- Initially I thought I observed slightly decreased performance while running `npm run e2e:server` and `e2e:client` but it seemed to shake out after the first few tests. Someone might wish to pull down the branch locally to confirm, if we think that might be a cause for concern.


itsalaidbacklife (approved):
👌

PR #434, opened 2023-06-06
feat: update CI flow to run tests in parallel

### LINKED ISSUE #429: [DevEx]: Parallelize Cypress tests in CI
### Improvement Summary

We should configure the pr-actions CI setup to run cypress tests in parallel containers

### Detailed Description

Github actions allows [unlimited minutes for public repos](https://github.com/orgs/community/discussions/26054), so it makes sense to optimize the amount of linear time that a CI run takes to run, rather than minimizing container-minute expenditure. We should configure Cypress to run tests in parallel in CI ([docs](https://docs.cypress.io/guides/continuous-integration/github-actions#Parallelization)) in `.github/workflows/pr-actions.yml`.

We should follow Cypresses recommendation regarding the separation of installation and testing setup per the docs. We may need to tinker with the number of containers to use, but let's try **10 containers to start** and see how quickly that runs.

itsalaidbacklife:
## Issue number

Relevant issue number (e.g. #404): 
Resolves #429 

This reconfigures our CI workflow to run the tests in parallel on separate VM's

## Please check the following

- For New Features:


.github/workflows/pr-actions.yml
```diff
@@ -48,37 +48,61 @@ jobs:
...
 
   # Run Cypress End to End Tests
-  test-e2e:
-    name: 'Test: End to End'
-    runs-on: ubuntu-latest
-    needs: test-unit
-    # stop the job if it runs over 20 minutes to prevent a hanging process from eating minutes
-    timeout-minutes: 45
+  install:
+    needs: draft
```
itsalaidbacklife, line 58: I reworked the sequencing so that the draft check runs first as the only prerequisite, and then lint, unit, and e2e tests run in parallel. 

Before the rationale was to cancel the whole run early when we fail on lint or unit tests, but now that the entire suite completes in less time, I think it's reasonable to start the e2e tests without waiting for linting and unit tests, just to cut down on the wait time

.github/workflows/pr-actions.yml
```diff
@@ -48,37 +48,61 @@ jobs:
...
+          # Disable running of tests within install job
+          runTests: false
+          build: npm run build
+
+      - name: Save build folder
+        uses: actions/upload-artifact@v3
+        with:
+          name: build
+          if-no-files-found: error
+          path: assets
```
itsalaidbacklife, line 76: This is the secret sauce that allows the parallel containers to share the single build of the frontend without repeating the build step

.github/workflows/pr-actions.yml
```diff
@@ -48,37 +48,61 @@ jobs:
...
+  cypress-run:
+    runs-on: ubuntu-22.04
+    needs: install
+    strategy:
+      # don't fail the entire matrix on failure
+      fail-fast: false
+      matrix:
+        # run copies of the current job in parallel
+        containers: [1, 2, 3, 4, 5]
```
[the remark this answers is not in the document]
itsalaidbacklife: 5 looks to be the right number with this setup; increasing to as many as 16 containers didn't significantly reduce runtime

.github/workflows/pr-actions.yml
```diff
@@ -48,37 +48,61 @@ jobs:
...
+        with:
+          name: build
+          if-no-files-found: error
+          path: assets
+  cypress-run:
+    runs-on: ubuntu-22.04
+    needs: install
+    strategy:
+      # don't fail the entire matrix on failure
+      fail-fast: false
```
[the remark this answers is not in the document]
itsalaidbacklife: With `fail-fast` set to false in the workflow file:

1. All containers ran their allotted test cases and total CI runtime was ~7:30 min. 
2. The container with the failing test used the cypress-fail-fast lib to vacuously quick-pass all the tests after the failing one so essentially only the one container ends early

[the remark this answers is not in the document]
itsalaidbacklife: Setting `fail-fast: true` did not seem to reduce the total runtime at all in case of a failure, so it probably makes sense to leave it as false and let the tests run their course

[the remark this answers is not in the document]
seriouslysean: Seems like it might be time to just remove it since the benefit won’t be as great now.

.github/workflows/pr-actions.yml
```diff
@@ -48,39 +48,15 @@ jobs:
...
-          record: true
-        env:
-          VITE_API_URL: http://localhost:1337
-          # pass GitHub token to allow accurately detecting a build vs a re-run build
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+    uses: ./.github/workflows/test-e2e.yml
+    needs: draft
+    secrets:
+      record-key: d1f04332-8bb9-4570-90e4-7c29b9510ee3
```
itsalaidbacklife, line 61: Forgive me @seriouslysean for I have sinned. While this is a general no-no, it will enable contributions from forks to reap the benefits of parallelization and of recorded test runs which I think will cut down on admin time and generally improve devX.

.github/workflows/test-e2e.yml
itsalaidbacklife, line 1: I pulled the test job into it's own workflow file so that the main pr-actions one could be leaner and just describe which jobs it calls rather than including their logical guts. I think this is easier to read

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "cuttle",
-  "version": "4.3.29",
+  "version": "4.3.30",
```
itsalaidbacklife, line 3:
```suggestion
  "version": "4.3.35",
```

package-lock.json
```diff
@@ -1,12 +1,12 @@
...
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "cuttle",
-      "version": "4.3.29",
+      "version": "4.3.30",
```
itsalaidbacklife, line 9:
```suggestion
      "version": "4.3.35",
```

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "4.3.33",
+  "version": "4.3.34",
```
itsalaidbacklife, line 4:
```suggestion
  "version": "4.3.35",
```

itsalaidbacklife (commented):
Bumping package versions to 4.3.35

PR #438, opened 2023-06-10
Bug #401 Allowing one-off on itself if seven and last card in deck

### LINKED ISSUE #401: [Bug]: Playing last 7's one off with no cards left in deck, allows second turn.
### Bug Summary

Playing a 7 with one card left in the deck, to reveal another 7, playing it’s one off with no cards now left in the deck, allows a second turn

### Detailed Description

If there is one 7 left in the deck, and you play a 7 that is already in your hand to reveal the last 7 in the deck and then play that 7's one off you will get a second turn. Currently the backend only checks that there is a card left in the deck and not if it's the card currently selected to be played.  A special condition test will also be needed.

`api/controllers/game/seven/untargeted-one-off.js` is responsible for handling this attempted move and is likely where the fix should be made.

`tests/e2e/specs/in-game/sevens.spec.js` is where the test should be added.

Similar special condition issue: 
https://github.com/cuttle-cards/cuttle/issues/121

Similar special condition test-case:
https://github.com/cuttle-cards/cuttle/pull/174

### Expected Behavior

The one-off should not be allowed, just as if you try to play a 7 when the deck is empty.

Haviles04:
## Issue number

Resolves #401 

## Please check the following

- For New Features:

- Fixed bug allowing one-off to be played on last card in deck
- Added logic to disable one-off option if deck is empty or if trying to play one off on last 7 in deck
- Added test cases for bug fix as well as disabling one off option
- Updated casing on error message of Fours to match casing on other errors

image

### DIFF
--- a/src/components/GameView/MoveChoiceOverlay.vue
+++ b/src/components/GameView/MoveChoiceOverlay.vue
@@ -80,7 +80,7 @@ export default {
       required: true,
     },
     cardSelectedFromDeck: {
-      type: Boolean,
+      type: Object,
       default: null,
     },
   },
--- a/src/components/GameView/MoveChoiceOverlay.vue
+++ b/src/components/GameView/MoveChoiceOverlay.vue
@@ -142,14 +142,14 @@ export default {
           oneOffDisabled = true;
           oneOffDisabledExplanation = "Can't be played with empty deck";
         }
-        return {
-          displayName: 'One-Off',
-          eventName: 'oneOff',
-          moveDescription: this.selectedCard.ruleText,
-          disabled: oneOffDisabled,
-          disabledExplanation: oneOffDisabledExplanation,
-        };
       }
+      return {
+        displayName: 'One-Off',
+        eventName: 'oneOff',
+        moveDescription: this.selectedCard.ruleText,
+        disabled: oneOffDisabled,
+        disabledExplanation: oneOffDisabledExplanation,
+      };
     },
 
     targetedOneOffMove() {
--- a/src/components/GameView/MoveChoiceOverlay.vue
+++ b/src/components/GameView/MoveChoiceOverlay.vue
@@ -53,7 +53,7 @@ export default {
     MoveChoiceCard,
     GameCard,
   },
-  emits:['points', 'faceCard', 'scuttle', 'jack', 'oneOff', 'targetedOneOff', 'cancel'],
+  emits: ['points', 'faceCard', 'scuttle', 'jack', 'oneOff', 'targetedOneOff', 'cancel'],
   props: {
     modelValue: {
       type: Boolean,
@@ -131,12 +131,27 @@ export default {
       };
     },
     oneOffMove() {
+      let oneOffDisabled = this.allMovesAreDisabled;
+      let oneOffDisabledExplanation = this.disabledText;
+      //Check deck while playing 7s
+      if (this.selectedCard.rank === 7) {
+        if (!this.$store.state.game.topCard) {
+          oneOffDisabled = true;
+          oneOffDisabledExplanation = 'Cannot play one-off with an empty deck';
+        } else if (
+          this.selectedCard.id === this.$store.state.game.topCard.id &&
+          !this.store.state.game.secondCard
+        ) {
+          oneOffDisabled = true;
+          oneOffDisabledExplanation = 'Cannot play one-off on last card in deck';
+        }
+      }
       return {
         displayName: 'One-Off',
         eventName: 'oneOff',
         moveDescription: this.selectedCard.ruleText,
-        disabled: this.allMovesAreDisabled,
-        disabledExplanation: this.disabledText,
+        disabled: oneOffDisabled,
+        disabledExplanation: oneOffDisabledExplanation,
       };
     },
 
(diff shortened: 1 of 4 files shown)


tests/e2e/specs/in-game/sevens.spec.js
```diff
@@ -1487,4 +1487,44 @@ describe('Playing sevens at the end of the deck', () => {
...
+      deck: [Card.FIVE_OF_DIAMONDS],
+    });
+    cy.drawCardOpponent();
+    cy.get('#deck').click();
+    cy.drawCardOpponent();
+    cy.get('[data-player-hand-card=7-0]').click();
+    cy.get('[data-move-choice=oneOff').should('have.class', 'v-card--disabled');
+  });
+
+  it('Cannot play one-off on last card (7) in deck after playing 7 one off - special case', () => {
```
itsalaidbacklife, line 1511:
```suggestion
  it('Cannot play last card of deck as 7 one-off when chaining 7's', () => {
```

tests/e2e/specs/in-game/sevens.spec.js
```diff
@@ -1487,4 +1487,44 @@ describe('Playing sevens at the end of the deck', () => {
...
+      p1FaceCards: [],
+      topCard: Card.SIX_OF_HEARTS,
+      secondCard: Card.SEVEN_OF_HEARTS,
+      deck: [],
+    });
+    cy.drawCardOpponent();
+    cy.get('#deck').click();
+    cy.passOpponent();
+    cy.get('[data-player-hand-card=7-0]').click();
+    cy.get('[data-move-choice=oneOff').should('have.class', 'v-card--disabled');
```
itsalaidbacklife, line 1528: This case seems really similar to the above one, where there are no cards in the deck and therefore we can't play a 7. Based on the logic you updated in the `MoveChoiceCard`, I expected this test case to be the case where the top card of the deck is a 7 and there's no second card, with a 7 in the player's hand to confirm that they can play the 7 in their hand, but that the 7 on the deck then can't be played as it's one-off (because there's no second card)

tests/e2e/specs/in-game/sevens.spec.js
itsalaidbacklife: I think this test case should be reworked to test the logic around chaining 7's

src/components/GameView/MoveChoiceOverlay.vue
```diff
@@ -131,12 +131,27 @@ export default {
...
+      let oneOffDisabled = this.allMovesAreDisabled;
+      let oneOffDisabledExplanation = this.disabledText;
+      //Check deck while playing 7s
+      if (this.selectedCard.rank === 7) {
+        if (!this.$store.state.game.topCard) {
+          oneOffDisabled = true;
+          oneOffDisabledExplanation = 'Cannot play one-off with an empty deck';
+        } else if (
+          this.selectedCard.id === this.$store.state.game.topCard.id &&
+          !this.store.state.game.secondCard
```
itsalaidbacklife, line 143: I don't think this is currently covered by your test cases. Instead the first chunk of logic around the missing topCard looks to be tested twice. Let's rework the 2nd test case to cover this situation

itsalaidbacklife (commented):
Looking good. I think the 2nd test case should be revised to test the situation where the last card in the deck is a 7 and the player tries to play it while resolving a 7 from their hand

tests/e2e/specs/in-game/sevens.spec.js
```diff
@@ -1487,4 +1487,45 @@ describe('Playing sevens at the end of the deck', () => {
...
+      p1FaceCards: [],
+      topCard: Card.SIX_OF_HEARTS,
+      secondCard: Card.SEVEN_OF_HEARTS,
+      deck: [],
+    });
+    cy.drawCardOpponent();
+    cy.get('[data-player-hand-card=7-0]').click();
+    cy.get('[data-move-choice=oneOff').should('not.have.class', 'v-card--disabled').click();
+    cy.resolveOpponent();
+    cy.get('#deck').click();
```
itsalaidbacklife, line 1528: Don't you need to specifically click the 7 of hearts in the deck, not just the deck itself?
```suggestion
    cy.get('[data-top-card=7-2]').click();
```

itsalaidbacklife: huh! This works I think because basically the top card is in the middle of the deck, but in general the data-top-card selector is better b/c it's more specific (and works when both the top and second card are revealed)

tests/e2e/specs/in-game/sevens.spec.js
```diff
@@ -1487,4 +1487,45 @@ describe('Playing sevens at the end of the deck', () => {
...
+      deck: [Card.FIVE_OF_DIAMONDS],
+    });
+    cy.drawCardOpponent();
+    cy.get('#deck').click();
+    cy.drawCardOpponent();
+    cy.get('[data-player-hand-card=7-0]').click();
+    cy.get('[data-move-choice=oneOff').should('have.class', 'v-card--disabled');
+  });
+
+  it("Cannot play last card of deck as 7 one-off when chaining 7's", () => {
```
itsalaidbacklife, line 1511: I know I suggested this name but it's nice to avoid the apostrophe so we can use single quotes
```suggestion
  it('Cannot play last card of deck as 7 one-off when chaining sevens', () => {
```

itsalaidbacklife (approved):
Nice work, good fix

Issue #413, opened 2023-05-23
[DevEx]: Error logging (rollbar) - Log error if the server rejects your move with "It's not your turn" if the client thought it _was_ your turn


itsalaidbacklife:
### Improvement Summary

If the client thinks it is the players' turn, and the player makes a move that's rejected by the server because it is *not* the player's turn, we should log this as an error

### Detailed Description

After the implementation of #368 , we will have an error logging service configured. From there, we can start logging known and unknown errors. This specific case is a situation where there's a discrepancy between the client's state and the server's state, in particular about who's turn it is.

Logging this specific issue with rollbar (the error logging service) will offer insight into how frequently it happens and perhaps why.

itsalaidbacklife:
Future errors to log: 

1. Race conditions in interdependent requests e.g. move-making, subscribe, ready
2. populateGame() errors (see [comment](https://github.com/cuttle-cards/cuttle/pull/445/files#r1241243261) from #445

Issue #188, opened 2022-10-28
[DevEx]: Replace backend hooks and services with consolidated helpers


itsalaidbacklife:
### Improvement Summary

Sailsjs supports three ways of defining re-usable utilities: hooks, services, and helpers. Hooks and services are no longer recommended; we should move them all over to helpers

### Detailed Description

See the [sails docs on helpers](https://sailsjs.com/documentation/concepts/helpers) for reference. Each one will include some metadata describing it in addition to the function that gets called.

Replacing existing utilities will involve both redefining them in the new helper file and also updating all consumers of them to call them as a helper. As a part of the update we should remove all imports and references to those existing hooks and services.

PR #445, opened 2023-06-17
Feature #111 refactor populate game

### LINKED ISSUE #111: [Feature]: Refactor `.populateGame()` method
### Feature Summary

The `.populateGame()` backend method should be cleaned up and refactored

### Detailed Description

`gameService.populateGame()` is the function that creates the data structure sent in the socket events and consumed by the client. It could be improved by

1. Replacing `var` with `const` and `let`
2. Allowing it to be called with any number of players in the game
3. Leveraging `async`/`await`
4. Updating the `tempGame` function to more programatically grab all the attributes off of the `game` it's passed like
```
Object.entries(game).forEach(([key, val]) => {
    if (key === 'players') {
        this[key] = [p0, p1];
        return;
    }
    this[key] = val;
});
```
(note that this can't be done for `tempPlayer` as the player object needs to have at least the `encryptedPassword` field omitted)

Haviles04:
## Issue number

Relevant issue number (e.g. #404): Resolves #111 

## Please check the following

- For New Features:

- Refactored .populateGame to use async await, clean up the way it read and fail fast on logic
- Refactored .tempGame to programmatically set key value pairs

### DIFF
--- a/api/services/gameService.js
+++ b/api/services/gameService.js
@@ -47,30 +47,18 @@ function tempUser(usr, points) {
   this.faceCards.sort(comapreByRankThenSuit);
 }
 
-function tempGame(game, spectators, p0, p1) {
-  this.id = game.id;
-  this.players = [p0, p1];
-  this.spectatingUsers = spectators;
-  this.deck = game.deck;
-  this.scrap = game.scrap;
-  this.topCard = game.topCard;
-  this.secondCard = game.secondCard;
-  this.oneOff = game.oneOff;
-  this.oneOffTarget = game.oneOffTarget;
-  this.twos = game.twos;
-  this.log = game.log;
-  this.chat = game.chat;
-  this.id = game.id;
-  this.turn = game.turn;
-  this.passes = game.passes;
-  this.resolving = game.resolving;
-  this.lastEvent = game.lastEvent;
-  this.result = game.result;
-  this.isRanked = game.isRanked;
-  this.p0 = game.p0;
-  this.p1 = game.p1;
-  this.p0Ready = game.p0Ready;
-  this.p1Ready = game.p1Ready;
+function tempGame(game, spectatingUsers, p0, p1) {
+  Object.entries(game).forEach(([key, val]) => {
+    if(key === 'players'){
+      this[key] = [p0, p1];
+      return;
+    }
+    if(key === 'spectatingUsers'){
+      this[key] = spectatingUsers;
+      return;
+    }
+    this[key] = val;
+  });
 }
 module.exports = {
   GameResult,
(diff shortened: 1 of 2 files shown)


api/services/gameService.js
```diff
@@ -106,62 +94,38 @@ module.exports = {
...
       }
-    });
-  }, //End populateGame()
+      //find users
+      const p0 = await userService.findUser({ userId: game.players[0].id });
+      const p1 = await userService.findUser({ userId: game.players[1].id });
+      const spectatingUsers = await fetchSpectatorUsernames(game.id);
+      //find points
+      const p0Points = await cardService.findPoints({ userId: p0.id });
+      const p1Points = await cardService.findPoints({ userId: p1.id });
```
itsalaidbacklife, line 120: These are all independent of each other so we can await them as a single Promise.all() and destructure the assignments here to get the same result query in in parallel instead of in series.

One slight tweak is that the cardService calls need adjusting to grab the user ids directly from the game like we do to get the users themselves (instead of off the user objects from the above queries)

itsalaidbacklife (commented):
Nice one. Let's use Promise.all() to parallelize the queried

api/services/gameService.js
```diff
@@ -34,43 +34,33 @@ async function fetchSpectatorUsernames(gameId) {
...
+    }
+    if(key === 'points'){
+      this[key] = points;
+      return;
+    }
+    this[key] = val;
+  });
+  this.hand.sort(compareByRankThenSuit);
+  this.points.sort(compareByRankThenSuit);
+  this.faceCards.sort(compareByRankThenSuit);
```
Haviles04, line 49: The issue states 'note that this can't be done for tempPlayer as the player object needs to have at least the encryptedPassword field omitted'. I believe by just returning on the encrypted password it will omit it while still programmatically grabbing the rest.

api/services/gameService.js
```diff
@@ -34,43 +34,33 @@ async function fetchSpectatorUsernames(gameId) {
...
+  Object.entries(game).forEach(([key, val]) => {
+    if(key === 'players'){
+      this[key] = [p0, p1];
+      return;
+    }
+    if(key === 'spectatingUsers'){
+      this[key] = spectatingUsers;
+      return;
+    }
+    this[key] = val;
```
itsalaidbacklife, line 62: I know this is the approach prescribed by the issue, but upon reflection I think it might actually be clearer to use destructuring and spread syntax to achieve omitting/manipulating certain specific properties. 

First to get on the same page about the purpose of this function i.e. why not just use the game object itself without this transform, it's to replace the `players` attribute because by default when sails queries for the game and populates it's players, the players themselves don't get populated with their hands, faceCards, etc. So this function combines the main `game` object with the deeply populated versions of the players (p1 and p0). With that in mind, I think the *only* change we need to make to the game is to reassign its players, so we can probably simplify this. As I write it, I can think of a couple approaches we can take, and I'm not actually sure what is best. Let's discuss them together. @seriouslysean penny for your thoughts as well.

itsalaidbacklife: The simplest approach is to mutate the `game` argument in-place and return it. 

```
function tempGame(game, spectatingUsers, p0, p1) {
  game.players = [p0, p1];
  game.spectatingUsers = spectatingUsers;
  return game;
}

```

This has the advantage of being dead-simple, and the disadvantage of mutating an argument, which in principle can have unintended side effects. Specifically it doesn't just return the desired data, it updates the variable `game` to match the same thing as well. This is probably fine for Cuttle because it's not necessarily bad that `game` become the fully populated game, but generally it's a no-no because you can easily get confused why variables are changing

itsalaidbacklife: What might be better would be to return a new object that copies the game properties with spread syntax and overrides the specific new ones that we want to change:

```
function tempGame(game, spectatingUsers, p0, p1) {
  return {
    ...game,
    players: [p0, p1],
    spectatingUsers,
  };
}
```

This is probably better because `game` remains unchanged and we just get a new object returned without cross-contamination. It's 2 more lines and requires understanding spread syntax + the fact that the properties specified on the bottom take precedence, but it's still succinctly clear

itsalaidbacklife: Alternatively we could deep copy the game and then mutate properties on the copy:

```

function tempGame(game, spectatingUsers, p0, p1) {
  const res = _.cloneDeep(game);
  res.players = [p0, p1];
  res.spectatingUsers = spectatingUsers;
  return game;
}

```
This fixes the argument mutation but I think the 2nd way is cleaner , as this requires a lodash method

api/services/gameService.js
```diff
@@ -34,43 +34,33 @@ async function fetchSpectatorUsernames(gameId) {
...
+  Object.entries(usr).forEach(([key,val]) => {
+    if(key === 'encryptedPassword'){
+      return;
+    }
+    if(key === 'points'){
+      this[key] = points;
+      return;
+    }
+    this[key] = val;
+  });
```
itsalaidbacklife, line 46: You're totally right that we can take the same strategy here as with tempGame. I'd like to explore an approach that returns a new object rather than one that uses `this` as I think it's more clear. Then there are open questions about how to strip the encryptedPassword. Here's one way:

```suggestion
const res = {
  ...usr,
  points,
};
delete res.encryptedPassword;

```
Then we'd do the sorting below on `res` and return `res` at the end

itsalaidbacklife: An alternative would be to use destructuring to pull the encryptedPassword off:

```
const {encryptedPassword, ...userDat} = usr;
const res = {
  ...userData,
  points
};

```

I think `delete` is probably clearer for being more explicit though

itsalaidbacklife: Also let's rename the argument from `usr` to `user`. I thought I was being clever skipping a letter but it's annoying when you don't know when and whether it's usr vs user so just using user everywhere would be clearer

api/services/gameService.js
```diff
@@ -34,43 +34,33 @@ async function fetchSpectatorUsernames(gameId) {
 
 // Used to create fully populated game
 function tempUser(usr, points) {
```
itsalaidbacklife, line 36: I don't like the names tempUser or tempGame, but I'm not sure what to call them. For one I think these methods will be cleaner if they return a new object rather than using `this` for the class-like assignment approach. With that in mind, it probably makes sense for the function names to be verbs describing the fact that they format data, rather than object-like names. How about `formatPlayerData()` and `formatGameData()`?

itsalaidbacklife: I'm open to other suggestions. I'd like to make this more clear.

itsalaidbacklife (commented):
This is a great implementation of the approach documented in the issue but I've got some thoughts on how we can do better. Details in the comments

api/services/gameService.js
Haviles04: I think the second or the third one are good, second is cleaner, third is slightly more obvious what is happening. On the third one I think we could use structuredClone() instead of cloneDeep this way you don't need lodash for it.

api/services/gameService.js
```diff
@@ -106,62 +96,39 @@ module.exports = {
...
+   populateGame: async function(options) {
+    try{
+      if (!options) {
+        if (!Object.hasOwnProperty.call(options, 'gameId') && !typeof options.gameId === 'number') {
+          throw new Error({message: 'gameId is required and must be a number'});
         }
-        reject({ message: 'gameId is required and must be a number' });
-      } else {
-        reject({ message: 'Cannot populate Game without GameId (options had no gameId)' });
+        throw new Error({ message: 'Cannot populate Game without GameId (options had no gameId)' });
```
itsalaidbacklife, line 105: So this sortof highlights the awkwardness of using an object param when the only data needed is a single integer. It would probably be better if this were just `gameId` instead of `options`. That would require updating every instance where we call `populateGame()` to just pass the gameId directly without wrapping it in an object, but then  the validation here would be a simpler single call to `Number.isIngeger(gameId)`

What do you think?

api/services/gameService.js
Haviles04: I think those work, the only others i can really think of might be addUserDetails and addGameDetails

api/services/gameService.js
itsalaidbacklife: Relatedly, as documented in #188, we want to completely remove all the services and hooks to replace them with helpers, since all three serve the same purpose and helpers is the newer approach. So if we *do* opt to update the param to be a single int instead of an object and therefore update every call to populateGame(), it might make sense to go a little further and rewrite this function as a helper. Granted that increases the scope of this PR, so it might make sense to push both of those changes into a separate issue. What do you think?

itsalaidbacklife (commented):
Another thought on the param signature and it's impact on input validation

api/services/gameService.js
itsalaidbacklife: Yes let's go with approach 2

api/services/gameService.js
Haviles04: I like the idea of removing the object param, I think it will close off one extra step that people may miss. I think we should handle this is #188 since we're going to have to go back to every .populateGame to do that issue anyways. I'm happy to pick up that issue too 👨‍💻

api/services/gameService.js
Haviles04: Honestly in this case, do we even need this function? I think we could just return it at the end of populateGame().

api/services/gameService.js
```diff
@@ -15,7 +15,7 @@ const GameResult = Object.freeze({
  * @param {rank: number, suit: number} card1
  * @param {rank: number, suit: number} card2
  */
-function comapreByRankThenSuit(card1, card2) {
+function compareByRankThenSuit(card1, card2) {
```
itsalaidbacklife, line 18: Heh. Good catch ty

api/services/gameService.js
```diff
@@ -106,62 +79,39 @@ module.exports = {
...
+        cardService.findPoints({ userId: game.players[0].id }),
+        cardService.findPoints({ userId: game.players[1].id })
+      ]);
+      // then format results
+      const populatedP0 = formatPlayerData(p0, p0Points);
+      const populatedP1 = formatPlayerData(p1, p1Points);
+      return {...game, players: [populatedP0, populatedP1], spectatingUsers};
+    } catch(err){ 
+      return err.message;
+     }
```
itsalaidbacklife, line 113: So the current implementation errors (with `reject()`) if there is an error and leaves it up to the caller to sort it out. We could keep that the same here by just omitting the try/catch block altogether. I think it's reasonable in this case to let the caller's manage the error handling. We can add it back with some proper error logging after completing #413

itsalaidbacklife (approved):

PR #448, opened 2023-06-28
Rematch (Resolves #93)

### LINKED ISSUE #93: [Feature]: Enable replaying the same opponent (Rematch)
### Feature Summary

GameOver dialogs should let you choose to replay your opponent 

### Detailed Description

It should provide the option to replay your opponent. It should display when/whether the opponent has said yes or no (if no, navigate home). If both players agree to a rematch, they should immediately be dropped into a new game with the same ranked setting, and switch who is p0 or p1.

lihaobhsfer:
Resolves #93

## Please check the following

- For New Features:

Some todo items
- [x] more test cases for "Leave game"
- [x] investigate window.location.reload() and see if there are ways to avoid a page refresh
- [x] update minor version
- [x] fix UI warnings
- [x] remove previous game after new one starts
- [x] test when opponent requests rematch first - needs another command to just do `joinRematch`
- [x] check if spectator mode still works
- [x] remove .only

### DIFF
--- a/api/controllers/game/join-rematch.js
+++ b/api/controllers/game/join-rematch.js
@@ -6,7 +6,7 @@ module.exports = async function (req, res) {
     const { usr: userId } = req.session;
     const { oldGameId } = req.body;
 
-    const [ user, game ] = await Promise.all([
+    const [ user, oldGame ] = await Promise.all([
       User.findOne({ id: userId }),
       Game.findOne({ id: oldGameId }),
     ]);


Haviles04 (commented):
I think it might be helpful if we stick to either play-again or rematch, we have rematchview, and the API file is play-again. We could also take out th  "wants to play again" and just do p0Rematch, p1Rematch

api/controllers/game/play-again.js
```diff
@@ -0,0 +1,128 @@
...
+
+          return Promise.all([game, p0, p1, ...updatePromises]);
+        })
+        .then(function getPopulatedGame(values) {
+          return gameService.populateGame({ gameId: values[0].id });
+        })
+        .then(function publish(fullGame) {
+          Game.publish([fullGame.id], {
+            verb: 'updated',
+            change: 'Initialize',
```
itsalaidbacklife, line 112: We've switched this to lowercase now
```suggestion
            change: 'initialize',
```

api/controllers/game/join-and-start-new-game.js
```diff
@@ -0,0 +1,18 @@
...
+module.exports = async function (req, res) {
+  try {
+    const { newGameId, pNum } = req.body;
+    const game = await Game.findOne({ id: newGameId }).populate('players');
+    Game.subscribe(req, [game.id]);
+
+    req.session.game = game.id;
+    req.session.pNum = pNum;
+
+    return res.ok();
```
itsalaidbacklife, line 13: I think it makes sense to emit the game's lastevent here so any client who missed it can get the last event an reconstruct the current game state

api/controllers/game/join-and-start-new-game.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Join the two players into the new game
+ */
+module.exports = async function (req, res) {
+  try {
+    const { newGameId, pNum } = req.body;
```
itsalaidbacklife, line 6: Instead of reading the game ID and pNum from the request body, I think it would make sense to query for the user by their player id (in their session) and read the game id and pNum them from that record in the database. That way we could use a request to this endpoint as a fallback in case you miss the socket event that has your game id and this one could give you what you need without clientside params for the data the client could be missing

api/controllers/game/play-again.js
```diff
@@ -0,0 +1,128 @@
...
+      (updatedGame.p1WantsToPlayAgain && gameUpdates.p0WantsToPlayAgain);
+
+    if (bothWantToPlayAgain) {
+      const shouldNewGameBeRanked = updatedGame.isRanked;
+      const newGame = await gameAPI.createGame(game.name, shouldNewGameBeRanked);
+
+      // Switch pNums
+
+      const newP0 = await User.findOne({ id: updatedGame.p1 });
+      const newP1 = await User.findOne({ id: updatedGame.p0 });
```
itsalaidbacklife, line 40: These can be queried parallel

api/controllers/game/join-and-start-new-game.js
itsalaidbacklife, line 1: How about calling this `joinRematch`

itsalaidbacklife (commented):
Great work! This is coming together very nicely

src/components/GameView/GameOverDialog.vue
```diff
@@ -40,6 +40,9 @@
             </div>
           </div>
         </div>
+        <p>
+          {{ opponentWantsToPlayAgain ? 'Opponent wants to play again' : '' }}
+        </p>
```
Haviles04, line 45:
```suggestion
        <p v-if="opponentWantsToPlayAgain">
         'Opponent wants to play again'
        </p>
```

api/controllers/game/play-again.js
```diff
@@ -0,0 +1,130 @@
...
+module.exports = async function (req, res) {
+  try {
+    const { game: gameId, pNum } = req.session;
+    const game = await Game.findOne({ id: gameId }).populate('players');
+    const gameUpdates = {};
+    if (pNum === 0 || pNum === '0') {
+      gameUpdates.p0Rematch = true;
+    } else {
+      gameUpdates.p1Rematch = true;
+    }
```
Haviles04, line 14:
```suggestion
    const gameUpdates = { [`p${pNum}Rematch`]:true };

```

I'm not sure if this is to complicated for our sake, @itsalaidbacklife  what do you think?

Haviles04 (commented):
This is awesome, can't wait to see it come to fruition!

api/controllers/game/ready.js
```diff
@@ -1,3 +1,5 @@
+const gameService = require('../../services/gameService');
+
```
itsalaidbacklife, line 2: Sails exposes the services as globals by default so we don't need an explicit import here. Granted in general I would prefer explicit imports, but keep it consistent with the rest of the app for now
```suggestion
```

lihaobhsfer: I'd actually prefer keeping the import since it allows me to click into the function to see its definition in VSCode. It comes in handy sometimes and provides better devx.

api/controllers/game/rematch.js
```diff
@@ -0,0 +1,64 @@
...
+    const { game: gameId, pNum } = req.session;
+    const game = await Game.findOne({ id: gameId }).populate('players');
+    const gameUpdates = {};
+    if (pNum === 0 || pNum === '0') {
+      gameUpdates.p0Rematch = true;
+    } else {
+      gameUpdates.p1Rematch = true;
+    }
+
+    await Game.updateOne({ id: game.id }).set(gameUpdates);
```
Haviles04, line 18:
```suggestion

    await Game.updateOne({ id: game.id }).set({`[p${pNum}Rematch`]:true});
```

Haviles04: Upon further thought I think we could break this down even further to a one-liner, unless there is a reason to explicitly show what's happening.

src/store/modules/game.js
```diff
@@ -211,6 +215,13 @@ export default {
         state.p1Ready = !state.p1Ready;
       }
     },
+    updateRematch(state, pNum) {
+      if (pNum === 0) {
+        state.p0Rematch = !state.p0Rematch;
+      } else {
+        state.p1Rematch = !state.p1Rematch;
+      }
```
Haviles04, line 223:
```suggestion
     state[`p${pNum}Rematch`] = !state[`p${pNum}Rematch`];
```

api/controllers/game/ready.js
itsalaidbacklife: I hear you and I think it's a pain  point worth discussing further. I don't particularly like the idea of it being imported explicitly some of the time and not others, though because an inconsistent style is harder to follow. So we could potentially roll it out as an app-wide change but I didn't want to increase the scope of this pr

lihaobhsfer: Agreed it's better to make it an app-wide change. I'll keep them for now for debugging purposes and remove them before release

itsalaidbacklife (commented):
Commenting to checkpoint the review

api/controllers/game/rematch.js
lihaobhsfer: `gameUpdates` is referenced later in this block so I'll declare it explicitly

src/components/GameView/GameOverDialog.vue
```diff
@@ -53,6 +56,14 @@
       >
         Go Home
       </v-btn>
+      <v-btn
```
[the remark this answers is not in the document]
lihaobhsfer: makes sense. We don't track decline for now. This probably means another set up events/endpoints when either player hits "go home"?

src/components/GameView/GameOverDialog.vue
```diff
@@ -40,6 +40,9 @@
             </div>
           </div>
         </div>
+        <p>
+          {{ opponentWantsToPlayAgain ? "Opponent wants to play again" : ""}}
+        </p>
```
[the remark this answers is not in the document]
lihaobhsfer: Added on the UI. needs a way to test this. Right now due to how the `rematchOpponent` command works, the tests currently always let the player click rematch and then fire the cy command. I'll see if a second command that just does the `joinRematch` part would fix this

api/controllers/game/join-and-start-new-game.js
lihaobhsfer: Update the `join-rematch` API, not quite sure if it fully captures what you mean in the comment

api/controllers/game/join-rematch.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * Join the two players into the new game
+ */
+module.exports = async function (req, res) {
+  try {
+    const { usr: userId, game: gameId } = req.session;
+    const user = await User.findOne({ id: userId });
+    const oldGame = await Game.findOne({ id: gameId });
```
itsalaidbacklife, line 8: These two could be done in parallel
```suggestion
    const [ user, oldGame ] = await Promise.all([
      User.findOne({ id: userId }),
      Game.findOne({ id: gameId })
    ]);
```

api/controllers/game/join-rematch.js
```diff
@@ -0,0 +1,29 @@
...
+    req.session.game = game.id;
+    req.session.pNum = user.pNum;
+
+    const gameUpdates = {
+      lastEvent: { change: 'joinRematch' },
+    };
+    await Game.updateOne({ id: game.id }).set(gameUpdates);
+    Game.publish([game.id], {
+      change: 'joinRematch',
+      game,
```
itsalaidbacklife, line 22: This endpoint is used after the cards are dealt right? I think we'll want to populate the game that's sent in the socket message so it has the full game data

itsalaidbacklife (commented):
Check pointing a few comments. This is looking good

src/components/GameView/GameOverDialog.vue
```diff
@@ -41,33 +41,51 @@
...
+          @click="goHome"
+        >
+          Go Home
+        </v-btn>
+        <v-btn
+          color="primary"
+          variant="flat"
+          data-cy="gameover-rematch"
+          @click="rematch"
+        > Rematch </v-btn>
```
itsalaidbacklife, line 64: Let's do the newlines like on the first button
```suggestion
          @click="rematch"
        >
          Rematch
        </v-btn>
```

src/views/RematchView.vue
```diff
@@ -0,0 +1,169 @@
...
+    GameUnavailableView,
+    UsernameToolTip,
+    BaseSnackbar,
+  },
+  data() {
+    return {
+      leavingGame: false,
+      showSnackbar: false,
+      snackBarMessage: '',
+      snackColor: 'error',
```
itsalaidbacklife, line 65: Since this never changes the snack color, it could hard code the `error` color into the template instead of using a reactive variable
```suggestion
```

src/views/RematchView.vue
```diff
@@ -0,0 +1,169 @@
...
+    waitingForGameToStart() {
+      return true;
+    },
+  },
+  async mounted() {
+    if (this.isSpectating && !this.$store.state.game.id) {
+      let { gameId } = this.$router.currentRoute.value.params;
+      gameId = Number(gameId);
+      if (!Number.isInteger(gameId)) {
+        await this.$store.dispatch('requestUnsubscribeFromGame');
```
itsalaidbacklife, line 82: If the gameId isn't an integer then I don't think we'll need to unsubscribe from it because it won't be a valid game that you could be subscribed to in the first place. I do think redirecting home as you're doing makes sense though
```suggestion
```

src/views/RematchView.vue
```diff
@@ -0,0 +1,169 @@
...
+        this.$router.push(ROUTE_NAME_HOME);
+        return;
+      }
+      this.$store.dispatch('requestSpectate', Number(gameId));
```
itsalaidbacklife, line 86: You've already casted this to number above
```suggestion
      this.$store.dispatch('requestSpectate', gameId);
```

src/views/RematchView.vue
```diff
@@ -0,0 +1,169 @@
...
+
+    if (!this.$store.state.auth.authenticated) {
+      this.$store.commit('setMustReauthenticate', true);
+    }
+    document.documentElement.style.setProperty('--browserHeight', `${window.innerHeight / 100}px`);
+    window.addEventListener('resize', () => {
+      // We execute the same script as before
+      let vh = window.innerHeight * 0.01;
+      document.documentElement.style.setProperty('--browserHeight', `${vh}px`);
+    });
```
itsalaidbacklife, line 97: Since this is all logic that's DOM specific I see why it belongs in `mounted()` but maybe we should separate this from the above data logic so that the earlier stuff is all in `created()` which will fire earlier and the DOM manipulation here is the only piece in `mounted()`

itsalaidbacklife (commented):
Checkpoint with another chunk of review

api/controllers/game/concede.js
```diff
@@ -1,6 +1,8 @@
 module.exports = async function (req, res) {
   try {
-    const winner = (req.session.pNum + 1) % 2;
+    // Read pNum from req body first, for testing purposes
+    const pNum = req.body.pNum ?? req.session.pNum;
```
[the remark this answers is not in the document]
lihaobhsfer: Currently the only place we provide a `pNum` from the request is from test cases where we need to set up the game as p0 or p1 or spectator. If we remove this tests won't work.

src/components/GameView/GameOverDialog.vue
lihaobhsfer: Linter doesn't allow me to put it in new lines

src/views/RematchView.vue
lihaobhsfer: I'll also see if we don't need this and the `mounted` method at all

api/controllers/game/concede.js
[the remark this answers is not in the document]
itsalaidbacklife: I think we can resolve this by updating the test to make the request from sessions that have the pnum we're conceding from. I'll take a look

src/routes/game/components/GameOverDialog.vue
```diff
@@ -52,35 +52,60 @@
...
+          color="surface-1"
+          variant="flat"
+          data-cy="gameover-go-home"
+          :loading="leavingGame"
+          @click="goHome"
+        >
+          Go Home
+        </v-btn>
+        <v-btn
+          color="primary"
```
itsalaidbacklife, line 76: I think we want to disable this button when the opponent has declined.
```suggestion
          :disabled="opponentDeclinedRematch("
          color="primary"
```

src/routes/rematch/RematchView.vue
```diff
@@ -0,0 +1,175 @@
...
+      <div id="game-menu-wrapper" class="d-flex flex-column flex-sm-row align-center">
+        <GameMenu :is-spectating="false" />
+      </div>
+
+      <BaseSnackbar
+        v-model="showSnackbar"
+        :message="snackBarMessage"
+        :color="snackColor"
+        data-cy="game-snackbar"
+        @clear="clearSnackBar"
```
itsalaidbacklife, line 19: Let's put the snackbar at the bottom of the page like we do elsewhere

api/controllers/game/play-again.js
```diff
@@ -0,0 +1,128 @@
...
+    if (pNum === 0 || pNum === '0') {
+      gameUpdates.p0WantsToPlayAgain = true;
+    } else {
+      gameUpdates.p1WantsToPlayAgain = true;
+    }
+
+    await Game.updateOne({ id: game.id }).set(gameUpdates);
+
+    // In case other player updated the game at the same time
+    const updatedGame = await Game.findOne({ id: game.id }).populate('players');
```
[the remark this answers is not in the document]
lihaobhsfer: Marking as resolved since we have a separate ongoing discussion on possible solutions.

src/stores/game.js
```diff
@@ -13,22 +13,21 @@ function queenCount(player) {
...
-  const sortP1 = (cards) => player.pNum === myPnum ? cards?.sort(compareByRankThenSuit) : cards;
+  const sortP1 = (cards) => (player.pNum === myPnum ? cards?.sort(compareByRankThenSuit) : cards);
   return {
     ...player,
-    hand: sortP1(player.hand)?.map((card) => createGameCard(card)),
-    points: sortP1(player.points)?.map((card) => createGameCard(card)),
-    faceCards: sortP1(player.faceCards)?.map((card) => createGameCard(card))
+    hand: sortP1(player.hand)?.map((card) => createGameCard(card)) ?? [],
+    points: sortP1(player.points)?.map((card) => createGameCard(card)) ?? [],
+    faceCards: sortP1(player.faceCards)?.map((card) => createGameCard(card)) ?? [],
```
lihaobhsfer, line 27: TODO: see if the fallbacks are necessary now that we are populating the full game after rematch

tests/e2e/support/commands.js
```diff
@@ -411,6 +411,24 @@ Cypress.Commands.add('playPointsSpectator', (card, pNum) => {
...
+});
+
+Cypress.Commands.add('rematchSpectator', (pNum) => {
+  io.socket.get('/game/rematch', { pNum }, function handleResponse(res, jwres) {
+    if (jwres.statusCode !== 200) {
+      throw new Error(jwres.body.message);
+    }
+    return jwres;
+  });
+});
```
itsalaidbacklife, line 430: These look at the same as `concedeOpponent()` and `rematchOpponent()`, no? Are they necessary?

src/routes/game/components/GameOverDialog.vue
```diff
@@ -52,35 +52,61 @@
           </div>
         </div>
       </template>
+      <p v-if="opponentWantsToRematch">
+        {{ gameStore.opponent.username }} wants to rematch
+      </p>
+      <p v-if="opponentDeclinedRematch">
+        {{ gameStore.opponent.username }} declined rematch and left the game
+      </p>
```
itsalaidbacklife, line 60: Let's add testing selectors that we can use to identify these during testing
```suggestion
      <p v-if="opponentWantsToRematch" data-cy="opponent-wants-rematch">
        {{ gameStore.opponent.username }} wants to rematch
      </p>
      <p v-if="opponentDeclinedRematch" data-cy="opponent-declined-rematch">
        {{ gameStore.opponent.username }} declined rematch and left the game
      </p>
```

tests/e2e/specs/in-game/spectating.spec.js
```diff
@@ -360,3 +366,77 @@ describe('Spectating Games', () => {
...
+  it('Spectate unranked games with rematch', function () {
+    // 1st game: Opponent concedes
+    cy.recoverSessionOpponent(playerTwo);
+    cy.concedeOpponent();
+    assertVictory();
+    cy.log('rematch player2');
+
+    cy.rematchOpponent({ rematch: true });
+
+    cy.wait(1000);
```
itsalaidbacklife, line 394: The purpose of this `cy.wait()` is to make sure that the `cy.rematchOpponent()` call is correctly processed before we move onto the `cy.window().its().then()` call starts making assertions on the DOM, basically to prevent the race condition of checking the store before the server has processed the opponent's rematch request. 


I think we can avoid the `cy.wait()` call by using an assertion on the DOM about what we expect to happen based on the above `cy.rematchOpponent()` call. That's considered the best practice because DOM assertions automatically retry until they succeed or timeout so that if it only takes say 300 milliseconds for the DOM assertion to pass we don't have to wait the full 1s before proceed.

In this case that should mean seeing the text that the opponent wants a rematch, right?
 so we could do this
```suggestion
    cy.get('[data-cy="opponent-wants-rematch"]').should('be.visible');
```

That way the assertion about the opponent wanting a rematch will retry until it passes (or times out after 4 seconds) so that the test can proceed earlier if the server takes less than 1s to respond. Check out the [cypress docs](https://docs.cypress.io/api/commands/wait#Time) on `cy.wait()` for more reading about the general idea

itsalaidbacklife: This should apply to every use of `cy.wait()` that's been added I think, though the specific DOM assertion will depend on the step we are waiting for

itsalaidbacklife (commented):
Awesome! This is so close. I have a suggestion for how we can clean up the tests and I'd like to explore if we can collapse the new database changes into a single column. Details are in the comments

api/controllers/game/rematch.js
```diff
@@ -0,0 +1,79 @@
...
+        game: updatedGame,
+        pNum,
+        gameId: newGame.id,
+        newGame: newFullGame,
+      });
+
+      return res.ok({ newGameId: newGame.id });
+    }
+
+    return res.ok();
```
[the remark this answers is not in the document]
lihaobhsfer: Makes sense to fail fast. I'll need to check the condition. It should be if "either one doesn't want to rematch" and we should make sure we only count `false` but not `null`s

api/models/Game.js
```diff
@@ -33,6 +33,20 @@ module.exports = {
...
+    p1Rematch: {
+      type: 'boolean',
+      allowNull: true,
+    },
+    rematchGame: {
+      model: 'game',
+    },
+    rematchOldGame: {
+      model: 'game',
+    },
```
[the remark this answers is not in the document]
lihaobhsfer: how do we know if `game` is a new or old game? 
Keeping two columns would make it easier to understand, following a general `prev` and `next` semantics of a sequence

api/models/User.js
```diff
@@ -28,6 +28,14 @@ module.exports = {
...
       isIn: [0, 1],
     },
+    rematchOldPNum: {
+      type: 'number',
+      allowNull: true,
+      isIn: [0, 1],
+    },
+    rematchOldGame: {
+      model: 'game',
+    },
```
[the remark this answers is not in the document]
lihaobhsfer: I'm not entirely sure whether `p0` and `p1` are persisted after gameover. I can double check

tests/e2e/support/commands.js
lihaobhsfer: Good catch these are not needed anymore

lihaobhsfer (commented):
replied to some comments inline

api/models/User.js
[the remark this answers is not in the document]
lihaobhsfer: removed `rematchOldPNum`

lihaobhsfer:
sql migrations
`Game`
```sql
-- Adding new columns
ALTER TABLE "Game"
ADD COLUMN "p0Rematch" BOOLEAN,
ADD COLUMN "p1Rematch" BOOLEAN,
ADD COLUMN "rematchGame" INTEGER;

-- Adding a foreign key constraint
ALTER TABLE "Game"
ADD CONSTRAINT "Game_rematchGame_fkey"
FOREIGN KEY ("rematchGame") REFERENCES "Game"("id");
```

`User`
```sql
-- Adding new column
ALTER TABLE "User"
ADD COLUMN "rematchOldGame" INTEGER;

-- Adding a foreign key constraint
ALTER TABLE "User"
ADD CONSTRAINT "User_rematchOldGame_fkey"
FOREIGN KEY ("rematchOldGame") REFERENCES "Game"("id");
```

api/controllers/game/rematch.js
```diff
@@ -0,0 +1,79 @@
...
+    const currentMatch = await Match.findOne({ id: updatedGame.match });
+    const shouldNewGameBeRanked = currentMatch?.winner ? false : updatedGame.isRanked;
+
+    const { p0: newP1Id, p1: newP0Id } = updatedGame;
+
+    Game.publish([game.id], {
+      change: 'rematch',
+      game: updatedGame,
+      pNum,
+    });
```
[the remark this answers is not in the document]
lihaobhsfer: We currently rely on the `rematch` event to trigger a redirect on the UI so this is necessary

api/controllers/game/join-rematch.js
```diff
@@ -0,0 +1,33 @@
...
+
+    const user = await User.findOne({ id: userId });
+    const oldGame = await Game.findOne({ id: user.rematchOldGame });
+
+    const newGameId = oldGame.rematchGame;
+    const game = await gameService.populateGame({ gameId: newGameId });
+    Game.subscribe(req, [game.id]);
+
+    req.session.game = game.id;
+    req.session.pNum = user.pNum;
```
itsalaidbacklife, line 16: I _think_ we could get rid of the `User.rematchOldGame` column by queries for the old game either based on your existing session depending on when that's set, or using a req param. Basically as long as we have the id of one game or the other, (old or new) and as long as we know whether it's the old game or the new game, we could find the other one without data on the user.

If we have the new game id (i.e. the rematch id) and want the old game, we could:
```
const oldGame = await Game.findOne({ rematchOldGame: rematchGameId });
```

itsalaidbacklife (commented):
Really nice. I have a few issues I'd like to write up for future revisions, and only one substantive change to request to see if we can skip the column on the user

tests/e2e/specs/in-game/spectating.spec.js
```diff
@@ -360,3 +366,75 @@ describe('Spectating Games', () => {
...
+        cy.expect(game.p0Rematch).to.be.null;
+        cy.expect(game.p1Rematch).to.be.true;
+      });
+
+    cy.recoverSessionOpponent(playerOne);
+
+    cy.wait(1000);
+
+    cy.rematchOpponent({ rematch: true });
+    cy.wait(1000);
```
itsalaidbacklife, line 408: Todo: see if we can avoid these by improving the promise management of `cy.recoverSessionOpponent()`

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -706,3 +706,391 @@ describe('Creating And Updating Ranked Matches', () => {
...
+    // 2nd game: Player concedes
+    cy.get('#game-menu-activator').click();
+    cy.get('#game-menu').should('be.visible').get('[data-cy=concede-initiate]').click();
+    cy.get('#request-gameover-dialog').should('be.visible').get('[data-cy=request-gameover-confirm]').click();
+
+    assertLoss();
+
+    cy.log('Opponent requests rematch first');
+    cy.rematchOpponent({ rematch: true });
+    cy.wait(1000);
```
itsalaidbacklife, line 1058: let's replace these `cy.wait()`s with DOM assertions

src/routes/rematch/RematchView.vue
itsalaidbacklife, line 1: Let's write up an issue to create a dialog that appears when your opponent declines a rematch while you're here so that you're informed and can leave

lihaobhsfer: Created here https://github.com/cuttle-cards/cuttle/issues/784

api/controllers/game/join-rematch.js
```diff
@@ -0,0 +1,36 @@
...
+ * Join the two players into the new game
+ */
+module.exports = async function (req, res) {
+  try {
+    const { usr: userId } = req.session;
+    const { oldGameId } = req.body;
+
+    const user = await User.findOne({ id: userId });
+
+    const oldGame = await Game.findOne({ id: oldGameId });
```
itsalaidbacklife, line 11: Now that these are independent, we can request them in parallel
```suggestion
    const [ user, game ] = await Promise.all([
      User.findOne({ id: userId }),
      Game.findOne({ id: oldGameId }),
    ]);
```

api/controllers/game/rematch.js
```diff
@@ -0,0 +1,74 @@
...
+    const { usr: userId, pNum } = req.session;
+    const { gameId, rematch } = req.body;
+
+    try {
+      await Game.findOne({ id: gameId }).populate('players');
+    } catch (err) {
+      if (rematch === false) {
+        return res.ok();
+      }
+    }
```
itsalaidbacklife, line 16: Is this stale? It looks like the query is not used and then we immediately requery for the game anyway
```suggestion
```

src/routes/game/components/GameOverDialog.vue
```diff
@@ -55,20 +55,45 @@
...
+      </div>
     </template>
   </BaseDialog>
+  <BaseSnackbar
+    v-model="showSnackbar"
+    :message="snackBarMessage"
+    :color="snackColor"
+    data-cy="game-over-snackbar"
+    @clear="clearSnackBar"
+  />
```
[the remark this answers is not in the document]
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/790

src/stores/game.js
```diff
@@ -806,5 +826,19 @@ export const useGameStore = defineStore('game', {
...
     },
+    async requestRematch({ gameId, rematch = true }) {
+      return new Promise((resolve, reject) => {
+        io.socket.get('/game/rematch', { gameId, rematch }, (res, jwres) => {
+          return this.handleGameResponse(jwres, resolve, reject);
+        });
+      });
+    },
+    async requestJoinRematch({ oldGameId }) {
+      return await new Promise((resolve, reject) => {
```
itsalaidbacklife, line 837: We generally don't need to await the returned promise
```suggestion
      return new Promise((resolve, reject) => {
```

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -706,3 +706,418 @@ describe('Creating And Updating Ranked Matches', () => {
...
+          startTime: dayjs().subtract(1, 'hour').valueOf(),
+          endTime: dayjs().subtract(1, 'hour').valueOf(),
+        };
+
+        cy.loadMatchFixtures([oldMatchBetweenPlayers, currentMatchWithDifferentOpponent]);
+      });
+    // Log in as playerOne
+    cy.loginPlayer(playerOne);
+    cy.setupGameAsP0(true, true);
+  });
```
itsalaidbacklife, line 749:
```suggestion
  });
  
```

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -706,3 +706,418 @@ describe('Creating And Updating Ranked Matches', () => {
...
+        cy.expect(p0Rematch).to.eq(null);
+        cy.expect(p1Rematch).to.eq(null);
+      });
+
+    // 2nd game: Player concedes
+    cy.get('#game-menu-activator').click();
+    cy.get('#game-menu').should('be.visible').get('[data-cy=concede-initiate]').click();
+    cy.get('#request-gameover-dialog').should('be.visible').get('[data-cy=request-gameover-confirm]').click();
+
+    assertLoss();
```
itsalaidbacklife, line 819:
```suggestion
    assertLoss();

```

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -706,3 +706,418 @@ describe('Creating And Updating Ranked Matches', () => {
...
+
+    // 3rd game: Ends via requested stalemate
+    // Request stalemate
+    cy.get('#game-menu-activator').click();
+    cy.get('#game-menu').should('be.visible').get('[data-cy=stalemate-initiate]').click();
+    cy.get('#request-gameover-dialog').should('be.visible').get('[data-cy=request-gameover-confirm]').click();
+    cy.get('#waiting-for-opponent-stalemate-scrim').should('be.visible');
+    // Opponent confirms
+    cy.stalemateOpponent();
+    assertStalemate();
```
itsalaidbacklife, line 863:
```suggestion
    assertStalemate();
    
```

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -706,3 +706,418 @@ describe('Creating And Updating Ranked Matches', () => {
...
+
+      // Confirm game was set to unranked
+      cy.request('http://localhost:1337/game').then((res) => {
+        // Sort games by updatedAt asc
+        const games = res.body.sort((game1, game2) => game1.updatedAt - game2.updatedAt);
+        expect(games.length).to.eq(5, 'Expected 5 games');
+        expect(games[4].isRanked).to.eq(false, 'Expected last game to be set to unranked after completion');
+      });
+    });
+  });
```
itsalaidbacklife, line 940:
```suggestion
  });
  
```

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -706,3 +706,418 @@ describe('Creating And Updating Ranked Matches', () => {
...
+        const { p1Rematch } = game;
+
+        cy.expect(p0Rematch).to.eq(true);
+        cy.expect(p1Rematch).to.eq(null);
+      });
+    cy.get('#waiting-for-game-to-start-scrim').should('be.visible');
+    cy.get('[data-cy=leave-unstarted-game-button]').should('be.visible').click();
+
+    cy.url().should('not.include', '/game');
+  });
```
itsalaidbacklife, line 970:
```suggestion
  });
  
```

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -706,3 +706,418 @@ describe('Creating And Updating Ranked Matches', () => {
...
+  it('Creates a match when two players play a ranked game for the first time this week, leave game during rematch, button should be disabled', function () {
```
itsalaidbacklife, line 971:
```suggestion
  it('Disables the rematch button in the GameOverDialog when opponent declines a rematch before the player makes a choice', function () {
```

tests/e2e/specs/in-game/spectating.spec.js
```diff
@@ -360,3 +366,87 @@ describe('Spectating Games', () => {
...
+    // Sign up players
+    cy.signupOpponent(playerOne).as('playerOneId');
+    cy.signupOpponent(playerThree).as('playerThreeId'); // spectator
+    // Opponent will be player 2 (the last one we log in as)
+    cy.signupOpponent(playerTwo).as('playerTwoId');
+
+    // Log in as playerOne
+    cy.loginPlayer(playerOne);
+    cy.setupGameAsSpectator();
+  });
```
itsalaidbacklife, line 384:
```suggestion
  });

```

itsalaidbacklife (approved):
Really nice work. This is such a cool feature. I'm going to commit some minor cleanup from my suggestions directly and then merge!

api/controllers/game/join-rematch.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * Join the two players into the new game
+ */
+module.exports = async function (req, res) {
+  try {
+    const { usr: userId } = req.session;
+    const { oldGameId } = req.body;
+
+    const [ user, game ] = await Promise.all([
```
itsalaidbacklife, line 9:
```suggestion
    const [ user, oldGame ] = await Promise.all([
```

tests/e2e/specs/in-game/spectating.spec.js
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/795

itsalaidbacklife:
Final migration run for merge
```
-- Adding new columns
ALTER TABLE "game"
ADD COLUMN IF NOT EXISTS "p0Rematch" BOOLEAN,
ADD COLUMN IF NOT EXISTS "p1Rematch" BOOLEAN,
ADD COLUMN IF NOT EXISTS "rematchGame" INTEGER;

-- Adding a foreign key constraint
ALTER TABLE "game"
ADD CONSTRAINT "game_rematchGame_fkey"
FOREIGN KEY ("rematchGame") REFERENCES "game"("id") on delete set null;
```

seriouslysean:
@itsalaidbacklife so we ever save these migrations in a migrations folder or something with the PR I’d? Something like migrations/720.sql or something would be helpful.

PR #449, opened 2023-07-03
Feature #447 directing users to login / signup

### LINKED ISSUE #447: [Feature]: Direct new users to Signup and returning users to Login pages on initial navigation
### Feature Summary

New users need to use the signup page and returning users need the login page. We should use a form of persistent key/value storage to allow the application to identify whether the unauthenticated user has previously authenticated and direct them to the appropriate page accordingly.

### Detailed Description

The persistent storage variable (which could be called something like) `isReturningUser` should be consumed during the `mustBeAuthenticated()` navigation guard in `src/router.js` so that if a user is not authenticated when navigating to a protected page e.g. `HomeView`, they direct to `LoginView` if `isReturningUser` is `true`, and to `SignupView` if it's false.

One question is whether the persistent variable `isReturningUser` should be stored in the `cookie` or in `window.localStorage`. Discussion is welcome on the approach

Haviles04:
## Issue number

Relevant issue number (e.g. #404): Resolves #447 

## Please check the following

- For New Features:

Added localStorage item of 'returning Users' that is consumed by the router to send the user to either /login or /signup based on if they have visited the site before


src/views/HomeView.vue
```diff
@@ -166,7 +166,8 @@ export default {
       return this.$vuetify.display.mdAndDown ? 'small' : 'medium';
     },
   },
-  async mounted() {
+  async created() {
+    localStorage.setItem('returningUser', true);
```
seriouslysean, line 170: Do you need to set this every time, or should we only be setting it if it doesn't exist? I don't mind either way, this is less complicated and it likely doesn't matter. What this does have, though, is a bug for when localStorage doesn't exist or errors out.  Check out the implementation in `src/components/CreateGameDialog.vue` via `setRankedPreference`. It uses try/catch to avoid any errors.

What I recommend is creating a utility to set/get a `localStorage` value and use that as a cuttle implementation of LS with graceful fallback in the event of an error. It should be used both when you check the value, and when you set the value.

seriouslysean (dismissed):
I have something like this in mind:

LS getter
```
function getLSValue(key) {
    try {
        return localStorage.getItem(key);
    } catch(e) {
        return null;
    }
}
```

LS setter
```
function setLSValue(key, value) {
    try {
        localStorage.setItem(key, value)
    } catch {
        // Error handling
    }
}
```

Bonus, you could use Jest to verify they work 😎

src/views/HomeView.vue
Haviles04: Where should these functions live?

tests/e2e/specs/out-of-game/authentication.spec.js
itsalaidbacklife, line 1: I think the other tests are failing because the assertions are out of date with the new logic (i.e. the app used to go to #/login when you were unauthenticated but now sometimes it's #/signup)

itsalaidbacklife (commented):
This is coming along nicely. Going to be a nice QoL improvement and boost for the new user experience in particular

src/views/HomeView.vue
itsalaidbacklife: Then the created hook here would dispatch the auth store module's `setIsReturningUser()` action here, and the navigation guard in `src/router.js` would get it's value using the `getIsReturningUser()` store value

src/store/modules/auth.js
```diff
@@ -27,6 +28,7 @@ async function handleLogin(context, username, password, signup = false) {
     }
     await reconnectSockets();
     // If the response was successful, the user is logged in
+    context.dispatch('setReturningUser');
```
Haviles04, line 31: How do we feel about calling this action right here in the handleLogin function instead of the created() on the home view?

src/store/modules/auth.js
```diff
@@ -158,5 +165,20 @@ export default {
     reconnectSocket() {
       io.socket.reconnect();
     },
+    getReturningUser(context) {
+      const { isReturningUser } = context.state;
+      if (isReturningUser === null) {
+        const val = getLocalStorage('returningUser');
```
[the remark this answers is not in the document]
itsalaidbacklife: At that point let's pull the LS name variables into the utility file and import them here and in the CreateGameDialog

itsalaidbacklife (commented):
Couple more comments

src/router.js
```diff
@@ -19,11 +20,15 @@ export const ROUTE_NAME_SIGNUP = 'Signup';
 export const ROUTE_NAME_STATS = 'Stats';
 export const ROUTE_NAME_STATS_SEASON = 'StatsBySeason';
 
-const mustBeAuthenticated = (to, from, next) => {
+const mustBeAuthenticated = async (to, from, next) => {
```
itsalaidbacklife, line 23: I don't think this needs to be async as the action itself is only making local storage calls which is synchronous

```suggestion
const mustBeAuthenticated = (to, from, next) => {
```

src/router.js
```diff
@@ -19,11 +20,15 @@ export const ROUTE_NAME_SIGNUP = 'Signup';
...
   if (store.state.auth.authenticated) {
     return next();
   }
-  return next('/login');
+  await store.dispatch('getReturningUser');
```
itsalaidbacklife, line 27: This action is synchronous
```suggestion
store.dispatch('getReturningUser');
```

src/router.js
```diff
@@ -23,7 +24,11 @@ const mustBeAuthenticated = (to, from, next) => {
   if (store.state.auth.authenticated) {
     return next();
   }
-  return next('/login');
+  store.dispatch('getReturningUser');
+  if (store.state.auth.isReturningUser) {
```
itsalaidbacklife, line 28: localstorage can only store strings so I'd pref that this be a more explicit check for 'true'. Also since the method returns the value I think it may be nicer to capture it in a variable here
```suggestion
  const isReturningUser = store.dispatch('getReturningUser');
  if (isReturningUser === 'true') {
```

src/store/modules/auth.js
```diff
@@ -158,5 +169,20 @@ export default {
...
     },
+    getReturningUser(context) {
+      const { isReturningUser } = context.state;
+      if (isReturningUser === null) {
+        const val = getLocalStorage(LS_IS_RETURNING_USER_NAME);
+        context.commit('setIsReturningUser', val);
+        return val;
+      }
+    },
+    setReturningUser(context) {
```
itsalaidbacklife, line 180: Can we rename these to `getIsReturningUser` and `setIsReturningUser`?

src/router.js
```diff
@@ -23,7 +24,11 @@ const mustBeAuthenticated = (to, from, next) => {
   if (store.state.auth.authenticated) {
     return next();
   }
-  return next('/login');
+  const isReturningUser = store.dispatch('getReturningUser');
```
itsalaidbacklife, line 27: The dispatch calls need to be updated to match the action rename
```suggestion
  const isReturningUser = store.dispatch('getIsReturningUser');
```

src/store/modules/auth.js
```diff
@@ -27,6 +32,7 @@ async function handleLogin(context, username, password, signup = false) {
     }
     await reconnectSockets();
     // If the response was successful, the user is logged in
+    context.dispatch('setReturningUser');
```
itsalaidbacklife, line 35:
```suggestion
    context.dispatch('setIsReturningUser');
```

itsalaidbacklife (approved):
Really nice work. The end result should be a more seamless onboarding experience -- and we've got a more generalized localstorage setup now for going forward

PR #454, opened 2023-07-18
Fix: Resolve 'Error Subscribing' flaky failures in CI tests

### LINKED ISSUE #453: [Bug]: Test flake -- Error subscribing opponent to game occasionally happens during `cy.setupGameAsP0()` in CI
### Bug Summary

Tests, especially one-off tests, fail somewhat randomly in the `beforeEach()` hook in CI due to an error when subscribing an opponent to the game.

### Detailed Description

The issue looks to be happening in the `cy.subscribeOpponent()` command in `test/e2e/support/commands.js`, which is executed during the `beforeEach()` hook inside of the higher-level command `cy.setupGameAsP0()`. 

There may be an issue where the `io` instance inside the commands file is not properly authenticating

### Current Behavior

Tests sporadically fail in CI during the `beforeEach()` hook.

### Expected Behavior

Tests pass consistently.

### Steps to Reproduce

The problem is inconsistent so we don't have consistent reproduction steps, however it looks to be the case that you can replicate locally as follows:
1. Boot the server and client and open cypress
2. Run a test
3. Restart the server
4. Run a test that calls `setupGameAsP0()` e.g. any of the tests in `tests/e2e/specs/in-game/basicMoves.spec.js` and it will fail when the opponent is supposed to join the game with `Error Subscribing`

It is possible that the issue is different in CI, however, as there is no indication that the server is restarting during these test runs and some pass prior to the random failure.

See these prior CI run logs for examples of the failure in CI:
https://github.com/cuttle-cards/cuttle/actions/runs/5534156665/jobs/10098765469

https://github.com/cuttle-cards/cuttle/actions/runs/5581962323/jobs/10200730269

itsalaidbacklife:
**WIP: DO NOT MERGE** (opening PR to execute CI runs to test the bug)

## Issue number

Relevant issue number (e.g. #404): Resolves #453 

## Please check the following

- For New Features:

### DIFF
--- a/tests/e2e/support/commands.js
+++ b/tests/e2e/support/commands.js
@@ -97,7 +97,12 @@ Cypress.Commands.add('setupGameAsP1', (alreadyAuthenticated = false, isRanked = 
     if (!alreadyAuthenticated) {
       cy.signupOpponent(opponentOne);
     }
-    cy.subscribeOpponent(gameSummary.gameId);
+    try {
+      cy.subscribeOpponent(gameSummary.gameId);
+    } catch {
+      cy.recoverSessionOpponent(opponentOne);
+      cy.subscribeOpponent(gameSummary.gameId);
+    }
     cy.readyOpponent();
     cy.window()
       .its('cuttle.app.config.globalProperties.$store')
--- a/tests/e2e/support/commands.js
+++ b/tests/e2e/support/commands.js
@@ -204,7 +204,7 @@ Cypress.Commands.add('subscribeOpponent', (id) => {
         if (jwres.statusCode === 200) {
           return resolve();
         }
-        console.log(`\n\nError subscribing opponent to game ${id}`, jwres);
+        cy.log(jwres);
         return reject(new Error(`error subscribing opponent to game ${id}`, {cause: jwres}));
       },
     );
--- a/tests/e2e/support/commands.js
+++ b/tests/e2e/support/commands.js
@@ -210,11 +210,12 @@ Cypress.Commands.add('subscribeOpponent', (id) => {
       {
         id,
       },
-      function handleResponse(res, jwres) {
+      function handleResponse(_res, jwres) {
         if (jwres.statusCode === 200) {
           return resolve();
         }
-        return reject(new Error(jwres));
+        const errMessage = jwres?.message ?? jwres?.error ?? 'Unknown error';
+        return reject(new Error(`Error subscribing opponent to game ${id}: ${errMessage}`));
       },
     );
   });
--- a/tests/e2e/support/commands.js
+++ b/tests/e2e/support/commands.js
@@ -75,7 +75,12 @@ Cypress.Commands.add('setupGameAsP0', (alreadyAuthenticated = false, isRanked = 
     if (!alreadyAuthenticated) {
       cy.signupOpponent(opponentOne);
     }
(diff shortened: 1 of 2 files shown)


itsalaidbacklife:
Finally got cypress to log the error and it's indeed a 403 on the request to subscribe, which only has the `isLoggedIn` policy:

```
  Ace One-Offs
    (Attempt 1 of 3) Plays an Ace to destroy all point cards
    (Attempt 2 of 3) Plays an Ace to destroy all point cards
    1) "before each" hook for "Plays an Ace to destroy all point cards"


  0 passing (7s)
  1 failing

  1) Ace One-Offs
       "before each" hook for "Plays an Ace to destroy all point cards":
     Error: [ResponseFromSails]  -- Status: 403  -- Headers: [object Object]  -- Body: Forbidden
     ```

File: tests/e2e/support/commands.js [hunk 1]
```diff
@@ -75,7 +75,12 @@ Cypress.Commands.add('setupGameAsP0', (alreadyAuthenticated = false, isRanked =
     if (!alreadyAuthenticated) {
       cy.signupOpponent(opponentOne);
     }
-    cy.subscribeOpponent(gameSummary.gameId);
+    try {
+      cy.subscribeOpponent(gameSummary.gameId);
+    } catch {
+      cy.recoverSessionOpponent(opponentOne);
+      cy.subscribeOpponent(gameSummary.gameId);
+    }
```
itsalaidbacklife, line 83: So the problem we're seeing in CI is sometimes the request to subscribe fails because the `io` instance registers as unauthenticated. This feels like a kludge of a fix because it's 1-time operation (though we probably don't want to loop it), so idk maybe it's fine? 

@seriouslysean what do you think? If we're gonna do it, I can at least roll it out into `setupGameAsP1()`

tests/e2e/support/commands.js
```diff
@@ -204,7 +209,7 @@ Cypress.Commands.add('subscribeOpponent', (id) => {
         if (jwres.statusCode === 200) {
           return resolve();
         }
-        return reject(new Error('error subscribing'));
+        return reject(new Error(jwres));
```
itsalaidbacklife, line 212: This was how I finally got cypress to display the error in the terminal output. What sucks is I haven't been able to get it to show both a detailed message and the error in the response. Not sure what's the best practice here. Ideally I'd like it to send a specific message indicating where we are in the control flow and also to display the server's response so we know the status code and message

tests/e2e/support/commands.js
seriouslysean: Can you explain to me a little more about when it occurs? Is it every time, sometimes, unpredictable? Marine some sort of ab race condition where it’s connecting too soon?

itsalaidbacklife: At this point it seems to happen once per CI run in a random one-off spec, mainly sixes and aces

itsalaidbacklife: I wrote up #455 to investigate the root cause

seriouslysean: As long as this doesn't cause other problems with it, I'm fine with this for now. It might be better to have a poll try until it succeeds for up to 10 seconds but if this passes and is a test only then whatever. 👍🏼

seriouslysean (approved):

PR #457, opened 2023-07-20
Add Translations via Vue i18n

### LINKED ISSUE #456: [Feature]: Add Localizations/Translations
### Feature Summary

Implement Translations

### Detailed Description

`vue-i18n` would be my suggestion for the implementation.

seriouslysean:
## Issue number

Relevant issue number: Resolves #456

## Please check the following

- For New Features:

cuttle-i18n


package.json
```diff
@@ -15,6 +15,7 @@
     "sails.io.js": "1.2.1",
     "socket.io-client": "2.3.0",
     "vue": "^3.2.39",
+    "vue-i18n": "^9.2.2",
```
seriouslysean, line 18: Need this in order to localize the content in Vue.

src/App.vue
```diff
@@ -3,6 +3,15 @@
...
       <router-view />
+
+      <!-- temporary locale changer -->
+      <div class="locale-changer">
+        <select v-model="$i18n.locale">
+          <option v-for="(lang, i) in langs" :key="`Lang${i}`" :value="lang">
+            {{ lang }}
+          </option>
+        </select>
+      </div>
```
seriouslysean, line 14: This will let you switch between active locales. We'd need to move the locales in to some sort of config, or better yet just use the keys of valid translations once they're completed. Something like:
```
Object.keys(messages);
```

src/App.vue
```diff
@@ -14,6 +23,11 @@ export default {
   components: {
     NavigationDrawer,
   },
+  computed: {
+    langs() {
+      return ['en', 'es'];
+    },
+  }
```
seriouslysean, line 30: This would be dynamic based on the locales we support. We should also be adding the lang to the html of the app for accessibility.

src/components/NavigationDrawer.vue
```diff
@@ -44,6 +46,14 @@ import {
...
 export default {
   name: 'NavigationDrawer',
+  setup() {
+    // Vuetify has its own translation layer that isn't very good
+    // It seems to conflict with the namespace of vue-i18n so we need to import it at the component
+    // level and utilize it this way with a composable. There may be another more global way but
+    // I haven't found anything just yet
+    const { t } = useI18n();
+    return { $t: t };
+  },
```
seriouslysean, line 56: This isn't the worst, but does give me a little big of pause. I think there's a way to get the `$t` global in all components. While I don't want to load EVERY translation up front, I think some of them should be global (shared text from buttons that are reused, etc).

src/components/NavigationDrawer.vue
```diff
@@ -57,25 +67,25 @@ export default {
     pageLinks() {
       const rules = [
         {
-          text: 'Rules',
+          text: this.$t('global.rules'),
```
seriouslysean, line 70: Once we have them all organized, we can add inline to a component this way which is really nice.

src/i18n.js
```diff
@@ -0,0 +1,61 @@
...
+      collapseMenu: 'Collapse Menu',
+    },
+    login: {
+      noAccount: `Don't have an account?`,
+      haveAccount: `Already have an account?`,
+      title: 'What is Cuttle?',
+      quote: '"Cuttle is a sharp, fast game built entirely on excellent mechanics. It is the sort of game - had I known about it in college - I would have worn decks ragged through play."',
+      quoteCite: 'Richard Garfield - Creator of Magic: The Gathering',
+      paragraph1: `Cuttle is a 2 player battle card game played with a standard 52-card deck of cards. It has the strategic nuance of trading card games like Magic, with the elegant balance of a standard deck--and you can play it for free! Test your mettle in the deepest cardgame under the sea!`,
+      paragraph2: `Be the first to score 21 points in this explosive battle of wits. Mount a valiant offense while disrupting your opponent with dastardly tricks. Do you have what it takes to become the Lord of the Deep?`,
```
seriouslysean, line 25: This is my least favorite part of this setup and needs to change. I don't have a ton of experience with large setups of translating a lot of text, but when we have multiple paragraphs of the "same content", there has to be a better way to enable proper spacing and still put the content on the page.

Maybe some sort of markdown component we can pass markdown in to? Something like this I would imagine for a string called `loginPageContent` or whatever:

```
## What is Cuttle?

Cuttle is a 2 player battle card game played with a standard 52-card deck of cards. It has the strategic nuance of trading card games like Magic, with the elegant balance of a standard deck--and you can play it for free! Test your mettle in the deepest cardgame under the sea!

Be the first to score 21 points in this explosive battle of wits. Mount a valiant offense while disrupting your opponent with dastardly tricks. Do you have what it takes to become the Lord of the Deep?
```

Where markdown would automatically convert the `##` in to a heading, and then automatically wrap the paragraphs in p tags. It would also make contributing translations much easier imo.

src/App.vue
```diff
@@ -3,6 +3,15 @@
     <navigation-drawer />
     <v-main>
       <router-view />
+
+      <!-- temporary locale changer -->
+      <div class="locale-changer">
+        <select v-model="$i18n.locale">
+          <option v-for="(lang, i) in $i18n.availableLocales" :key="`${i}-${lang}`" :value="lang">
```
seriouslysean, line 10: Can loop available locales instead of hardcoding them.

src/components/Global/MarkdownContent.vue
```diff
@@ -0,0 +1,25 @@
+<script setup>
```
seriouslysean, line 1: This component will handle any markdown string and convert it to html, then `v-html` it in to a div with the class of markdown. The benefit is that we can use markdown in the localizations, and then style the resulting html. This does make it a littler harder to use vuetify but I think it's ok for blocks of text that should be colocated anyway.

src/components/Global/MarkdownContent.vue
```diff
@@ -0,0 +1,25 @@
...
+<template>
+  <div class="markdown" v-html="text"></div>
+</template>
+
+<style lang="scss">
+.markdown {
+  p {
+    padding-bottom: 16px;
+  }
+}
```
seriouslysean, line 24: We can also target the tags insider markdown easily-- this fixes a spacing bug on the homepage. Eventually we should move these to `em`s though.

image

src/components/NavigationDrawer.vue
```diff
@@ -44,6 +46,14 @@ import {
 
 export default {
   name: 'NavigationDrawer',
+  setup() {
+    // Vuetify has its own translation layer that isn't very good
+    // It seems to conflict with the namespace of vue-i18n so we need to import it at the component
+    // level and utilize it this way with a composable. There may be another more global way but
+    // I haven't found anything just yet
+    const { t } = useI18n();
+    return { t };
```
seriouslysean, line 55: `$` is a reserved prefix, so we need to use `t` instead which is fine with me.

src/i18n.js
```diff
@@ -0,0 +1,23 @@
+// https://vue-i18n.intlify.dev/
+
+import { createI18n } from 'vue-i18n';
+
+// TODO we should lazyload non global translations on a per page basis
+import en from '@/translations/en.json';
+import es from '@/translations/es.json';
+import fr from '@/translations/fr.json';
```
seriouslysean, line 8: Also added french. Viva la France!

image

src/translations/es.json
```diff
@@ -0,0 +1,22 @@
+{
```
seriouslysean, line 1: Another benefit of this is that we can pass them through AI (like ChatGPT) really quickly to translate. Is it perfect? No, but it's pretty damn good.

src/App.vue
seriouslysean: We'll also need a better place for this to live, and maybe some way to set the initial language based on the user's locale. That last one can be a follow up, though.

src/App.vue
seriouslysean: Moved to using the global `$i18n.availableLanguages`.

src/App.vue
itsalaidbacklife: That makes sense. I'll leave this comment open until we write up the issue so we remember to doc it

seriouslysean:
@itsalaidbacklife set to full PR to see if it passes the tests. I still want to add lazy loading of the locales, and a selector for language in the lefthand menu.

src/i18n.js
```diff
@@ -0,0 +1,26 @@
+// https://vue-i18n.intlify.dev/
+
+import { createI18n } from 'vue-i18n';
+
+// TODO we should lazyload non global translations on a per page basis
+// https://vue-i18n.intlify.dev/guide/advanced/lazy.html
+import en from '@/translations/en.json';
+// import es from '@/translations/es.json';
+// import fr from '@/translations/fr.json';
```
seriouslysean, line 9: Turning off Spanish and French for now since we don't use them.

seriouslysean (commented):
@itsalaidbacklife turned off the other locales and removed the locale switcher, so now this only lays the groundwork for i18n and "localizes" the login and signup page.

src/views/LoginView.vue
```diff
@@ -137,44 +137,34 @@
...
     <section>
       <v-container fluid class="quote text-h5 text-center d-flex flex-column align-center">
-        <blockquote>
-          "Cuttle is a sharp, fast game built entirely on excellent mechanics. It is the sort of game - had I
-          known about it in college - I would have worn decks ragged through play"
-        </blockquote> 
-        <br> 
-        <cite>Richard Garfield - <br>Creator of Magic: The Gathering</cite>
+        <blockquote>{{ t('login.quote') }}</blockquote>
+        <br>
```
seriouslysean, line 147: I don't like this, it should really be done with CSS but I left it so I didn't break anything else.

src/views/LoginView.vue
```diff
@@ -137,44 +137,34 @@
...
      ///BlockQuote Section////
     /////////////////////////-->
     <section>
       <v-container fluid class="quote text-h5 text-center d-flex flex-column align-center">
-        <blockquote>
-          "Cuttle is a sharp, fast game built entirely on excellent mechanics. It is the sort of game - had I
-          known about it in college - I would have worn decks ragged through play"
-        </blockquote> 
-        <br> 
-        <cite>Richard Garfield - <br>Creator of Magic: The Gathering</cite>
```
seriouslysean, line 151: I did remove this br here. If we need a line break, we need to make a new localized slug, or an interpolated slug so we can dynamically add the break tag. Barring that, I just removed it since it's more of a CSS thing.

utils/config-utils.js
```diff
@@ -9,7 +9,8 @@ export function initCuttleGlobals(app) {
   const cuttle = {
     version,
     // Expose app for debugging/testing in lower envs
-    app: test ? app : null,
+    // app: test ? app : null,
+    app,
```
itsalaidbacklife, line 13: Why this change? Was this just for local development? I don't think we want to add `app` to the window in prod

package.json
itsalaidbacklife, line 1: Shall we bump up a minor version?

utils/config-utils.js
seriouslysean: Oops, sorry! Forgot I was testing this locally to make sure i18n was in the global scope.

seriouslysean:
@itsalaidbacklife I'm considering this the basis for internationalization support. I recommend some follow up tickets that I recommend the Cuttle dev community help with:
1. Reorganize the slug naming
```
/locales
    /en
        global.json
        home.json
        stats.json
        game.json
    /es
        global.json
        home.json
        stats.json
        game.json
```
2. Create a localization dropdown to switch between locales easily -- I have the code, we just need to create a component around it and find a placement that works for all pages. This may work better after a persistent header is implemented or maybe we can even tie it to routes eventually like https://cuttle.cards/, https://cuttle.cards/es/, etc.
```
<select v-model="$i18n.locale">
  <option v-for="(lang, i) in langs" :key="`Lang${i}`" :value="lang">
    {{ lang }}
  </option>
</select>
```
3. Figure out an organized method to access/load localizations by locale and route. For instance:
    - Loading the homepage would load all `en` localization information (as primary, and a fallback) along with all the homepage localizations only, but not any other page like stats or game view
    - Loading the stats page directly would load all `en` localization information (as primary, and a fallback) along with all the stats localizations only
    - Clicking a locale in the dropdown would load all global locale information for the selected locale, along with page specific data.
    - (optional) Loading the `/es/` path on the homepage would load all english homepage data, all spanish homepage data and set english as the fallback.

itsalaidbacklife (approved):
Great work. This sets us up to roll our translations once we have the navigation redesigned to support language selection

PR #458, opened 2023-07-20
meta image

Mal402:
## Issue number

Relevant issue number (e.g. #404): Resolves #

## Please check the following

- For New Features:


index.html
```diff
@@ -6,6 +6,22 @@
...
+
+    
+    <meta property="og:url" content="https://www.cuttle.cards/" />
+    <meta property="og:title" content="Cuttle Cards Online" />
+    <meta property="og:type" content="website" />
+    <meta property="og:image" content="/public/img/cuttle_meta_img.jpg/" />
+    <meta property="og:description"
+        content="Cuttle is a 2 player battle card game played with a standard 52-card deck of cards. It has the strategic nuance of trading card games like Magic, with the elegant balance of a standard deck--and you can play it for free!" />
+
+    <meta name="twitter:site" content="https://unacog.com/">
```
itsalaidbacklife, line 18: Not sure what these do but they should point to the cuttle twitter, right?
```suggestion
    <meta name="twitter:site" content="https://twitter.com/cuttle_cards/">
```

index.html
```diff
@@ -6,6 +6,22 @@
...
+    <meta name="twitter:title" content="Unacog About">
```
itsalaidbacklife, line 19: What's this one for?

index.html
```diff
@@ -6,6 +6,22 @@
     <meta name="viewport" content="width=device-width,initial-scale=1.0" />
     <link rel="icon" href="/favicon.ico" />
     <title>Cuttle</title>
+
+    
+    <meta property="og:url" content="https://www.cuttle.cards/" />
```
seriouslysean, line 11: We need some sort of env for the site url instead of hardcoding it.

index.html
```diff
@@ -6,6 +6,22 @@
...
+    <meta property="og:title" content="Cuttle Cards Online" />
+    <meta property="og:type" content="website" />
+    <meta property="og:image" content="/public/img/cuttle_meta_img.jpg/" />
+    <meta property="og:description"
+        content="Cuttle is a 2 player battle card game played with a standard 52-card deck of cards. It has the strategic nuance of trading card games like Magic, with the elegant balance of a standard deck--and you can play it for free!" />
```
seriouslysean, line 16: This should likely be shared somewhere or possibly set in vite as template vars.

index.html
seriouslysean: Yes, likely just copied from  somewhere else. I would prefer to have an env for this too.

Mal402: let me fix these, my mistake. 

We were suppose to replace https://unacog.com/ with https://cuttle.cards - this was from the code i pulled it from. 

Content should not be content="https://twitter.com/cuttle_cards/"> 
but content="https://cuttle.cards"> 

Twitter:site

index.html
```diff
@@ -6,6 +6,22 @@
...
+
+    <meta name="twitter:site" content="https://unacog.com/">
+    <meta name="twitter:title" content="Unacog About">
+    <meta name="twitter:description"
+        content="Cuttle is a 2 player battle card game played with a standard 52-card deck of cards. It has the strategic nuance of trading card games like Magic, with the elegant balance of a standard deck--and you can play it for free!">
+    <meta name="twitter:card" content="summary_large_image">
+    <meta name="twitter:image:src" content="/public/img/cuttle_meta_img.jpg/">
```
seriouslysean, line 23: Name should use hyphens imo and be named more clearly. Something like `cuttle-twitter-banner`, promo image, etc.

seriouslysean (changes requested):
Thanks for the contribution! Needs a little work before it goes in, though.

index.html
Mal402: This was a copy paste mistake! It should be chagned to cuttle.card with the https

index.html
Mal402: yeah sorry - some sloppiness from my project :) please make any necessary adjustments.

index.html
Mal402: This does not point to cuttle twitter. it instead lets the OG protocol know what the site is about. Facebook made the protocol to make sharing more media rich. Twitter implemented their own standard and that's what this is pointing to. Discord uses the twitter meta data and so we need to put it in there if we want to have nice image and rich description when sharing the link on discord.

index.html
itsalaidbacklife: @seriouslysean from what I can tell, vue doesn't natively support dynamic `<meta>` tags. It looks like if we want to make it programatic we could use the [unhead lib](https://unhead.harlanzw.com/guide/getting-started/setup) which seems to be the closest thing to a standard approach I can find.

That said, the main reason for making this programatic is basically to be able to reuse the content values for the og (facebook) and twitter properties, right? Curious whether it's worth a dependency to manage that complexity. Penny for your thoughts

index.html
itsalaidbacklife: Okay thanks for explaining. I haven't previously done much with metadata so now I see that `twitter` and `og` are essentially separate prefixes/protocols for defining the same properties in ways that will work across different media. Thanks for the tip.So this should be
```suggestion
    <meta name="twitter:site" content="https://cuttle.cards/">
```

index.html
itsalaidbacklife: Gotcha. Then I think we just want 
```suggestion
    <meta name="twitter:title" content="Cuttle">
```

index.html
itsalaidbacklife: @seriouslysean you mean for the image filename? If it's the same img file for the `og` and `twitter` protocols, maybe we can stick with a generic name (which @Mal402 we do generally use kebab-case capitalization for):

`cuttle-summary-banner.jpg` maybe?

index.html
seriouslysean: You could also use something like https://vue-meta.nuxtjs.org/

I think the benefit is that you’ll be able to easily swap based on the page type without too much overhead. So rules, homepage, etc could all have their own content.

this is particularly nice if you’re planning on building this out and making localized pages behind a context path like /es or /fr, etc.

itsalaidbacklife: I think that makes sense and could reasonably be a follow up issue. First step is just getting the static metadata in there.

When we have a clearer vision for what page-specific metadata should be, we can implement `useHead()` to make it dynamic

index.html
```diff
@@ -6,6 +6,20 @@
     <meta name="viewport" content="width=device-width,initial-scale=1.0" />
     <link rel="icon" href="/favicon.ico" />
     <title>Cuttle</title>
+    <meta property="og:url" content="https://www.cuttle.cards/" />
+    <meta property="og:title" content="Cuttle" />
+    <meta property="og:type" content="website" />
+    <meta property="og:image" content="/public/img/cuttle-summary-banner.jpg/" />
```
itsalaidbacklife, line 12: @Mal402 I gave you the wrong advice earlier. We should skip /public
```suggestion
    <meta property="og:image" content="/img/cuttle-summary-banner.jpg/" />
```

index.html
```diff
@@ -6,6 +6,20 @@
...
+    <meta property="og:description"
+        content="Cuttle is a 2 player battle card game played with a standard 52-card deck of cards. It has the strategic nuance of trading card games like Magic, with the elegant balance of a standard deck--and you can play it for free!" />
+
+    <meta name="twitter:site" content="https://www.cuttle.cards/">
+    <meta name="twitter:title" content="Cuttle">
+    <meta name="twitter:description"
+        content="Cuttle is a 2 player battle card game played with a standard 52-card deck of cards. It has the strategic nuance of trading card games like Magic, with the elegant balance of a standard deck--and you can play it for free!">
+    <meta name="twitter:card" content="cuttle-summary-banner-image">
+    <meta name="twitter:image:src" content="/public/img/cuttle-summary-banner.jpg/">
```
itsalaidbacklife, line 21: Same here
```suggestion
    <meta name="twitter:image:src" content="/img/cuttle-summary-banner.jpg/">
```

index.html
itsalaidbacklife: Twitter recommends a char limit of 200 for the content, so let's reword this
```suggestion
        content="Cuttle is a battle card game played with a standard 52-card deck of cards. The strategic nuance of trading card games, with the elegant balance of a standard deck--and you can play it for free!" />
```

index.html
itsalaidbacklife: @seriouslysean I created #462 to track this

itsalaidbacklife: @seriouslysean I created #462 to track this

itsalaidbacklife (approved):
@Mal402 this is great! I've got a couple remaining changes requested in the comments (I told you the wrong thing for image urls, and I'd like to shorten the description to < 200 chars).

Other than that, we have a system where we bump the "version" field in package.json every time we merge an update. This will be a "patch" version (3rd number), so can you please pull from the main branch and merge that into your local branch, then go into `package.json` and increase the version from 4.4.2 to 4.4.3 (one patch version). Commit that and push along with the changes in the comments and this is good to go!

Nice work on this; it will be a great QoL improvement that makes Cuttle easier to share :)

seriouslysean:
@itsalaidbacklife are we updating package lock also 😬

itsalaidbacklife:
> @itsalaidbacklife are we updating package lock also 😬 

Actually no. package-lock updates whenever we `npm install` which means we can use it to tell when was the last time we updated deps, which I think is handy. So we manually update package.json and leave package-lock until we install deps

seriouslysean:
@itsalaidbacklife  we should talk about that at some point. It’s recommended you keep those two synced. See https://docs.npmjs.com/cli/v9/commands/npm-version which would also make tags for us, and we might want to also start doing “releases” of the app when it’s stable in GitHub.

seriouslysean (approved):
Leaving final approval to @itsalaidbacklife

itsalaidbacklife:
> @itsalaidbacklife  we should talk about that at some point. It’s recommended you keep those two synced. See https://docs.npmjs.com/cli/v9/commands/npm-version which would also make tags for us, and we might want to also start doing “releases” of the app when it’s stable in GitHub.

Yeah good call. I'd never found anything definitive on it so I'd made up that policy a while back with no real basis -- we should discuss. Releases is an interesting concept though I wonder how to think about them when were releasing every merge

itsalaidbacklife (approved):
Nice work! Thanks for your contribution

Issue #446, opened 2023-06-23
[Bug]: Close buttons have wrong style in TargetingOverlay, RulesDialog, and ScrapDialog


itsalaidbacklife:
### Bug Summary

We frequently use an “X” button in dismissible dialogs and overlays, and several of them have incorrect styles 

### Detailed Description

The ScrapDialog, TargetSelectionOverlay, and RuleaDialog all have the incorrect button styles for this X button. They should look like the X in the MoveChoiceOverlay: no background color and an X icon. Instead they have a white background, and the TargetSelectionOverlay specifically has no icon visible.

To fix it we should revise the close buttons in these components to match the style from the MoveChoiceOverlay

### Current Behavior

They currently look like this
![IMG_4491](https://github.com/cuttle-cards/cuttle/assets/6520704/81d16806-9b9f-417e-b108-8e70ffa5b355)

![IMG_4490](https://github.com/cuttle-cards/cuttle/assets/6520704/57dbbfbd-713c-4f0d-95b4-68b49f13cadb)
![IMG_4489](https://github.com/cuttle-cards/cuttle/assets/6520704/b81c98bb-4fe8-4284-bbf6-4b0caecc8b73)

### Expected Behavior

They should look like this
![IMG_4492](https://github.com/cuttle-cards/cuttle/assets/6520704/3dcd309a-21a5-4c2a-b260-fa14c679392d)

### Steps to Reproduce

1.
2.
3.

web-dev-kushal:
I want to work on this issue. Pls do assign me

itsalaidbacklife:
> I want to work on this issue. Pls do assign me
> 
> 

Have at it! Feel free to ask any questions about the issue here. If you'd like help setting up, you can chat with me and the other devs on our [discord](https://discord.gg/dS5BRjaHrm)

Mal402:
Hey, I think this is something that I can take on!

kawaljen:
Hey can I work on this one ?

itsalaidbacklife:
> Hey can I work on this one ?

Yes please!

kawaljen:
![Screenshot from 2024-01-21 10-33-05](https://github.com/cuttle-cards/cuttle/assets/7325003/9276b0ef-eebf-4b47-a54b-1e35f53dc3de)

kawaljen:
[...]

PR #467, opened 2023-07-26
close btn fixes

Mal402:
## Issue number

Relevant issue number: #

## Please check the following

- For New Features:

### DIFF
--- a/src/components/GameView/ScrapDialog.vue
+++ b/src/components/GameView/ScrapDialog.vue
@@ -8,7 +8,7 @@
     <v-card id="scrap-dialog">
       <v-card-title class="d-flex justify-space-between">
         <h1>Scrap Pile</h1>
-        <v-btn icon data-cy="close-scrap-dialog-x" @click="show = false">
+        <v-btn icon data-cy="close-scrap-dialog-x" variant="text" @click="show = false">
           <v-icon icon="mdi-close" size="large" />
         </v-btn>
       </v-card-title>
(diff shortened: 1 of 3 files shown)


src/components/GameView/ScrapDialog.vue
```diff
@@ -24,7 +24,6 @@
       <v-card-actions class="d-flex justify-end my-2">
         <v-btn
           color="primary"
-          variant="outlined"
```
itsalaidbacklife, line 27: This one can still be outlined. It's really the icon buttons we want to update here

src/components/GameView/TargetSelectionOverlay.vue
```diff
@@ -18,7 +18,7 @@
         Cancel
       </v-btn>
     </div>
-    <v-btn icon data-cy="cancel-target-mobile" @click="$emit('cancel')">
+    <v-btn icon data-cy="cancel-target-mobile" variant=“text” @click="$emit('cancel')">
```
itsalaidbacklife, line 21: How does this one look now? Could you add a screenshot of the new result? I'm wondering if we should change the color in addition to the button style or if it's good as-is. What do you think?

itsalaidbacklife (commented):
Nice work, this is almost ready. I've got one change request and one question in the PR comments, and then there are three quick chores to do:

1) The pr failed CI at the linting step, which means the code styles are wrong (in this case because line length spilled over a threshold). To fix it just run `npm run lint:fix` and commit + push the results
2) Please bump the `version` in package.json a patch version i.e. to 4.4.5
3) in the PR description, where it says Relevant issue: Resolves #, please update it to Resolves #446, so that the issue is automatically closed once this PR is merged (generally most of our PR's address a preexisting issue and should reference it in this portion of the description to keep the issue tracking up to date).

Thanks for your help!

itsalaidbacklife:
Closing as inactive

Issue #464, opened 2023-07-24
[Feature]: Display site metrics for usage (games played and number of unique players) in stats page


itsalaidbacklife:
### Feature Summary

It would be helpful to have easier visibility into the site's main usage metrics: the number of games played, and the number of unique players, on a weekly basis. We could display this data in the 'Stats' page under a new sub page called 'Site Usage'

### Detailed Description
This is a follow up to #468 which is the backend implementation to provide that data

The two most salient metrics for the usage of cuttle.cards are:

1. The number of unique players in a week
2. The number of games played in a week

We could create a line graph with lines for these two values relative to the selected season. We could put it in the 'Stats' page and add tabs underneath the dates for the selected season that say:
Leaderboard (this would be the current version of the stats page content)
Site Usage (this would be where we put the new chart)

We could use [vue chartjs](https://vue-chartjs.org/guide/) for the charts and update the existing backend endpoint for calculating stats data so that it includes information on the number of games played and the number of unique users. This should include both ranked and unranked games and only list the number of users, not the actual usernames. The chart should use our new primary and secondary colors.

PR #469, opened 2023-07-26
Feat/usage stats backend

### LINKED ISSUE #468: [Feature]: Backend for #464 -- Site usage stats
### Feature Summary

Prerequisite for #464 -- this is the backend issue to update the StatsController so that it delivers data on the number of unique players and the number of games played for each week of each season.

### Detailed Description

In order to create visibility into how the site is growing and being used, we want to create charts that showcase the weekly usage via two metrics:
1. Number of unique users who played games
2. Total count of games played

This is the backend issue for implementing:
1. The test logic necessary to seed completed game data
2. A test that asserts the expected response object has the correct data, given the seeds
3. The StatsController update to send the correct data

Each `Season` returned by the backend in `StatsController.getStats()` should have two additional attributes:
1. `gameCounts: Array<int>`
2. `uniquePlayersPerWeek: Array<int>`

where each index in the array corresponds to the week number within that season (starting at 0) and the value is the count of games played or unique players for that week

itsalaidbacklife:
## Issue number

Relevant issue number (e.g. #404): Resolves #468 

This adds data to the season objects sent by the backend from the /stats endpoint. Specifically it adds
`uniquePlayersPerWeek: int[]` and `gameCounts: int[]` which are each a list of the counts in each week i.e. `gameCounts[2]` is the number of games played in the 3rd week of the season. This will let us create a dashboard chart with this data per season when we complete #464 

## Please check the following

- For New Features:


api/controllers/StatsController.js
```diff
@@ -96,20 +102,76 @@ function transformSeasonToDTO(season) {
...
+*  firstPlace: int (playerId)
+*  secondPlace: int (playerId)
+*  thirdPlace: int (playerId)
+*  fourthPlace: int (playerId)
+*  rankings: Map<int: playerId, {
+*    username: string
+*    matches: Map<int: weekNum, Array<{opponent: string, result: Result}>>
+*  }>
+*  gameCounts: Array<int>
+*  uniquePlayersPerWeek: Array<Set<int: playerIds>>
```
itsalaidbacklife, line 139: The `uniquePlayerCountsPerWeek` uses a `Set()` for each week to de-dupe the players by their id, then converts to the size of that set for the response sent back clientside. This makes for relatively efficient processing without sending usernames or ids to the client beyond the ones used for the ranked leaderboard

api/controllers/StatsController.js
```diff
@@ -130,6 +192,21 @@ module.exports = {
           }
         }
       });
+
+      games.forEach((game) => {
+        const relevantSeason = seasons.find((season) => {
+          // Find season this match took place during
+          return dayjs(game.updatedAt).isBetween(dayjs(season.startTime), dayjs(season.endTime));
+        });
```
itsalaidbacklife, line 200: I'd like to DRY this up since it's used in for both games and matches above, and ideally refactor it to use a binary search so that these loops each run in O(Nlog(N)) instead of `O(N^2)`

Unfortunately I don't see a way to reduce the repetition of 

```
        if (!relevantSeason) {
          return;
        }
```

which both also use

api/controllers/StatsController.js
```diff
@@ -118,18 +180,34 @@ module.exports = {
...
-            const player2 = idToUserMap.get(match.player2);
-            if (player1 && player2) {
-              // Player 1
-              addMatchToRankings(relevantSeason, match, player1, player2);
-              // Player 2
-              addMatchToRankings(relevantSeason, match, player2, player1);
-            }
+          if (!relevantSeason) {
+            return;
+          }
```
itsalaidbacklife, line 185: Swapping to early return to reduce indentation & to be more consistent with the rest of the app

api/controllers/StatsController.js
```diff
@@ -118,18 +180,34 @@ module.exports = {
...
+            addMatchToRankings(relevantSeason, match, player2, player1);
           }
         }
       });
+
+      games.forEach((game) => {
+        const relevantSeason = seasons.find((season) => {
+          // Find season this match took place during
+          return dayjs(game.updatedAt).isBetween(dayjs(season.startTime), dayjs(season.endTime));
+        });
```
itsalaidbacklife, line 201: It might make sense to pull the games update and the match updates into their own functions to make the actual controller endpoint a bit more lean, idk. @seriouslysean how clearly does this read to you?

api/controllers/StatsController.js
seriouslysean: Marking this so I can look later :)

byarmis: Can this be moved to be a part of the match logic above?  Since you already have the `relevantSeason` for a match, you can then associate the match's games with that season as well.  Would also DRY it up since you're only looping through everything once

itsalaidbacklife: @byarmis the issue there is that we're looping through the matches in one spot and the games in another, so it's not clear to me how to combine them. `relevantSeason` is relative to whichever game or match you're looking at i.e. it's the season in which that game or match took place. So it would be different for the games vs the matches.

Granted we could query for all the games associated with each match and process those with a nested for loop, but the issue there is that we only have matches for ranked games so that would skip all the unranked ones. That's why we have the query for all finished games as separate from the matches and why the loop is processed as a sibling rather than a nested loop: the games include ranked and unranked while the matches are only ranked

itsalaidbacklife: This speaks to a suggestion you made earlier @byarmis to associate every game (including unranked) with a match. That might clean up this processing, (though it introduces its own complexities), but it would entail a larger rework

byarmis: Ahh that's a good point, I didn't remember that matches only happened for ranked games.

It doesn't appear to be the most common paradigm used in this code base, but you could also maybe have a seasons class that, as an instance variable, has a list of seasons and a `.findSeason()` method that takes in a date and returns a season so the sorting and binary searching only need to happen in one place.

itsalaidbacklife: Yeah I'd definitely like to sort the seasons and replace this `.find()` with a binary search. Since getting the seasons requires a query we want to make sure we only query for the seasons once and that this search method just processes them in memory.

My plan was to define it as a function that takes the sorted array of seasons and the date as arguments and to define it as a sails helper since that's generally how we've organized the reusable backend utilities. A static method on a singleton class is an interesting idea though.

api/controllers/StatsController.js
```diff
@@ -6,6 +6,40 @@ dayjs.extend(isBetween);
...
+ * @param {*} timeStamp - Desired time in millis since epoch
+ * @returns the season | null if none is found
+ */
+function getSeasonFromTimeStamp(sortedSeasons, timeStamp) {
+  let currentIndex;
+  let minIndex = 0;
+  let maxIndex = sortedSeasons.length - 1;
+  // Sanity check in case list is unsorted or start & endTimes are out of whack
+  let numTries = 0;
+  while (numTries < sortedSeasons.length) {
```
itsalaidbacklife, line 21: So this condition feels silly because so far as I can tell, there is actually no way that the while loop exits without returning `return currentSeason` or `return null`. I think this could actually have a constant condition, but it sketches me out to use `while (true)`.

It seems like a natural exit condition to use `while (minIndex < maxIndex)` except the way the loop is written atm we don't set the `currentIndex` or `currentSeason` until the beginning of the next loop (and each loop ends just with setting `minIndex` or `maxIndex`. I think that's relatively clean b/c each loop only winds up in one path and ultimately returns or sets one variable, but maybe it would make sense to set and check `currentIndex` + `currentSeason` at the end of the loop rather than the beginning so we could replace the block

```
    // If min = max and we haven't found it, there is no matching season
    if (minIndex === maxIndex) {
      return null;
    }
```

with an exit based on the while's condition using `while (minIndex < maxIndex)`

itsalaidbacklife (commented):
Did some refactoring and leaving additional comments for further thought

api/controllers/StatsController.js
itsalaidbacklife: I've pulled the season lookup into  `getSeasonFromTimestamp()` above to DRY this up and utilize binary search

api/controllers/StatsController.js
```diff
@@ -96,20 +102,76 @@ function transformSeasonToDTO(season) {
...
+    const startTime = dayjs(season.startTime);
+    const endTime = dayjs(season.endTime);
+    const numWeeks = endTime.diff(startTime, 'week');
+    for (let i=0; i<numWeeks; i++) {
+      res.gameCounts.push(0);
+      res.uniquePlayersPerWeek.push(new Set());
+    }
+    return res;
+  });
+}
```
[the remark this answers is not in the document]
itsalaidbacklife: Done. I pulled the logic for initializing the `uniquePlayersPerWeek` and `gameCounts` arrays into the `getSeasonsWithoutRankings` helper. This help keeps the type signatures more consistent so we aren't adding properties at different stages (now there's the one type used during processing and the transformed version returned by the API), plus we skip an extra loop by consolidating the initialization in the one `getSeasonsWithoutRankings()` helper rather than looping there and also in the controller

api/controllers/StatsController.js
```diff
@@ -6,6 +6,40 @@ dayjs.extend(isBetween);
 /////////////
 // Helpers //
 /////////////
+/**
+ * Find the season in which a given timestamp occurred using binary search
+ * @param {*} sortedSeasons - Seasons sorted (OLDEST FIRST)
```
itsalaidbacklife, line 11: So it feels kinda strange to me to require that the seasons be sorted in DESC chronological order (oldest first), but that's what the api has been delivering in the end and what the frontend expects in order to put the oldest/current season in the top of the drop-down list. 

I initially had implemented this function so that it expected the seasons to be sorted in ASC chronological order ([commit](https://github.com/cuttle-cards/cuttle/commit/ded2b2195a4362b970e18fee2c049128ee746ebb)), but that meant that we had to loop again to transform the array back to DESC order before displaying it (in that commit I did it clientside in the page's `created()` hook), which just seems unnecessary given that the only reason to sort it ASC was because that seemed more "natural" for a binary search. 

But given that this function is only called in this controller I'm a context where were ultimately going to deliver the seasons in DESC chronological order, I think it actually reads better this way and also thereby skips the need to flip the sort order back and forth. It just has the quirks that the `sortedSeasons` param needs to be sorted in DESC chronological order (which it already is when returned by the `getSeasonsWithoutRankings()` helper and that the higher/lower cases below decrement `maxIndex` and increment `minIndex` respectively, rather than the other way around.

api/helpers/get-seasons-without-rankings.js
```diff
@@ -12,14 +12,25 @@ module.exports = {
...
+          gameCounts: [],
+          uniquePlayersPerWeek: [],
         };
+        // initialize gameCounts and uniquePlayersPerWeek
+        const startTime = dayjs(season.startTime);
+        const endTime = dayjs(season.endTime);
+        const numWeeks = endTime.diff(startTime, 'week');
+        for (let i=0; i<numWeeks; i++) {
+          res.gameCounts.push(0);
+          res.uniquePlayersPerWeek.push(new Set());
```
itsalaidbacklife, line 31: `uniquePlayersPerWeek` uses a `Set` of the player's ids who played each week so that we can efficiently dedupe them when processing the games. When transforming the processed data to return it as the api response, we convert the sets to their integer size to get the actual player counts

src/views/StatsView.vue
```diff
@@ -154,6 +154,7 @@ export default {
         this.error = true;
         return;
       }
+      // Sort most recent season first
```
itsalaidbacklife, line 157: This is stale
```suggestion
```

tests/e2e/fixtures/statsFixtures.js
```diff
@@ -326,6 +326,90 @@ const seasonThreeMatches = [
   },
 ];
 
+const gameFixtures = [
```
itsalaidbacklife, line 329: Should probably add some additional games to test some edge cases:
1) game in a different season
2) game before first season
3) game after last season

itsalaidbacklife (commented):
Couple explanations and todos

api/controllers/StatsController.js
```diff
@@ -6,6 +6,40 @@ dayjs.extend(isBetween);
...
+ * @param {*} timeStamp - Desired time in millis since epoch
+ * @returns the season | null if none is found
+ */
+function getSeasonFromTimeStamp(sortedSeasons, timeStamp) {
+  let currentIndex;
+  let minIndex = 0;
+  let maxIndex = sortedSeasons.length - 1;
+  // Sanity check in case list is unsorted or start & endTimes are out of whack
```
seriouslysean, line 19: I need a little more context regarding this bit of logic, when and how do they get out of whack?

api/controllers/StatsController.js
seriouslysean: I would love to talk through this more and see if there's a better data fetch we could do or a less flakey loop to get the data you're looking for.

api/controllers/StatsController.js
```diff
@@ -6,6 +6,40 @@ dayjs.extend(isBetween);
...
+  let numTries = 0;
+  while (numTries < sortedSeasons.length) {
+    currentIndex = Math.floor((minIndex + maxIndex) / 2);
```
seriouslysean, line 22: Is this just getting the middle result? Is there not a better way to get the current season? I might also be reading this wrong, PRing on my phone from vacation. 😎

api/controllers/StatsController.js
```diff
@@ -6,6 +6,40 @@ dayjs.extend(isBetween);
...
+    const currentSeason = sortedSeasons[currentIndex];
+    if (timeStamp >= currentSeason.startTime && timeStamp <= currentSeason.endTime) {
+      return currentSeason;
+    }
+    // If min = max and we haven't found it, there is no matching season
+    if (minIndex === maxIndex) {
+      return null;
+    }
+    // Target time is before current season
+    if (timeStamp < currentSeason.startTime) {
```
seriouslysean, line 32: Would it be better to allow a query with timestamp on the records so we don't need to do that logic in JavaScript?

seriouslysean (commented):
More later, I only got through a single function lol

api/controllers/StatsController.js
byarmis: Is it possible to move the minIndex / maxIndex updating logic to the beginning of the loop and change the conditional to be `while (minIndex < maxIndex)`?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "4.4.4",
+  "version": "4.4.5",
```
byarmis, line 4: Nit: this change would be a minor change, not a patch change

```suggestion
  "version": "4.5.0",
```

tests/e2e/fixtures/statsFixtures.js
byarmis: Speaking of  testing, and I don't recall if this is set up, but I think `getSeasonFromTimeStamp` is ripe for some unit tests just because there are so many different cases:

* no seasons
* 2 seasons (which is tested here)
* 3 seasons
* game before the first season
* game after the last season

Most of those cases probably aren't possible, but having the backend error out for such an avoidable reason feels silly.  Setting up all of these in `gameFixtures` is possible, but I imagine can get annoying and the tests end up kinda removed from the thing that they're testing

api/controllers/StatsController.js
itsalaidbacklife: So realistically it shouldn't. So first the purpose of this function is to identify in which season a given timestamp occurred ie which season was this game or season in? 

The function should return the relevant season if there is one and `null` otherwise. To do that, it makes several assumptions, namely that every season's time `startTime` is before its `endTime` and that the `sortedSeasons` is sorted in descending chronological order.

That should always be the case unless one or more of the seasons get defined incorrectly in the database or if this function gets called in an additional place with incorrectly sorted args. If either of those weird situations happen, we want it to just return `null` to ensure we never infinitely loop. That said, my "sanity check" here doesn't look like it would ever fire because the conditional 

```
if (minIndex === maxImdex) {
  return null;
}
```
should always happen first (since it takes log_2(sortedSeasons.length) loops for min and max index to meet, but hitting this `numTries` limit requires as many loops as the length of the list, which we can't hit before the other check. Which means the while loop could just be `while (true)`, but even though I can't imagine a way for that to infinite loop, that still makes me nervous

api/controllers/StatsController.js
itsalaidbacklife: Yeah it just calculates true middle index between `minIndex` and `maxIndex`. I can't think of a simpler way than averaging and using `Math.floor()` to truncate to int, but I'm open to suggestions

api/controllers/StatsController.js
itsalaidbacklife: Actually there is overhead that comes with each individual query e.g. network latency since the database and app server live on different vm's and it is generally more performant to do fewer, batched queries for more data then process it with serverside code than it is to repeatedly make smaller queries that each require minimal serverside processing.

As a rule of thumb, rather than querying from inside a loop, you want to query once and loop through the results. That's why this controller method queries for the seasons once at the top of the api func and then passes the array of seasons as an argument here to prevent repeated queries

api/controllers/StatsController.js
itsalaidbacklife: > Is it possible to move the minIndex / maxIndex updating logic to the beginning of the loop and change the conditional to be `while (minIndex < maxIndex)`?

Yeah I think that would make more sense. It might have slightly more redundancy since atm the beginning of the first loop initialized currentIndex whereas I think we'd need to initialize it outside the loop to make this change but I think it would read more clearly if the exit condition were explicitly in the loop in that way, yeah

package.json
itsalaidbacklife: I think since this is just the backend update and there's no user-facing change yet that it's better represented as a patch. I'd consider the front end implementation a minor update when this data is actually visualized for users

byarmis: If we're getting technical, changes to what data an API returns _are_ user-facing.  In the case of the API, the "user" is the frontend.

If there is something written to interface with the post-change version of the API, it might not function with the pre-change version of the API

According to [SemVer](https://semver.org/):

> Given a version number MAJOR.MINOR.PATCH, increment the:

> 1. MAJOR version when you make incompatible API changes
> 2. MINOR version when you add functionality in a backward compatible manner
> 3. PATCH version when you make backward compatible bug fixes

The changes you're proposing are backwards compatible (older consumers of the API won't need to change anything), but you're adding functionality and not just fixing bugs

api/controllers/StatsController.js
itsalaidbacklife: I think I've worked out a reasonable compromise. @byarmis I think the conditional should be inclusive rather than exclusive:
```
while (minIndex <= maxIndex)
```

because if the updates to the `minIndex` and `maxIndex` come after the assignment and checks of `currentSeason`, then we need to do the exact same checks _before the loop_ in order to check the _first_ value of `currentSeason` (i.e. the halfway point of the list). By using `<=` instead of `<` and checking `currentSeason` at the beginning of the loop, but setting `currentIndex` at the _end_ of the loop, I think we get the best of both worlds since all the checking is still done inside the loop without needing to separately handle the first case/iteration, and we're still able to end the loop if the indices ever meet without finding the target season.

package.json
itsalaidbacklife: Fair point; it's always helpful to come back to the documentation. That said, as I understand it, the 'by the book' definitions of SemVer don't really apply to applications in the first place, right? They were clearly written for tooling and dependencies. What does backwards compatible even mean for an application like Cuttle? Would it mean "being able to use an out-of-date client with the deployed API?"

That's a pretty weird use case (especially for the primary differentiator between version numbers), but it's the most coherent I can think of. Is that what you have in mind? So the rule would essentially be:

1. Every bugfix is a patch unless it requires changing the api in such a way that an out of date client... couldn't use the application? Would still experience the bug? In which case it becomes a major version change? I'm fuzzy on whether 'add functionality' is limited to new features or whether this also applies to bugs, but it sounds like every bugfix is either a patch or a major version from the above. Is that right?
2. Every new feature is a minor version change (including for example a single line of css, right?) unless it requires api changes in which case it's a major version? I'm still fuzzy on what would constitute 'incompatible' changes and whether I am right to hone in specifically on the idea of someone using a previous iteration of the frontend to connect directly to the current backend as the defining principle or if there is something else to use.

In part what's counterintuitive about this is that some bug fixes feel 'bigger' than some 'features' and visa versa. But I'd love arrive at a clearer and more consistent definition and guideline I can use for defining the version numbers, because I'll readily admit that I've been shooting from the hip. So this is helpful discourse! I think the key thing for me would be unpacking the notion of 'compatibility' in the context of applications, as well as wrapping my head around who the 'user' is given that some changes are dev-specific e.g. documentation, refactoring, linting, testging setup etc, and others are obviously player-centric. How do you frame your thinking about these issues @byarmis?

seriouslysean: I’m with @byarmis — it’s completely possible (albeit very unlikely) that an old client would connect to the app. Unless we added an out of version check to the backend and had it return a 400 level error message, it’s still possible.

that said, I’m all for tightening up our own definition and while I do think there’s some level of guesswork we can certainly define and document our own ideas and rails.

tests/e2e/fixtures/statsFixtures.js
itsalaidbacklife: I also updated the cypress test so that it loads a fixture that includes:
1. A game before the first season
2. A game after the last season
3. Games in each season across multiple weeks

And asserts that the endpoint responds with the expected data. This isn't as thorough as the unit testing we can get in #475 , but it does address the most essential cases imo

PR #470, opened 2023-07-31
Feat #395 out of game dialogs

### LINKED ISSUE #395: [Feature]: Apply the new BaseDialog colors & styles (chocolate and cream) to the out-of-game dialogs
### Feature Summary

#389 revised all the in-game dialogs by defining a new BaseDialog component that the dialogs in the game consume to standardize their look and feel. Now we should update the styles of the out-of-game dialogs e.g. CreateGameDialog to match

### Detailed Description

This is the style of our new in-game dialogs:
image

They use the surface-1 (chocolate) and surface-2 (cream) colors from our client/plugins/vuetify.js theme definition. 

The following dialogs in the application do not yet use the BaseDialog component and therefore have a bland black and white look that doesn't match the application:

1. CreateGameDialog
2. RulesDialog
3. StatsScoringDialog

Here's how the CreateGameDialog looks, for comparison
image

Updating these remaining dialogs will improve the look and feel of the application and give users a better and more coherent sense for the brand of the site.

For implementation, look at the definition of the `client/components/Global/BaseDialog.vue` component and look at how it is used in the FourDialog to see how to apply it in these remaining three out-of-game dialogs. Notably:

1. id and title (the heading) are set via props
2. The dialog body is set via a template tag filling the #body slot (`<template #body></template>`)
3. The dialog buttons are set via a template tag filling the #actions slot (`<template #actions></template>`)

Haviles04:
## Issue number

Relevant issue number (e.g. #404): Resolves #395 

## Please check the following

- For New Features:

image

image

image


src/components/GameView/GameMenu.vue
```diff
@@ -113,8 +110,20 @@ export default {
...
   },
   methods: {
+    closeMenu() {
+      this.showGameMenu = false;
+    },
+    openRulesDialog() {
+      this.showRulesDialog = true;
+      this.showConcedeDialog = false;
+      this.showStalemateDialog = false;
+      this.showGameMenu = false;
```
itsalaidbacklife, line 125: This makes sense and is consistent with how we have the other dialogs, but I don't love the overlap between the various opening and closing functions, given that the dialogs are mutually exclusive and we have to close all the other ones every time we open something.

I wonder if it would make sense to use a single string-values variable for which dialog is open and use computed props to consume it for each individual dialog. Something like:

```
data() {
  return {
    shownDialog: ''; // '' | 'rules' | 'concede' | 'stalemate'
  }
},
computed: {
  showRuleaDialog: {
    get: () => this.shownDialog === 'rules',
    set(val) {
      this.shownDialog = val ? 'rules' : '';
    }
  },
  // one computed prop like above per dialog
}

```
Then each of the dialogs could v-model its respective computed prop. I think we might also want to in-line the @ click handlers to assign to the variable in the html at that point rather than defining a specific method for opening each dialog, opening a dialog would only require assigning a value to one variable.

This might actually be more total lines of code so idk if it's better but the idea is that when opening a dialog we could just assign a value to the one piece of data like `this.shownDialog = 'rules';` and that would take care of turning all the others off because they all read from the same variable. @Haviles04 what do you think?

src/components/RulesDialog.vue
```diff
@@ -1,239 +1,216 @@
...
-            class="rules-button"
-            color="primary"
-            variant="outlined"
-            :size="buttonSize"
-          >
-            Rules
-          </v-btn>
-        </slot>
-      </span>
+  <base-dialog max-width="1250">
```
itsalaidbacklife, line 2: Does the max-width assignment work here? I'd be surprised if it did anything bc I don't think BaseDialog pays attention to it

src/components/StatsScoringDialog.vue
```diff
@@ -1,90 +1,90 @@
...
+    :id="`rankedScoring`"
+    max-width="1250"
+  >
     <!-- Activator -->
     <template #activator="{ props }">
       <slot name="button">
         <v-btn
           v-bind="props"
-          color="primary"
+          color="surface-2"
```
itsalaidbacklife, line 13: This is the activator button that's used in the StatsView, right? I don't think we need to update this color yet until we rework the stats view itself
```suggestion
          color="primary"
```

src/components/RulesDialog.vue
```diff
@@ -1,239 +1,216 @@
...
   props: {
     show: {
       type: Boolean,
       default: false,
     },
-    hideActivator: {
-      type: Boolean,
-      default: false,
-    },
   },
```
itsalaidbacklife, line 201: Is the `show` prop actually used? I don't see it in the base-dialog instance so I'm wondering if this actually opens atm?

itsalaidbacklife (commented):
Nice work. A few nitpicks and a possible rework of the variables in GameMenu

src/components/RulesDialog.vue
Haviles04: This makes it bigger, without it it's around 650px max

src/components/CreateGameDialog.vue
```diff
@@ -117,3 +125,13 @@ export default {
...
 </script>
+
+<style scoped>
+.dialog-card {
+  /* Stuck using important because vuetify applies it to these styles for cards */
+  color: rgba(var(--v-theme-surface-2)) !important;
+  border-radius: 12px !important;
+  opacity: 0.92;
+  border: 4px solid rgba(var(--v-theme-surface-2));
+}
```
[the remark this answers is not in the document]
Haviles04: Nope, removed this.

src/components/GameView/GameMenu.vue
Haviles04: I like this, I don't think we need the setter in the computed value, we can just do `return this.shownDialog === 'rules' 
and set shownDialog inline like `shownDialog = 'rules'` in the html

itsalaidbacklife: The only funky thing about this is that we're using `v-model` which is a 2-way binding that is literally shorthand for 

```
:modelValue="shownDialog"
@update:modelValue="shownDialog = $event"
```

Any of these components trying to close themselves by emitting an `update:modelValue` events, it will be ignored and throw an error because `shownDialog` is a computed prop without a setter

src/components/CreateGameDialog.vue
```diff
@@ -117,3 +125,5 @@ export default {
   },
 };
 </script>
+
+
```
itsalaidbacklife, line 129:
```suggestion
```

src/components/RulesDialog.vue
```diff
@@ -1,239 +1,205 @@
...
-      <v-card-actions class="d-flex justify-end">
-        <v-btn variant="outlined" color="primary" @click="close">Got It!</v-btn>
-      </v-card-actions>
-    </v-card>
-  </v-dialog>
+        <h4>Where can I play Cuttle?</h4>
+        <p>
+          You can play Cuttle anywhere you have a deck of cards and a friend! If you want to play online,
+          there is only one place:
+          <a href="https://www.cuttle.cards" target="_blank">www.cuttle.cards</a>
```
itsalaidbacklife, line 177: Can we style this link a lighter blue? It's hard to see
image

itsalaidbacklife: It also might be nice to set this dialog to `scrollable` so that you can see the actions at the bottom all the time ([docs](https://vuetifyjs.com/en/api/v-dialog/#props-scrollable))

itsalaidbacklife (commented):
Looking good. Last couple nitpicks

src/components/RulesDialog.vue
Haviles04: Nice, scrollable sounds good.... As for the link, I feel like that section is a little redundant, at that point you're already in the app playing a game and checking the rules.

itsalaidbacklife: Yeah fair point we can remove it

src/components/StatsScoringDialog.vue
```diff
@@ -1,96 +1,101 @@
 <template>
-  <v-dialog v-model="show">
+  <base-dialog
+    v-model="show"
+    title="Ranked Scoring"
+    :id="`rankedScoring`"
```
itsalaidbacklife, line 5: This is a simple string so I think we can just remove the v-bind and make it a literal
```suggestion
    id="rankedScoring"
```

src/components/GameView/GameMenu.vue
```diff
@@ -113,21 +107,29 @@ export default {
         ? 'The game will end and your opponent will win.'
         : 'If your opponent agrees, the game will end in a stalemate. The request will cancel if the opponent declines or either player makes a move';
     },
+    buttonSize() {
+      return this.$vuetify.display.mdAndDown ? 'small' : 'medium';
+    },
+    showRulesDialog:{
+      get(){
+      return this.shownDialog === 'rules';},
```
itsalaidbacklife, line 115:
```suggestion
        return this.shownDialog === 'rules';
      },
```

src/components/GameView/GameMenu.vue
```diff
@@ -113,21 +107,29 @@ export default {
...
+      set(val){
+      this.shownDialog = val ? 'rules' : '';
```
itsalaidbacklife, line 117:
```suggestion
        this.shownDialog = val ? 'rules' : '';
```

src/components/GameView/GameMenu.vue
```diff
@@ -113,21 +107,29 @@ export default {
         ? 'The game will end and your opponent will win.'
         : 'If your opponent agrees, the game will end in a stalemate. The request will cancel if the opponent declines or either player makes a move';
     },
+    buttonSize() {
+      return this.$vuetify.display.mdAndDown ? 'small' : 'medium';
+    },
+    showRulesDialog:{
+      get(){
+      return this.shownDialog === 'rules';},
+      set(val){
```
itsalaidbacklife, line 116:
```suggestion
      set(val) {
```

src/components/GameView/GameMenu.vue
```diff
@@ -113,21 +107,29 @@ export default {
         ? 'The game will end and your opponent will win.'
         : 'If your opponent agrees, the game will end in a stalemate. The request will cancel if the opponent declines or either player makes a move';
     },
+    buttonSize() {
+      return this.$vuetify.display.mdAndDown ? 'small' : 'medium';
+    },
+    showRulesDialog:{
+      get(){
```
itsalaidbacklife, line 114:
```suggestion
    showRulesDialog: {
      get() {
```

itsalaidbacklife (approved):

PR #481, opened 2023-08-05
Feat/usage stats backend

### LINKED ISSUE #468: [Feature]: Backend for #464 -- Site usage stats

### LINKED ISSUE #480: [Bug]: Usage stats calculation crashed server
### Bug Summary

After merging #469, production crashed when computing the site usage stats. This was reverted in PR #478, but the feature should be fixed and re-added.

### Detailed Description

Per Issue #468 , we want to calculate site usage stats: games played and unique players per week. The initial implementation in PR #469 crashed production, but we would still like this functionality. We should diagnose the problem, fix it, and re-add the stats calculations to the /stats endpoint

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Steps to Reproduce

1.
2.
3.

itsalaidbacklife:
## Issue number

Relevant issue number (e.g. #404): Resolves #480 and resolves #468

This adds backend calculations of the number of games played and number of unique players for each week. It also fixes the null pointer exception introduced in the previous implementation #469 by dealing with the arrays of player and game counts in a more type safe manner.

## Please check the following

- For New Features:


api/controllers/StatsController.js
```diff
@@ -96,40 +131,61 @@ function transformSeasonToDTO(season) {
...
+        }
+
+        const weekNum = dayjs(game.updatedAt).diff(relevantSeason.startTime, 'week');
+        if (!relevantSeason.gameCounts[weekNum]) {
+          relevantSeason.gameCounts[weekNum] = 0;
+        }
+        relevantSeason.gameCounts[weekNum]++;
+        if (!relevantSeason.uniquePlayersPerWeek[weekNum]) {
+          relevantSeason.uniquePlayersPerWeek[weekNum] = new Set();
+        }
```
itsalaidbacklife, line 185: @seriouslysean and @byarmis these nullish checks and re-initialization logic are the key to fixing the null pointer exception that crashed prod when I merged #469. Basically the Spades 2022 season started at 8:30pm on a Wednesday and ended at 6:00pm on a Wednesday 11 weeks later, which means the difference in weeks betwen start and end time as computed by `dayjs` is actually 10, even though it is only 2.5 hours short of being 11 weeks long. So any game in that 11th week would cause an NPE when trying to add the player ids to the set for that week on the lines below, which crashed the server. 

The `gameCounts` calculation was also wrong b/c it would get NaN and keep trying to increment and failing, but the `.add()` below is what actually crashed the server.

I considered trying to update the initialization of the seasons in `get-seasons-without-rankings` to somehow round to accommodate this, but I think it's actually more defensive and type safe to just make sure that any value in the array has a value before we update it here.

itsalaidbacklife: Actually maybe it does make more sense to just update the initialization in the helper. Per the `dayjs` [docs](https://day.js.org/docs/en/display/difference) we can make diff output a float instead of truncating, which would let us round up. That should solve the problem too, maybe more elegantly

tests/e2e/fixtures/statsFixtures.js
```diff
@@ -3,10 +3,9 @@ const dayjs = require('dayjs');
 const seasonFixtures = [
   {
     name: 'Clubs 2022',
-    // startTime: 1642642200,
-    startTime: dayjs('2022-01-19', 'yyyy-mm-dd').valueOf(),
-    // endTime: 1674178200,
-    endTime: dayjs('2022-02-18', 'yyyy-mm-dd').valueOf(),
+    startTime: dayjs('2022-01-19 20:30').valueOf(),
+    // 4 weeks long, minus 2.5 hours
+    endTime: dayjs('2022-02-16 18:00').valueOf(),
```
itsalaidbacklife, line 8: Updating the season start and end times to be just under 4 weeks long is how I got the test case to replicate the failure

tests/e2e/fixtures/statsFixtures.js
```diff
@@ -326,6 +321,179 @@ const seasonThreeMatches = [
...
+  {
+    name: 'Diamonds Week 1, Game 1',
+    status: false,
+    p0Ready: true,
+    p1Ready: true,
+    passes: 0,
+    turn: 12,
+    isRanked: false,
+    result: 0,
+    updatedAt: dayjs('2022-02-20').valueOf(),
```
itsalaidbacklife, line 433: This makes the game in the 4th (shortened) week of Clubs 2022 which is what causes the crash without updated handling/initalization logic

api/helpers/get-seasons-without-rankings.js
```diff
@@ -12,14 +12,25 @@ module.exports = {
...
           secondPlace: season.secondPlace?.username || null,
           thirdPlace: season.thirdPlace?.username || null,
           fourthPlace: season.fourthPlace?.username || null,
+          gameCounts: [],
+          uniquePlayersPerWeek: [],
         };
+        // initialize gameCounts and uniquePlayersPerWeek
+        const startTime = dayjs(season.startTime);
+        const endTime = dayjs(season.endTime);
+        const numWeeks = Math.ceil(endTime.diff(startTime, 'week', true));
```
itsalaidbacklife, line 28: This computation essentially caused a bug that crashed prod because seasons that were just shorter than a given number of weeks were being truncated here. So Spades 2022 is 11 weeks long except it ended at a time of day 2.5 hours earlier than it started, so without adding `true` to the end of the `.diff()` call, dayjs was truncating and computing it to be 10 weeks long, causing an NPE downstream when we try to compute stats for week 11 (since games did happen that week).

This way we compute the number of weeks in a way that rounds up instead of down, ensuring there's enough weeks represented for all the data associated with that season.

I'd initially implemented a fix in the `StatsController` when we relate a given game to a given week, but I think it's cleaner to do it here. 

Here's my initial writeup of the problem and solution: https://github.com/cuttle-cards/cuttle/pull/481/files#r1285100829

tests/e2e/specs/out-of-game/stats.spec.js
```diff
@@ -243,4 +252,26 @@ describe('Stats Page', () => {
...
+      expect(clubs2022.uniquePlayersPerWeek[1]).to.eq(4);
+
+      const diamonds2022 = seasons.find((season) => season.name === 'Diamonds 2022');
+      expect(diamonds2022).not.to.be.undefined;
+      // Week 1 stats
+      expect(diamonds2022.gameCounts[0]).to.eq(1);
+      expect(diamonds2022.uniquePlayersPerWeek[0]).to.eq(2);
+      // Week 2 stats
+      expect(diamonds2022.gameCounts[1]).to.eq(2);
+      expect(diamonds2022.uniquePlayersPerWeek[1]).to.eq(2);
```
itsalaidbacklife, line 274: Todo: add assertions for week 4 of clubs 2022

itsalaidbacklife (commented):
Add todo comment

tests/e2e/specs/out-of-game/stats.spec.js
itsalaidbacklife: done

api/controllers/StatsController.js
```diff
@@ -1,11 +1,40 @@
...
+      minIndex = currentIndex + 1;
+    // Target time is after current season
+    } else {
+      maxIndex = currentIndex - 1;
+    }
+    currentIndex = currentIndex = Math.floor((minIndex + maxIndex) / 2);
+  }
+  return null;
+}
+
```
Haviles04, line 37:
```suggestion
function getSeasonFromTimeStamp(sortedSeasons, timeStamp) {
  if (sortedSeasons.length <= 0) {
    return null;
  }
  return sortedSeasons.find(
    ({ startTime, endTime }) => startTime <= timeStamp && timeStamp <= endTime
  );
}

```

Haviles04: I'm not sure if there's a particular reason for using Binary search here, but I think the .find method could work.

tests/e2e/specs/out-of-game/stats.spec.js
```diff
@@ -243,4 +252,34 @@ describe('Stats Page', () => {
...
+      expect(clubs2022.uniquePlayersPerWeek[1]).to.eq(4);
+      // Week 3 stats
+      expect(clubs2022.gameCounts[2]).to.eq(0);
+      expect(clubs2022.uniquePlayersPerWeek[2]).to.eq(0);
+      // Week 3 stats
+      expect(clubs2022.gameCounts[3]).to.eq(1);
+      expect(clubs2022.uniquePlayersPerWeek[3]).to.eq(2);
+
+      // Diamonds 2022 stats
+      const diamonds2022 = seasons.find((season) => season.name === 'Diamonds 2022');
```
Haviles04, line 275:
```suggestion
      const diamonds2022 = seasons.find(({name}) => name === 'Diamonds 2022');
```

tests/e2e/specs/out-of-game/stats.spec.js
```diff
@@ -243,4 +252,34 @@ describe('Stats Page', () => {
       cy.get('[data-cy=season-select]').should('contain', seasonTwo.name);
     });
   });
+
+  it('Sends the counts of games played and unique players for each week of each season', () => {
+    cy.request('http://localhost:1337/stats').then(({body: seasons}) => {
+      // Clubs 2022 stats
+      const clubs2022 = seasons.find((season) => season.name === 'Clubs 2022');
```
Haviles04, line 259:
```suggestion
      const clubs2022 = seasons.find(({name}) => name === 'Clubs 2022');
```

Haviles04 (commented):
This looks great, I'm not sure what our thoughts are on destructuring are, but I suggested some spots we can use it.

api/controllers/StatsController.js
Haviles04: Okay, so It seems like binary search is faster than the .find method since it's already sorted. However I wonder if the performance outweighs the complexity/readability of the code

itsalaidbacklife: This was originally a `.find()` call, but I refactored it to binary search for performance reasons. `.find()` is a linear search through the entire array, which takes up to n loops, where n is the array length. And since we’re looping through the matches (and later the games), using .find takes m*n loops at max ie O(n^2). 

Binary search cuts down the number of iterations because the assumption of having sorted input means we can check the halfway point and cut the list in half every time, making the max number of checks log_(n) giving the algorithm O(nlog_2(n)) which should be faster

itsalaidbacklife: The performance vs readability trade off is a fair question and it may not be worth it to use binary search, but since this code loops through every game and every match in the site's history I think it makes sense to try to optimize it for speed

seriouslysean: This is the one time readability is less important :)

Issue #473, opened 2023-08-03
[Bug]: ScrapDialog has old dialog style (doesn't use `BaseDialog`)


itsalaidbacklife:
### Bug Summary

All the in game dialogs at this point are supposed to consume the BaseDialog component to give them a consistent style. The ScrapDialog fails to do this, and looks out of place as a result.

### Detailed Description

Generally our dialogs rendered with `BaseDialog` have a consistent style that has our chocolate and cream theme colors like so:
![image](https://github.com/cuttle-cards/cuttle/assets/6520704/acfb76d5-1c33-4b56-9ba6-00d2aa143e35)

But the ScrapDialog (for viewing the scrap) currently does _not_ use the BaseDialog and so it is white and black:
![image](https://github.com/cuttle-cards/cuttle/assets/6520704/306415bd-0bb8-481b-8342-dacc9a5bacb3)

We should update the ScrapDialog.vue component so that it instantiates a `basedialog` instead of the raw `v-dialog` in order to apply the appropriate styles

### Steps to Reproduce

1.
2.
3.

haaarsh619:
Hello there ! I'm a first-timer on open source and I would love to participate.  can I give it a shot ?

itsalaidbacklife:
@haaarsh619 Yes of course! If you have any questions about how to approach the issue, feel free to ask them in this thread. Or if you'd prefer help in real-time, you can join our [discord](https://discord.gg/6vQCFa6nvP) and use the #code channel to chat with me and the other devs

itsalaidbacklife:
Resolved by #496

PR #483, opened 2023-08-06
Fix issue: #473

haaarsh619:
Fix issue: #473


itsalaidbacklife:
This has failed ci due to linting and the tests, so it needs a bit more work. To fix the linting errors you can run `npm run lint:fix` and commit the resulting changes.

The test failure looks to be that the activator for the scrap dialog isn't being found (no element with `id="scrap-dialog"` on the page) which will take a bit more digging. Have you opened the app in a game and been able to open the scrap dialog? 

Can you comment with screenshots of the gameView and then the open dialog? I will take a deeper look tomorrow to see if I can help you troubleshoot it

src/components/GameView/ScrapDialog.vue
```diff
@@ -1,47 +1,36 @@
 <template>
-  <v-dialog v-model="show" max-width="650" scrollable>
+  <base-dialog v-model="show" title="Scrap Pile" id="scrap-pile-dialog">
```
itsalaidbacklife, line 2: The tests specifically look for an element with id="scrap-dialog", so the name change is causing the current test failure. It's simplest to just keep the id the same as it was (if we had to change it though, we would also need to change the tests to look for this new id)
```suggestion
  <base-dialog v-model="show" title="Scrap Pile" id="scrap-dialog">
```

itsalaidbacklife (commented):
This is close. The cause of the test failure is the id change, which is a quick fix

src/components/GameView/ScrapDialog.vue
```diff
@@ -1,47 +1,55 @@
...
-          />
-        </div>
-      </v-card-text>
-      <v-card-actions class="d-flex justify-end my-2">
-        <v-btn
-          color="primary"
-          variant="outlined"
-          data-cy="close-scrap-dialog-button"
+        <v-btn 
+          icon
```
itsalaidbacklife, line 12: While you are at it, per #446, let's fix the button style to remove the background circle
```suggestion
          icon
          variant="text"
```

itsalaidbacklife (approved):
Nice work. Good use of a slot for the title close button. I'm requesting one additional change to the close button style related to another issue, and the. Can you please merge from main to ensure this is up to date and then update the `version` field in `package.json` a minor version to 4.6.0? Then this is good to go

PR #486, opened 2023-08-07
Feat: Initial docker support

seriouslysean:
## Issue number

Relevant issue number: Resolves #

## Please check the following

- For New Features:

1. Install [Docker Desktop](https://www.docker.com/products/docker-desktop/)
1. Run `npm run docker:start`
2. Run `npm run start:devtools` in another tab
   - If you don't do this, you'll see port `8098` failures in the browser console when the devtools fail to connect
3. Go to http://localhost:8080 in your browser


seriouslysean:
@itsalaidbacklife I wonder if it makes sense to only dockerize a prod version that uses a proper db instead of `sails-disk` -- we might be able to use that for testing instead of the dev build.

package.json
```diff
@@ -57,6 +57,8 @@
   "main": "app.js",
   "private": true,
   "scripts": {
+    "docker:start": "docker-compose -f ./docker/docker-compose.yml up",
+    "docker:stop": "docker-compose -f ./docker/docker-compose.yml down",
```
itsalaidbacklife, line 61: @seriouslysean What do you think about naming these commands `"start:docker"` and `"stop:docker"` to be more consistent with the other boot commands?

itsalaidbacklife, line 61: We should update docs/setup-development.md to describe the use of docker. In particular, I think that we'll want to  to describe:
1. The fact that docker is optional for local development
2. That you need to install [docker compose](https://docs.docker.com/desktop/install/mac-install/)
3. The start and stop commands
4. How to open the app in your browser once the container is up (I can't currently figure out how to do this 😅

itsalaidbacklife: Also when do you need to rebuild the container? It looks like frontend changes hot reload, but backend changes won't (similar to local development now), so if you are working in docker and make a backend change you'll need to shut down the server with a keyboard interrupt or the stop command and then rerun `npm run docker:start`

vite.config.js
```diff
@@ -9,6 +9,10 @@ import vuetify from 'vite-plugin-vuetify';
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
+const PROXY_HOST = process.env.CUTTLE_DOCKERIZED === true
+  ? 'http://localhost:1337'
+  : 'http://cuttle-server:1337';
```
itsalaidbacklife, line 14: What is going on with this? My understanding is that it's using vite to proxy requests from localhost:8080 to either http://localhost:1337 or to http://cuttle-server:1337 if we're using docker?

I'm fuzzy on what this does and when it's used. I tried going to http://cuttle-server:1337 while the container was up and it didn't respond, so I'm not sure how the container name actually relates to the url if at all

package.json
seriouslysean: I was originally going to do that, but we are going to end up with more docker commands and I'd prefer to colocate them under the docker namespace.

vite.config.js
seriouslysean: This is a really interesting nuance, but you have to consider vite is running in the context of docker, and inside of docker the way the services talk to each other is via reference-- so `http://cuttle-server:1337/` actually works when the vue docker image is the one running it. This allows vite to proxy requests from vue (via docker) to sails (via docker) using docker networking.

package.json
seriouslysean: There's still an issue with this. I can't seem to get the client to talk to the server despite having the right ports forwarded, etc.

sternj:
Running the dockerized version, I'm seeing a lot of requests from `cuttle-client` to `localhost:1337` when I think it should be aiming at `cuttle-server:1337`

sternj:
```
cuttle-client             |
cuttle-client             | 5:28:05 PM [vite] http proxy error at /user/status:
cuttle-client             | Error: connect ECONNREFUSED 127.0.0.1:1337
cuttle-client             |     at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1300:16)
cuttle-client             | 5:31:24 PM [vite] http proxy error at /user/signup:
cuttle-client             | Error: connect ECONNREFUSED 127.0.0.1:1337
cuttle-client             |     at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1300:16)
cuttle-client             | 5:31:37 PM [vite] http proxy error at /user/signup:
cuttle-client             | Error: connect ECONNREFUSED 127.0.0.1:1337
cuttle-client             |     at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1300:16)
```

config/sockets.js
```diff
@@ -25,7 +25,7 @@ module.exports.sockets = {
    * > https://sailsjs.com/docs/reference/web-sockets/socket-client           *
    *                                                                          *
    ***************************************************************************/
-  // transports: [ 'websocket' ],
+  transports: ['websocket'],
```
seriouslysean, line 28: I don't know if we explicitly needed this but we should leave this on to make sure we always are using websockets specifically.

I'd really like some more context on this setting but it does make things more predictable.

docker/docker-compose.yml
```diff
@@ -0,0 +1,73 @@
+version: "3.8"
```
seriouslysean, line 1: New docker compose version T better support to level name property.

docker/docker-compose.yml
```diff
@@ -0,0 +1,73 @@
...
+
+# Naming the project allows us to use it when deleting containers so we can target only ones
+# generated by this project
+# https://docs.docker.com/compose/environment-variables/envvars/#compose_project_name
+name: cuttle
```
seriouslysean, line 6: This will prefix all of our containers for easier management via commands.

docker/docker-compose.yml
```diff
@@ -0,0 +1,73 @@
...
+
+# Share volumes between images, see https://stackoverflow.com/a/62873977
+#  - https://docs.docker.com/compose/compose-file/07-volumes/
+#  - https://docs.docker.com/compose/compose-file/compose-file-v3/#extension-fields
+x-app-volume:
```
seriouslysean, line 11: This allows a top level config for multiple services to share. In this case the cuttle src directory.

This may not be enough since it won't watch any root file changes like the vite config but I think that's acceptable for now.

docker/docker-compose.yml
```diff
@@ -0,0 +1,73 @@
...
+      # Allow all connections, no password required (not recommended)
+      # - POSTGRES_HOST_AUTH_METHOD=trust
+    volumes:
+      - ./data/pgdata:/var/lib/postgresql/data
+    restart: always
+
+  server:
+    container_name: server
+    # Required to be able to hit http://server:1337 from the browser via the client container
+    hostname: server
```
seriouslysean, line 36: This took a lot of trial and error but it's required to access the server via localhost from the frontend client in the browser.

docker/docker-compose.yml
```diff
@@ -0,0 +1,73 @@
...
+      - "1337:1337"
+    volumes: *app-volume
+    depends_on:
+      - database
+
+  client:
+    container_name: client
+    environment:
+      - CUTTLE_DOCKERIZED=true
+      - VITE_API_URL=http://localhost:1337
```
seriouslysean, line 61: May need to set this to http://server:1337

docker/docker-compose.yml
```diff
@@ -0,0 +1,73 @@
...
+      - VITE_BUILD_OUTPUT=assets
+    # No env file because docker is different :)
+    # env_file:
+    build:
+      context: ..
+      dockerfile: ./docker/.Dockerfile
+    command: npm run start:client
+    ports:
+      - "8080:8080"
```
seriouslysean, line 70: Allows us to hit the normal localhost:8080 port in our browser.

docker/docker-compose.yml
```diff
@@ -0,0 +1,73 @@
...
+    volumes: *app-volume
+    depends_on:
+      - server
```
seriouslysean, line 73: Chaining these ensures they spin up in a logical order; db, server, client

package.json
```diff
@@ -59,6 +59,15 @@
   "main": "app.js",
   "private": true,
   "scripts": {
+    "docker:build": "docker compose -f ./docker/docker-compose.yml build --no-cache",
```
seriouslysean, line 62: Build the images and containers

package.json
```diff
@@ -59,6 +59,15 @@
...
+    "docker:clean:containers": "docker rm $(docker ps -qa --no-trunc --filter \"status=exited\") || true",
+    "docker:clean:images": "docker rmi $(docker images --filter \"dangling=true\" -q --no-trunc) || true",
+    "docker:clean:volumes": "docker volume rm $(docker volume ls -qf dangling=true) || true",
+    "docker:clean": "npm run docker:clean:containers;npm run docker:clean:images;npm run docker:clean:volumes",
+    "docker:down": "docker stop $(docker ps -q)",
+    "docker:nuke": "docker rmi $(docker images -f=\"reference=cuttle*\" -q --no-trunc) || true; npm run docker:clean",
```
seriouslysean, line 68: Utilities to help with as lot of random docker stuff. Removing dangling containers, images and volumes, shutting down docker and removing all Cuttle specific images outright in case of any weird problems.

src/components/Global/MarkdownContent.vue
```diff
@@ -1,6 +1,6 @@
 <script setup>
 import { parse } from 'marked';
-import { computed, defineProps } from 'vue';
```
seriouslysean, line 3: This is built in at compile time and no longer required.

src/plugins/sails.js
```diff
@@ -38,9 +38,10 @@ export const reconnectSockets = () => {
 
 // Configure socket connection url for dev environments
 if (!import.meta.env.PROD) {
-  io.sails.url = import.meta.env.VITE_API_URL || 'localhost:1337';
+  io.sails.url = import.meta.env.VITE_API_URL || 'http://localhost:1337';
```
seriouslysean, line 41: Aligning our usage and structure everywhere we assign these values.

vite.config.js
```diff
@@ -2,55 +2,77 @@ import { dirname, resolve } from 'path';
...
-    proxy: {
-      '/game': {
-        target: 'http://localhost:1337',
-        changeOrigin: true,
-      },
-      // Required for the health response to work on the client
-      '/health': {
-        target: 'http://localhost:1337',
-        changeOrigin: true,
+export default defineConfig(({ mode }) => {
```
seriouslysean, line 13: Converted this to a function so we can use envs to change the data.

vite.config.js
```diff
@@ -2,55 +2,77 @@ import { dirname, resolve } from 'path';
...
+  // Load env file based on `mode` in the current working directory.
+  // Set the third parameter to '' to load all env regardless of the `VITE_` prefix.
+  // See https://vitejs.dev/config/#using-environment-variables-in-config
+  const env = loadEnv(mode, process.cwd(), '');
+  let HOST_SERVER_URL = 'http://localhost:1337';
+  if (env.CUTTLE_DOCKERIZED === 'true') {
```
seriouslysean, line 19: If we're dockerized, we need to change a few values so the client proxies to the server properly.

vite.config.js
```diff
@@ -2,55 +2,77 @@ import { dirname, resolve } from 'path';
...
+        },
       },
-      '/test': {
-        target: 'http://localhost:1337',
-        changeOrigin: true,
+      // Watching doesn't work on windows, so we need to use polling -- this does lead to high CPU
+      // usage though, which is a bit of a bummer. Should probably make this conditional at some
+      // point, see https://v3.vitejs.dev/config/server-options.html#server-watch
+      watch: {
+        usePolling: true,
```
seriouslysean, line 68: I would like to not need to keep this but it's useful for Linux (read: docker), and will fix the issue of watchers not working in Windows under WSL.

Maybe we can fork it by OS?

seriouslysean (commented):
@itsalaidbacklife it's ALIVE! I got everything working, but I did turn off Postgres for now since I switched back to using dev.

We probably want a docker specific sails env, a docker fork in the dev env or to move back to using staging for the db adapter.

config/env/staging.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
    ***************************************************************************/
 
   models: {
-    migrate: 'safe',
+    migrate: 'alter',
```
seriouslysean, line 26: @itsalaidbacklife this was key to getting postgres working on docker, but I'm not entirely sure why. I guess it allows it to alter an already established db but we'll need to chat through this.

itsalaidbacklife: It's a setting that tells sails what to do when booting to make the database match what's expected. There are three options:
`safe` - preserve the database as-is and crash if data is misshapen
`alter` - leave existing rows but add necessary tables and columns where possible 
`drop` - delete everything and create all the tables from scratch 

Here are the [docs](https://sailsjs.com/documentation/concepts/models-and-orm/model-settings#migrate)

docker/docker-compose.yml
itsalaidbacklife: It sounds like it works without that change?

seriouslysean: I think it's good but want to double check one more time to make sure it's fully compatible.

docker/.Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM node:18.12.1
+
+WORKDIR /cuttle/
+
+ENV NPM_CONFIG_LOGLEVEL=notice
+
+RUN apt-get update
```
ollien, line 7: It's good practice, btw, to collapse these RUN lines where possible; this saves space by reducing the number of layers

```Dockerfile
RUN apt-get-update && \
         apt-get clean
```

docker/.Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+RUN apt-get clean
+
+# Install app dependencies
+# A wildcard is used to ensure both package.json AND package-lock.json are copied
+# where available (npm@5+)
+COPY package*.json .
```
ollien, line 13: You copy this in the next line. Though, I will say, IME I like to do something like this

```Dockerfile
COPY package*.json
RUN npm ci
COPY . .
```
This allows you to rebuild the image without rebuilding depndencies if only the code changes

docker/.Dockerfile
seriouslysean: @ollien yeah totally, I also wanted to make sure we're running on npm 9.8 if possible to stop our package lock from jumping between versions 2 and 3.

I'll make that change here but I still want another pass eventually to pull in more files so it makes a new build when things change in root files, and to reduce the number of layers as you said.

Thanks for the feedback!

docker/docker-compose.yml
```diff
@@ -0,0 +1,73 @@
...
+    container_name: server
+    # Required to be able to hit http://server:1337 from the browser via the client container
+    hostname: server
+    environment:
+      - CUTTLE_DOCKERIZED=true
+      - DATABASE_URL=postgresql://cuttlesworth:p4ssw0rd!@database:5432/cuttle
+      # Can set this to staging for the postgres database to work in docker but may need to modify
+      # the onlyAllowOrigins setting in `config/env/staging.js` file
+      - NODE_ENV=staging
+      - VITE_API_URL=http://localhost:1337
```
seriouslysean: Only Cypress needs it for e2e purposes, so I'm removing it from server and client.

docker/.Dockerfile
seriouslysean: Great point, thanks 👍🏼

ollien: Yep - so the idea is that the layer cache is only invalidated if the filesystem is modified as a result of the COPY operation; since `node_modules` is `.docker-ignore`'d, it won't get overwritten in the container.

https://docs.docker.com/build/cache/

Of course, do check my work on this (an easy test is to just change the code (but not `package.json` or `package-lock.json` and re-run `docker build`, and then make sure that the build doesn't re-run the `npm ci` step). That said, I've done this in a number of repos to great success :)

docker/docker-compose.yml
seriouslysean: This was a lie, we need it for the sails socket.io instance url and server:1337 won't work... BUTTTTTT, Vite still needs the server:1337 url for the proxy since that runs on the server side from the client (I know that's wacky, but I'll document it so it's more clear.

docker/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM node:18.12.1
+
+WORKDIR /cuttle/
+
+ENV NPM_CONFIG_LOGLEVEL=notice
+
+RUN apt-get update && \
```
byarmis, line 7: Not sure this step is needed since you don't install or update anything

docker/docker-compose.yml
```diff
@@ -0,0 +1,73 @@
...
+    # env_file:
+    build:
+      context: ..
+      dockerfile: ./docker/Dockerfile
+    command: npm run start:server
+    ports:
+      - "1337:1337"
+    volumes: *app-volume
+    depends_on:
+      - database
```
byarmis, line 55: As a heads up, when the Docker container considers itself ready and when the database is ready to accept connections are two different things.  [Random SO post](https://stackoverflow.com/questions/35069027/docker-wait-for-postgresql-to-be-running) with a potential solution.

It might not cause any issues for people but might make tests behave strangely

docs/docker/docker.md
```diff
@@ -0,0 +1,47 @@
...
+  - [Postgres Container](#postgres-container)
+  - [Diagram](#diagram)
+
+
+## Local Development
+
+1. Install [Docker Desktop](https://www.docker.com/products/docker-desktop/)
+1. Run `npm run docker:start`
+2. Run `npm run start:devtools` in another tab
+   - If you don't do this, you'll see port `8098` failures in the console
```
itsalaidbacklife, line 16:
```suggestion
   - If you don't do this, you'll see port `8098` failures in the browser console when the devtools fail to connect
```

docs/docker/docker.md
```diff
@@ -0,0 +1,47 @@
...
+
+- Runs Sails.js
+  - Exposed via http://localhost:1337
+- Hostname of `server:1337`
+- Connects to Postgres via port `5432`
+
+### Postgres Container
+
+- Runs Postgres
+  - Exposed via http://localhost:1337
```
itsalaidbacklife, line 40: This runs on 5432 I think
```suggestion
  - Exposed via http://localhost:5432
```

docs/setup-development.md
```diff
@@ -1,4 +1,10 @@
-## Project setup and development
+# Project Setup and Development
+
+## Docker Development
+
+See [Docker](./docker/docker.md) Instructions.
```
itsalaidbacklife, line 5: Maybe add that it's optional?
```suggestion
Cuttle can optionally be run with docker for maximal consistency. See [Docker](./docker/docker.md) Instructions.
```

docs/docker/docker.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Local Development
+
+1. Install [Docker Desktop](https://www.docker.com/products/docker-desktop/)
+1. Run `npm run docker:start`
+2. Run `npm run start:devtools` in another tab
+   - If you don't do this, you'll see port `8098` failures in the console
+3. Go to http://localhost:8080 in your browser
+
+You can connect to the db via `postgresql://cuttlesworth:p4ssw0rd!@database:5432/cuttle` locally, and use something like [DBeaver](https://dbeaver.io/).
```
itsalaidbacklife, line 19:
```suggestion
You can connect to the db via `postgresql://cuttlesworth:p4ssw0rd!@database:5432/cuttle` locally, and query or manage migrations with a database administration tool like [DBeaver](https://dbeaver.io/).
```

package.json
itsalaidbacklife, line 68: How is this one different from `clean`?

itsalaidbacklife (approved):
Stellar work. This is a huge win. I've got a couple questions and some nitpicks but the essential pieces are in place

docker/docker-compose.yml
seriouslysean: We actually need this as localhost due to the sails plugin using it on the client side. We may need to fork where we use this eventually but for now it works as written.

package.json
seriouslysean: This has been resolved as of the latest changes.

vite.config.js
```diff
@@ -2,55 +2,77 @@ import { dirname, resolve } from 'path';
...
+  // Load env file based on `mode` in the current working directory.
+  // Set the third parameter to '' to load all env regardless of the `VITE_` prefix.
+  // See https://vitejs.dev/config/#using-environment-variables-in-config
+  const env = loadEnv(mode, process.cwd(), '');
+  let HOST_SERVER_URL = 'http://localhost:1337';
+  if (env.CUTTLE_DOCKERIZED === 'true') {
+    HOST_SERVER_URL = 'http://server:1337';
+    console.log('');
+    console.log(`Running Cuttle in DOCKER, setting server url to "${HOST_SERVER_URL}"`);
+    console.log('');
```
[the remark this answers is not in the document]
itsalaidbacklife: Yeah imo newlines are simpler

package.json
itsalaidbacklife: i.e. even kill things that are actively running?

docker/docker-compose.yml
itsalaidbacklife: I need a deeper look to wrap my head around this, but that's not blocking atm

package.json
seriouslysean: Nah, you'd still need to stop them as far as I understand. We could maybe preface this with a docker down but I think we shouldn't do that for safety as a final barrier lol.

docker/docker-compose.yml
seriouslysean: It's the vite proxy, it runs via node which is server side to do the proxy operations, thus you can use the internal hostname of the docker container in how they talk to one another. If it's a client (browser) side operation, you need to use the exposed localhost urls.

PR #487, opened 2023-08-07
Fix: Session test failures (WIP)

seriouslysean:
## Issue number

Relevant issue number: Resolves #

## Please check the following

- For New Features:

### DIFF
--- a/config/policies.js
+++ b/config/policies.js
@@ -17,8 +17,8 @@
  */
 
 const conditionalIsLoggedIn = process.env.CUTTLE_ENV === 'test'
-  ? ''
-  : 'isLoggedIn';
+  ? []
+  : ['isLoggedIn'];
 
 module.exports.policies = {
   statsController: {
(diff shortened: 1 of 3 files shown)


config/policies.js
```diff
@@ -33,44 +37,44 @@ module.exports.policies = {
   },
 
   'game/reconnect': ['isLoggedIn', 'isInGame'],
-  'game/create': ['isLoggedIn', 'hasGameName'],
+  'game/create': [conditionalIsLoggedIn, 'hasGameName'],
```
seriouslysean, line 40: Could possibly try a different adjustment by adding an opponent specific routes for mock testing that uses the normal route and doesn't have a logged in check.

seriouslysean:
@itsalaidbacklife I think this bug has to do with managing multiple sessions from a singular backend.

https://docs.cypress.io/guides/references/trade-offs

Cypress doesn’t support multiple browsers for our multiple user case, which makes me think we need to mock the cases rather than try and reproduce them in game.

This is why bypassing the auth check in tests only had this pass as far as I can tell. It would be worth going back to verify the tests that failed always had to do with multiple users being registered and actioned on.

Consider our most egregious example, the spectator tests:
- Register the player and log them in
- Express sets the session with the player data
- Register an opponent and log them in
- Express replaces the session data with the opponent 
- Register a spectator and it does it again

The issue then seems to be that the original user isn’t authed properly and gets a 403 when we try and take an action. It would be interesting to try and verify the session ids locally.

That said, there’s a few things I think we can try as ignoring the auth status seems pretty bad, even if it does work. 😬 
- mock non player signups and set them via api calls only, to avoid the socket collisions — we’d still need to blast status to the player somehow
    - This seems like the beat option in terms of standards and stability
- Allow server to manage multiple users in a singular browser session — May work but goes against how the app is actually used
- Add query params to the non player requests to allow them to have slightly different logic (not replace session, etc)
- Bypass auth checks for non player users (will still need a way to not override the express session, maybe some sort of non player middleware in server routes)

seriouslysean:
@Haviles04 🤘🏼 https://github.com/cuttle-cards/cuttle/pull/490

itsalaidbacklife:
Closing for now as this may have been resolved by dependency updates

PR #492, opened 2023-08-08
feat: added scrolling to login

### LINKED ISSUE #485: [Feature]: Add landing page video + background image to login page and auto-scroll to the login form when mounting `LoginView` in the Login mode.
### Feature Summary

Currently the Signup and Login pages are very similar except the signup page has an additional screen-height worth of content at the top to showcase the board background and an introductory video. The difference between the pages can be jarring when navigating between signup and login. To remedy this we should make them have the same content and use programatic scrolling to achieve the effect of directing returning users immediately into the login form.

### Detailed Description

Signup and Login are both handled by the `LoginView` page, which conditionally renders this top section only when users are signing in, under the assumption that returning users do not wish to see introductory content every time:
image

However this causes jumpy, inconsistent navigation when switching between signup and login:

https://github.com/cuttle-cards/cuttle/assets/6520704/3bc20290-072d-4b34-b9a7-e7107e5097d4

In order to make the navigation smoother and the layout more consistent between these versions of the same page, we should remove the conditional rendering of the top section so that this page  _always_ displays the top introductory section (with the game board and video) above the form for signup/login.

In order to enable returning users to directly log in without needing to manually scroll past this introductory section, we should then make it so that the LoginView page programatically scrolls to the login form when the component mounts when you are visiting `/#/login` but not when you are visiting `/#/signup`. This way new users still see the intro content front and center, and returning users can easily ignore it, but the page layout will be more consistent between signup and login, making navigation between them smoother and more intuitive.

Haviles04:
## Issue number

- Resolves #485

## Please check the following

- For New Features:

Removed the conditional rendering and focused login on mounted if the router is /login

### DIFF
--- a/src/views/LoginView.vue
+++ b/src/views/LoginView.vue
@@ -3,17 +3,13 @@
     <!--//////////////////////////
      /////Welcome Section /////
     //////////////////////////-->
-    <section v-if="!isLoggingIn">
+    <section>
       <v-container fluid id="welcome-container" class="welcomeContainer">
         <nav class="d-flex justify-space-between align-center mb-2">
-          <img class="cardLogo" src="/img/loginView/logo-cards-behind.svg">
-          <v-btn
-            variant="text"
-            class="text-h6"
-            @click ="$refs.usernameInput.focus()"
-          >
+          <img class="cardLogo" src="/img/loginView/logo-cards-behind.svg" />
+          <v-btn variant="text" class="text-h6" @click="scrollAndFocusLogin()">
             {{ t('global.signup') }}
-            <v-icon icon="mdi-account-circle" color="white" class="ml-2"/>
+            <v-icon icon="mdi-account-circle" color="white" class="ml-2" />
           </v-btn>
         </nav>
 
@@ -24,7 +20,7 @@
             <base-video source="https://www.youtube.com/embed/qOqkNbhMdsI" />
             <v-btn
               class="rulesBtn w-100 mt-8"
-              size='x-large'
+              size="x-large"
               to="/rules"
               color="newPrimary"
               data-cy="rules-link"
@@ -40,13 +36,13 @@
      /////Login Section ///////
     //////////////////////////-->
 
-    <section>
-      <v-container id="login-container" class="container">
+    <section ref="loginContainer">
+      <v-container id="login-container" class="container"   >
         <img id="logo" alt="Cuttle logo" src="/img/logo.png" />
         <v-row>
           <!-- Left side form -->
-          <v-col id="username-login-form" :cols="$vuetify.display.mdAndUp ? 8 : 12" >
-            <p class="text-h4 text-center font-weight-medium mb-10 ">
+          <v-col id="username-login-form" :cols="$vuetify.display.mdAndUp ? 8 : 12">
+            <p class="text-h4 text-center font-weight-medium mb-10">
               {{ formHeaderText }}
             </p>
 
@@ -86,14 +82,13 @@
 
               <div
                 id="login-button-container"
-                class="d-flex flex-column flex-md-row
-                  justify-space-between align-center flex-wrap"
+                class="d-flex flex-column flex-md-row justify-space-between align-center flex-wrap"
               >
                 <v-btn
                   class="px-16"
                   :loading="loading"
                   :disabled="!isFormValid"
-                  :color=" isFormValid ? 'newPrimary' : 'disabled'"
+                  :color="isFormValid ? 'newPrimary' : 'disabled'"
                   type="submit"
                   size="x-large"
                   text-color="white"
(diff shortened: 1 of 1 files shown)


src/views/LoginView.vue
```diff
@@ -269,6 +269,10 @@ export default {
     scrollAndFocusLogin() {
       this.$refs.loginContainer.scrollIntoView();
       this.$refs.usernameInput.focus();
+    },
+    directToSignup() {
+      this.$router.push({ name: ROUTE_NAME_SIGNUP });
+      this.scrollAndFocusLogin();
```
Haviles04: So I was unsure if I should do two different functions or if one would be better, 

```
    scrollAndFocusLogin(signup) {
      if(signup){
      this.$router.push({ name: ROUTE_NAME_SIGNUP });
      }
      this.$refs.loginContainer.scrollIntoView();
      this.$refs.usernameInput.focus();
    }, 
```

src/views/LoginView.vue
```diff
@@ -268,69 +260,80 @@ export default {
...
       return val.length >= 8 || 'Password must contain at least eight characters';
     },
+    scrollAndFocusLogin() {
+      this.$refs.loginContainer.scrollIntoView();
+      this.$refs.usernameInput.focus();
+    },
+    directToSignup() {
+      this.$router.push({ name: ROUTE_NAME_SIGNUP });
+      this.scrollAndFocusLogin();
+    }
```
Haviles04, line 276: I feel like we could make this one function but I'm unsure if a boolean or string value is better.

```     scrollAndFocusLogin(signup) {
      if (signup) {
        this.$router.push({ name: ROUTE_NAME_SIGNUP });
      }
      this.$refs.loginContainer.scrollIntoView();
      this.$refs.usernameInput.focus();
    }, ```

File: src/views/LoginView.vue [hunk 3]
```diff
@@ -3,17 +3,13 @@
...
       <v-container fluid id="welcome-container" class="welcomeContainer">
         <nav class="d-flex justify-space-between align-center mb-2">
-          <img class="cardLogo" src="/img/loginView/logo-cards-behind.svg">
-          <v-btn
-            variant="text"
-            class="text-h6"
-            @click ="$refs.usernameInput.focus()"
-          >
+          <img class="cardLogo" src="/img/loginView/logo-cards-behind.svg" />
+          <v-btn variant="text" class="text-h6" @click="directToSignup">
```
Haviles04, line 10:
```suggestion
          <v-btn variant="text" class="text-h6" @click="scrollAndFocusLogin(true)">
```

src/views/LoginView.vue
```diff
@@ -40,13 +36,13 @@
      /////Login Section ///////
     //////////////////////////-->
 
-    <section>
-      <v-container id="login-container" class="container">
+    <section ref="loginContainer">
+      <v-container id="login-container" class="container"   >
```
itsalaidbacklife, line 40: White space 
```suggestion
      <v-container id="login-container" class="container">
```

src/views/LoginView.vue
itsalaidbacklife: Actually I think the button in the top right of the page should jump you to the form without changing the url i.e. if you're on the login page it should keep you on login and if you're on the signup page it should keep you on signup.

In that case I think we don't need the conditional logic here at all, just scroll the form into view and focus the username

src/views/LoginView.vue
```diff
@@ -268,69 +260,80 @@ export default {
...
+    },
+    directToSignup() {
+      this.$router.push({ name: ROUTE_NAME_SIGNUP });
+      this.scrollAndFocusLogin();
+    }
+  },
+ mounted() {
+    if (this.isLoggingIn) {
+      this.scrollAndFocusLogin();
+    }
```
itsalaidbacklife, line 281: How is the UX of this auto scroll on page mount? Can you share a video?

src/views/LoginView.vue
itsalaidbacklife: Instead of changing the page to signup let's make this button say login or signup depending on the route and just scroll and focus without changing the page

itsalaidbacklife (changes requested):
Nice work. I'm requesting a functional change to the Sign Up button at the top of the page that I think will simplify things. Can you also bump package.json a minor version?

src/views/LoginView.vue
Haviles04: Posted in the PR description

src/views/LoginView.vue
```diff
@@ -203,19 +197,17 @@ export default {
...
     isLoggingIn() {
       return this.$route.name === ROUTE_NAME_LOGIN;
     },
     formHeaderText() {
-      return this.isLoggingIn
-        ? this.t('login.formHeaderLogin')
-        : this.t('login.formHeaderSignup');
+      return this.isLoggingIn ? this.t('login.formHeaderLogin') : this.t('login.formHeaderSignup');
     },
     buttonText() {
```
Haviles04, line 212: I think since now this is being used in two places a different name might be better.. maybe loginOrSignupText

Haviles04: it just occurred to me that the second use of this is also a button... disregard...LOL

PR #493, opened 2023-08-08
[DexEx] #374 reorganize status and results

### LINKED ISSUE #374: [DevEx]: Reorganize `status` and `result` database columns

Haviles04:
## Issue number

- Resolves #374 

## Please check the following

- For New Features:

This is a rework of the database architecture.

- Added winner column with user table as foreign key
- Renamed current status to "old_status" and created new status column (integer)
     * 1 - CREATED (game has been created but has not started)
     * 2 - STARTED (game has been started but has not yet finished)
     * 3 - FINISHED (game has been completed)
     * 4 - ARCHIVED (game has been removed from the list without being started)
- Wrote and tested database migration posted below
- Fixed associated test that expected status to be a boolean, and added expections of a winner when new status === completed

POSTGRES MIGRATION
```  
 ALTER TABLE game RENAME status TO old_status;
 ALTER TABLE game ADD status INTEGER;
 ALTER TABLE game ADD winner INTEGER REFERENCES "user" ON DELETE CASCADE;
 UPDATE game SET status = 3 WHERE result =1 OR result =0 OR  result =2;
 UPDATE game SET status = 2 WHERE result =-1;
 UPDATE game SET winner = game.p0 WHERE result = 0;
 UPDATE game SET winner = game.p1 WHERE result = 1;
 UPDATE game SET winner = null WHERE result = 2; 
 UPDATE game SET status = 4 WHERE log = '{}' and result = -1;
 UPDATE game SET status = 4 WHERE "updatedAt" < EXTRACT(epoch from current_date - interval '1 DAY') * 1000 AND result = -1;
```


api/controllers/game/subscribe.js
```diff
@@ -35,7 +35,7 @@ module.exports = function (req, res) {
         } else {
           pNum = (game.players[0].pNum + 1) % 2;
           await Game.updateOne({ id: game.id }).set({
-            status: false,
+            status: gameService.GameStatus.CREATED,
```
Haviles04, line 38: Is this needed considering this should be set when the game created and then not changed until both players are ready? Was it here as a safety?

Haviles04:
@itsalaidbacklife Should we add some sort of incomplete enum like we used to?

api/controllers/game/stalemate.js
```diff
@@ -30,15 +30,16 @@ module.exports = async function (req, res) {
...
       };
       updatePromises.push(gameService.clearGame({ userId }));
     }
 
     updatePromises.push(Game.updateOne({ id: gameId }).set(gameUpdates));
     await Promise.all(updatePromises);
 
-    if (victory.gameOver && gameUpdates.result === gameService.GameResult.STALEMATE && game.isRanked) {
+    if (victory.gameOver && gameUpdates.status === gameService.GameStatus.FINISHED && game.isRanked) {
       victory.currentMatch = await sails.helpers.addGameToMatch(game);
```
Haviles04, line 43: Not to self, do we need to update victory here as well

Haviles04: We do not, because `addGameToMatch` sets its own winner to it's own model

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -555,10 +558,13 @@ describe('Creating And Updating Ranked Matches', () => {
...
-        cy.expect(results[0]).to.eq(0);
-        cy.expect(results[1]).to.eq(0);
-        cy.expect(results[2]).to.eq(2);
+        const results = game.currentMatch.games.map(({status, winner}) => ({status, winner}));
+        cy.expect(results[0].status).to.eq(3);
+        cy.expect(results[0].winner).to.eq(this.playerOneId);
+        cy.expect(results[1].status).to.eq(3);
+        cy.expect(results[1].winner).to.eq(this.playerTwoId);
+        cy.expect(results[2].status).to.eq(3);
+        cy.expect(results[2].winner).to.eq(null);
```
Haviles04, line 567: I think there is something funky here `cy.expect(results[1]).to.eq(0);` should mean p0 won the this game, but `   cy.expect(results[1].winner).to.eq(this.playerTwoId);` passes

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -519,9 +521,10 @@ describe('Creating And Updating Ranked Matches', () => {
     cy.window()
       .its('cuttle.app.config.globalProperties.$store.state.game')
       .then((game) => {
-        const results = game.currentMatch.games.map((g) => g.result);
-        cy.expect(results[0]).to.eq(0);
-        cy.expect(results[1]).to.eq(0);
+        const results = game.currentMatch.games.map(({status, winner}) => ({status, winner}));
+        cy.expect(results[0].status).to.eq(3);
+        cy.expect(results[0].winner).to.eq(this.playerOneId);
+        cy.expect(results[1].status).to.eq(3);
```
Haviles04, line 527: we should add another `cy.expect(results[0].winner).to.eq(this.playerOneId);`

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -472,7 +472,7 @@ describe('Creating And Updating Ranked Matches', () => {
     cy.loginPlayer(playerOne);
     cy.setupGameAsP0(true, true);
   });
-  it('Creates a match when two players play a ranked game for the first time this week', function () {
+  it.only('Creates a match when two players play a ranked game for the first time this week', function () {
```
Haviles04, line 475: remover this `.only` when finished

src/store/modules/gameList.js
Haviles04, line 1: I felt like importing gameService here felt a little weird so I decided to just use the correlating numbers.

tests/e2e/fixtures/statsFixtures.js
Haviles04, line 1: I think it might be helpful to change this up a little, this way not every game is a p0 win.

tests/e2e/specs/in-game/gameOver.spec.js
Haviles04, line 1: I think some of these expectations can be refactored a little bit to DRY it up. Maybe a different PR though.

Haviles04 (commented):
This one was alot of files! A ton of fun, and my head would be spinning without cypress telling me each of the spots that I missed refactoring something.

tests/e2e/specs/in-game/gameOver.spec.js
Haviles04: Maybe we just dont need to check the status every time

api/helpers/add-game-to-match.js
```diff
@@ -35,13 +35,13 @@ module.exports = {
...
           if (priorGame.p0 === relevantMatch.player1) {
             numPlayer1Wins++;
           } else {
             numPlayer2Wins++;
           }
-        } else if (priorGame.result === 1) {
+        } else if (priorGame.winner === priorGame.p1) {
           if (priorGame.p1 === relevantMatch.player1) {
             numPlayer1Wins++;
           } else {
```
itsalaidbacklife, line 47: Now that game.winner replaces game.result I think we can significantly refactor this whole block. The wonky conditional logic around p0 and p1 was all about cross referencing which player was which and who won by comparing the result (indicating whether p0 or p1 won) against which player was p0 or p1 in the game, against which player is p1 vs p2 in the match. 

This can be much much simpler now. The match wants to know how many games each player has one. So just loop through the games and compare the winner attribute to the match's player1 and player2

itsalaidbacklife: Writing the logic here the first time is what inspired me to write this issue up because the use of result (as opposed to the winner's id) made this so convoluted that I wanted to rework the database to clean up this and related logic

api/helpers/is-game-full.js
itsalaidbacklife, line 1: Neat helper. Where is it used? Is the idea to identify games that are available to join? I'm wondering whether we need to specify the other statuses besides started and whether we need to care about the log and lastEvent at all

api/hooks/customGameHook/index.js
```diff
@@ -27,7 +28,7 @@ module.exports = function gameHook() {
     },
     findOpenGames: function () {
       return new Promise(function (resolve, reject) {
-        Game.find({ status: true })
+        Game.find({ status: gameService.GameStatus.CREATED })
```
itsalaidbacklife, line 31: I think we want this to also filter out the games that have 2+ players

src/components/GameListItem.vue
```diff
@@ -20,7 +20,7 @@
         <v-btn
           v-if="!isSpectatable"
           v-bind="buttonAttrs"
-          :disabled="!status"
+          :disabled="status >= 2"
```
itsalaidbacklife, line 23: I think it's really anything != 1

src/store/modules/gameList.js
itsalaidbacklife: The ideal I think would be to define the enum in a central place that the gameService imports that the frontend could also import without using the gameService itself (since gameService is for the backend).

That file would have to use commonjs/es5/require() syntax to be usable in the backend but the frontend should be able to use it still

src/views/HomeView.vue
```diff
@@ -172,7 +172,7 @@ export default {
       this.showSnackBar = false;
     },
     handleSubscribeError(gameId, message) {
-      this.$store.commit('updateGameStatus', {id: gameId, newStatus: false});
+      this.$store.commit('updateGameStatus', {id: gameId, newStatus: 2});
```
itsalaidbacklife, line 175: If we can get the enum into the frontend this will be another place to use it

tests/e2e/specs/in-game/gameOver.spec.js
itsalaidbacklife: Maybe a helper function to make common assertions with params for just the differing parts?

itsalaidbacklife (commented):
Great work! This is coming together nicely. Mostly small tweak requests, with a heavier look at add-game-to-match and the use of status in the frontend

api/helpers/is-game-full.js
```diff
@@ -15,7 +15,8 @@ module.exports = {
 
   fn: ({ game }, exits) => {
 
-    const gameIsFull = !game.status || 
+    const gameIsFull =
```
seriouslysean, line 18: @itsalaidbacklife @Haviles04 woukd it be better to eventually have an endpoint that returns what we need as part of the response? It's possible we still need this helper but consider this:

Something like /api/games that would return a json object of all games and each game object could have a series of keys that denotes the status; isFull, etc, bawd on the frontend's needs.

There could also be endpoints for the initial available games for the game list view like /api/games/available or even /api/game/id (though I can't recall if something like this exists since I'm on my phone).

It wouldn't devalue the use-case of this helper, but might help simplify some of the code elsewhere.

api/helpers/is-game-full.js
Haviles04: This is used in subscribe.js

utils/status-utils.js
Haviles04, line 1: I tried to import this in game services, but it fails without using `module.exports`, and the store wont accept `module.exports` leading me to put this in two places which I hate. Open for suggestions on this.

seriouslysean: You'd have to convert it to es5 syntax using module.exports and import it that way in to both files. Tricky bit of juggling until we transpile the client and server, it's something I'd like to do for devex though.

Haviles04: How do you go about importing it? I've tried 
```
import {GameStatus}
import { GameStatus }
import * as GameStatus

Comment (seriouslysean) [reply utils/status-utils.js:1]: I think you need to convert the file to module.exports and use a require statement perhaps. I can take a look tomorrow.

Comment (seriouslysean) [reply utils/status-utils.js:1]: This is all you need:
image

```
diff --git a/api/services/gameService.js b/api/services/gameService.js
index 5ba9f39..51ea8db 100644
--- a/api/services/gameService.js
+++ b/api/services/gameService.js
@@ -1,14 +1,6 @@
 const userService = require('../../api/services/userService.js');
 
-/**
- * Game status states
- */
-const GameStatus = Object.freeze({
-  CREATED: 1,
-  STARTED: 2,
-  FINISHED: 3,
-  ARCHIVED: 4,
-});
+const GameStatus = require('../../utils/status-utils.js');
 
 /**
  * @returns int <= 0 if card1 is lower rank or same rank & lower suit
diff --git a/src/components/GameListItem.vue b/src/components/GameListItem.vue
index 43244d3..7c07e03 100644
--- a/src/components/GameListItem.vue
+++ b/src/components/GameListItem.vue
@@ -51,7 +51,7 @@
 </template>
 
 <script>
-import { GameStatus } from '../../utils/status-utils';
+import { GameStatus } from '_/utils/status-utils';
 
 export default {
   name: 'GameListItem',
diff --git a/src/views/HomeView.vue b/src/views/HomeView.vue
index 11a7326..8ca7df3 100644
--- a/src/views/HomeView.vue
+++ b/src/views/HomeView.vue
@@ -133,7 +133,7 @@ import { mapState } from 'vuex';
 import GameListItem from '@/components/GameListItem.vue';
 import CreateGameDialog from '@/components/CreateGameDialog.vue';
 import BaseSnackbar from '@/components/Global/BaseSnackbar.vue';
-import { GameStatus } from '../../utils/status-utils';
+import { GameStatus } from '_/utils/status-utils';
 
 const TABS = {
   PLAY: 'play',
diff --git a/utils/status-utils.js b/utils/status-utils.js
index 22f0a7f..dc0f37d 100644
--- a/utils/status-utils.js
+++ b/utils/status-utils.js
@@ -5,4 +5,6 @@ const GameStatus = Object.freeze({
   ARCHIVED: 4,
 });
 
-export { GameStatus } ;
\ No newline at end of file
+module.exports = {
+  GameStatus,
+};
```

You need the right syntax for `module.exports = {}` first, and while not required you should be using the aliases we set up for imports in the codebase.

Comment (seriouslysean) [reply utils/status-utils.js:1]: For reference, https://github.com/vitejs/vite/discussions/2741

Comment (seriouslysean) [reply utils/status-utils.js:1]: I don't care enough to worry about colocating them with that blocker in mind, but maybe we can think through it more if we decide to move the server to es6+ or whatever in the future.

Comment (Haviles04) [reply utils/status-utils.js:1]: Are we against putting this in a JSON file like the comment suggests?

Comment (seriouslysean) [reply utils/status-utils.js:1]: It kind of defeats the purpose since you still need a way to import on the server and client, though it does share the data. If you really want to share it you'll need a json file, and then a client import to proxied export and a server version also, maybe in separate utils files you can then import from.

File: api/helpers/is-game-full.js (hunk 10 above)
Comment (Haviles04) [reply api/helpers/is-game-full.js:18]: I like this idea, @itsalaidbacklife should we write up an issue on this?

File: utils/status-utils.js
Comment (Haviles04) [reply utils/status-utils.js:1]: So, I put this information into a JSON file and I was able to import it in both the backend and the frontend, The object was frozen anyway, so I think using a JSON object is a decent trick to work here.

File: api/services/gameService.js [hunk 11]
```diff
@@ -120,17 +112,20 @@ module.exports = {
...
+    const p0Wins = userService.checkWin({ user: p0 });
+    const p1Wins = userService.checkWin({ user: p1 });
     if (p0Wins || p1Wins) {
       res.gameOver = true;
       const gameUpdates = {};
       if (p0Wins) {
         res.winner = 0;
-        gameUpdates.result = GameResult.P0_WINS;
+        gameUpdates.winner = p0.id;
+        gameUpdates.status = GameStatus.FINISHED;
```
itsalaidbacklife, line 124: We can DRY this up by moving these status updates into the above block for where either player wins  since it's the same regardless of which player

src/store/modules/gameList.js
```diff
@@ -1,12 +1,13 @@
...
 import { io } from '@/plugins/sails.js';
+import GameStatus  from '../../../utils/GameStatus.json';
 
 class GameSummary {
   constructor(obj) {
     this.id = obj.id ? obj.id : null;
     this.name = obj.name ? obj.name : null;
     this.numPlayers = Object.prototype.hasOwnProperty.call(obj, 'players') ? obj.players.length : 0;
-    this.status = Object.prototype.hasOwnProperty.call(obj, 'status') ? obj.status : false;
+    this.status = Object.prototype.hasOwnProperty.call(obj, 'status') ? obj.status : GameStatus.STARTED;
```
itsalaidbacklife, line 10: I like that this makes games default to unavailable to join as a player but it means they could be spectate or, even though the status being missing means something is wrong. I think it makes sense for them to default to archived instead so that if this data is missing the game doesn't appear anywhere 
```suggestion
    this.status = Object.prototype.hasOwnProperty.call(obj, 'status') ? obj.status : GameStatus.ARCHIVED;
```

src/store/modules/gameList.js
```diff
@@ -30,7 +31,7 @@ export default {
       if (gameIndex < 0 || gameIndex > state.openGames.length) {
         return;
       }
-      const [ startedGame ] = state.openGames.splice(gameIndex, 1);
+      const [startedGame] = state.openGames.splice(gameIndex, 1);
```
itsalaidbacklife, line 34: We should update startedGame's status now that we can track started state
```suggestion
      const [startedGame] = state.openGames.splice(gameIndex, 1);
      startedGame.status = GameStatus.STARTED;
```

src/store/modules/gameList.js
```diff
@@ -44,6 +45,7 @@ export default {
       const updatedGame = state.openGames.find((game) => game.id === data.id);
       if (updatedGame) {
         updatedGame.status = data.newStatus;
+        updatedGame.numPlayers = data.numPlayers;
```
itsalaidbacklife, line 48: I'm looking at the `HomeView` call of this function and it doesn't look like it specifies the `numPlayers`. Is the caller of this mutation now expected to provide that?

api/helpers/is-game-full.js
itsalaidbacklife: @seriouslysean is the suggestion to make it so that all games are sent, regardless of status, and the frontend filters the ones that are joinable? Or to make it so that there is a different endpoint for getting the games that are playable vs the ones that are spectatable (atm both are served by the same endpoint but queried with separate helpers)

utils/GameStatus.json
itsalaidbacklife, line 1: Nice move using json for commonjs and es module compatibility. I think this is the most elegant solution to that problem we've see

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -511,7 +480,7 @@ describe('Home - Create Game', () => {
       .then((games) => {
         expect(games.length).to.eq(1, 'Expect exactly 1 game in store');
         expect(games[0].numPlayers).to.eq(0, 'Expect no players in gameLists game in store, but found some');
-        expect(games[0].status).to.eq(true, 'Expect game to have status true');
+        expect(games[0].status).to.eq(1, 'Expect game to have status CREATED');
```
itsalaidbacklife, line 483: Let's import the game status into the test and use the enum to assert values

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -472,7 +485,7 @@ describe('Creating And Updating Ranked Matches', () => {
...
```
itsalaidbacklife, line 488:
```suggestion
  it('Creates a match when two players play a ranked game for the first time this week', function () {
```

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -16,6 +16,19 @@ function goHomeJoinNewGame() {
   cy.log('Joined new game successfully');
 }
 
+function validateMatch(match, length, p1, p2) {
+  expect(match.player1.id).to.eq(p1);
+  expect(match.player2.id).to.eq(p2);
+  expect(match.startTime).to.be.greaterThan(0);
+  expect(match.endTime).to.eq(null);
```
itsalaidbacklife, line 23: It's surprising that endTime is always null. Are we only calling this on matches that are in progress? Or are matches not getting end times set when they should?

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -16,6 +16,19 @@ function goHomeJoinNewGame() {
...
+  expect(match.games.length).to.eq(length);
+}
+
+function validateWin({status, winner}, gameWinner) {
```
itsalaidbacklife, line 27: So this validates that a game has the expected status and winner?
```suggestion
function validateGameResult({status, winner}, expectedWinner) {
```

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -16,6 +16,19 @@ function goHomeJoinNewGame() {
   cy.log('Joined new game successfully');
 }
 
+function validateMatch(match, length, p1, p2) {
```
itsalaidbacklife, line 19: For consistency maybe this could be
```suggestion
function validateMatchResult(match, length, p1, p2) {
```

itsalaidbacklife (changes requested):
Great work. This is getting close. My next step is to test the migration, but most of my comments are minor cleanup

api/helpers/is-game-full.js
seriouslysean: I think discrete endpoints make the most sense to me so that you're siloing responses to what matters for that particular request. If throughput isn't an issue one response with a key like `gameType` or similar might be ok, too. I'm really fine with whatever as long as we're making it more clear for the engineers working on this stuff. Code clarity and separation of concerns matters a lot imo, especially for open source.

api/hooks/customGameHook/index.js
```diff
@@ -7,7 +8,7 @@ module.exports = function gameHook() {
       return new Promise(function (resolve, reject) {
         Game.create({
           name: gameName,
-          status: true,
+          status: gameService.GameStatus.CREATED,
```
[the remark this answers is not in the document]
seriouslysean: If it's easy enough, go for it, but don't let scope creep conflate an already large PR imo.

src/store/modules/gameList.js
seriouslysean: I would prefer archived also, if the app isn't sure just don't even give the option to let the user do anything with it.

utils/GameStatus.json
seriouslysean: It's certainly fine for now, but long term I don't like it if we move to storing anything other than enums in shared files. You can only use this with string essentially, any sort functions or other logic won't work. We should probably look in to transpiling sails so we can import across both and share more going forward -- outside of a larger rewrite in something like nuxt (which I still think is a good long term goal).

api/helpers/add-game-to-match.js
```diff
@@ -35,19 +35,12 @@ module.exports = {
...
 
       for (const priorGame of [game, ...relevantMatch.games]) {
-        if (priorGame.result === 0) {
-          if (priorGame.p0 === relevantMatch.player1) {
+        if (!priorGame.winner) continue;
+          if (priorGame.winner === relevantMatch.player1) {
             numPlayer1Wins++;
           } else {
             numPlayer2Wins++;
           }
```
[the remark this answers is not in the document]
Haviles04: Ohhhh, we can run the two ifs on the same indenting because it's not nested! good catch!!

api/helpers/ find-spectatable-games.js
```diff
@@ -6,13 +6,11 @@ module.exports = {
...
 
   fn: async (_, exits) => {
-
     const recentUpdateThreshhold = dayjs().subtract(5, 'minute').valueOf();
 
     try {
       const games = await Game.find({
-        status: false, // not open to join
-        result: -1, // not finished yet
+        status:  gameService.GameStatus.STARTED,
```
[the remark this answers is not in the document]
Haviles04: I feel like there has to be a way that we can query the table to check the length of the array of players, I just haven't found it yet. I've tried stackoverflow and chatGPT but I've been unsuccesful so far, I'll keep trying to find a good solution for this because we do it in another spot as well.

api/helpers/is-game-full.js
```diff
@@ -15,7 +15,8 @@ module.exports = {
...
+      game.status === gameService.GameStatus.STARTED ||
```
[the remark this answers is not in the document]
Haviles04: Good call on this, you're right.

api/services/gameService.js
Haviles04: Nice catch!

src/store/modules/gameList.js
Haviles04: I don't think this is needed actually

tests/e2e/specs/in-game/gameOver.spec.js
Haviles04: The match isn't over yet, I dont think it is needed in every check probably just after the match was created and then it was copy pasted into the rest of the spots that checked.  I can refactor this to remove the end time and then just check for it after the match is created and then ended, that would reduce our code even slightly more

tests/e2e/specs/in-game/gameOver.spec.js
itsalaidbacklife: remove the .only
```suggestion
  it('Creates a match when two players play a ranked game for the first time this week', function () {
```

tests/e2e/specs/in-game/gameOver.spec.js
```diff
@@ -16,6 +16,18 @@ function goHomeJoinNewGame() {
...
 
+function validateMatchResult(match, length, p1, p2) {
+  expect(match.player1.id).to.eq(p1);
+  expect(match.player2.id).to.eq(p2);
+  expect(match.startTime).to.be.greaterThan(0);
+  expect(match.games.length).to.eq(length);
+}
+
+function validateGameResult({status, winner}, expectedWinner) {
+  cy.expect(status).to.eq(3);
```
itsalaidbacklife, line 27: Can we import the `GameStatus` in this file and compare to it?
```suggestion
  cy.expect(status).to.eq(GameStatus.FINISHED);
```

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -534,7 +516,7 @@ describe('Home - Create Game', () => {
       .then((games) => {
         expect(games.length).to.eq(1, 'Expect exactly 1 game in store');
         expect(games[0].numPlayers).to.eq(0, 'Expect no players in gameLists game in store, but found some');
-        expect(games[0].status).to.eq(true, 'Expect game to have status true');
+        expect(games[0].status).to.eq(GameStatus.CREATED, 'Expect game to have status CREATED'); //status = created
```
itsalaidbacklife, line 519: Comment is no longer needed
```suggestion
        expect(games[0].status).to.eq(GameStatus.CREATED, 'Expect game to have status CREATED');
```

tests/e2e/specs/out-of-game/stats.spec.js
```diff
@@ -45,10 +45,12 @@ function setup() {
       const matches = matchesFixture.map(transformMatchFixture);
 
       const games = gameFixtures.map((game) => {
+        const winner = game.winner === game.p0 ? this[game.p0] : this[game.p1];
```
itsalaidbacklife, line 48: What about games that stalemate? I think we need to allow for the possibility that `game.winner` is undefined, which is how we'd represent a stalemate. I can't think of a super clean way to accommodate both the null check and the comparison to p0 vs p1, but maybe we can just use a conditional
```suggestion
        let winner = null;
        if (game.winner) {
          winner = game.winner === game.p0 ? this[game.p0] : this[game.p1];
        }
```

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -2,6 +2,8 @@ import { assertSnackbarError } from '../../support/helpers';
 import { Card } from '../../fixtures/cards';
 import { myUser, opponentOne, opponentTwo, playerOne, playerTwo } from '../../fixtures/userFixtures';
 import { SnackBarError } from '../../fixtures/snackbarError';
+import GameStatus from '../../../../utils/GameStatus.json';
+import dayjs from 'dayjs';
```
itsalaidbacklife, line 6: Let's keep the external lib dependencies at the top, so let's put `dayjs` first

itsalaidbacklife (approved):
Very nicely done

itsalaidbacklife:
Final version of the migration ran:
```
-- Delete games with missing users
delete from game 
where 
	p0 not in (
		select id
		from "user" 
	) or p1 not in (
		select id 
		from "user" 
	);

-- Set status and winner column values from old status and result columns
ALTER TABLE game RENAME status TO old_status;
ALTER TABLE game ADD status INTEGER;
ALTER TABLE game ADD winner INTEGER REFERENCES "user" ON DELETE CASCADE;
UPDATE game SET status = 3 WHERE result = 1 OR result = 0 OR  result = 2;
UPDATE game SET status = 4 WHERE result = -1;
UPDATE game SET winner = game.p0 WHERE result = 0;
UPDATE game SET winner = game.p1 WHERE result = 1;
UPDATE game SET winner = null WHERE result = 2; 
alter table game drop column "result";
alter table game drop column "old_status";
```

PR #500, opened 2023-08-10
Resolved 484

### LINKED ISSUE #484: [Feature]: Add loading state (skeleton loader) to Stats Leaderboard
### Feature Summary

The StatsLeaderboard currently has a `loading` prop that is given a value but never used. We should conditionally render a skeleton loader when this prop is true (and the table when it is false) to clarify that the data is loading.

### Detailed Description

See the Vuetify docs on [skeleton loaders](https://vuetifyjs.com/en/components/skeleton-loaders/) for details on how to implement them. Essentially the `StatsLeaderboard` component should use `v-if` to conditionally render a skeleton loader that looks like the leaderboard table when the `loading` prop is true, and use `v-else` to show the data table when the data has finished loading.

This is what the table currently looks like
image

This is what it looks like while the data is loading
image

Instead of this blank screen, we should show a skeleton loader on the leaderboard section, and additionaly put the `v-select` in the `StatsView` page (used to select a season) into a loading state by setting its `loading` prop to the existing `dataIsLoading` variable

haaarsh619:
Resolves #484 

## Please check the following

- ✅ Successfully run the stats.spec.js test
- ✅  Is the code formatted properly ?

### DIFF
--- a/src/plugins/vuetify.js
+++ b/src/plugins/vuetify.js
@@ -1,5 +1,6 @@
 import * as components from 'vuetify/components';
 import * as directives from 'vuetify/directives';
+import { VSkeletonLoader } from 'vuetify/labs/components';
 import { createVuetify } from 'vuetify';
 import { aliases, mdi } from 'vuetify/iconsets/mdi';
 
@@ -43,7 +44,10 @@ const cuttleTheme = {
 };
 
 export default createVuetify({
-  components,
+  components: {
+    ...components,
+    VSkeletonLoader
+  },
   directives,
   icons: {
     defaultSet: 'mdi',
--- a/src/views/LoginView.vue
+++ b/src/views/LoginView.vue
@@ -60,7 +60,7 @@
                 class="mt-4"
                 variant="solo"
                 :rules="usernameRules"
-                :dense="$vuetify.display.mdAndDown ? true : false"
+                :density="$vuetify.display.mdAndDown ? true : false && 'compact'"
                 autocomplete="username"
                 data-cy="username"
               />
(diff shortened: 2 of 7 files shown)


src/views/LoginView.vue
```diff
@@ -60,7 +60,7 @@
                 class="mt-4"
                 variant="solo"
                 :rules="usernameRules"
-                :dense="$vuetify.display.mdAndDown ? true : false"
+                :density="$vuetify.display.mdAndDown ? true : false && 'compact'"
```
itsalaidbacklife, line 63: I think these are not correct. The general idea is that the `dense` prop was boolean (true meant smaller) and it's now replaced by a `density` prop which takes a string value of `'compact'` or `'default'`. We want this to be compact on small devices so what we actually want is
```suggestion
                :density="$vuetify.display.mdAndDown ? 'compact' : false && 'default'"
```

src/views/LoginView.vue
```diff
@@ -72,7 +72,7 @@
                 class="my-4"
                 variant="solo"
                 :rules="passwordRules"
-                :dense="$vuetify.display.mdAndDown ? true : false"
+                :density="$vuetify.display.mdAndDown ? true : false && 'compact'"
```
itsalaidbacklife, line 75: Same here
```suggestion
                :density="$vuetify.display.mdAndDown ? 'compact' : false && 'default'"
```

src/plugins/vuetify.js
```diff
@@ -1,5 +1,6 @@
 import * as components from 'vuetify/components';
 import * as directives from 'vuetify/directives';
+import { VSkeletonLoader } from 'vuetify/labs/components';
```
itsalaidbacklife, line 3: I didn't actually realize `v-skeleton-loader` was an unstable component when writing up issue #484, so this is a little bit sketchy but I think it is okay as the skeleton loader is being officially released in the upcoming 3.4.0 (Blackguard) Vuetify release.

@seriouslysean commenting to let you know I goofed on this one but I think it's still an improvement so I've written 502 to track upgrading and removing the Vuetify labs import once it releases

itsalaidbacklife (approved):
Nice work. The load time got slower after we started calculating additional stats on this page so having a proper loading state will improve the UX significantly

PR #501, opened 2023-08-10
Issue Resolves 499

### LINKED ISSUE #499: [Bug]: Discord button not responsive on mobile portrait
### Bug Summary

Discord button is not responsive on mobile portrait

### Detailed Description

The discord button in `LoginView.vue`  is not responsive on mobile portrait and the text and icon overflow. 
Screenshot_20230810-142803

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Steps to Reproduce

1.
2.
3.

haaarsh619:
Resolves #499

### DIFF
--- a/src/views/LoginView.vue
+++ b/src/views/LoginView.vue
@@ -108,7 +108,8 @@
               </div>
 
               <v-btn
-                class="w-100 my-10 text-h6 h-auto py-2"
+                class="w-100 my-10  h-auto py-2"
+                :class="$vuetify.display.xs ? 'text-subtitle-2' : 'text-h6'"
                 size="large"
                 color="newSecondary"
                 href="https://discord.com/invite/9vrAZ8xGyh"
--- a/src/views/LoginView.vue
+++ b/src/views/LoginView.vue
@@ -108,8 +108,7 @@
               </div>
 
               <v-btn
-                class="w-100 my-10  h-auto py-2"
-                :class="$vuetify.display.xs ? 'text-subtitle-2' : 'text-h6'"
+                class="w-100 my-10 text-subtitle-2 text-sm-subtitle-1 text-md-h6 h-auto py-2"
                 size="large"
                 color="newSecondary"
                 href="https://discord.com/invite/9vrAZ8xGyh"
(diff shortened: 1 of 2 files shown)


itsalaidbacklife (approved):
Looks good. Can you pull from main and bump the version a patch number?

src/views/LoginView.vue
```diff
@@ -108,7 +108,8 @@
               </div>
 
               <v-btn
-                class="w-100 my-10 text-h6 h-auto py-2"
+                class="w-100 my-10  h-auto py-2"
```
Haviles04, line 111: I think we can use vuetifys breakpoints here `text-xs-h4`

https://vuetifyjs.com/en/styles/text-and-typography/

Haviles04 (commented):
Nice!

src/views/LoginView.vue
itsalaidbacklife: Ah yeah so this could replace the dynamic class binding below and keep the classes all together. I think that would be better

src/views/LoginView.vue
```diff
@@ -108,7 +108,8 @@
...
+                :class="$vuetify.display.xs ? 'text-subtitle-2' : 'text-h6'"
```
itsalaidbacklife, line 112: So per @Haviles04 's suggestion we'd have
```suggestion
                class="w-100 my-10  h-auto py-2 text-subtitle-2 text-md-h6"
```
Which we'd then likely need to lint for line length. @haaarsh619 I think this is nicer

itsalaidbacklife (commented):
Requesting refactoring to leverage Vuetify typography breakpoint classes and remove the dynamic style binding

src/views/LoginView.vue
haaarsh619: oh i tried this but i forgot it have mobile first approach and i did something like  this,
```
text-h6  text-xs-subtitle-2 
```
 and this wont work but now i get it what was my  mistake. 
 thanks i will implement you suggestion.

itsalaidbacklife (approved):

PR #509, opened 2023-08-15
Feature #498 Make Scrap Dialog Persistant

### LINKED ISSUE #498: [Feature]: Make ScrapDialog scrollable and not persistent
### Feature Summary

The `ScrapDialog` was recently reimplemented and restyled with the `BaseDialog` component, which improved the overall look, feel, and consistency of the component. We could further improve the UX of this component by making it scrollable (so the header and bottom buttons always show) and removing the `persistent` attribute (so clicking outside the dialog closes it).

### Detailed Description

Currently the `ScrapDialog` contents can become taller than the Dialog height when there are lots of cards, and the scrolling happens inside the entire dialog, hiding the header, or the buttons at the bottom depending on where you scroll.

Top of the dialog:
image

Bottom of the dialog:
image

It would look nicer if the dialog title and the "Close" button at the bottom were always visible and the scrollable area were the middle of the dialog, with the dropdown and cards. We could achieve this by updating the `ScrapDialog` so that it sets the `scrollable` prop in its `BaseDialog` instance like
```
<base-dialog v-model="show" id="scrap-dialog" scrollable>
```

Which would look like this:
image

Additionally, because the `ScrapDialog` is optional for the player to use (unlike most in game dialogs which require the player to make a move), it should _not_ be `persistent`. Currently, clicking outside the dialog causes it to pulse but stay open, indicating "no you have to do something". It would be smoother if this instead closed the `ScrapDialog`. 

This is the current behavior:

https://github.com/cuttle-cards/cuttle/assets/6520704/cf1be625-8c79-4d82-96ef-b49c3235c1d3

The current behavior is what it is because the `BaseDialog` always sets the child `<v-dialog>`'s `persistent` prop to true, making all dialogs persistent i.e. staying open when the user clicks outside of it. To fix this, we should make the `BaseDialog` configurable in this way by defining a new `persistent` prop in `BaseDialog` with a default value of `true` and then having `BaseDialog` set the

```
<v-dialog :persistent="persistent">
```

so that it passes the prop down into the v-dialog, allowing the existing consumers of the `BaseDialog` to remain `persistent` as desired while letting us update the `ScrapDialog` so that it's constituent `BaseDialog` is not persistent but is scrollable like
```
<BaseDialog :persistent="false">
```

Haviles04:
## Issue number

- Resolves #498 

## Please check the following

- For New Features:

- Added scrollable = true to `ScrapDialog.vue`
- Changed persistant to a prop defaulting to true in `BaseDialog.vue`

### DIFF
--- a/src/components/GameView/ScrapDialog.vue
+++ b/src/components/GameView/ScrapDialog.vue
@@ -1,5 +1,10 @@
 <template>
-  <base-dialog v-model="show" id="scrap-dialog">
+  <base-dialog
+    v-model="show"
+    id="scrap-dialog"
+    scrollable="true"
+    :persistent="false"
+  >
     <template #activator="{ props }">
       <span v-bind="props">
         <slot name="activator" />
--- a/src/components/Global/BaseDialog.vue
+++ b/src/components/Global/BaseDialog.vue
@@ -1,7 +1,7 @@
 <template>
   <v-dialog
     v-model="show"
-    persistent
+    :persistent="persistent"
     max-width="650"
     :scrollable="scrollable"
     elevation
@@ -58,6 +58,10 @@ export default {
     opacity:{
       type: Number,
       default: .92,
+    },
+    persistent: {
+      type: Boolean,
+      default: true
     }
   },
   computed: {


itsalaidbacklife (approved):
Looks good. Please bump us a patch version

itsalaidbacklife (approved):
Nice work

PR #510, opened 2023-08-16
Feat/usage chart

### LINKED ISSUE #464: [Feature]: Display site metrics for usage (games played and number of unique players) in stats page

itsalaidbacklife:
This implements a `BaseLineChart` and a specialized `StatsUsageChart` component to render the usage stats (games played and unique players) at the bottom of the stats page. The graph looks like this
image

## Issue number

- Resolves #464 

## Please check the following

- For New Features:


src/components/BaseLineChart.vue
```diff
@@ -0,0 +1,40 @@
...
+      type: Array,
+      required: true,
+    },
+  },
+  data() {
+    return {
+      chartData: {
+        labels: this.labels,
+        datasets: this.dataSets,
+      },
```
itsalaidbacklife, line 33: This should be computed. I think we also won't need the `key` once that's fixed

package-lock.json
```diff
@@ -1,13 +1,14 @@
 {
   "name": "cuttle",
-  "version": "4.6.2",
+  "version": "4.6.3",
```
itsalaidbacklife, line 3: Should bump to match package.json at least b/c this installed deps
```suggestion
  "version": "4.7.0",
```

src/components/BaseLineChart.vue
```diff
@@ -0,0 +1,40 @@
...
+  name: 'BarChart',
+  components: { Line },
+  props: {
+    labels: {
+      type: Array,
+      required: true,
+    },
+    dataSets: {
+      type: Array,
+      required: true,
```
itsalaidbacklife, line 25: Should somehow communicate the types of these. Labels is supposed to be an array of strings for the x-axis labels and datasets matches the [chartjs dataset type](https://www.chartjs.org/docs/latest/general/data-structures.html). This would be a great usecase for typescript, but since that's not configured maybe we should use a vaildator, or even just a comment above the prop. @seriouslysean any thoughts on how to clarify these?

src/components/BaseLineChart.vue
```diff
@@ -0,0 +1,40 @@
...
+        labels: this.labels,
+        datasets: this.dataSets,
+      },
+      chartOptions: {
+        responsive: true,
+      },
+    };
+  },
+};
+</script>
```
itsalaidbacklife, line 40:
```suggestion
</script>

```

src/components/StatsUsageChart.vue
itsalaidbacklife, line 1: This is the specific/opinionated component that inputs a season and renders the chart for its counts of unique players and games played per week

src/components/StatsUsageChart.vue
```diff
@@ -0,0 +1,48 @@
+<template>
+  <base-line-chart :data-sets="dataSets" :labels="labels" :key="season.id" />
```
itsalaidbacklife, line 2: I threw a `key` on this because data wasn't updating reactively when setting another season but I think that's just because I was setting the chart data in component `data()` nonreactively instead of as a comptuted prop
```suggestion
  <base-line-chart :data-sets="dataSets" :labels="labels" />
```

src/components/StatsUsageChart.vue
```diff
@@ -0,0 +1,48 @@
...
+      type: Object,
+      required: true,
+    },
+  },
+  computed: {
+    theme() {
+      return this.$vuetify.theme.themes.cuttleTheme.colors;
+    },
+    labels() {
+      return this.season.gameCounts.map((_el, index) => `Week ${index + 1}`);
```
itsalaidbacklife, line 24: Labels on the X axis go 'Week 1', 'Week 2', etc for as many weeks as are included

src/components/StatsUsageChart.vue
```diff
@@ -0,0 +1,48 @@
...
+        {
+          label: 'Unique players',
+          data: this.uniquePlayers,
+          borderColor: this.theme.newSecondary,
+        },
+      ];
+    }
+  },
+};
+</script>
```
itsalaidbacklife, line 48:
```suggestion
</script>

```

tests/e2e/specs/out-of-game/stats.spec.js
```diff
@@ -283,3 +283,11 @@ describe('Stats Page', () => {
     });
   });
 });
+
+describe('Usage stats', () => {
+  beforeEach(setup);
+
+  it('Displays the site usage stats in the stats usage page', () => {
```
itsalaidbacklife, line 290: I have been trying to work out how to test the chart and I'm stumped. Chartjs renders the chart in a `<canvas>` element which is kindof a black box, so there's little to be done as far as reading from the DOM. Atm we have a test on the data sent from the backend, (which this should be combined with), and that might be the best we can do. @seriouslysean penny for your thoughts.

src/views/StatsView.vue
```diff
@@ -76,6 +76,12 @@
           </h2>
           <stats-leaderboard :loading="loadingData" :season="selectedSeason" />
         </template>
+
+        <!-- Usage stats -->
+        <div id="usage-stats-section">
+          <h2 class="text-h2 mt-8 mb-4">Site Usage</h2>
+          <stats-usage-chart :season="selectedSeason" />
```
itsalaidbacklife, line 83: I like that because the `<stats-usage-chart>` is specialized for the specific line graphs we're showing, we can just send it a season object and the `StatsView` needs 0 additional logic, just to pass the already-selected season

src/views/StatsView.vue
```diff
@@ -9,7 +9,7 @@
   <template v-else>
     <div id="stats-wrapper">
       <!-- Upper Service -->
-      <section id="upper-surface" class="pa-8 mb-6">
+      <section id="upper-surface" class="px-8 pt-8 mb-6">
```
itsalaidbacklife, line 12: This removes the bottom padding on the top section (below the dates for the selected season) to change this:
image

Into this
image

itsalaidbacklife (commented):
Notes on this first draft

src/components/BaseLineChart.vue
```diff
@@ -0,0 +1,40 @@
...
+  <Line
+    id="my-chart-id"
+    :options="chartOptions"
+    :data="chartData"
+  />
+</template>
+
+<script>
+import { Line } from 'vue-chartjs';
+import { Chart as ChartJS, Title, Tooltip, Legend, PointElement, LineElement, CategoryScale, LinearScale } from 'chart.js';
```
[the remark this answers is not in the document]
itsalaidbacklife: Confirmed that we're now using all of these (after I added a title)

package.json
```diff
@@ -1,25 +1,27 @@
...
-    "vue-i18n": "^9.2.2",
+    "vuex": "^4.1.0",
     "vue-router": "^4.2.4",
     "vuetify": "^3.3.11",
-    "vuex": "^4.1.0"
+    "vue-i18n": "^9.2.2",
+    "vue-chartjs": "^5.2.0",
+    "chart.js": "^4.3.3",
+    "socket.io-client": "2.3.0",
+    "marked": "^5.1.1"
```
itsalaidbacklife, line 24: I reordered deps so that backend deps go first, then deps used in both the backend + frontend i.e. lodash, then frontend deps, putting all the vue stuff at the top of frontend. This doesn't really matter but I think it's a bit easier to scan now

tests/e2e/specs/out-of-game/stats.spec.js
```diff
@@ -253,6 +253,15 @@ describe('Stats Page', () => {
...
   });
 
+});
+
+describe('Usage stats', () => {
+  beforeEach(setup);
+
+  it('Displays the site usage stats in the stats usage page', () => {
+    cy.get('#usage-stats-section').find('#usage-stats-chart');
+  });
```
itsalaidbacklife, line 263: @seriouslysean per our convo, I've grouped the existing test on the data sent from the backend with this new one that just asserts that the chart appears

PR #515, opened 2023-08-17
Issue #513: Implemented fallback value for color prop of BaseSnackbar…

### LINKED ISSUE #513: [DevEx]: Refactor `BaseSnackbar`'s `color` prop to have default value of `'error'`
### Improvement Summary

The `BaseSnackbar` component is used to configure and style snackbars to have a consistent look and feel. It accepts a `color` prop to configure the snackbar's background color, but all of the current consumers of `BaseSnackbar` use the same `color="error"`. It would be more DRY to keep the prop but give it a default value of `'error'` so all consumers could omit the prop.

### Detailed Description

This way all the consumers of `BaseSnackbar`, namely `ReuathenticateDialog`, `GameView`, `HomeView`, and `LoginView` can remove the `color` prop form `BaseSnackbar`, but if we encounter a usecase for a non-error snackbar, we can configure the color by overriding the `color`

swebe3qn:
… and removed the color prop from its consumers.

## Issue number

- Resolves #513

## Please check the following

- For New Features:


src/components/Global/BaseSnackbar.vue
```diff
@@ -7,7 +7,7 @@
   <v-snackbar
     :modelValue="modelValue"
     @update:modelValue="$emit('update:modelValue', $event)"
-    :color="color"
+    :color="color || 'error'"
```
itsalaidbacklife, line 10: This is the right idea and should work, but the more traditional and semantic way to set a default value for the prop is to configure the prop definition itself. So we could keep this line as is
```suggestion
    :color="color"
```

Then update the props definition on line 34, so that instead of using the primitive array-of-strings syntax , we could use the object syntax which lets us specify a default value or whether a prop is required. Check the [vue docs](https://vuejs.org/guide/components/props.html#prop-validation) on prop validation. 

This way we could specify that `modelValue`, `message`, and `data-cy` are `required` props, but `color` instead has `default: 'error'`

itsalaidbacklife (commented):
Nice work! This should do the trick, but in order to be more consistent and explicit in the prop definitions, I'm requesting that the default value be moved from the template code into a rework of the prop definitions to use the object syntax. Feel free to ask any questions in the comments

swebe3qn:
Thanks for your feedback! Just pushed another commit.

itsalaidbacklife (commented):
Nice work, this looks good. Last step before merge: can you pull and merge main into your branch and then update the `version` field in `package.json‘ a patch version from 4.7.0 to 4.7.1?

swebe3qn:
Done.

itsalaidbacklife (approved):
Nice work; thank you for your contribution! If you'd like to contribute further, please feel free to grab any issue that interests you. Or if you'd like to chat with myself or the other devs about finding an issue you'd like to work on, you can join us on [discord](https://discord.gg/6vQCFa6nvP)

PR #526, opened 2023-08-21
[DevEx]: Apply eslint rule for element-attribute-order

### LINKED ISSUE #459: [DevEx]: Apply eslint rule for element-attribute-order to enforce consistent directive/attribute ordering in html
### Improvement Summary

The Vue Style guide recommends a specific [order for attributes in html tags](https://v2.vuejs.org/v2/style-guide/?redirect=true#Element-attribute-order-recommended). We've loosely followed this but we could make this explicit and automatic by using the corresponding [eslint rule](https://eslint.vuejs.org/rules/attributes-order.html#vue-attributes-order).

### Detailed Description

This rule applies to html in vue templates to define the order that attributes (including directives) should appear e.g.

```
<!-- ✓ GOOD -->
<button id="someId" v-model="someVar" @click="someFunc" />

<!-- ✗ BAD -->
<button @click="someFunc" v-model="someVar"  id="someId" />
```

We should configure the rule in `.eslintrc.js` so that it causes an error when violated and correspondingly update all the template code that breaks the rule so everything passes without errors.

kawaljen:
## 459
[DevEx]: Apply eslint rule for element-attribute-order to enforce consistent directive/attribute ordering in html

Relevant 459 (https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
- Resolves #459

## Please check the following

- [ ] Do the tests still pass? Yes
- [ ] Is the code formatted properly? Yes
- For New Features:
  - [ ] Have tests been added to cover any new features or fixes? No
  - [ ] Has the documentation been updated accordingly? No

The changes are mainly format change


src/components/AwardCard.vue
```diff
@@ -20,8 +20,14 @@
...
   props: {
-    username: String,
-    place: Number,
+    username: {
+      type: String,
+      default: '',
+    },
+    place: {
+      type: String,
+      default: '',
```
itsalaidbacklife, line 29: I think the change to object syntax is great. This was previously a Number, was it changed deliberately? I also think that we should make `place` a required prop instead of giving it a default value, because it doesn't really make sense to use this component without specifying a particular place
```suggestion
      type: Number,
      required: true,
```

kawaljen: oups yes sorry, wasn't done intentionally

itsalaidbacklife (approved):
Nice work, this cleans up a fair bit of the issues around code style standardization.

Please pull from main, address the PR comments, and update the `version` field in `package.json` from `4.7.2` to `4.7.3` and this is good to go.

src/components/GameView/ChooseTwoDialog.vue
```diff
@@ -50,6 +51,7 @@ export default {
     oneOff: {
       type: Object,
       requried: true,
+      default: null,
```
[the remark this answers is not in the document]
seriouslysean: Required is misspelled, fixing that would also likely resolve the error.

itsalaidbacklife:
@kawaljen this is almost ready. There is one test failure and one linting problem left to fix:

1) the linting problem is in `src/components/BaseLineChart.vue`. The issue is that the name "Line" is reserved and the more restrictive lint rules added here now disallow the use of that component name on line 17. To fix it, we can Alia's the import to call it LineChart and then use that in the component s section
```
import {Line: LineChart} from 'vue-chartjs';
export default BaseLineChart {
  components: {
    LineChart,
  }
}

```

2) the test failure is in `tests/e2e/specs/out-of-game/home.spec.js` and it's because the placeholder text for empty game lists no longer has extraneous white space. We should change the assertion on the placeholder text in 
```
it('Displays placeholder text when no games are available:')
```
To remove the spaces on either side of the text.

Those two fixes should do it and then we're good to go. Nice work on this one

kawaljen:
Ok for the lint error

But for the cypress error, it is actually the opposite, the formatting adds white space..
Should I use contain instead ?
    it('Displays placeholder text when no games are available', () => {
      cy.get('[data-cy=text-if-no-game]').should('contain.text', 'No Active Games');
      cy.contains('p', 'No Active Games');
    });

(but from what I understand it is doing the same thing twice, should I use only of them )

> @kawaljen this is almost ready. There is one test failure and one linting problem left to fix:
> 
> 1. the linting problem is in `src/components/BaseLineChart.vue`. The issue is that the name "Line" is reserved and the more restrictive lint rules added here now disallow the use of that component name on line 17. To fix it, we can alias the import to call it LineChart and then use that in the component s section
> 
> ```
> import {Line: LineChart} from 'vue-chartjs';
> export default BaseLineChart {
>   components: {
>     LineChart,
>   }
> }
> ```
> 
> 2. the test failure is in `tests/e2e/specs/out-of-game/home.spec.js` and it's because the placeholder text for empty game lists no longer has extraneous white space. We should change the assertion on the placeholder text in
> 
> ```
> it('Displays placeholder text when no games are available:')
> ```
> 
> To remove the spaces on either side of the text.
> 
> Those two fixes should do it and then we're good to go. Nice work on this one

Ok for the lint error

But for the cypress error, it is actually the opposite, the formatting adds white space..
Should I use contain instead ?
    it('Displays placeholder text when no games are available', () => {
      cy.get('[data-cy=text-if-no-game]').should('contain.text', 'No Active Games');
      cy.contains('p', 'No Active Games');
    });

(but from what I understand it is doing the same thing twice, should I use only of them )

itsalaidbacklife:
> Ok for the lint error
> 
> 
> 
> But for the cypress error, it is actually the opposite, the formatting adds white space..
> 
> Should I use contain instead ?
> 
>     it('Displays placeholder text when no games are available', () => {
> 
>       cy.get('[data-cy=text-if-no-game]').should('contain.text', 'No Active Games');
> 
>       cy.contains('p', 'No Active Games');
> 
>     });
> 
> 
> 
> (but from what I understand it is doing the same thing twice, should I use only of them )
> 
> 

Yes you are correct the bottom two assertions are equivalent and we should remove one. And yes we want substring matching so that the white space is ignored

itsalaidbacklife (approved):
Great work. I know this increased in scope from the initial ticket and took significant effort to finish. Way to see it through. This will help keep the codebase much more consistent

Issue #351, opened 2023-03-12
[DevEx]: Define new `hasGameId` policy and apply it to the endpoints where req.body is expected to have a gameId


itsalaidbacklife:
### Improvement Summary

Policies allow us to validate requests before they reach controller logic, to ensure they're valid at a surface level before the controller code runs.

### Detailed Description

To define the policy, create a new file `api/policies/hasGameId.js` which follows the format of the other policies by throwing an error if `req.body.gameId` doesn't exist or isn't a number or isn't an integer and otherwise returns `next()` to proceed to the next policy/endpoint.

Then map this policy to any of the ones that need to specify a gameId in `config/policies.js` and update the endpoints where appropriate to remove any validation they do to confirm that the `gameId` exists

itsalaidbacklife:
Closed by #527

PR #527, opened 2023-08-22
Issue #351

haaarsh619:
Issue number #351


api/policies/hasGameId.js
```diff
@@ -0,0 +1,19 @@
...
+ * @description :: Only allows requests that contain a 'gameId' parameter, which is a number
+ * @docs        :: http://sailsjs.org/#!/documentation/concepts/Policies
+ *
+ */
+module.exports = function (req, res, next) {
+  // User is allowed, proceed to the next policy,
+  // or if this is the last policy, the controller
+  if (Object.hasOwnProperty.call(req.body, 'gameId')) {
+    if (typeof req.body.gameId === 'number') return next();
+  }
```
itsalaidbacklife, line 14: The original policies were written before optional chaining was available, but now that we've updated to node < 14, we can consolidate this to:
```suggestion
  if (typeof req?.body?.gameId === 'number') return next();
```

itsalaidbacklife (approved):
Nice work, looks good. Requesting 2 nitpicks and this should be set

itsalaidbacklife (commented):
This is most of the way there, but this doesn't actually _apply_ the policy to the endpoints. The last step is to update `config/policies.js` which defines which policies are applied to which endpoints so that `'game/subscribe'` and `'game/spectate'` both apply the `hasGameId` policy in addition to the existing `isLoggedIn` policy.

Other than that, please pull from `main` and bump the version field in `package.json` from `4.7.0` to `4.7.1`. Nice work!

PR #528, opened 2023-08-22
add key to chart.

### LINKED ISSUE #523: [Bug]: StatsUsageChart title is not reactive
### Bug Summary

The StatsUsageChart displayed on the StatsView page shows a chart with metrics for the number of games played and the number of unique players each week, with a title like”Cuttle played in Diamonds 2022”, where “Diamonds 2022” is the name of the selected Season being viewed. The chart data correctly updates when the user selects a new season at the top of the page, but the chart name remains out of date, showing the old chart title with the previously selected season.

### Detailed Description

This is likely a reactivity issue that has to do with how the StatsView, StatsUsageChart, and BaseLineChart pass data as props. We should try updating the chart id as well as its title so the canvas id includes the season id and then update the test in the stats.spec.js that checks that the chart appears to also confirm that it has the correct id (as a proxy for the title, which we likely can’t assert because it lives in the canvas element)

### Current Behavior

The chart title stays the same when selecting a new season even though the chart data correctly updates

### Expected Behavior

The chart title updates to reflect the newly selected season

### Steps to Reproduce

1. Go to https://cuttle.cards/#/stats
2. Look at the chart underneath “Site Usage”, note the general shape of the data and the chart title e.g. “Amount of Cuttle played in Hearts 2023”
3. At the top of the page select a different season e.g. Diamonds 2023.
4. Note how the chart data changes but the chart title does not

jerryh-jr:
Resolves issue with title not reactively updating when chart changes.

Attempted multiple resolutions to resolve. 

- Using a watcher could update the title in chartOptions, but the chart would not update to show the new title.

Bug link from the vuechartjs page is https://github.com/apertureless/vue-chartjs/issues/1031

- This Resolves #523 by binding a key to the statsUsageChart. 

## Please check the following

- For New Features:


src/components/StatsUsageChart.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <base-line-chart
+    :key="season.id"
```
seriouslysean, line 3: I would love to see if this could be moved to the base chart component so any other usage benefits from it.

@itsalaidbacklife it might also be better to switch away from this library altogether and either directly use it or try some alternatives. There are a lot of reactivity problems with this one in its current iteration.

seriouslysean (commented):
Looks ok to me, would like to see the key moved first but otherwise a sound solution.

I you mentioned alternatives but it would be interesting to see if we can reproduce the reactivity issues in stackblitz or similar.

src/components/StatsUsageChart.vue
seriouslysean: I figured out the issue and you don't need a key to solve it-- we were missing a part yesterday when we paired on it. The issue is more about reactivity and when values update with how we're passing props than anything else. Consider the following in `src/components/StatsUsageChart.vue`:
```
<base-line-chart
  id="usage-stats-chart"
  :title="title"
  :data-sets="dataSets"
  :labels="labels"
/>
```

This is a totally valid usage and passes a few props. So then we move on to `src/components/BaseLineChart.vue`:
```
<Line
  :id="instanceId"
  :options="chartOptions"
  :data="chartData"
/>
```

This sets the id (which needs a little work because the default isn't logical and is in fact invalid, but I digress) and also sets the options and data. The data IS reactive, but the chartOptions are not-- why? Lets look at how they're defined:

```
computed: {
  chartData() {
    return {
      labels: this.labels,
      datasets: this.dataSets,
    };
  },
},
```

We have a computed props for the options and those update every single time. Computed props will properly update when a change to the prop is detected. This is reflected in the chart when the lines change as the seasons are changed.

The chartOptions don't update though, and how are they defined?

```
data() {
  return {
    chartOptions: {
      responsive: true,
      plugins: {
        title: {
          text: this.title,
          display: !!this.title,
        },
      },
    },
  };
},
```

At first glance, this looks right-- they're using local component state and as we all know local component state is reactive _when the value changes_. The issue is that the prop for title is changing, but it's never actually re-setting the local state via a setter. It's just changing and leaving the state alone. You need to actually convert this object to a computed prop in order for reactivity to start working again. This means you don't need the `:key` prop and are able to benefit from the reactivity for performance reasons.

-----

In `BaseLineChart.vue` you should also convert the id to use a new static property on the `created` hook to use the unique id of the component when not set so you benefit from a truly unique id if it's not set.

On the element:
```
:id="instanceId"
```
Prop validation:
```
props: {
  id: {
    type: String,
    required: false,
  },
```
Static property:
```
created() {
  this.instanceId = this.id || this._uid;
},
```

Let me know if you have any questions -- and nice work. This was a tricky one!

seriouslysean (changes requested):

src/components/StatsUsageChart.vue
jerryh-jr: Hey Sean, thanks for taking a look into this and helping resolve it. I was able to put in your suggestion and get it pushed up.

After testing it looks like it works perfect. Let me know if any other changes need to be made.

src/components/BaseLineChart.vue
```diff
@@ -36,26 +36,27 @@ export default {
       required: true,
     },
   },
+  created() {
+    this.instanceId = this.id || this._uid;
```
itsalaidbacklife, line 40: Where does `this._uid` come from?

seriouslysean: It's the internal Vue component ID, so it'll be unique. :) might be worth verifying multiple instances are unique though, which they should be.

itsalaidbacklife: It looks like it's not officially supported by Vue and considered an unstable feature:  https://github.com/vuejs/vue/issues/5886

Maybe we should use crypto or another lib to generate it?

seriouslysean: Fine with me, I'm also fine not using it and requiring the id passed as a prop. Just need safety that it'll always be unique.

seriouslysean: You also have the benefit of no server side component at the moment, but it would be worth considering it since for non game pages we might want to move there eventually for SEO and caching reasons.

itsalaidbacklife: Yeah I think making it a required prop could be simpler and skip the need for this created hook and auto generating a default. @jerryh-jr let's nix the `created()` hook and make `id` a `required` prop.

itsalaidbacklife (approved):
Looks great! Can you merge from main and bump the version in package.json a patch version to 4.7.3?

seriouslysean (approved):
Once the bump goes in this is good to go, nice work! 👍

jerryh-jr:
@seriouslysean @itsalaidbacklife package version changed.

PR #530, opened 2023-08-26
Newnavbar

haaarsh619:


itsalaidbacklife (commented):
Nice work. This will be a key piece of the app redesign, that opens doors to localization and other improvements to the look and feel.

src/App.vue
```diff
@@ -1,14 +1,16 @@
 <template>
   <v-app id="app">
-    <navigation-drawer v-if="showNav" />
+    <the-top-navbar :isSmallDevice="isSmallDevice" v-if="showNav" />
```
itsalaidbacklife, line 3: This will start being fixed automatically by the linter after #526 is merged, but v-if goes before props according to the vue style guide
```suggestion
    <the-top-navbar v-if="showNav" :isSmallDevice="isSmallDevice" />
```

src/components/TheTopNavbar.vue
```diff
@@ -0,0 +1,75 @@
...
+              :to="page"
+              exact
+              :data-nav="text"
+            />
+          </v-list>
+        </v-menu>
+        <img
+          id="logo"
+          alt="Cuttle logo"
+          src="/public/img/cuttle_logo_text_white.svg"
```
itsalaidbacklife, line 32: I believe we should omit /public (see other logo instances e.g. in the login page)
```suggestion
          src="/img/cuttle_logo_text_white.svg"
```

src/composables/navLink.js
```diff
@@ -0,0 +1,35 @@
...
+    } 
+      return [
+        { text: t('global.play'), icon: 'play', page: { name: ROUTE_NAME_HOME } },
+        { text: t('global.stats'), icon: 'chart-bar', page: { name: ROUTE_NAME_STATS } },
+        { text: t('global.rules'), icon: 'information', page: { name: ROUTE_NAME_RULES } },
+      ];
+    
+  });
+
+  const menuItems = [{ text: t('global.logout'), icon: 'logout', page: { name: ROUTE_NAME_LOGOUT } },];
```
itsalaidbacklife, line 32: I think this should be reactive i.e. make it computed like the pageLinks so that if you change language preference it updates

src/components/TheTopNavbar.vue
```diff
@@ -0,0 +1,75 @@
...
+  <v-toolbar data-cy="nav-drawer" color="primary">
+    <v-toolbar-title>
+      <div class="d-flex flex-md-row flex-row-reverse align-center justify-space-between" style="cursor: pointer">
+        <v-menu  transition="slide-x-transition">
+          <template #activator="{ props }">
+            <v-btn
+              data-cy="user-menu"
+              class="hidden-xs-only d-flex text-body-1"
+              v-bind="props"
+              variant="text"
```
itsalaidbacklife, line 11: Does this always show the hamburger, including on large devices? The design intended to hide the hamburger menu on large devices. I'm open to discussion on that point. What do you think is better and why?

haaarsh619: its the menu which open when click on username.

according to the figma file on larger devices we have to click on username to see log out option thats why it is visible for both large and small devices.

src/components/TheTopNavbar.vue
```diff
@@ -0,0 +1,75 @@
...
+          {{ text }}
+        </v-tab>
+      </v-tabs>
+    </v-toolbar-items>
+  </v-toolbar>
+
+</template>
+
+<script>
+
```
itsalaidbacklife, line 59:
```suggestion
```

src/composables/navLink.js
```diff
@@ -0,0 +1,36 @@
...
+      return [
+        { text: t('global.play'), icon: 'play', page: { name: ROUTE_NAME_HOME } },
+        { text: t('global.stats'), icon: 'chart-bar', page: { name: ROUTE_NAME_STATS } },
+        { text: t('global.rules'), icon: 'information', page: { name: ROUTE_NAME_RULES } },
+      ];
+    
+  });
+  const menuItems = computed(() => {
+    return [{ text: t('global.logout'), icon: 'logout', page: { name: ROUTE_NAME_LOGOUT } },];
+  });
```
itsalaidbacklife, line 33: Upon second thought, because the user menu options are simple and only used in one place (unlike the page links, which are used in the top and bottom nav), I think we should remove `menuItems` from this composable and instead make `TheUserMenu` its own component that defines this list in itself

src/components/TheTopNavbar.vue
```diff
@@ -0,0 +1,75 @@
...
+              v-for="({ text, icon, page }, i) in menuItems"
+              :key="i"
+              :prepend-icon="`mdi-${icon}`"
+              :title="text"
+              :to="page"
+              exact
+              :data-nav="text"
+            />
+          </v-list>
+        </v-menu>
```
itsalaidbacklife, line 28: Because this menu is going to get additional functionality (e.g. localization settings) in the future, I'd like to pull it out into its own component called `TheUserMenu` which will manage its own list of menu options and the functionality for clicking them etc

src/components/TheTopNavbar.vue
```diff
@@ -0,0 +1,75 @@
...
+export default {
+name: 'TheTopNavbar',
+props: {
+    isSmallDevice: {
+      type: Boolean,
+      required: true,
+    },
+  },
+  setup() {
+    const { pageLinks, menuItems } = usePageLinks();
```
itsalaidbacklife, line 71: Since we're pulling `TheUserMenu` into its own component and it's the only consumer of `menuItems` we can remove `menuItems` from the composable and define that logic in the component itself

src/composables/navLink.js
```diff
@@ -0,0 +1,36 @@
...
+import { useStore } from 'vuex';
+import {
+  ROUTE_NAME_HOME,
+  ROUTE_NAME_LOGIN,
+  ROUTE_NAME_RULES,
+  ROUTE_NAME_STATS,
+  ROUTE_NAME_LOGOUT
+} from '@/router.js';
+
+export default function usePageLinks() {
```
itsalaidbacklife, line 12: We could use a more explicit rule regarding the use of default vs named exports. I had expected the composable to use a named export because the examples in the Vue docs tend to use named exports, but I don't actually have a specific reason in mind to prefer one over the other for composables, given that they generally export a single function per file.

@haaarsh619 and @seriouslysean which do you think we should use?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "4.8.1",
+  "version": "4.8.2",
```
itsalaidbacklife, line 4: Let's use a minor version since this is a new feature. The rule has been fuzzy, but the general idea has been:
1) patch for bug fixes, docs, style, and refactoring 
2) minor version for features
3) major version for sweeping architectural changes
```suggestion
  "version": "4.9.0",
```

itsalaidbacklife (commented):
Requesting to pull TheUserMenu into its own component, a couple nitpicks, and one open question about import style

src/composables/navLink.js
haaarsh619: i am not sure but i think if we are exporting multiple things than we can use named export and for single function than we can use defaults export.

src/components/TheTopNavbar.vue
```diff
@@ -0,0 +1,55 @@
...
+      <v-tabs class="pa-2">
+        <v-tab
+          variant="text"
+          v-for="({ text, icon, page }, i) in pageLinks"
+          :data-cy="text"
+          :key="i"
+          :title="text"
+          :to="page"
+        >
+          <v-icon :icon="`mdi-${icon}`" />
```
seriouslysean, line 26: This icon also needs a title or text to denote what it is. Alternatively you can use aria hidden to have screen readers ignore it so that the descriptive text doesn't matter.

src/components/TheTopNavbar.vue
```diff
@@ -0,0 +1,55 @@
...
+  components: {
+    TheUserMenu
+  },
+  props: {
+    isSmallDevice: {
+      type: Boolean,
+      required: true,
+    },
+  },
+  setup() {
```
seriouslysean, line 50: Also convert this to use script setup rather than the options API.

src/components/TheTopNavbar.vue
```diff
@@ -0,0 +1,55 @@
...
+import { getPageLinks } from '../composables/navLink.js';
+import TheUserMenu from '../components/TheUserMenu.vue';
+
+export default {
+name: 'TheTopNavbar',
+  components: {
+    TheUserMenu
+  },
+  props: {
+    isSmallDevice: {
```
seriouslysean, line 45: It might be better to provide this from app and then inject here where you need it if you want the shortened name, alternatively it might be ok just to use the vuetify namespace since they add $vuetify to every component.

src/components/TheTopNavbar.vue
```diff
@@ -0,0 +1,55 @@
+<template>
+  <v-toolbar data-cy="nav-drawer" color="primary">
+    <v-toolbar-title>
+      <div class="d-flex flex-md-row flex-row-reverse align-center justify-space-between" style="cursor: pointer">
+        <TheUserMenu />
```
seriouslysean, line 5: Switch name to user menu.

src/components/TheUserMenu.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+          <v-menu  transition="slide-x-transition">
+          <template #activator="{ props }">
+            <v-btn
+              data-cy="user-menu"
+              class="hidden-xs-only d-flex text-body-1"
+              v-bind="props"
+              variant="text"
+            >
+              <v-icon icon="mdi-account-circle" />
```
seriouslysean, line 10: Aria hidden maybe on this too, or descriptive text.

src/components/TheUserMenu.vue
```diff
@@ -0,0 +1,38 @@
...
+              {{$store.state.auth.username}}
+            </v-btn>
+          </template>
+          <v-list density="compact" class="bg-surface-2 text-surface-1">
+            <v-list-item
+              v-for="({ text, icon, page }, i) in menuItems"
+              :key="i"
```
seriouslysean, line 17: Is index enough here?

src/components/TheUserMenu.vue
```diff
@@ -0,0 +1,38 @@
...
+          </v-list>
+        </v-menu>
+</template>
+
+<script>
+import { getMenuItems }  from '../composables/navLink.js';
+
+export default {
+name: 'TheUserMenu',
+  setup() {
```
seriouslysean, line 33: Switch to script setup also since these are all net new.

seriouslysean (commented):
@itsalaidbacklife first pass on my own thoughts, leaving final approval and code level decisions to you, and the UX of course. 🙂

src/components/TheBottomNav.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <v-bottom-navigation
```
seriouslysean, line 2: This should probably be wrapped in a footer tag for semantics.

src/components/TheTopNavbar.vue
```diff
@@ -0,0 +1,55 @@
+<template>
+  <v-toolbar data-cy="nav-drawer" color="primary">
```
seriouslysean, line 2: Wrapped in a header tag perhaps.

src/components/TheTopNavbar.vue
seriouslysean: I also like the name TheHeader better if we need to stick with it.

src/components/TheBottomNav.vue
```diff
@@ -0,0 +1,30 @@
...
+    </v-btn>
+  </v-bottom-navigation>
+</template>
+
+<script>
+import { getPageLinks } from '../composables/navLink.js';
+
+export default {
+name: 'TheBottomNav',
+  setup() {
```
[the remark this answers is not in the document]
seriouslysean: I also like the name TheFooter better if we need to stick with it.

src/components/TheBottomNav.vue
itsalaidbacklife: I wonder if Vuetify injects one?

src/components/TheTopNavbar.vue
itsalaidbacklife: I discussed this with @seriouslysean and based on the [Vue style guide for singleton component instances](https://v3.ru.vuejs.org/style-guide/#single-instance-component-names-strongly-recommended) we'll stick with TheUserMenu

src/components/TheTopNavbar.vue
itsalaidbacklife: I think that we want to set a role, aria-label, and aria-hidden="false". 
```suggestion
          <v-icon
            :icon="`mdi-${icon}`"
            :aria-label="text"
            aria-hidden="false"
            role="img" 
          />
```

The label is for screen readers to know what the icon's about, aria-hidden="false" ensures it appears for screen readers, and role="img" tells screen readers that this icon represents an image

seriouslysean: To play devil's advocate the icon doesn't add anything for a screen reader that the text doesn't so just hiding it is less work imo.

src/components/TheTopNavbar.vue
itsalaidbacklife: After taking a closer look at the Vue style guide re: [singleton instance components](https://vuejs.org/style-guide/rules-strongly-recommended.html#single-instance-component-names), these components shouldn't have props, so as @seriouslysean let's remove the `isSmallDevice` prop and instead compute that directly from within this component.

In order to keep the names more brief, as @seriouslysean has suggested, let's rename this `TheHeader` and rename `TheBottomNav` to `TheFooter`.

We are switching to `<script setup>` syntax, so going forward we are going to ask that all new components (including these two) be implemented in the `<script setup>` style.

So:
1. Let's rename this file to `TheHeader.vue`
2. Let's convert it to `<script setup>`
3. Let's remove the `isSmallDevice` prop and compute it inside this component with Vuetify's [useDisplay() composable](https://vuejs.org/style-guide/rules-strongly-recommended.html#single-instance-component-names)

src/components/TheUserMenu.vue
itsalaidbacklife: I hope to write up a style with directions for icons and to create issues to update all of our existing icons with aria attributes, but yes we want a label, aria-hidden="false" and role="img"
```suggestion
              <v-icon
                icon="mdi-account-circle"
                aria-label="Open user menu"
                aria-hidden="false"
              />
```

src/components/TheUserMenu.vue
itsalaidbacklife: It seems okay to me, though I'm open to alternatives.

src/components/TheUserMenu.vue
```diff
@@ -0,0 +1,38 @@
...
+    const menuItems = getMenuItems();
+    return { menuItems };
+  },
+};
+</script>
```
itsalaidbacklife, line 38: Prefer newlines at ends of files
```suggestion
</script>

```

src/components/TheUserMenu.vue
itsalaidbacklife: Seconded. I'll write this up as a part of the style guide, but all new components are going to be written with `<script setup>` and we'll start converting the existing ones over in time

src/components/TheUserMenu.vue
```diff
@@ -0,0 +1,38 @@
...
+              :to="page"
+              exact
+              :data-nav="text"
+            />
+          </v-list>
+        </v-menu>
+</template>
+
+<script>
+import { getMenuItems }  from '../composables/navLink.js';
```
itsalaidbacklife, line 29: Prefer to use @ to refer to imports relative to src so that if we move this component the imports don't break
```suggestion
import { getMenuItems }  from '@/composables/navLink.js';
```

src/composables/navLink.js
```diff
@@ -0,0 +1,38 @@
...
+    ];
+  });
+}
+
+export function getMenuItems() {
+  const { t } = useI18n();
+
+  return computed(() => {
+    return [{ text: t('global.logout'), icon: 'logout', page: { name: ROUTE_NAME_LOGOUT } }];
+  });
```
itsalaidbacklife, line 37: This list will exclusively be consumed by `TheUserMenu` so let's remove it from the composable and put it there

src/components/TheBottomNav.vue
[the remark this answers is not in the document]
itsalaidbacklife: Seconded. Please rename this to `TheFooter.vue` and switch it to the `<script setup>` syntax

src/components/TheBottomNav.vue
```diff
@@ -0,0 +1,30 @@
...
+    grow
+  >
+    <v-btn
+      variant="text"
+      v-for="({ text, icon, page }, i) in pageLinks"
+      :key="i"
+      :title="text"
+      :to="page"
+    >
+      <v-icon>{{`mdi-${icon}`}}</v-icon>
```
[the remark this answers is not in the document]
itsalaidbacklife: Same with the other icons
```suggestion
      <v-icon
        :icon="`mdi-${icon}`"
        :aria-label="text"
        aria-hidden="false"
        role="img"
      />
```

src/components/TheTopNavbar.vue
itsalaidbacklife: Yes let's convert to script setup and rename to `TheHeader.vue`

src/App.vue
```diff
@@ -1,14 +1,16 @@
 <template>
   <v-app id="app">
-    <navigation-drawer v-if="showNav" />
+    <the-top-navbar v-if="showNav" :isSmallDevice="isSmallDevice" />
     <v-main>
       <router-view />
     </v-main>
+    <the-bottom-nav v-if="showNav && isSmallDevice" />
```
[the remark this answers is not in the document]
itsalaidbacklife: Per our discussion and the vue style guide, we'll name this `<the-footer>`

src/App.vue
```diff
@@ -1,14 +1,16 @@
 <template>
   <v-app id="app">
-    <navigation-drawer v-if="showNav" />
+    <the-top-navbar v-if="showNav" :isSmallDevice="isSmallDevice" />
```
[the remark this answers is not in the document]
itsalaidbacklife: Per our discussion and the vue style guide, we'll name it `<the-header>`

src/composables/navLink.js
itsalaidbacklife: @haaarsh619 I'm going to do a deeper dive into the vue docs recommendations but for now let's follow the examples of the docs and use named exports for composables and have them return objects that we can destructure in the consumers

itsalaidbacklife (commented):
This is very close. I think one more pass will do it. 

Per @seriouslysean 's review, the high level things left are:
1. accessibility attrs for icons
2. Renaming `TheTopNavbar` to `TheHeader` and `TheBottomNav` to `TheFooter`
3. Removing the `isSmallDeviceProp` from `TheHeader` and computing it locally with vuetify's `useDisplay()` composable
4. Converting all the new components to use `<script setup>`


Really nice work, this is going to be a huge UX improvement

itsalaidbacklife (approved):
Very nicely done. This will be a significant UX improvement and the basis for a complete redesign of the out of game experience

src/components/TheTopNavbar.vue
itsalaidbacklife: We can normalize this in a future app-wide accessibility update

PR #544, opened 2023-09-02
541 undefined username

### LINKED ISSUE #541: [Bug]: hasValidUsername policy errors when `req.body.username` is null.
### Bug Summary

The `hasValidUsername` policy is intended to validate whether requests that require the client to send a `username` in the request body do so (rejecting the requests otherwise). It has a type safety issue where if the `username` is missing from the request body, it actually errors and sends a 500 response, rather than correctly returning a ` badRequest()` 400 response.

### Detailed Description

Here is the policy:

```
module.exports = function (req, res, next) {
  const username = req.body.username.trim(); // this errors if req.body.username is undefined (can't .trim() undefined)

  if (!username) {
    return res.badRequest({ message: 'Please provide a non-empty username' });
  }

  if (typeof username !== 'string' || !username.match(/^[\w.@-]+$/)) {
    return res.badRequest({ message: 'Your username must contain only letters or numbers' });
  }
  req.body.username = username;
  return next();
};
```

We can fix this with optional chaining in the first line of the policy function
```
const username = req.body.username?.trim();
```

### Current Behavior

Sending a request to signup without a username causes the backend to send a 500 response.

### Expected Behavior

Sending a request to signup with no username causes the backend to send a 400 response.

### Steps to Reproduce

1.
2.
3.

ABertaud:
## Issue number

- Resolves #541 

## Please check the following

- For New Features:

### DIFF
--- a/api/policies/hasValidUsername.js
+++ b/api/policies/hasValidUsername.js
@@ -8,7 +8,7 @@
  *
  */
 module.exports = function (req, res, next) {
-  const username = req.body.username.trim();
+  const username = req.body.username?.trim();
 
   if (!username) {
     return res.badRequest({ message: 'Please provide a non-empty username' });
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "cuttle",
-  "version": "4.8.0",
+  "version": "4.8.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cuttle",
-  "version": "4.8.1",
+  "version": "5.0.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "cuttle",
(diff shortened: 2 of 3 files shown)


seriouslysean (approved):
Good defensive add, leaving final approval to @itsalaidbacklife. 👍 

Thanks for the contribution!

itsalaidbacklife (approved):
This looks great, thank you. Normally we bump the version in package.json rather than package-lock.json. Can you bump us there from 5.0.0 to 5.0.1? 

package-lock.json should either be left alone or updated to 5.0.1 to match package.json

seriouslysean:
@itsalaidbacklife I'll complain about this forever, but we should be using npm version which will bump both package files and tag the build on merge. Package lock is meant to be a locked version of the package file. https://docs.npmjs.com/cli/v9/commands/npm-version

Maybe we can document it better and add npm bump scripts to avoid the tag if you prefer.

ABertaud:
Alright, I used `npm version <new_version>` which updated both `package.json` and `package-lock.json`.

itsalaidbacklife (approved):
Nicely done

PR #556, opened 2023-09-03
feat: Run tests against dockerized app

### LINKED ISSUE #534: [DevEx]: Update CI to run tests against dockerized app
### Improvement Summary

As of #468, we have an MVP setup for a dockerized version of the app. We should modify our CI setup so that the cypress tests run against this dockerized version of the app to improve consistency and reduce test flake.

### Detailed Description

We have seen a number of issues with test flake in CI such as #471 and #521 which threaten the precision of the test suite (i.e. false failing results make it harder to detect a real failure). To combat this inconsistency, we should run the tests against a dockerized version of the app, which behaves more predictably. This should ensure at minimum that failures in CI are replicable locally, making it easier to troubleshoot.

To start, the tests should run against a version of the `server` container that:
1. is in `development` mode (using mocked database and session store)
2. has built the app and serves it on localhost:1337

itsalaidbacklife:
## Issue number

- Resolves #534 

## Please check the following

- For New Features:

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -10,29 +10,29 @@ on:
 
 jobs:
   # Install & build frontend
-  install:
-    runs-on: ubuntu-22.04
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v3
+  # install:
+  #   runs-on: ubuntu-22.04
+  #   steps:
+  #     - name: Checkout
+  #       uses: actions/checkout@v3
 
-      - name: Cypress install
-        uses: cypress-io/github-action@v5
-        with:
-          # Disable running of tests within install job
-          runTests: false
-          build: npm run build
+  #     - name: Cypress install
+  #       uses: cypress-io/github-action@v5
+  #       with:
+  #         # Disable running of tests within install job
+  #         runTests: false
+  #         build: npm run build
 
-      - name: Save build folder
-        uses: actions/upload-artifact@v3
-        with:
-          name: build
-          if-no-files-found: error
-          path: assets
+  #     - name: Save build folder
+  #       uses: actions/upload-artifact@v3
+  #       with:
+  #         name: build
+  #         if-no-files-found: error
+  #         path: assets
 
   cypress-run:
     runs-on: ubuntu-22.04
-    needs: install
+    # needs: install
     strategy:
       # don't fail the entire matrix on failure
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/test-e2e.yml
```diff
@@ -10,29 +10,29 @@ on:
...
-        with:
-          # Disable running of tests within install job
-          runTests: false
-          build: npm run build
+  #     - name: Cypress install
+  #       uses: cypress-io/github-action@v5
+  #       with:
+  #         # Disable running of tests within install job
+  #         runTests: false
+  #         build: npm run build
```
itsalaidbacklife, line 24: I completely disabled cacheing to start just so I could figure out how to get it to even boot and run against the dockerized app. I'm pleasantly surprised to see that this still runs the entire suite in ~11min. 

We could likely cut that down significantly with a proper build cache

.github/workflows/test-e2e.yml
```diff
@@ -43,22 +43,23 @@ jobs:
...
           parallel: true
-          start: npm start
+          start: npm run docker:start
           config-file: cypress.config.js
-          # wait for 2 minutes for the server to respond
-          wait-on: 'http://localhost:1337'
-          wait-on-timeout: 120
+          config: baseUrl=http://localhost:8080
+          # wait for 5 minutes for the server to respond
+          wait-on: 'http://localhost:8080'
```
itsalaidbacklife, line 61: Hard coded the CI e2e to point to localhost:8080 since the config file inherits the cypress BASE_URL from the VITE_API_URL when available. Otherwise atm this would try to run the tests against localhost:1337, which failed because the client wasn't built or served on it. I tried running the build step in the `server` service config before booting the backend, but it was crashing the service for some reason.

I think ultimately that running the tests against a built version of the client, served by the `server` would be a better representation of the real application (that's how we've been doing it). But to make it work we'd need to get the `server` service to actually build the app

docker/docker-compose.yml
```diff
@@ -15,20 +15,20 @@ x-app-volume:
...
+  #     - 5432:5432
+  #   environment:
+  #     - POSTGRES_DB=cuttle
+  #     - POSTGRES_USER=cuttlesworth
+  #     - POSTGRES_PASSWORD=p4ssw0rd!
+  #     # Allow all connections, no password required (not recommended)
+  #     # - POSTGRES_HOST_AUTH_METHOD=trust
+  #   volumes:
+  #     - ./data/pgdata:/var/lib/postgresql/data
+  #   restart: always
```
itsalaidbacklife, line 31: I turned off the database service altogether b/c I haven't figured out how to make it dynamically configurable yet. It would be nice to be able to have different modes with and without postgres and with the server in different NODE_ENV's

docker/docker-compose.yml
```diff
@@ -39,7 +39,7 @@ services:
       - DATABASE_URL=postgresql://cuttlesworth:p4ssw0rd!@database:5432/cuttle
       # Can set this to staging for the postgres database to work in docker but may need to modify
       # the onlyAllowOrigins setting in `config/env/staging.js` file
-      - NODE_ENV=staging
+      - NODE_ENV=development
```
itsalaidbacklife, line 42: Switched to development so that we use the mock db which currently works better for testing in CI

itsalaidbacklife:
Closing this as it isn't a priority atm

Issue #425, opened 2023-05-28
[DevEx]: Move all testing-specific endpoints into the TestController


itsalaidbacklife:
### Improvement Summary

Currently some of the testing-specific endpoints (the ones only allowed during `development` or `staging`) are in the TestController, and some are in the `api/controllers/game` folder. We should consolidate all of them into the TestController for consistency and simplicity.

### Detailed Description

We should take all of the backend actions/endpoints which have the `'developmentOrStagingOnly'` policy applied to them in `config/policies.js` and move them into `api/controllers/TestController`.  While we are at it, we should update the urls to be more consistent and remove unnecessary testing-specific endpoints. For the unnecessary endpoints we should delete the files in api/controllers/game/<unnecessary_endpoint_name>.js, the policy mapping in `config/policies.js`, and the route mapping in `config/routes.js`

In particular:
1. game/stackDeck is no longer used and can be deleted 
2. game/clear is no longer used and can be deleted
3. `/game/loadGameFixture` should be remapped to `/test/loadGameFixture` in `config/routes.js` and the corresponding endpoint file should be moved from api/controllers/game/load-fixture` to `api/test/load-game-fixture.js`

PR #559, opened 2023-09-05
Issue 425 fix

BrandonCarp:
Issue https://github.com/cuttle-cards/cuttle/issues/425

Relevant https://github.com/cuttle-cards/cuttle/issues/425

Resolves https://github.com/cuttle-cards/cuttle/issues/425
Please check the following
[Y] Do the tests still pass? (see [Run the Tests](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#run-the-tests))
[Y] Is the code formatted properly? (see [Linting (Formatting)](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#linting-formatting))
For New Features:
[NA] Have tests been added to cover any new features or fixes?
[Y] Has the documentation been updated accordingly?
Please describe additional details for testing this change
The backend actions/endpoints which have 'developmentOrStagingOnly' policy applied to them were moved into api/controllers/TestController. While also deleting the previous files they were in.


config/routes.js
```diff
@@ -89,14 +89,14 @@ module.exports.routes = {
...
 
   // Testing helpers
   '/test/wipeDatabase': 'TestController.wipeDatabase',
   '/test/badSession': 'TestController.setBadSession',
   '/test/loadSeasonFixture': 'TestController.loadSeasonFixture',
   '/test/loadMatchFixtures': 'TestController.loadMatchFixtures',
   '/test/loadFinishedGameFixtures': 'TestController.loadFinishedGameFixtures',
+  '/test/stackDeck': 'TestController.stackDeck',
+  '/test/loadGameFixture': 'TestController.loadGameFixture',
+  '/test/clearGame': 'TestController.clear',
```
itsalaidbacklife, line 101: This does seem like a good change (making the urls more consistent with the other test-specific ones), but it also means that we need to change all references to these urls. Look for all the consumers of
```
/game/stackDeck
/game/loadFixture
/game/clear
```

Looking at them now, both `stackDeck` and `clearGame` are actually unused in the entire app. So we can delete those two altogether, (both the files themselves, + the route and policy mappings).

Whereas the `'/game/LoadFixture'` route is currently in use in `tests/e2e/support/commands.js` in the `loadGameFixture()` command. We need to make sure that this function is updated so that it calls the updated endpoint url i.e. `'/test/loadGameFixture'`
```suggestion
  '/test/loadGameFixture': 'TestController.loadGameFixture',
```

api/controllers/TestController.js
```diff
@@ -54,4 +54,174 @@ module.exports = {
...
+        return res.badRequest(err);
+      });
+  },
+
+  clearGame: async function (req, res) {
+    return gameService.clearGame({ userId: req.session.usr }).then(function postClear() {
+      return res.ok();
+    });
+  },
+
```
itsalaidbacklife, line 87: Turns out these aren't used. We can remove them both (but keep `loadGameFixture()`
```suggestion
```

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "5.1.3",
+  "version": "5.1.5",
```
itsalaidbacklife, line 4: This should increment just one patch version to 5.1.4
```suggestion
  "version": "5.1.4",
```

Same applies to `package-lock.json`

It looks like the version bump was run twice

itsalaidbacklife: You can either revert the last commit or just change these files by hand

itsalaidbacklife (commented):
Version was bumped twice but otherwise it's ready

package-lock.json
```diff
@@ -1,12 +1,12 @@
...
   "name": "cuttle",
-  "version": "5.1.3",
+  "version": "5.1.5",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "cuttle",
-      "version": "5.1.3",
+      "version": "5.1.5",
```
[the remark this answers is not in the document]
seriouslysean: Can't just change these here, running the version script also creates a tag so that may need to be addressed too. 

It's possible you needed to delete tag, reset the version and run it again. @itsalaidbacklife  we may need to test this case and add documentation on how to recover from it.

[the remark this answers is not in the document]
BrandonCarp: Sorry just getting to this, I was curious about that. Ive never worked on another persons codebase so things like these I was curious of the possible problems that could happen. Yeah Ryan told me how to fix it. Appreciate you dudes

itsalaidbacklife:
@BrandonCarp are you still interested in finishing this? It needs the merge conflicts fixed and then it should be ready.

Issue #560, opened 2023-09-06
[Feature]: Add 4px margin between icons and tab labels in `TheHeader`


itsalaidbacklife:
### Feature Summary

`TheHeader` component has tabs that appear in the upper right, with icons and labels for each page e.g. 'Play'. The icons are too close to the text and would look better with a 4px margin between them. 

### Detailed Description

See the image below.
![image](https://github.com/cuttle-cards/cuttle/assets/6520704/62b880c8-d662-47ab-9d2d-ae7796cfd3b2)

We should add 4px of margin between the icons and the text in the tabs inside `src/components/TheHeader.vue` so the elements have room to breathe.

DilushaHesaranga:
### Hey there !!!

_Do you still meed help with this issue : )_
**If yes, leave a comment** 

> I am glad to work with this issue.

itsalaidbacklife:
@DilushaHesaranga have it it! Feel free to ask any questions you have about the issue here, or if you'd prefer real-time support you can join us on [discord](https://discord.gg/zZujC7Tp5j) where you can chat with myself and the other devs

DilushaHesaranga:
Sure

> @DilushaHesaranga have it it! Feel free to ask any questions you have about the issue here, or if you'd prefer real-time support you can join us on [discord](https://discord.gg/zZujC7Tp5j) where you can chat with myself and the other devs

**Sure!! I will fix them and reach you soon : )**

PR #561, opened 2023-09-06
I added 4px of margin between the icons and the text in the tabs inside  src/components/TheHeader.vue

DilushaHesaranga:
src/components/TheHeader.vue

## Issue number

 [https://github.com/cuttle-cards/cuttle/issues/560](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

## Please check the following

- [x] Do the tests still pass?
- [x] Is the code formatted properly? 
- For New Features:

## Here's how I modified your code to achieve this:

_Add a CSS class to your <v-tabs> element. For example, let's call it "tab-with-margin"
Define the CSS style for this class in your component's <style> section_

_.tab-with-margin .v-tab selects the individual tabs inside the <v-tabs> component and makes them flex containers so that the icon and text are aligned horizontally._
_.tab-with-margin .v-tab .v-icon selects the icons within the tabs and adds a right margin of 4px._

**In this code, I've added the "tab-with-margin" class to the <v-tabs> element and defined CSS styles to add a 4px margin**

### DIFF
--- a/src/components/TheHeader.vue
+++ b/src/components/TheHeader.vue
@@ -15,7 +15,7 @@
         </div>
       </v-toolbar-title>
       <v-toolbar-items v-if="!mobile" class="hidden-xs-only">
-        <v-tabs class="pa-2">
+        <v-tabs class="pa-2 tab-with-margin">
           <v-tab
             v-for="({ text, icon, page }, i) in pageLinks"
             :key="i"
@@ -46,3 +46,14 @@ import { useDisplay } from 'vuetify';
 const { mobile } = useDisplay();
 const pageLinks = getPageLinks();
 </script>
+
+<style scoped>
+.tab-with-margin .v-tab {
+  display: flex;
+  align-items: center;
+}
+
+.tab-with-margin .v-tab .v-icon {
+  margin-right: 4px;
+}
+</style>


src/components/TheHeader.vue
```diff
@@ -46,3 +46,14 @@ import { useDisplay } from 'vuetify';
...
+
+<style scoped>
+.tab-with-margin .v-tab {
+  display: flex;
+  align-items: center;
+}
+
+.tab-with-margin .v-tab .v-icon {
+  margin-right: 4px;
+}
```
itsalaidbacklife, line 58: How does this look with and without the center-aligned flex?

My thought was just to put the margin directly in the `<v-icon>` html with the vuetify helper class like

```

            <v-icon
              :icon="`mdi-${icon}`"
              class="mr-1"
              :aria-label="text"
              aria-hidden="false"
              role="img"
            />
```

itsalaidbacklife (commented):
Can you send a screenshot of how the tabs look with and without `align-items: center`? I wasn't expecting to vertically center the tabs, but I'm open to it if it looks better. Other than that, could you please update the `version` field in `package.json` and `package-lock.json` to set both of them to `5.1.2`?

src/components/TheHeader.vue
DilushaHesaranga: okay Give me some time time and I'll reach you surely

DilushaHesaranga:
ok .
On it

PR #564, opened 2023-09-08
Feature/require version tag

seriouslysean:
## Issue number

- Resolves #

## Please check the following

- For New Features:

### DIFF
--- a/.github/actions/package.json
+++ b/.github/actions/package.json
@@ -1,0 +1,15 @@
+{
+  "name": "cuttle-actions",
+  "version": "0.1.0",
+  "description": "GitHub Actions for Cuttle",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "private": true,
+  "license": "MIT",
+  "dependencies": {
+    "@actions/core": "^1.6.0",
+    "@actions/github": "^5.0.0"
+  }
+}
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -37,10 +37,18 @@ jobs:
           exit 1
       - run: echo "PR is not Draft"
 
+  # Require a team label
+  version-label:
+    uses: ./.github/workflows/run-npm-script.yml
+    needs: draft
+    with:
+      label: "Check PR Version Label"
+      script: "check-version-label"
+
   # Check Linting status
   lint:
     uses: ./.github/workflows/run-npm-script.yml
-    needs: draft
+    needs: version-label
     with:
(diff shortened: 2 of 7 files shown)


seriouslysean:
Test PR against my local cuttle, https://github.com/seriouslysean/cuttle/pull/3

-----

With a missing `version-*` tag

image

https://github.com/seriouslysean/cuttle/actions/runs/6123456351/job/16621380606

With a valid `version-*` tag

image

https://github.com/seriouslysean/cuttle/actions/runs/6123478350/job/16621451117

.github/actions/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "cuttle-actions",
```
seriouslysean, line 2: We need a separate `package.json` here in order to install the `@actions` libs we need without muddying the normal cuttle `package.json`.

.github/workflows/pr-actions.yml
```diff
@@ -37,26 +37,34 @@ jobs:
           exit 1
       - run: echo "PR is not Draft"
 
+  # Require a team label
+  version-label:
+    uses: ./.github/workflows/run-npm-script.yml
+    needs: draft
+    with:
+      label: "Check PR Version Label"
+      script: "check-version-label"
```
seriouslysean, line 46: If this fails, none of the other tests will run which is a nice benefit imo.

.github/actions/package.json
itsalaidbacklife: Is that considered a best practice? It's an interesting idea, I've just never heard of it

.github/workflows/pr-actions.yml
itsalaidbacklife: Thinking further about this for our ideal flow. Adding a label doesn't trigger the PR, right? So every PR that gets created will initially fail this check, not run the tests, and then after the label is added, we'll need to manually re-run the tests.

I think it might be better if this check were required by the branch protection policy, but if it ran both on 'normal' pr events like the rest and then also on the 'label' event where we can check if the PR has a version label like in this [SO](https://stackoverflow.com/questions/62325286/run-github-actions-when-pull-requests-have-a-specific-label).

That way the tests can still run when the PR is opened, we can apply the label without needing to re-run the whole test suite, and then it'll check for the version label so that can pass

seriouslysean: Want to take a crack at that since you have the permissions? I'm fine letting this simmer, it's not a hard requirement at the moment.

itsalaidbacklife:
Closing this because our approach is changing to automating the bump instead of requiring it

PR #568, opened 2023-09-10
DevEx #537 Replace VueX with Pinia

### LINKED ISSUE #537: [DevEx]: Replace Vuex with Pinia
### Improvement Summary

Cuttle uses Vuex for state management. Now that the app has been updated to use Vue 3, we can replace Vuex with its officially supported replacement: pinia.

### Detailed Description

[Pinia](https://pinia.vuejs.org/introduction.html) is the official state management library for the Vue ecosystem. It is essentially an upgrade to its predecessor, Vuex, which Cuttle still uses. The central value proposition for switching to pinia is a simpler and more type-safe api for calling methods on the stores as while vuex relies on calling `.dispatch()` and `.commit()` and passing function names as strings, pinia allows store consumers to directly call methods on the store instance.

Upgrading will essentially require updating all of the vuex modules in `src/store/modules` to be pinia stores and updating the component references to those stores to replace the `.dispatch()` and `.commit()` calls with regular method calls on their respective functions e.g. `this.$store.dispatch(‘requestLogin’)` will become `this.authStore.requestLogin()`. Additionally, most of the store mutations will no longer be necessary, as store consumers can directly mutate variable state in the store without a mutation function. The only mutations worth keeping will be the complex ones that update data in a nuanced way, like the `game` store’s `updateGame()` mutation or the `gameList` store’s `gameStarted()` mutation. See the [pinia migration guide](https://pinia.vuejs.org/cookbook/migration-vuex.html) for additional guidance.

For the initial implementation, we will stick with the Vue options API that the components already use to minimize the scope of this change. See the [pinia docs on options api usage](https://pinia.vuejs.org/cookbook/options-api.html) for details and examples.

Haviles04:
## Issue number

- Resolves #537

## Please check the following

- For New Features:

## Please describe additional details for testing this change\

Migrated entire application store from VueX to Pinia.
-  Changed store modules to flat file structure 
- Converted old modules to new functional pinia modules
- Removed Actions that only set to a new value, and set it explicitly where the action was used
- Changed store references throughout entire application to utilize new `...mapStores(useStore)` function
- Added store to window object on `mounted()` inside `App.vue`
- Changed entire test suite to utilize new window store `cy.window().its('storeName').then((store) => logic)`


itsalaidbacklife (commented):
Nice start. This is looking good so far. I left a couple comments, all minor. I think it may make sense to do one module and update its references ie get the references to the auth store module next

src/stores/auth.js
```diff
@@ -1,18 +1,19 @@
...
+} from '../../utils/local-storage-utils.js';
 
 // TODO Figure out how to reconsolidate this with backend
 const getPlayerPnumByUsername = (players, username) => {
   const pNum = players.findIndex(({ username: pUsername }) => pUsername === username);
   return pNum > -1 ? pNum : null;
 };
 
-async function handleLogin(context, username, password, signup = false) {
+async function handleLogin(username, password, signup = false) {
```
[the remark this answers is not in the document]
Haviles04: Yea, I actually wasn't sure how this was going to work out. I meant to ask you. I like moving it to actions 🤙

src/stores/auth.js
```diff
@@ -1,80 +1,45 @@
...
-  },
-  mutations: {
-    authSuccess(state, username) {
-      state.authenticated = true;
-      state.username = username;
+  }),
+  actions: {
+    authSuccess(username) {
+      this.state.authenticated = true;
+      this.state.username = username;
```
Haviles04, line 23: Need to remove state here

src/stores/auth.js
```diff
@@ -99,24 +64,22 @@ export default {
           },
           function handleResponse(res, jwres) {
             if (jwres.statusCode === 200) {
-              context.commit('setMustReauthenticate', false);
+              this.mustReauthenticate = false;
               const pNum =
-                res.pNum ?? getPlayerPnumByUsername(context.rootState.game.players, context.state.username);
+                res.pNum ?? getPlayerPnumByUsername(this.rootState.game.players, this.state.username);
```
Haviles04, line 69: Need to figure out rootState here

Haviles04:
I made a bunch of progress on this, I changed all the vueX store modules to pinia, updated router component, updated socket event files, navLink composables, gameView, and loginView..

I'm unsure if I'm reading this correctly or not, but pinia guide says not to useStore outside of a function outside of a component, this is causing me to have to declare the store in every function inside `/plugins/sockets/connectivityEvents.js`
https://pinia.vuejs.org/core-concepts/outside-component-usage.html

src/App.vue
```diff
@@ -31,7 +34,15 @@ export default {
   '$route.meta'({hideNavigation}){
     this.showNav = !hideNavigation;
   }
- }
+  },
+  created() {
+    //Pass store to window object on testing
+    if (window.cypress !== null) {
```
itsalaidbacklife, line 40: I think it's a capital C and that it will be undefined rather than null when missing
```suggestion
    if (window.Cypress) { 
```

src/components/GameView/GameDialogs.vue
```diff
@@ -99,57 +89,57 @@ export default {
     // TODO: Refactor and combine usage in GameView.vue
     mustReauthenticate: {
       get() {
-        return this.$store.state.auth.mustReauthenticate;
+        return this.authStore.mustReauthenticate;
       },
       set(val) {
-        this.$store.commit('setMustReauthenticate', val);
+        this.authStore.setMustReauthenticate(val);
```
itsalaidbacklife, line 95: It probably makes sense to remove any former mutation functions that just set a single variable and replace them with inline assignments
```suggestion
        this.authStore.mustReauthenticate = val;
```

src/components/GameView/FourDialog.vue
```diff
@@ -64,7 +67,7 @@ export default {
       },
     },
     hand() {
-      return this.$store.state.game.players[this.$store.state.game.myPNum].hand;
+      return this.gameStore.players[this.gameStore.myPNum].hand;
```
itsalaidbacklife, line 70: We can use the player getter to dry this up 
```suggestion
      return this.gameStore.player.hand;
```

src/components/GameView/GameDialogs.vue
```diff
@@ -69,23 +69,13 @@ export default {
...
-    ...mapGetters([
-      'discarding',
-      'player',
-      'opponent',
-      'opponentPointTotal',
-      'opponentQueenCount',
-      'playerWins',
-    ]),
+    ...mapStores(useGameStore),
+    ...mapStores(useAuthStore),
```
itsalaidbacklife, line 73: mapStores can input multiple stores
```suggestion
    ...mapStores(useGameStore, useAuthStore),
```

src/components/GameView/GameDialogs.vue
```diff
@@ -69,23 +69,13 @@ export default {
...
     game() {
       // TODO: Figure out a better way to do this, mapping the whole module is a
       // bit unusual-- the usages can probably be changed to only need a subset,
       // or moved to some sort of store method or getter
-      return this.$store.state.game;
+      return this.gameStore;
```
itsalaidbacklife, line 78: It feels a little weird to have this computed prop just to alias the gameStore that we already mapped. How about instead we replace the references to `game` with `gameStore`?

src/components/GameView/GameMenu.vue
```diff
@@ -145,24 +148,24 @@ export default {
...
+      await this.gameStore.requestConcede();
       this.shownDialog = '';
     },
     async requestStalemate() {
       this.loading = true;
       try {
-        await this.$store.dispatch('requestStalemate');
-        this.$store.commit('setWaitingForOpponentToStalemate', true);
+        await this.gameStore.requestStalemate();
+        this.gameStore.setWaitingForOpponentToStalemate(true);
```
itsalaidbacklife, line 158: This looks like another method that just sets a single value; if so, let's remove it
```suggestion
        this.gameStore.waitingForOpponentToStalemate = true;
```

itsalaidbacklife (commented):
Awesome work. Commenting with my initial thoughts on the files I got through

src/App.vue
```diff
@@ -31,7 +34,15 @@ export default {
...
   }
- }
+  },
+  created() {
+    //Pass store to window object on testing
+    if (window.Cypress !== null) { 
+    window.gameStore = useGameStore();
+    window.authStore = useAuthStore();
+    window.gameListStore = useGameListStore();
+    }
```
Haviles04, line 44: I'm wondering if it would be better to do 
```
window.store = { 
 gameStore : useGameStore(),
 authStore : useAuthStore(),
 gameListStore: useGameListStore(),
 };

itsalaidbacklife: Interesting idea. Ultimately it doesn't matter much either way, but I think it's more in the spirit of pinia to keep it flat like you already have it since that's closer to how you'd grab them with script setup. And really, what's the point in an additional layer of property access?

Haviles04: Cool, my thought was maybe putting it inside window.store would keep them all together, but you right it doesn't really matter. :)

seriouslysean: It might still be better to keep these in the cuttle namespace regardless of the store nesting, right now you're polluting the window object.

itsalaidbacklife (commented):
Looking really good. Left a couple nitpick comments

itsalaidbacklife (approved):
Very nicely done. This will be a significant QoL improvement for DevEx

seriouslysean (approved):
Major QOL, nice work 👍

PR #571, opened 2023-09-17
Feat: Enable Language Translation

### LINKED ISSUE #548: [Feature]: Enable language selection (translation/internationalization/localization) from `TheUserMenu`
### Feature Summary

Cuttle has the basics of a translation setup in place, but there is no UI method available for selecting your language. We should update `TheUserMenu` component to include options for language selection.

### Detailed Description

This update is dependent on #519 being completed (which reworks the navigation and creates TheUserMenu).

Clicking your username in the upper-left corner of `TheHeader` opens `TheUserMenu` which currently only displays your username and a button to log out:
image

We should update this menu (`TheUserMenu`) to additionally show your currently selected language (using i18n) and add a secondary menu that shows the available languages to select:
image

This should render the selection from the available list of languages stored in `$i18n` in a similar way to this initial implementation, which was removed due to navigational challenges (there was nowhere to put it): https://github.com/cuttle-cards/cuttle/pull/457/commits/c027bf3e5fe8b1a2f7482f1e76f03224d017e854

Selecting the language should set `$i18n.locale` so as to automatically translate the pages that have translation available.

haaarsh619:
## Issue number #548 

 Resolves #548 

## Please check the following

- For New Features:

### DIFF
--- a/src/components/TheHeader.vue
+++ b/src/components/TheHeader.vue
@@ -17,10 +17,10 @@
       <v-toolbar-items v-if="!mobile" class="hidden-xs-only">
         <v-tabs class="pa-2">
           <v-tab
-            v-for="({ text, icon, page }, i) in pageLinks"
+            v-for="({ text, icon, page, cy_name }, i) in pageLinks"
             :key="i"
             variant="text"
-            :data-cy="text"
+            :data-cy="cy_name"
             :title="text"
             :to="page"
           >
--- a/src/components/TheHeader.vue
+++ b/src/components/TheHeader.vue
@@ -17,10 +17,10 @@
       <v-toolbar-items v-if="!mobile" class="hidden-xs-only">
         <v-tabs class="pa-2">
           <v-tab
-            v-for="({ text, icon, page, cy_name }, i) in pageLinks"
+            v-for="({ text, icon, page, cyName }, i) in pageLinks"
             :key="i"
             variant="text"
-            :data-cy="cy_name"
+            :data-cy="cyName"
             :title="text"
             :to="page"
           >
--- a/src/i18n.js
+++ b/src/i18n.js
@@ -4,22 +4,22 @@ import { createI18n } from 'vue-i18n';
 
 // TODO we should lazyload non global translations on a per page basis
 // https://vue-i18n.intlify.dev/guide/advanced/lazy.html
-import English from '@/translations/en.json';
-import Español from '@/translations/es.json';
-import Française from '@/translations/fr.json';
+import en from '@/translations/en.json';
+import es from '@/translations/es.json';
+import fr from '@/translations/fr.json';
 
 const messages = {
- English,
- Español,
- Française,
+ en,
+ es,
+ fr,
 };
 
 const i18n = createI18n({
   legacy: false,
   globalInjection: true,
-  locale: 'English',
+  locale: 'en',
   // https://vue-i18n.intlify.dev/guide/essentials/fallback.html
-  fallbackLocale: 'English',
+  fallbackLocale: 'en',
   messages,
 });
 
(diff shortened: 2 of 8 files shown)


src/i18n.js
```diff
@@ -4,22 +4,22 @@ import { createI18n } from 'vue-i18n';
 
 // TODO we should lazyload non global translations on a per page basis
 // https://vue-i18n.intlify.dev/guide/advanced/lazy.html
-import en from '@/translations/en.json';
-// import es from '@/translations/es.json';
-// import fr from '@/translations/fr.json';
+import English from '@/translations/en.json';
+import Español from '@/translations/es.json';
+import Française from '@/translations/fr.json';
```
seriouslysean, line 9: Don’t use accents in the code, stick to the locale abbreviations that I had before.

itsalaidbacklife: It seems like it would be nice if the languages had a display name in addition to the variable name that's used by the code. It seems like that's sortof baked into the syntax of the Vue i18n implementation, but it's strange to me that the options we display in the menu are seemingly limited to text that makes for reasonable variable names

haaarsh619: So how about i go with this Approch 
```
          <v-list-item
            v-for="(lang, i) in $i18n.availableLocales"
            :key="`${i}-${lang}`"
            :value="lang"
            @click="$i18n.locale = lang"
            :data-lang="lang"
          >
  -->        {{langValue[lang]}}
          </v-list-item>
```
```
const langValue = {en:'English', es:'Español'} 
```

seriouslysean: It’s a little trickier than that, the names in the drop-down menu would need to change by locale also so we’ll probably need a language specific locale map. Could maybe en put those as localized strings in the translation json files and use slugs.

itsalaidbacklife: To simplify, let's keep the original names for now and revisit them later

itsalaidbacklife (commented):
This looks good. I'd like to clean up the testing function and get some input from @seriouslysean about whether to translate testing selectors, then we can bump the version and merge

tests/e2e/specs/out-of-game/localization.spec.js
```diff
@@ -0,0 +1,37 @@
...
+  cy.get('[data-cy="language-menu"]').click();
+
+  //Selects the language 
+  cy.get('[data-lang="' + name + '"]').click();
+
+  // checks elements in navbar has changed language
+  cy.get('[data-cy="' + data.global.play + '"]').should('contain', data.global.play);
+  cy.get('[data-cy="user-menu"]').click();
+  cy.get('[data-nav="' + data.global.logout + '"]').should('contain', data.global.logout);
+}
```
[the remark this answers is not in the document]
haaarsh619: Yes because in this 
```
const menuItems = computed(() => {
  return [{ text: t('global.logout'), icon: 'logout', page: { name: ROUTE_NAME_LOGOUT } }];
});
```
this text is passing in data-nav so it is updating when language is changed.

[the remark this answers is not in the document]
seriouslysean: Honestly I’d leave the data selectors using the locale abbreviations so it’s not even something we need to translate at all and more consistent for testing.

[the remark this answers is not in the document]
haaarsh619: yes i did that in recent commit can you verify once.

src/components/TheHeader.vue
```diff
@@ -17,10 +17,10 @@
       <v-toolbar-items v-if="!mobile" class="hidden-xs-only">
         <v-tabs class="pa-2">
           <v-tab
-            v-for="({ text, icon, page }, i) in pageLinks"
+            v-for="({ text, icon, page, cy_name }, i) in pageLinks"
```
itsalaidbacklife, line 20: Let's use `camelCase` for variable and attribute names instead
```suggestion
            v-for="({ text, icon, page, cyName }, i) in pageLinks"
```

PR #573, opened 2023-09-23
Fix #560  Added 4px margin between icons and tab labels in TheHeader

### LINKED ISSUE #560: [Feature]: Add 4px margin between icons and tab labels in `TheHeader`

dushyantpant5:
## Issue number #560 

- Resolves #

## Please check the following

-  All tests ran successfully
- The code is formatted properly , according to the guidelines
- For New Features:
  - No tests added
  - No documentation update

Screenshot 2023-09-23 at 7 55 43 PM

### DIFF
--- a/src/components/TheHeader.vue
+++ b/src/components/TheHeader.vue
@@ -23,7 +23,7 @@
             :data-cy="cyName"
             :title="text"
             :to="page"
-            :style="{margin : '4px' }"
+            :style="{margin :'4px'}"
           >
             <v-icon
               :icon="`mdi-${icon}`"
--- a/src/components/TheHeader.vue
+++ b/src/components/TheHeader.vue
@@ -23,6 +23,7 @@
             :data-cy="cyName"
             :title="text"
             :to="page"
+            :style="{margin : '4px' }"
           >
             <v-icon
               :icon="`mdi-${icon}`"


src/components/TheHeader.vue
```diff
@@ -23,6 +23,7 @@
             :data-cy="cyName"
             :title="text"
             :to="page"
+            :style="{margin :'4px'}"
```
itsalaidbacklife, line 26: The intent of the issue is to add a margin between the icon and text, not between the tabs themselves. That means we should move this onto the icon itself and specifically add margin-right. 

Additionally, we normally favor using classes that utilize preexisting Vuetify styles ((docs:)[https://vuetifyjs.com/en/styles/spacing/]), in this case for a right margin of 4px (a single unit of Vuetify's spacing helpers), that would be 
```class="mr-1"```

File: src/components/TheHeader.vue [hunk 2]
```diff
@@ -23,6 +23,7 @@
...
           >
             <v-icon
               :icon="`mdi-${icon}`"
```
itsalaidbacklife, line 29: This is where we should put the margin class so that it goes between the icon and text

itsalaidbacklife (commented):
Thanks for your contribution. This is the right idea, though there is some fine tuning needed for it to be merged. Details are in the comments. If you have any questions, feel free to ask them in the comments

src/components/TheHeader.vue
dushyantpant5: I opened a new PR with required changes. Please check

Issue #549, opened 2023-09-03
[Feature]: Store language preference in local storage and set it when bootstrapping client


itsalaidbacklife:
### Feature Summary

When a user selects their preferred language, we should store this in localstorage using our `local-storage-utils` and then use this as their default language when they visit the site, until the user changes it.

### Detailed Description

This depends on #548 . Once we have a method for selecting the user's preferred language, we should track this preference in localstorage so that returning users can continue to view the site in their language of choice. 

To accomplish this, we can use Cuttle's existing `local-storage-utils` to get the user's previously selected language (if it is set) when the client starts (in `src/main.js`) and use that to set `$i18n`'s `locale` so that the app renders text in the user's selected language. 

In addition, when a user selects a language from `TheUserMenu`, we should store this preference in localStorage so that the next time the user visits the app, the app uses their previously selected language.

This can be done by storing a new value in localStorage called `preferredLocale` when the user selects a language, and reading from this value to set `$i18n`'s `locale` during app initialization

PR #574, opened 2023-09-23
feat: using preferred language from local storage

haaarsh619:
## Issue number #549 

Resolves [[Feature]: Store language preference in local storage and set it when bootstrapping client #549](https://github.com/cuttle-cards/cuttle/issues/549)
## Please check the following

- For New Features:

### DIFF
--- a/src/i18n.js
+++ b/src/i18n.js
@@ -13,13 +13,17 @@ const messages = {
  es,
  fr,
 };
-// i18n.global.locale.value = getLocalStorage('preferredLocale') ?? 'en';
+
+const preferredLocale = getLocalStorage('preferredLocale');
+const fallbackLocale = 'en';
+const locale = Object.keys(messages).includes(preferredLocale) ? preferredLocale : fallbackLocale; 
+
 const i18n = createI18n({
   legacy: false,
   globalInjection: true,
-  locale: getLocalStorage('preferredLocale') ?? 'en',
+  locale,
   // https://vue-i18n.intlify.dev/guide/essentials/fallback.html
-  fallbackLocale: 'en',
+  fallbackLocale,
   messages,
 });
 
--- a/src/i18n.js
+++ b/src/i18n.js
@@ -1,7 +1,7 @@
 // https://vue-i18n.intlify.dev/
 
 import { createI18n } from 'vue-i18n';
-
+import { getLocalStorage } from '_/utils/local-storage-utils';
 // TODO we should lazyload non global translations on a per page basis
 // https://vue-i18n.intlify.dev/guide/advanced/lazy.html
 import en from '@/translations/en.json';
@@ -13,11 +13,11 @@ const messages = {
  es,
  fr,
 };
-
+// i18n.global.locale.value = getLocalStorage('preferredLocale') ?? 'en';
 const i18n = createI18n({
   legacy: false,
   globalInjection: true,
-  locale: 'en',
+  locale: getLocalStorage('preferredLocale') ?? 'en',
   // https://vue-i18n.intlify.dev/guide/essentials/fallback.html
   fallbackLocale: 'en',
   messages,
--- a/src/main.js
+++ b/src/main.js
@@ -5,7 +5,7 @@ import router from '@/router';
 import store from '@/store/store';
 import i18n from '@/i18n';
 import { initCuttleGlobals } from '_/utils/config-utils';
-
+import { getLocalStorage } from '_/utils/local-storage-utils';
 import App from '@/App.vue';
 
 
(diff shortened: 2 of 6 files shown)


src/main.js
```diff
@@ -23,6 +23,9 @@ app.use(store);
 // Add localization to vue
 app.use(i18n);
 
+//Using Language from Loacal Storage if not found than by defualt English is selceted
```
seriouslysean, line 26:
```suggestion
// Use language from Local Storage if it exists
```

src/main.js
```diff
@@ -23,6 +23,9 @@ app.use(store);
...
+i18n.global.locale.value = getLocalStorage('preferredLocale') ?? 'en';
```
seriouslysean, line 27: I'm not sure this is the right way to do it. Would it be better to set the default initially when loading i18n in the first place instead of switching after the app initializes?

haaarsh619: like before app uses the i18n right.

seriouslysean: More like in here, https://github.com/cuttle-cards/cuttle/blob/main/src/i18n.js, as we're instantiating i18n.

We will also need to validate that it's a valid locale in case anyone changes the value. Maybe validate against the files that exist in the locale directory. @itsalaidbacklife, Penny for your thoughts.

itsalaidbacklife (commented):
This looks great. Could you add and/or update a test for this? Basically that it defaults to English and then that Lang preference is preserved on logout and reload?

src/main.js
```diff
@@ -20,6 +20,9 @@ app.use(vuetify);
 // Add vuex store to vue
 app.use(store);
 
+// Use language from Local Storage if it exists
+i18n.global.locale.value = getLocalStorage('preferredLocale') ?? 'en';
```
itsalaidbacklife, line 24: Per @seriouslysean's point let's fold this config into `@/i18n.js` itself on line 20 where we initialize the selected language. Then we can remove this line and the import on line 8 here

tests/e2e/specs/out-of-game/localization.spec.js
```diff
@@ -1,37 +1,56 @@
...
+    cy.get('[data-cy="user-menu"]').click();
+    cy.get('[data-nav="Log Out"]').should('contain', lang.global.logout);
+  };
 
-  // checks elements in navbar has changed language
-  cy.get('[data-cy="Play"]').should('contain', lang.global.play);
-  cy.get('[data-cy="user-menu"]').click();
-  cy.get('[data-nav="Log Out"]').should('contain', lang.global.logout);
-}
+  it('Should check translation for default', () => {
```
itsalaidbacklife, line 45:
```suggestion
  it('Should check translation for English (default)', () => {
```

src/main.js
```diff
@@ -20,6 +19,8 @@ app.use(vuetify);
 // Add vuex store to vue
 app.use(store);
 
+// Use language from Local Storage if it exists
+
```
itsalaidbacklife, line 23:
```suggestion
```

src/i18n.js
```diff
@@ -13,11 +13,11 @@ const messages = {
  es,
  fr,
 };
-
+// i18n.global.locale.value = getLocalStorage('preferredLocale') ?? 'en';
```
seriouslysean, line 16: Remove this comment

src/i18n.js
```diff
@@ -13,11 +13,11 @@ const messages = {
...
 const i18n = createI18n({
   legacy: false,
   globalInjection: true,
-  locale: 'en',
+  locale: getLocalStorage('preferredLocale') ?? 'en',
```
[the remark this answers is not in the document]
seriouslysean: It does seem like we'd want to validate that the language exists though before setting it (even if it'll automatically use the fallback). Something like this:
```
const preferredLocale = getLocalStorage('preferredLocale');
const fallbackLocale = 'en';
const locale = Object.keys(messages).contains(locale) ?? fallbackLocale;
const i18n = createI18n({ locale, fallbackLocale });
```

Then you could also inline `fallbackLocale` with the new variable.

seriouslysean (changes requested):
Really close, just one more change imo.

haaarsh619:
@seriouslysean Something like this 
```
const preferredLocale = getLocalStorage('preferredLocale');
const fallbackLocale = 'en';
const locale = Object.keys(messages).contains(preferredLocale) ? preferredLocale : fallbackLocale; 

const i18n = createI18n({
  legacy: false,
  globalInjection: true,
  locale,
  fallbackLocale,
  messages,
});
```

seriouslysean:
@haaarsh619 yep, as long as that works locally it look 👌🏼 to me.

itsalaidbacklife (approved):
Very well done

seriouslysean (approved):
This looks great, thanks @haaarsh619!

Issue #572, opened 2023-09-21
[Bug]: Discord Button Not Responsive  In login Page


haaarsh619:
### Bug Summary

Discord button is not responsive on mobile

### Detailed Description

The discord button in LoginView.vue is not responsive on mobile  the text and icon overflow.
it occurs when French or Spanish Language is Selected. This Issue occurs when language is Changed.

![image](https://github.com/cuttle-cards/cuttle/assets/72140070/3c0c2c16-4045-42b6-8fdc-b1588e485ff4)

### Steps to Reproduce

1. After Login Change the language from Menu  
2. Then Logout 
3. Then Scroll to the Login section there  is discord button text and icon in login section.

dushyantpant5:
@haaarsh619 , can I try it?

haaarsh619:
> @haaarsh619 , can I try it?

Yes sure, 
If you have doubts  or need any help you can also use the #code channel of our [discord](https://discord.gg/6vQCFa6nvP) .

dushyantpant5:
<img width="1147" alt="Screenshot 2023-09-23 at 9 38 45 PM" src="https://github.com/cuttle-cards/cuttle/assets/60127798/206e429a-1305-4c34-83ee-2a1c827fd1c1">

I fixed it with adding <span> and doing inline styling. Is there any style class in Vue to do it? Or this I commit it in this way only?

haaarsh619:
I am not sure can you commit and then @itsalaidbacklife will review it.

Shahidkhatrii:
Hi @haaarsh619 I would like to fix it if this issue is still open

itsalaidbacklife:
Hi @Shahidkhatrii there is an open PR #577 that addresses this. That will likely be merged soon so it would be better to pick up another issue. If you're interested in frontend try filtering for issues that are labeled with "frontend" and "good first issue". If you'd like to chat with myself or other devs about finding a suitable one, you can reach out on the #code channel of the Cuttle [discord](https://discord.gg/arWabYp8SC)

PR #577, opened 2023-09-24
Error #572 Fix Unresponsive Discord Button

dushyantpant5:
## #572 

Relevant [#572 ](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
- Resolves #

## Please check the following

- [x] Yes all the test passed
- [x] Code is properly formatted
- For New Features:
  - [x] No 
  - [x] No
Screenshot 2023-09-23 at 9 38 45 PM


src/views/LoginView.vue
```diff
@@ -117,7 +117,7 @@
                 target="_blank"
               >
                 <img class="discord" src="/img/loginView/logo-discord.svg">
-                {{ t('login.joinDiscord') }}
+                <span style="white-space: normal;">{{ t('login.joinDiscord') }}</span>
```
itsalaidbacklife, line 120: For css that doesn't have an existing global Vuetify style, we generally use a class in the html and then select it with css. Can you move this out of sn online style and into the css block? 

Can you also show a screenshot of before and after the change?

itsalaidbacklife (commented):
Nice work l. Let's use css instead of an in-line style

src/views/LoginView.vue
dushyantpant5: Okay will be working on it

dushyantpant5:
Screenshot 2023-09-24 at 6 02 46 PM

Result after adding classe

src/views/LoginView.vue
```diff
@@ -367,5 +367,9 @@ export default {
   .video-container__wrapper {
     width: 85%;
   }
+
+  .discordButton{
+    white-space: normal;
+  }
```
itsalaidbacklife, line 373: Is this intended to be inside the media query? Atm it only applies to small devices because it's in the media query. Would it make more sense to move it outside and apply this style for all devices?

itsalaidbacklife (commented):
Is this intended to be in the media query?

src/views/LoginView.vue
dushyantpant5: @itsalaidbacklife It was intended to be inside media query as the test is out of button in mobile view. Also if we will move that outside it will work the same.
Let me know if I have to make changes.

src/views/LoginView.vue
itsalaidbacklife: I think it makes more sense to move it outside the media query because we want it to linebreak on any device that needs it based on whatever the button size is. So yes, please move this above

dushyantpant5: @itsalaidbacklife  Made the changes , Please check

dushyantpant5:
@itsalaidbacklife , Please check the changes

itsalaidbacklife (commented):
Hi @dushyantpant5 nicely done. Can you please update this from main and then run `npm run version: patch` one time and push? That will update the version number. Then this is ready to merge!

dushyantpant5:
> Hi @dushyantpant5 nicely done. Can you please update this from main and then run `npm run version: patch` one time and push? That will update the version number. Then this is ready to merge!

Done with the task , @itsalaidbacklife

itsalaidbacklife (approved):
Nicely done. Thanks for your contribution!

PR #586, opened 2023-09-29
set the colors of the selected bottom nav elements to newPrimary

### LINKED ISSUE #583: [Feature]: Set color of of nav elements in `TheFooter.vue` to `'newPrimary'`
### Feature Summary

The navigation elements in `TheHeader` use `newPrimary` (pink) for the active element, but `TheFooter` does not, making it inconsistent. We should make `The

### Detailed Description

See images. `TheFooter.vue` navigation (used on small devices) does not color the selected element:
image

But `TheHeader.vue` does:
image

We should use the technique to set the color of the active link in `TheFooter.vue` to `'newPrimary'` which is the pink seen in `TheHeader`

aybeedee:
## Issue

Relevant [issue number](https://github.com/cuttle-cards/cuttle/issues/583)
- Resolves #583 

## Please check the following

- For New Features:
  - Has the documentation been updated accordingly: No need, minor UI change.

- Similar implementation used to set the color as is done in `TheHeader.vue`
- Can see result in image below
- **Note**: The primary color clashes a bit with the brown background in dark mode. For that, an option considered was to change the variant of the `v-btn` to `plain` instead of `text` to hide background. However, did not proceed with it since it also lowers the opacity of the text. More details [here](https://vuetifyjs.com/en/components/buttons/#variants).

footer-color-changes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cuttle",
-  "version": "5.5.1",
+  "version": "5.6.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "cuttle",
-      "version": "5.5.1",
+      "version": "5.6.0",
       "dependencies": {
         "chart.js": "^4.3.3",
         "connect-redis": "3.0.2",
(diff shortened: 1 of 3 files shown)


aybeedee:
bumped it

PR #601, opened 2023-09-29
Feat #552 Add localization to lobby view

### LINKED ISSUE #552: [Feature]: Add translations/localization/internationalization to `LobbyView`
### Feature Summary

With the foundations for localization in place in the app, we can start adding translations to individual pages by pulling their text out of the html and into the local files in `src/translations`. We can now add this to `LobbyView`.

### Detailed Description

Look at `src/views/LoginView` for an example of a page that is correctly using `$i18n` to localize/translate content. The user-visible strings have been removed from the `<template>` code and placed into a "login" section of the translation files in `src/translations`. 

The page then imports and consumes the `useI18n()` composable to grab the `t()` function used to render/translate text by pulling from the appropriate language's json file. This is done using a `setup()` block of the page like so:

```
  setup() {
    const { t } = useI18n();
    return { t };
  },
```

From there, we can grab a given string by name in the appropriate locale using `{{ t('login.joinDiscord') }}` since each translation json in this case has a `login` attribute for all the text in the login page, with a `joinDiscord` child attribute for the text to show in the 'Join our discord community' button.

We should do the same thing to the `LobbyView`, by extracting all the text in the page into a `lobby` section of each translation json file (e.g. `src/translations/en.json`) and then configuring `LobbyView` with the `useI18n()` composable and rendering all its text with `t()` to translate it to the user's selected locale.

Haviles04:
## Issue number

- Resolves #552 

## Please check the following

- For New Features:

Added translations to lobby view

### DIFF
--- a/src/components/LobbyPlayerIndicator.vue
+++ b/src/components/LobbyPlayerIndicator.vue
@@ -15,12 +15,13 @@
       class="ready-overlay text-truncate"
       :style="{ 'font-size': readyFontSize, 'padding-bottom': readyPadding }"
     >
-      READY
+      {{ t('lobby.ready') }}
     </div>
   </div>
 </template>
 
 <script>
+import { useI18n } from 'vue-i18n';
 export default {
   name: 'LobbyPlayerIndicator',
   props: {
@@ -33,9 +34,13 @@ export default {
       default: false,
     },
   },
+  setup() {
+    const { t } = useI18n();
+    return { t };
+  },
   computed: {
     message() {
-      return this.playerUsername || 'Invite';
+      return this.playerUsername || this.t('lobby.invite');
     },
     playerPadding() {
       return this.$vuetify.display.mdAndUp ? '72px' : '32px';
--- a/src/translations/en.json
+++ b/src/translations/en.json
@@ -23,5 +23,12 @@
     "content2": "Be the first to score 21 points in this explosive battle of wits. Mount a valient offense while disupting your opponent with dastardly tricks. Do you have what it takes to become the Lord of the Deep?",
     "readRules": "Read the Rules",
     "joinDiscord": "Join our discord community"
+  },
+  "lobby":{
+    "lobbyFor": "Lobby for",
+    "ready": "READY",
+    "unready": "UNREADY",
+    "exit": "EXIT",
+    "invite": "Invite"
   }
 }
--- a/src/translations/es.json
+++ b/src/translations/es.json
@@ -23,5 +23,12 @@
     "content2": "Sé el primero en alcanzar 21 puntos en esta explosiva batalla de ingenio. Monta una ofensiva valiente mientras interrumpes a tu oponente con trucos malvados. ¿Tienes lo necesario para convertirte en el Señor de las Profundidades?",
     "readRules": "Leer las reglas",
     "joinDiscord": "Únete a nuestra comunidad en Discord"
+  },
+  "lobby":{
+    "lobbyFor": "Lobby para",
+    "ready": "Preparado",
+    "unready": "NO PREPARADO",
+    "exit": "SALIR",
+    "invite": "Invitar"
   }
 }
--- a/src/translations/es.json
+++ b/src/translations/es.json
@@ -26,7 +26,7 @@
   },
   "lobby":{
     "lobbyFor": "Lobby para",
-    "ready": "Preparado",
+    "ready": "PREPARADO",
     "unready": "NO PREPARADO",
     "exit": "SALIR",
     "invite": "Invitar"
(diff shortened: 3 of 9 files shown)


src/translations/en.json
```diff
@@ -23,5 +23,12 @@
     "content2": "Be the first to score 21 points in this explosive battle of wits. Mount a valient offense while disupting your opponent with dastardly tricks. Do you have what it takes to become the Lord of the Deep?",
     "readRules": "Read the Rules",
     "joinDiscord": "Join our discord community"
+  },
+  "lobby":{
```
dan-sarich, line 27: Question: Is this verbiage used else where in the application? To avoid duplication, maybe it would be better to keep them at the global level instead of scoped only for the lobby.

src/translations/es.json
```diff
@@ -23,5 +23,12 @@
     "content2": "Sé el primero en alcanzar 21 puntos en esta explosiva batalla de ingenio. Monta una ofensiva valiente mientras interrumpes a tu oponente con trucos malvados. ¿Tienes lo necesario para convertirte en el Señor de las Profundidades?",
     "readRules": "Leer las reglas",
     "joinDiscord": "Únete a nuestra comunidad en Discord"
+  },
+  "lobby":{
+    "lobbyFor": "Lobby para",
+    "ready": "Preparado",
```
dan-sarich, line 29: Issue: Value for 'ready' is all caps in `en.json` Keep these casing consistent between locales.

src/translations/fr.json
```diff
@@ -23,5 +23,12 @@
     "content2": "Soyez le premier à marquer 21 points dans cette bataille explosive d'intelligence. Montez une offense vaillante tout en perturbant votre adversaire avec des astuces perfides. Avez-vous ce qu'il faut pour devenir le Seigneur des Profondeurs ?",
     "readRules": "Lire les règles",
     "joinDiscord": "Rejoignez notre communauté sur Discord"
+  },
+  "lobby":{
+    "lobbyFor": "Faire pression pour",
+    "ready": "prêt",
```
dan-sarich, line 29: Issue: Value for 'ready' is all caps in `en.js` Keep these casing consistent between locales.

src/views/LobbyView.vue
```diff
@@ -94,12 +99,10 @@ export default {
...
     },
     iAmReady() {
-      return this.gameStore.myPNum === 0
-        ? this.gameStore.p0Ready
-        : this.gameStore.p1Ready;
+      return this.gameStore.myPNum === 0 ? this.gameStore.p0Ready : this.gameStore.p1Ready;
     },
     readyButtonText() {
-      return this.iAmReady ? 'UNREADY' : 'READY';
+      return this.iAmReady ? this.t('lobby.unready') : this.t('lobby.ready');
```
dan-sarich, line 105: Issue: avoid duplication
```suggestion
      return this.t(this.iAmReady ? 'lobby.unready' : 'lobby.ready') 
```

src/translations/fr.json
Haviles04: Good call!

src/components/LobbyPlayerIndicator.vue
```diff
@@ -15,12 +15,13 @@
       class="ready-overlay text-truncate"
       :style="{ 'font-size': readyFontSize, 'padding-bottom': readyPadding }"
     >
-      READY
+      {{ t('lobby.ready') }}
```
seriouslysean, line 18: I wonder if we should do the caps via css, would avoid the need to capitalize or case in any specific way in the locale files.

seriouslysean (commented):
Just one question, otherwise ok with me and will leave final approval for @itsalaidbacklife.

src/translations/en.json
itsalaidbacklife: @dan-sarich good question. All the text that's been put under "lobby" here is specific to the lobby and does not appear elsewhere. The text that is shared across pages in the app e.g. the names of pages in the nav links are under the "global" section above. 

The lobby has very limited text overall and it happens not to have any significant overlap with text in other pages so everything here is specific to the lobby

src/components/LobbyPlayerIndicator.vue
itsalaidbacklife: Yeah it's an interesting question especially considering that some pages will want upper vs lower vs first letter upper cased. In reality the buttons already have default Vuetify css to uppercase them per material design spec so in that sense we could do whatever we want for button text.

itsalaidbacklife (approved):
Nicely done. We're chipping away at localization slowly but surely

itsalaidbacklife (approved):
Nicely done. We're chipping away at localization slowly but surely

PR #604, opened 2023-10-02
DevEx #189 refactor card names and rule text out of the database

### LINKED ISSUE #189: [DevEx]: Refactor card names and rule text out of the database
### Improvement Summary

There is no good reason to include the text associated with cards in the database. Instead the utility functions that generate the payloads sent to the client through the socket (i.e. `populateGame`) should use a sub-utility to append this information at runtime. That way we'll save database space and be able to more flexibly change this information in the future

### Detailed Description

This will involve updating the `Card` model, the `cardService` 's `createCard()` utility, and the `populateGame()` function which generates the DTO sent over the socket

### LINKED ISSUE #554: [Feature]: Translate the ruleText on each card (making language selection apply to the `MoveChoiceCard`s)
### Feature Summary

We should enable translation/localization/internationalization of the move choices that appear when you click a card.

### Detailed Description

This depends on #189. 

Once the `ruleText` for each card is generated clientside, rather than in the backend, we can apply translations to the Move Choices in the game (the options that are displayed when the user selects a card in their hand):
image

There are two essential places that need to be updated in order to translate this appropriately:

1. The `MoveChoiceOverlay` needs to correctly set the `displayName` of all the moves using `t()` from `$i18n` like the other pages
2. The `ruleText` on each card needs to be rendered with `t()` so that the description under each move i.e. what an Ace does is also translated

This will require updating the translation files in `src/translations` e.g. `src/translations/en.json` with a new attribute called `game` that contains all the translations for inside the game and adding a subsection `game.cards` for the description of the individual cards names and `ruleText`. Then the `game` store module can use these to create the `Card` objects that are consumed by the application so that all the cards are appropriately translated.

Haviles04:
## Issue number

- Resolves #189 
- Resolves #554

## Please check the following

- For New Features:

### DIFF
--- a/src/stores/game.js
+++ b/src/stores/game.js
@@ -31,7 +31,7 @@ class GameCard {
     this.suit = card.suit;
     this.rank = card.rank;
     this.name = str_rank + str_suit;
-    this.attachments = card.attachments.map(attachment => createGameCard(attachment));
+    this.attachments = card.attachments?.map(attachment => createGameCard(attachment));
   }
 }
 const createGameCard = (card) => {
--- a/src/stores/game.js
+++ b/src/stores/game.js
@@ -158,6 +158,7 @@ export const useGameStore = defineStore('game', {
   },
   actions: {
     updateGame(newGame) {
+      console.log(newGame.players);
       if (Object.hasOwnProperty.call(newGame, 'lastEvent')) {
         if (Object.hasOwnProperty.call(newGame.lastEvent, 'change')) {
           this.lastEventChange = newGame.lastEvent.change;
@@ -191,6 +192,8 @@ export const useGameStore = defineStore('game', {
         this.players = newGame.players.map((player) => ({
           ...player,
           hand: player.hand?.map((card) => new GameCard(card)) ?? player.hand,
+          points: player.points?.map((card) => new GameCard(card) ?? player.points),
+          faceCards: player.faceCards?.map((card) => new GameCard(card) ?? player.faceCards),
         }));
       if (Object.hasOwnProperty.call(newGame, 'spectatingUsers')) {
         this.spectatingUsers = newGame.spectatingUsers;
@@ -201,16 +204,16 @@ export const useGameStore = defineStore('game', {
         this.topCard = new GameCard(newGame.topCard);
       } else this.topCard = null;
 
-      if (Object.hasOwnProperty.call(newGame, 'secondCard'))
+      if (Object.hasOwnProperty.call(newGame, 'secondCard')) {
         this.secondCard = new GameCard(newGame.secondCard);
-      else this.secondCard = null;
+      } else this.secondCard = null;
 
       if (Object.hasOwnProperty.call(newGame, 'oneOff')) this.oneOff = cloneDeep(newGame.oneOff);
       else this.oneOff = null;
 
-      if (Object.hasOwnProperty.call(newGame, 'oneOffTarget'))
+      if (Object.hasOwnProperty.call(newGame, 'oneOffTarget')) {
         this.oneOffTarget = cloneDeep(newGame.oneOffTarget);
-      else this.oneOffTarget = null;
+      } else this.oneOffTarget = null;
 
       if (Object.hasOwnProperty.call(newGame, 'isRanked')) this.isRanked = newGame.isRanked;
       if (Object.hasOwnProperty.call(newGame, 'currentMatch')) this.currentMatch = newGame.currentMatch;
--- a/src/stores/game.js
+++ b/src/stores/game.js
@@ -50,6 +50,7 @@ class GameCard {
     this.rank = card.rank;
     this.name = str_rank + str_suit;
     this.ruleText = ruleText[card.rank - 1];
+    this.attachedTo = card.attachedTo;
   }
 }
 
--- a/src/stores/game.js
+++ b/src/stores/game.js
@@ -16,7 +16,6 @@ function queenCount(player) {
 
 class GameCard {
   constructor(card) {
-    if (!card) return;
     const str_rank =
       {
         1: 'A',
@@ -35,6 +34,10 @@ class GameCard {
     this.attachments = card.attachments;
   }
 }
+const createGameCard = (card) => {
+  if (!card) return null;
+  return new GameCard(card);
+};
 
 export const useGameStore = defineStore('game', {
   state: () => ({
@@ -141,7 +144,6 @@ export const useGameStore = defineStore('game', {
   },
   actions: {
     updateGame(newGame) {
-      console.log(newGame);
       if (Object.hasOwnProperty.call(newGame, 'lastEvent')) {
         if (Object.hasOwnProperty.call(newGame.lastEvent, 'change')) {
           this.lastEventChange = newGame.lastEvent.change;
@@ -164,9 +166,9 @@ export const useGameStore = defineStore('game', {
       if (Object.hasOwnProperty.call(newGame, 'turn')) this.turn = newGame.turn;
       if (Object.hasOwnProperty.call(newGame, 'chat')) this.chat = cloneDeep(newGame.chat);
       if (Object.hasOwnProperty.call(newGame, 'deck'))
-        this.deck = newGame.deck?.map((card) => new GameCard(card));
+        this.deck = newGame.deck?.map((card) => createGameCard(card));
       if (Object.hasOwnProperty.call(newGame, 'scrap'))
-        this.scrap = newGame.scrap?.map((card) => new GameCard(card));
+        this.scrap = newGame.scrap?.map((card) => createGameCard(card));
       if (Object.hasOwnProperty.call(newGame, 'log')) this.log = cloneDeep(newGame.log);
       if (Object.hasOwnProperty.call(newGame, 'name')) this.name = newGame.name;
       if (Object.hasOwnProperty.call(newGame, 'p0Ready')) this.p0Ready = newGame.p0Ready;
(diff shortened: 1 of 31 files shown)


src/stores/game.js
```diff
@@ -140,23 +179,30 @@ export const useGameStore = defineStore('game', {
...
       if (Object.hasOwnProperty.call(newGame, 'log')) this.log = cloneDeep(newGame.log);
       if (Object.hasOwnProperty.call(newGame, 'name')) this.name = newGame.name;
       if (Object.hasOwnProperty.call(newGame, 'p0Ready')) this.p0Ready = newGame.p0Ready;
       if (Object.hasOwnProperty.call(newGame, 'p1Ready')) this.p1Ready = newGame.p1Ready;
       if (Object.hasOwnProperty.call(newGame, 'passes')) this.passes = newGame.passes;
-      if (Object.hasOwnProperty.call(newGame, 'players')) this.players = cloneDeep(newGame.players);
+      if (Object.hasOwnProperty.call(newGame, 'players'))
+        this.players = newGame.players.map((player) => ({
+          ...player,
+          hand: player.hand?.map((card) => new GameCard(card)) ?? player.hand,
```
itsalaidbacklife, line 193: Also need their points and faceCards

src/stores/game.js
```diff
@@ -14,6 +14,45 @@ function queenCount(player) {
...
+      }[card.rank] ?? card.rank;
+    // Stringify Suit
+    const str_suit = ['♣️', '♦️', '♥️', '♠️'][card.suit];
+    this.createdAt = card.createdAt;
+    this.updatedAt = card.updatedAt;
+    this.id = card.id;
+    this.suit = card.suit;
+    this.rank = card.rank;
+    this.name = str_rank + str_suit;
+    this.ruleText = ruleText[card.rank - 1];
```
itsalaidbacklife, line 52: We also need the attachments, which should already exist on the card coming from the server

src/stores/game.js
```diff
@@ -140,23 +179,30 @@ export const useGameStore = defineStore('game', {
...
 
-      if (Object.hasOwnProperty.call(newGame, 'topCard')) this.topCard = cloneDeep(newGame.topCard);
-      else this.topCard = null;
+      if (Object.hasOwnProperty.call(newGame, 'topCard')) {
+        this.topCard = new GameCard(newGame.topCard);
+      } else this.topCard = null;
 
-      if (Object.hasOwnProperty.call(newGame, 'secondCard')) this.secondCard = cloneDeep(newGame.secondCard);
+      if (Object.hasOwnProperty.call(newGame, 'secondCard'))
+        this.secondCard = new GameCard(newGame.secondCard);
```
itsalaidbacklife, line 205: Let's wrap the if in {} since we need a newline

itsalaidbacklife (commented):
Great start

src/stores/game.js
```diff
@@ -14,6 +14,31 @@ function queenCount(player) {
...
+    this.suit = card.suit;
+    this.rank = card.rank;
+    this.name = str_rank + str_suit;
+    this.attachments = card.attachments;
+  }
+}
+const createGameCard = (card) => {
+  if (!card) return null;
+  return new GameCard(card);
+};
```
Haviles04, line 40: I'm unsure if this is a good way to do this, without this.. the ui shows some stuff that it shouldn't, specifically when looking for a `target` which = `store.oneOffTarget`, if we just run `if(!card) return` inside the constructor while the is `store.oneOffTarget = null` it becomes `{}` instead of `null` which causes ui elements to show on the `v-if="target"`

src/translations/en.json
```diff
@@ -30,5 +30,20 @@
...
+    "4": "Your opponent discards two cards of their choice from their hand",
+    "5": "Draw two cards from the deck",
+    "6": "Scrap all Royals and Glasses eights",
+    "7": "Play one of the top two cards of the deck and put the other back (both are revealed)",
+    "8": "Your opponent plays with an open hand (their cards are revealed to you)",
+    "9": "Return target card to its controller's hand. They can't play it next turn",
+    "10": "No effect",
+    "11": "Play on top of target point card to steal it",
+    "12": "Your other cards may only be targeted by scuttles",
+    "13": "Reduces the points you need to win. (1K: 14pts, 2K: 10pts, 3K: 5pts, 4K: 0pts)"
```
Haviles04, line 47: I thought it better to put it inside the `translations` files, this way both the back and front end can both read from the same spot. By doing this I set up the initial localization of the dialog components that use the rule text.

utils/game-utils.js
```diff
@@ -5,7 +5,20 @@ const getPlayerPnumByUsername = (players, username) => {
...
+  const str_rank =
+    {
+      1: 'A',
+      11: 'J',
+      12: 'Q',
+      13: 'K',
+    }[card.rank] ?? card.rank;
+  const str_suit = ['♣️', '♦️', '♥️', '♠️'][card.suit];
+  return str_rank + str_suit;
+};
```
Haviles04, line 18: Wondering if this is a good place for this to live, or if this should be in `gameService` and should be called in the api files as `gameService.getCardName(card)`

src/stores/game.js
```diff
@@ -14,6 +14,31 @@ function queenCount(player) {
...
+      }[card.rank] ?? card.rank;
+    // Stringify Suit
+    const str_suit = ['♣️', '♦️', '♥️', '♠️'][card.suit];
+    this.createdAt = card.createdAt;
+    this.updatedAt = card.updatedAt;
+    this.id = card.id;
+    this.suit = card.suit;
+    this.rank = card.rank;
+    this.name = str_rank + str_suit;
+    this.attachments = card.attachments;
```
itsalaidbacklife, line 34: Attachments are all cards so we need to map them here
```suggestion
    this.attachments = card.attachments.map(attachment => createGameCard(attachment));
```

src/stores/game.js
itsalaidbacklife: I think this is reasonable to keep the `null` cards e.g. `topCard` when it's missing to be `null` like they currently are

src/stores/game.js
```diff
@@ -140,31 +165,42 @@ export const useGameStore = defineStore('game', {
...
+        this.deck = newGame.deck?.map((card) => createGameCard(card));
+      if (Object.hasOwnProperty.call(newGame, 'scrap'))
+        this.scrap = newGame.scrap?.map((card) => createGameCard(card));
       if (Object.hasOwnProperty.call(newGame, 'log')) this.log = cloneDeep(newGame.log);
       if (Object.hasOwnProperty.call(newGame, 'name')) this.name = newGame.name;
       if (Object.hasOwnProperty.call(newGame, 'p0Ready')) this.p0Ready = newGame.p0Ready;
       if (Object.hasOwnProperty.call(newGame, 'p1Ready')) this.p1Ready = newGame.p1Ready;
       if (Object.hasOwnProperty.call(newGame, 'passes')) this.passes = newGame.passes;
-      if (Object.hasOwnProperty.call(newGame, 'players')) this.players = cloneDeep(newGame.players);
+      if (Object.hasOwnProperty.call(newGame, 'players'))
```
itsalaidbacklife, line 177: Let's use {} for all the multi-line ifs

src/stores/game.js
```diff
@@ -140,31 +165,42 @@ export const useGameStore = defineStore('game', {
...
+        this.players = newGame.players.map((player) => ({
+          ...player,
+          hand: player.hand?.map((card) => createGameCard(card)) ?? player.hand,
+          points: player.points?.map((card) => createGameCard(card) ?? player.points),
+          faceCards: player.faceCards?.map((card) => createGameCard(card) ?? player.faceCards),
```
itsalaidbacklife, line 182: There's something funky about these coalesces. As written, they should only trigger if `player.hand` etc is `null` or more likely `undefined` i.e. the player doesn't have a `.hand` at all (not just an empty list). And if that's the case, then we're coalescing to the same undefined value i.e. `player.hand` which we just said was undefined. So I think we can just leave them out out. 

Also the one for hand is currently coalescing outside the .map(), like I said above, but `.points()` and `.faceCards` are coalescing _inside_ the `.map()` which is weirder b/c it'll set an individual card instance to the entire list.

I _think_ we just want
```suggestion
          hand: player.hand?.map((card) => createGameCard(card)),
          points: player.points?.map((card) => createGameCard(card)),
          faceCards: player.faceCards?.map((card) => createGameCard(card)),
```

but we should sync up b/c the type safety is weird

utils/game-utils.js
itsalaidbacklife: I could go either way. Given that it tends to be called inside of string formatting, it's at least shorter when consumed this way, so I think it's okay

src/translations/en.json
itsalaidbacklife: I think that makes sense. I think this placement at the top level is okay as well, though we could consider putting them inside "global". It's got enough stuff that it seems reasonable to keep it at the top level, idk

itsalaidbacklife (approved):
Nice work. A couple nitpicks and this is ready. Go ahead and bump the version

src/stores/game.js
```diff
@@ -141,31 +166,45 @@ export const useGameStore = defineStore('game', {
...
+      } else this.topCard = null;
 
-      if (Object.hasOwnProperty.call(newGame, 'secondCard')) this.secondCard = cloneDeep(newGame.secondCard);
-      else this.secondCard = null;
+      if (Object.hasOwnProperty.call(newGame, 'secondCard')) {
+        this.secondCard = createGameCard(newGame.secondCard);
+      } else this.secondCard = null;
 
-      if (Object.hasOwnProperty.call(newGame, 'oneOff')) this.oneOff = cloneDeep(newGame.oneOff);
+      if (Object.hasOwnProperty.call(newGame, 'oneOff')) this.oneOff = createGameCard(newGame.oneOff);
```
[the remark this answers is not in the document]
seriouslysean: I suggest making an ESLint rule to disallow one line if statements, it'll be cleaner overall and we won't need to nitpick them.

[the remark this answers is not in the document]
itsalaidbacklife: for use of {}: https://github.com/cuttle-cards/cuttle/issues/620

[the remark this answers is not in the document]
itsalaidbacklife: For refactoring the `Object.ahsOwnProperty.call()`'s out of this func: https://github.com/cuttle-cards/cuttle/issues/621

itsalaidbacklife (approved):
Great work!

PR #605, opened 2023-10-02
Added aria attributes to all the icons

### LINKED ISSUE #542: [Feature]: Add aria attributes to all the icons
### Feature Summary

Cuttle uses icons semantically, meaning all the icons have an intended meaning. For accessibility purposes, these icons should therefore have certain metadata set. We should update the existing uses of icons to set this data accodingly.

### Detailed Description

Specifically, each `<v-icon>` in the application should set:
1. `aria-label="descriptive text based on what the icon is for"`
2. `aria-hidden="false"` (this ensures it is read on screen readers)
3. `role="img"` 

We should set these for every icon in the application, which you can find by doing a search for all instances of `<v-icon`

haaarsh619:
## Issue number #542 

- Resolves #542 

## Please check the following

- For New Features:

### DIFF
--- a/src/components/GameView/ScrapDialog.vue
+++ b/src/components/GameView/ScrapDialog.vue
@@ -20,7 +20,11 @@
           data-cy="close-scrap-dialog-x" 
           @click="show = false"
         >
-          <v-icon icon="mdi-close" size="large" />
+          <v-icon icon="mdi-close" size="large" 
+          aria-label="close icon"
+          aria-hidden="false"
+          role="img"           
+          />
         </v-btn>
       </div>
     </template>
--- a/src/components/GameView/ScrapDialog.vue
+++ b/src/components/GameView/ScrapDialog.vue
@@ -20,10 +20,12 @@
           data-cy="close-scrap-dialog-x" 
           @click="show = false"
         >
-          <v-icon icon="mdi-close" size="large" 
-          aria-label="close icon"
-          aria-hidden="false"
-          role="img"           
+          <v-icon
+            icon="mdi-close"
+            size="large" 
+            aria-label="close icon"
+            aria-hidden="false"
+            role="img"           
           />
         </v-btn>
       </div>
--- a/src/components/GameView/ScrapDialog.vue
+++ b/src/components/GameView/ScrapDialog.vue
@@ -17,15 +17,14 @@
           icon
           color="surface-1"
           elevation="0"
-          data-cy="close-scrap-dialog-x" 
+          data-cy="close-scrap-dialog-x"
+          aria-label="Close scrap dialog" 
           @click="show = false"
         >
           <v-icon
             icon="mdi-close"
             size="large" 
-            aria-label="close icon"
-            aria-hidden="false"
-            role="img"           
+            aria-hidden="true"          
           />
         </v-btn>
       </div>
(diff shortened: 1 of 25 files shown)


src/components/GameView/ScrapDialog.vue
```diff
@@ -20,7 +20,13 @@
           data-cy="close-scrap-dialog-x" 
           @click="show = false"
         >
-          <v-icon icon="mdi-close" size="large" />
+          <v-icon
+            icon="mdi-close"
+            size="large" 
+            aria-label="close icon"
+            aria-hidden="false"
+            role="img"
```
itsalaidbacklife, line 28:
```suggestion
            aria-hidden="true"
```

src/components/GameView/ScrapDialog.vue
```diff
@@ -20,7 +20,13 @@
           data-cy="close-scrap-dialog-x"
```
itsalaidbacklife, line 20:
```suggestion
          data-cy="close-scrap-dialog-x"
          aria-label="Close scrap dialog"
```

src/components/GameView/SpectatorListMenu.vue
```diff
@@ -10,7 +10,14 @@
...
         variant="text"
       >
-        <v-icon class="mr-1" icon="mdi-eye-outline" size="large" />
+        <v-icon
+          class="mr-1"
+          icon="mdi-eye-outline"
+          size="large"
+          aria-label="eye icon"
+          aria-hidden="false"
+          role="img"
```
itsalaidbacklife, line 19:
```suggestion
          aria-hidden="true"
```

src/components/GameView/SpectatorListMenu.vue
```diff
@@ -10,7 +10,14 @@
         icon
         variant="text"
```
itsalaidbacklife, line 11:
```suggestion
        variant="text"
        aria-label="View current spectators"
```

src/components/GameView/TargetSelectionOverlay.vue
```diff
@@ -19,7 +19,13 @@
...
     </div>
     <v-btn icon data-cy="cancel-target-mobile" @click="$emit('cancel')">
-      <v-icon icon="mdi-close" size="x-large" />
+      <v-icon
+        icon="mdi-close"
+        size="x-large"
+        aria-label="close icon"
+        aria-hidden="false"
+        role="img"       
+      />
```
itsalaidbacklife, line 28:
```suggestion
      <v-icon icon="mdi-close" size="x-large" aria-hidden="true" />
```

itsalaidbacklife (commented):
I've learned some about accessibility practices since writing up the issue and have reflected this in the comments. Basically most of our icons are aesthetic only because the context of their container clarifies (or should clarify) what the icon is for. 

So for example with buttons, if the button has text in it, we don't need a label at all because the text serves as the label. And if there is just an icon without text in the button, then we should label the button itself rather than the icon. 

We only need to label the icon itself when the icon is informational and doesn't have other context that clarifies what it is. Additionally for the labels themselves, rather than generically explaining the icon type I've suggested labels that attempt to specifically describe what they do in the context of Cuttle i.e. 'Close Scrap Dialog'.

These suggestions are captured in my comments

itsalaidbacklife (approved):
Looks good. Let's bump a minor version and this is ready

itsalaidbacklife (approved):

PR #652, opened 2023-10-09
Feat/automate version bump

### LINKED ISSUE #96: [Feature Request]: automatically increment package version on merge
### Feature Summary

Merging a pull request should automatically increment the version field in package.json based on a label

### Detailed Description

The package version has been haphazardly updated at somewhat random points. This should be standardized so that every merge updates either the major, minor, or patch  version, depending on the scope of the change. 

To automate this, we should define three new labels for PR: Major update, Minor update, and Patch. Every PR should be labeled with exactly one of those labels (failing CI otherwise), and merging a pr should increment the major, minor, or patch number of the version field in package.json into the base branch

itsalaidbacklife:
Automatically increment version number in main branch when PR is merged into main.

Looks for the labels 'version-patch', 'version-minor', and 'version-major' and runs the corresponding version bumping command then pushes to main.
## Issue number

- Resolves #96 

## Please check the following

- For New Features:


.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,30 @@
+on:
+  pull_request:
+    branches:
+      - main
+    types: [closed]
+jobs:
+  bump-version:
+    if: github.event.pull_request.merged == true
```
itsalaidbacklife, line 8: This is necessary to ensure that the workflow only runs when PRs are merged not when they are closed without merging

.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,30 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
itsalaidbacklife, line 12: Updating to use checkout@v4. We should apply this on our other workflows, but that can be a future update

.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          token: ${{ secrets.GH_PUSH_TOKEN }}
```
itsalaidbacklife, line 14: This was the trick to pushing to a protected branch: https://github.com/orgs/community/discussions/25305#discussioncomment-3247409

.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,30 @@
...
+        if: contains(github.event.pull_request.labels.*.name, 'version-major')
+        run: npm run version:major
+      - name: bump minor version
+        if: contains(github.event.pull_request.labels.*.name, 'version-minor')
+        run: npm run version:minor
+      - name: bump patch version
+        if: contains(github.event.pull_request.labels.*.name, 'version-patch')
+        run: npm run version:patch
+      - name: push
+        run: git push origin main
```
itsalaidbacklife, line 30: Notably these don't check the other labels. Each step independently runs and bumps the version if it finds a corresponding label, so if we apply major + minor + patch version labels, it will bump the package by all three. That could allow mistakes on our part, but it's simpler this way to start and we can see how it goes from here

tools/msg-bump-version.js
```diff
@@ -12,13 +12,12 @@ const msg = `
...
-  Every PR should increase the package file versions by either a
-  patch, minor, or major number, based on the version label on the
-  issue being addressed.
+  Thanks for contributing to cuttle.cards!
+  You *don't* need to update the version manually now;
+  it's been automated!
 
   See contributing docs for more information:
-    https://github.com/cuttle-cards/cuttle/blob/main/docs/CONTRIBUTING.md#make-changes
+    https://github.com/cuttle-cards/cuttle/blob/main/docs/CONTRIBUTING.md#(automated)-version-management
```
itsalaidbacklife, line 20: I'm not 100% how the () will work in the link to the specific header. I renamed the docs header section about this to ```
### (Automated) Version Management
```

so I'm guessing the parens will be included in the link thus the change to ....md#(automated)-version-management

We'll see if it works 🤷

.github/workflows/bump-version.yml
seriouslysean: You know me, I don't agree with that since you'd never be doing all three on a given PR. We also have that status check for labels in place that will verify we only have one so this is probably fine.

PR #653, opened 2023-10-09
Feat #618 animate threes

### LINKED ISSUE #618: [Feature]: Animate transition of resolving a three (revealed card moving into player hand)
### Feature Summary

You cannot currently see which card an opponent selected when they pick a card from the scrap with a 3. We should use an animated transition to clarify this interaction

### Detailed Description

Currently, when a player selects a card from the scrap with a 3 you can't see which card was selected, and the resolution looks similar to drawing a card:

https://github.com/cuttle-cards/cuttle/assets/6520704/edab2336-96ab-4f19-bc7a-4f6f88adf9e2

We should clarify this interaction by revealing the selected card on top of the scrap pile and transitioning it towards the hand of the player that played the three. To do this we should update the socket event sent via `Game.publish()` in `api/controllers/game/resolve-three.js` so that it includes the card being moved like

```
      Game.publish([fullGame.id], {
        change: 'resolveThree',
        game: fullGame,
        victory,
        chosenCardId: req.body.cardId
      });
```

That way the frontend can know which specific card to animate in the scrap. Then we should add data to the GameStore for `cardSelectedFromScrap` which we could use to render and animate the selected card. We should then define a new `processResolveThree` function in the `gameStore.js` store which behaves similarly to `processScuttle` by setting the new `cardSelectedFromScrap` value to the card that was just picked, then setting a timeout and after the delay, unsetting `cardSelectedFromScrap` and updating the rest of the game object so that card triggers its leave animation and disappears.

Then we should update the DOM in `GameView.vue` so that the Scrap area can temporarily display the `cardSelectedFromScrap` as a `<GameCard>`  and animate it up-and-right or down-and-right at a diagonal using new transition classes similar to the ones used on the hands, points etc. The card should have it's leave transition to be up-and-right if moving into the opponent hand and down-and-right if moving into the player hand. 

This way we will have a much smoother, clearer interaction that makes it easy to tell which card was selected when your opponent plays a 3.

Haviles04:
## Issue number

- Resolves #618 

## Please check the following

- For New Features:

 - Updated `resolve-threes` api endpoint to broadcast card pulled from scrap and the pNum of player pulling it
 - Added function and state in store to update new data
 - Added props and transitions to `GameCard` component

https://github.com/cuttle-cards/cuttle/assets/108551827/507ced8c-84e0-4ca8-ace4-5928e90bc060


src/components/GameView/GameCard.vue
```diff
@@ -79,6 +85,14 @@ export default {
       type: Boolean,
       default: false,
     },
+    isThrees: {
```
itsalaidbacklife, line 88: how about `isThreeTarget`
```suggestion
    isThreeTarget: {
```

src/components/GameView/GameCard.vue
```diff
@@ -6,6 +6,8 @@
       glasses: isGlasses,
       jack: isJack,
       frozen: isFrozen,
+      threesPlayer: isThrees && playerChoosingFromScrap,
+      threesOpponent: isThrees && !playerChoosingFromScrap
```
itsalaidbacklife, line 10: Let's make computed properties for the specific classes and just refer to them here like
```suggestion
      threesPlayer: isPlayerThreeTarget,
      threesOpponent: isOpponentThreeTarget
```

Then in our computed props we can have

```
computed: {
  isPlayerThreeTarget() {
    return this.isThreeTarget && this.playerChoosingFromScrap
  },
  isOpponentThreeTarget() {
    return this.isThreeTarget && !this.playerChoosingFromScrap;
  },
}
```

src/plugins/sockets/inGameEvents.js
```diff
@@ -53,6 +53,7 @@ export async function handleInGameEvents(evData) {
       break;
     case SocketEvent.RESOLVE_THREE:
       gameStore.updateGameThenResetPNumIfNull(evData.game);
+      gameStore.processResolveThree(evData.choiceAndPlayer);
```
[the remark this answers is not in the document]
itsalaidbacklife: I'm curious what the UX would look like the other way if the delay was short. You see the card and then it moves into the hand

itsalaidbacklife (commented):
Really nice work. I have one thought about potentially reworking where the game update happens though I'm not positive the change would be an improvement

src/plugins/sockets/inGameEvents.js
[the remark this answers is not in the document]
Haviles04: I tried putting the `updateGameThenResetPNumIfNull()` inside the `processResolveThrees` and was getting some wonky-ness with the deck animation

src/components/GameView/GameCard.vue
```diff
@@ -288,6 +308,38 @@ export default {
...
+    transform: translate(200px, -200px);
+    opacity: 0;
+  }
+}
+@keyframes slideOpponentThreePortrait {
+  100% {
+    transform: translateY(-200px);
+    opacity: 0;
+  }
+}
```
itsalaidbacklife, line 341: These look great, though they're a bit inconsistent with how we've done the transitions for the other cards. For the rest of them, we've used `<transition>` elements which basically apply certain classes when elements leave and certain styles when they enter. Then we essentially style the end states and give the elements `transition: all <some_duration>` and let the transition system do the rest. 

That said, this is pretty concise and I bet `<transition>` would take more code, so I'm not positive it's actually better to use the transition tag for this use case. @seriouslysean penny for your thoughts.

One thing I will say is we've named most of the transition classes after the directions they move things in, rather than the specific card they apply to. e.g. `.in-below-out-left` or `.in-above-out-below`. 

But that was with the intention of reusing those styles, which we might not actually do here since this somewhat specifically moves stuff from the scrap pile towards the hands. But if we did, we could maybe call them `out-up-and-left` and `out-down-and-right` or something. 

Again I'm not actually sure what's best here. It's interesting that this is different than the approach seen in the `<GameView>` but it's also a specific use case that may not be reused and it's clear and concise. Hmm!

Haviles04: I think switching it to a transition element makes sense, to keep everything in order.

src/components/GameView/GameCard.vue
```diff
@@ -288,6 +292,14 @@ export default {
...
 }
 
+.threesPlayer{
+  animation: slidePlayerThree 2s ease;
+}
+.threesOpponent{
+  animation: slideOpponentThree 2s ease;
+}
+
+
```
Haviles04, line 302: Remove this, not used anymore

Haviles04 (commented):
I think this is pretty ready to go, I just have to remove some unneed stuff

tests/e2e/specs/in-game/one-offs/3_threes.spec.js
```diff
@@ -137,10 +137,14 @@ describe('Playing THREEs', () => {
     cy.resolveThreeOpponent(Card.ACE_OF_SPADES);
 
     cy.get('#waiting-for-opponent-resolve-three-scrim').should('not.exist');
-
+    cy.get('[data-cy="scrap-chosen-card"]').should('be.visible');
```
itsalaidbacklife, line 140: Have you tried adding a not exists assertion after this? Without the 8 for points below
```suggestion
    cy.get('[data-cy="scrap-chosen-card"]').should('be.visible');
    cy.get('[data-cy="scrap-chosen-card"]').should('not.exist');
```

tests/e2e/specs/in-game/one-offs/3_threes.spec.js
```diff
@@ -137,10 +137,14 @@ describe('Playing THREEs', () => {
...
+    
+    cy.get('[data-player-hand-card=8-0]').click(); // ace of spades
```
itsalaidbacklife, line 142:
```suggestion
    cy.get('[data-player-hand-card=8-0]').click();
```

itsalaidbacklife (approved):
Really nicely done. This came together well, and it adds juice while clarifying the UX. Excellent work. I have one last shot in the dark to resolve the flake

itsalaidbacklife (approved):
🔥🔥🔥

Issue #622, opened 2023-10-06
[Feature]: Enable language selection from SignupView


itsalaidbacklife:
### Feature Summary

Now that it is possible to to select your preferred language, we should enable users to select a language before logging in. 

### Detailed Description

Currently users can select their preferred language from the `UserMenu` in `TheHeader` after they've logged in:
![image](https://github.com/cuttle-cards/cuttle/assets/6520704/bf8459e8-9586-471a-afe8-338d65ff103f)

But new users who have not yet created an account can't select their language of choice until after signing up. We should improve this by adding an element to the top of the signup page for selecting your preferred language:
![image](https://github.com/cuttle-cards/cuttle/assets/6520704/8df742f3-04b1-4e6f-8010-b0a3e3655da2)

This could open the same language selection menu that's currently in use in `TheUserMenu`:
![image](https://github.com/cuttle-cards/cuttle/assets/6520704/177597e0-e39d-4e0c-b813-584030b775f2)

To do this, we should refactor the language selection functionality into a new `TheLangSelectMenu` component that is consumed by both `TheUserMenu` (as it currently works) and now newly applied to the top of the page directly in `SignupView`.

The result of this would be the same ability to select languages via TheHeader in TheUserMenu in the top left of the app after logging in, while additionally being able to select your language directly from `SignupView`

JonasThers:
Looking into this atm

PR #654, opened 2023-10-09
Feat/622 Enable language selection before login

JonasThers:
Resolves: #622 

Screenshot 2023-10-09 at 22 46 54
Screenshot 2023-10-09 at 22 47 07
Screenshot 2023-10-09 at 23 02 47

I took the logic for language selection and put it in its own component, which is not nested in both `TheUserMenu` and `LoginView`

- For New Features:


src/components/LanguageSelection.vue
```diff
@@ -0,0 +1,43 @@
+<template>
+  <v-menu location="end">
+    <template #activator="{ props }">
+      <v-list-item
+        :class="isInUserMenu ? `bg-surface-2 text-surface-1` : ``"
```
seriouslysean, line 5: I don't like the idea of modifying classes like this just to have some slight differences. I also don't have a better idea that works with vuetify.

In other projects I've used scoped classes where the parent would set a class on the component and either override in the parent or in component with the class the parent put on it.

@itsalaidbacklife, any idea of a better way to do this that's a little more vuetify friendly?

seriouslysean (changes requested):
The changes look sound but I have a question about implementation choice.

src/components/LanguageSelection.vue
itsalaidbacklife, line 1: Let's rename this to `TheLanguageSelector` since we use the [single-instance-component-names](https://vuejs.org/style-guide/rules-strongly-recommended.html#single-instance-component-names) convention for singletons like this (components that appear only once per page) and because I think 'selector' makes it more clear that this is the component used to _do_ the selecting

src/components/LanguageSelection.vue
```diff
@@ -0,0 +1,43 @@
...
+      type: Boolean,
+      default: false,
+    }
+});
+
+const changeLocale = (lang) => {
+  locale.value = lang;
+  setLocalStorage('preferredLocale', lang);
+};
+</script>
```
itsalaidbacklife, line 43: Prefer one newline at the end of files
```suggestion
</script>

```

src/components/LanguageSelection.vue
itsalaidbacklife: @seriouslysean this seems like a good use case for the `variant` prop that would either be `"light"` or `"dark"` similar to the same prop that already exists in `TheUserMenu`, `TheHeader` and `TheFooter`. That would simply replacing the existing `isInUserMenu` prop with `variant` (which is expected to be 'light' or 'dark', perhaps using 'dark' as default.

As for the styles themselves I think classes applied within the component like this are reasonable. The parent sets the prop and the child can then consume the prop to do the styling. The main change then would be decoupling the prop name from the name `isInUserMenu` so that we can reuse the light and dark variants in other places e.g. the `GameView` at a future update. 

The only other thing I would change is refactoring this class calculation into a computed property so that the template code has less in it. That could work like:

```
const props = defineProps({
  variant: {
      type: string,
      default: 'dark',
    }
});
const activatorClass = computed(() => props.variant === 'dark' ? `bg-surface-2 text-surface-1` : '');
```
And in the template then we'd have

```
      <v-list-item
        :class="activatorClass"
```

JonasThers: Thanks for the great feedback, guys! I'll apply these changes asap

JonasThers:
Any idea how my changes in https://github.com/cuttle-cards/cuttle/pull/654/commits/2b3b7cfb7282eb38e27261a7eb2c9a2060efb1ff could cause this?
Screenshot 2023-10-11 at 0 00 33

seriouslysean:
> Any idea how my changes in https://github.com/cuttle-cards/cuttle/pull/654/commits/2b3b7cfb7282eb38e27261a7eb2c9a2060efb1ff could cause this?

This is a really annoying known issue with how sails is currently running alongside cypress. It's not from your PR.

The changes here look good and wouldn't cause that.

itsalaidbacklife (approved):
Nicely done. This is a solid improvement to the UX and refactoring the language selector into its own component is a nice bonus for devEx. Thanks for your contribution!

PR #664, opened 2023-10-11
Update router hash history to web history

premtw:
Resolves #672 
Resolves #675 

- Update Hash History to the Web history

PR #673, opened 2023-10-11
[Feature]: Enable wildcard routing in sails.js

### LINKED ISSUE #672: [Feature]: Enable wildcard routing in sails.js
### Feature Summary

Enable wildcard routing in sails.js

### Detailed Description

This will lay the groundwork for switching to the web router in vue router.

seriouslysean:
## Issue number

- Resolves #672

This is a prerequisite to https://github.com/cuttle-cards/cuttle/pull/664.

## Please check the following

- For New Features:

image

### DIFF
--- a/api/middlewares/serve-index.js
+++ b/api/middlewares/serve-index.js
@@ -1,0 +1,17 @@
+const { join } = require('path');
+
+const ASSETS_PATH = join(__dirname, '../../assets');
+
+function serveIndex(req, res, next) {
+  try {
+    const isFile = req.url.startsWith('/assets') || req.url.includes('.');
+    if (isFile) {
+      return next();
+    }
+    return res.sendFile('index.html', { root: ASSETS_PATH });
+  } catch (err) {
+    return res.badRequest(err);
+  }
+}
+
+module.exports = serveIndex;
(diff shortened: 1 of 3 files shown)


api/middlewares/serve-index.js
```diff
@@ -0,0 +1,17 @@
+const { join } = require('path');
+
+const ASSETS_PATH = join(__dirname, '../../assets');
+
+function serveIndex(req, res, next) {
+  try {
+    const isFile = req.url.startsWith('/assets') || req.url.includes('.');
+    if (isFile) {
+      return next();
+    }
```
seriouslysean, line 10: @itsalaidbacklife this is not the right way to do this, but I was proofing it out. As far as I can tell we _need_ to use custom middleware, or move the index.html file to a sails.js view.

I don't think that will work either though, because vite requires the index.html in the root so we'd need some sort of post build action to move the file which feels way weirder.

I'd like to talk to the sails maintainers and see if there's a better recommendation for a "catch all" route that a SPA would typically use.

This paves the way for the web router, and also would eventually allow us to move to vite as a middleware to consolidate the need to run two apps so that we would only need to run one (which would be way nicer).

seriouslysean:
Proved we can do it, actual solution is up in the web route ticket.

PR #686, opened 2023-10-14
Feat #619 Animate fours

### LINKED ISSUE #619: [Feature]: Animate transition of opponent discarding from 4 be revealing the selected cards
### Feature Summary

We should animate the transition of when an opponent discards from a 4 by revealing the discarded cards before they move downwards towards the scrap

### Detailed Description

This is a similar issue to #227 and #618 which clarify interactions via animating and revealing cards. 

Currently when you play a 4 and your opponent discards cards, you can see the _that_ the opponent discards (you see two cards move down from their hand) but not _which_ cards the opponent discards. You can only tell by looking at the scrap (or if you have glasses on the field).

Uploading Opponent discarding from 4.mov…

We should clarify this interaction by revealing the cards that the opponent selected for discard before transitioning them downward in the current manner. The cards should be revealed using a flip animation that flips them from card back to card front and then the revealed cards should be moved downward towards the scrap.

To accomplish this, we should update `controllers/api/game/resolve-four.js` so that the socket event sent with `Game.publish()` includes the ids of the cards selected for discard. This could be done as follows:
```
      Game.publish([fullGame.id], {
        change: 'resolveFour',
        game: fullGame,
        victory,
        discardedCardIds: cardsToScrap,
      });
```

From there, the frontend `gameStore.js` should be updated with a new piece of state that tracks the cards being discarded e.g. `discardedCards` which should be filled with the two cards being discarded by your opponent and then cleared out when we update the rest of the game.

Then we should add a new function `processOpponentDiscard()` which sets the `discardedCards` array and then sets a timeout and updates the rest of the game and clears out the `discardedCards`.

We should then update the `<GameCard>` component to enable a flipping transition from card back to card front, and update the `<GameView>` page so that the cards in the opponent hand are flipped and revealed when they are present in the `discardedCards` array. 

The full effect of this should be that when an opponent discards we flip and reveal the discarded cards in their hand before they animate downwards towards the scrap.

Notably, there are a few edge cases to consider:
1. The cards to not need to be flipped when the player has glasses
2. No cards need to be flipped when it is the current player discarding (only their opponent)
3. It is possible that only one card is discarded in which case only that card should be flipped.

Haviles04:
## Issue number

- Resolves #619 

## Please check the following

- For New Features:

- Updated API endpoint to added the discarded cards to the last event, as well as send the `cardsToScrap` in the socket message
- Added a state variable to the store `discardedCards`
- Dynamically showed suit and rank by checking if the cards id was inside `discardedCards`
- Added store function with timeout to delay store being updated.
- Added transition to flip card when changed
- The game will now only sort the players hand and not the opponents, this will also happen in the `updateGame` function in the store instead of in `populateGame` on the server side

https://github.com/cuttle-cards/cuttle/assets/108551827/430f4729-c9d9-4a81-9b4c-c1f4323669ce

### DIFF
--- a/src/routes/game/GameView.vue
+++ b/src/routes/game/GameView.vue
@@ -110,8 +110,8 @@
                     <GameCard
                       v-for="(card) in gameStore.opponent.hand"
                       :key="card.id"
-                      :suit="checkIsFoursDiscarded(card) && card.suit"
-                      :rank="checkIsFoursDiscarded(card) && card.rank"
+                      :suit="isBeingDiscarded(card) ? card.suit : undefined"
+                      :rank="isBeingDiscarded(card) ? card.rank : undefined"
                       data-opponent-hand-card
                       class="transition-all opponent-card-back-wrapper opponent-hand-card mx-2"
                     />
@@ -1090,7 +1090,7 @@ export default {
         this.$refs.logsContainer.scrollTop = this.$refs.logsContainer.scrollHeight;
       }
     },
-    checkIsFoursDiscarded(card) {
+    isBeingDiscarded(card) {
       return this.gameStore.discardedCards?.some(id => id === card.id); 
     }
   },
--- a/src/routes/game/GameView.vue
+++ b/src/routes/game/GameView.vue
@@ -110,6 +110,8 @@
                     <GameCard
                       v-for="(card, index) in gameStore.opponent.hand"
                       :key="index"
+                      :suit="checkIsFoursDiscarded(card) && card.suit"
+                      :rank="checkIsFoursDiscarded(card) && card.rank"
                       data-opponent-hand-card
                       class="transition-all opponent-card-back-wrapper opponent-hand-card mx-2"
                     />
@@ -1088,6 +1090,9 @@ export default {
         this.$refs.logsContainer.scrollTop = this.$refs.logsContainer.scrollHeight;
       }
     },
+    checkIsFoursDiscarded(card) {
+      return this.gameStore.discardedCards?.some(id => id === card.id); 
+    }
   },
 };
 </script>
--- a/src/routes/game/GameView.vue
+++ b/src/routes/game/GameView.vue
@@ -108,8 +108,8 @@
                     class="opponent-hand-wrapper transition-all"
                   >
                     <GameCard
-                      v-for="(card, index) in gameStore.opponent.hand"
-                      :key="index"
+                      v-for="(card) in gameStore.opponent.hand"
+                      :key="card.id"
                       :suit="checkIsFoursDiscarded(card) && card.suit"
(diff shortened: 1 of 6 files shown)


src/stores/game.js
```diff
@@ -172,6 +173,11 @@ export const useGameStore = defineStore('game', {
           this.cardChosenFromScrap = null;
           this.playerChoosingFromScrap = null;
         }
+        if (Object.hasOwnProperty.call(newGame.lastEvent, 'discardedCards')) {
+          this.discardedCards = newGame.lastEvent.discardedCards;
+        } else {
+          this.discardedCards = null;
+        }
```
Haviles04, line 180: I'm not sure this is actually needed, I think the code would work just fine without it, especially since we explicitly set `this.discardedCards` in the `processFours()` function in the store. But I put this as a safety so that it will null out `this.discardedCards` if it wasn't the last event to happen

itsalaidbacklife: Safety first, kids

src/routes/game/GameView.vue
```diff
@@ -1088,6 +1090,9 @@ export default {
         this.$refs.logsContainer.scrollTop = this.$refs.logsContainer.scrollHeight;
       }
     },
+    checkIsFoursDiscarded(card) {
```
itsalaidbacklife, line 1093:
```suggestion
    isBeingDiscarded(card) {
```

src/routes/game/GameView.vue
```diff
@@ -108,8 +108,10 @@
                     class="opponent-hand-wrapper transition-all"
                   >
                     <GameCard
-                      v-for="(card, index) in gameStore.opponent.hand"
-                      :key="index"
+                      v-for="(card) in gameStore.opponent.hand"
+                      :key="card.id"
+                      :suit="checkIsFoursDiscarded(card) && card.suit"
+                      :rank="checkIsFoursDiscarded(card) && card.rank"
```
itsalaidbacklife, line 114:
```suggestion
                      :suit="isBeingDiscarded(card) && card.suit"
                      :rank="isBeingDiscarded(card) && card.rank"
```

src/routes/game/components/GameCard.vue
```diff
@@ -52,11 +55,11 @@ export default {
   name: 'GameCard',
   props: {
     suit: {
-      type: Number,
+      type: [Number, Boolean],
```
itsalaidbacklife, line 58:
```suggestion
      type: Number,
```

src/routes/game/components/GameCard.vue
```diff
@@ -52,11 +55,11 @@ export default {
...
       default: undefined,
     },
     rank: {
-      type: Number,
+      type: [Number, Boolean],
```
itsalaidbacklife, line 62:
```suggestion
      type: Number,
```

PR #693, opened 2023-10-17
(#581) navigation disappears between 960 & 1280px

### LINKED ISSUE #581: [Bug]: App navigation disappears between 960px and 1280px
### Bug Summary

Devices between 960px and 1280px wide currently lose the nav links in `TheHeader` (top nav) but do not trigger opening `TheFooter`, leaving them unable to navigate between pages 

### Detailed Description

See screenshot:
image

Notice there are no nav links in the upper right corner as seen on larger devices (> 1280px wide):
image

And neither does TheFooter appear with nav links in the bottom of the page, as seen on smaller devices:
image

This is a problem because in `App.vue`, `TheFooter` is rendered conditionally based on the vuetify breakpoint `smAndDown` i.e. below 960px. In contrast, `TheHeader` conditionally renders the nav links in the `<v-toolbar-items>` based on vuetify's `mobile` breakpoint, which triggers below 1280px, leaving a gap between the two for devices between 960px and 1280px wide.

To solve this, we should make the use of breakpoints match. Since the `mobile` breakpoint is more inclusive than was previously anticipated, the best way to sync these is by updating `TheHeader.vue` so that it renders the toolbar based on the `smAndDown` breakpoint the same way that `App.vue` uses for `TheFooter.vue` (rather than its current use of `mobile`

### Current Behavior

Devices between 960px and 1280px wide have no navigational elements on the home, stats, and about pages.

### Expected Behavior

Nav links appear in TheHeader for devices > 960px and instead uses TheFooter for navigation for devices smaller than 960px

### Steps to Reproduce

1. Set your browser width to 1000px
2. open the homepage
3. notice there are no links in the header and the footer doesn't appear

Rafajrg21:
## Issue number

- Resolves #581

## Please check the following

- For New Features:

For testing you can set the window width to 1000px and it should show you the navigation options in the header, at 960px it should show it as well but when lowered to 959px or less, the navigation options in header should be hidden and it should show the navigation footer

### DIFF
--- a/src/components/TheHeader.vue
+++ b/src/components/TheHeader.vue
@@ -14,7 +14,7 @@
           >
         </div>
       </v-toolbar-title>
-      <v-toolbar-items v-if="!mobile" class="hidden-xs-only">
+      <v-toolbar-items v-if="!smAndDown" class="hidden-xs-only">
         <v-tabs class="pa-2">
           <v-tab
             v-for="({ text, icon, page, cyName }, i) in pageLinks"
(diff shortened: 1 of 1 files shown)


itsalaidbacklife (approved):
Nice work, thanks for the fix!

PR #699, opened 2023-10-20
[Feat #674] Joining lobby by URL

### LINKED ISSUE #674: [Feature]: Enable joining a lobby by sharing a link to your current game
### Feature Summary

It would be convenient to be able to invite someone directly to your game from the lobby view to avoid the need to tell them which specific game to join using its name. We should allow direct navigation to a game lobby to direct users to authenticate if necessary and then to join the game

### Detailed Description

The intended flow is that a user who is sitting in the lobby can copy their url bar and send it to another user who could then navigate directly to their game and take care of all the appropriate setup in the database. 

This has two main parts:

1. If a user navigates to a specific lobby without being logged in, we should redirect them to the `LoginView` and set a query param that the LoginView uses to redirect them _back to that game lobby_ after successful authentication
2. When the `LobbyView` is created, it should check in the `gameStore` to confirm that the user has the game data and has properly subscribed to the game. If the game data is _not_ present (i.e. the user directly navigated to this url without first going home and clicking the join button), then we should fire off the `gameStore.requestSubscribe()` action with the id from the lobby url. 

Notably for 2) we should use a loading state while the request to join the lobby is being processed (a v-skeleton loader would be appropriate) and if the request to join the game fails (e.g. another person already joined before you clicked the link), we should show a dialog that says 
"Oops

Could not join that game at this time. Go home to find another one!"

with a button to "Go Home" that directs the user to the HomeView.

This way users will be able to share links to their current lobby so other players can directly join their game, and we'll appropriately authenticate and join the game, and handle errors by going home.

psinha0:
## Feature #674 

- Resolves #674 

## Please check the following

- For New Features:

- Users can now copy the current lobby URL and share it as a mode of inviting others to play against them.

https://github.com/cuttle-cards/cuttle/assets/58553140/3806631f-93a5-4824-a3a9-f48439e29098

- Changed the router for `LobbyView` to go to `LoginView` whilst storing the lobby URL temporarily if the user is not logged in when accessing a game lobby URL.
- If a game does not exist in store `gameList`, redirect user back to the homepage.


JonasThers:
Great job!

Just an UX suggestion: How about a button, which could say something "Share link for game", which would copy the url to the clipboard? I figure it would make it more obvious that this is a possibility, but it's just a thought

src/stores/gameList.js
```diff
@@ -36,6 +36,10 @@ export const useGameListStore = defineStore('gameList', {
       startedGame.status = GameStatus.STARTED;
       this.spectateGames.push(startedGame);
     },
+    doesGameExist(gameId) {
+      const game = this.openGames.find((game) => game.id == gameId);
+      return (game) ? true : false;
+    },
```
Haviles04, line 42: I think we can one-line this

```suggestion
    doesGameExist(gameId) {
      return this.openGames.some(({id}) => id === gameId);
    },
```

src/routes/lobby/LobbyView.vue
```diff
@@ -169,6 +198,23 @@ export default {
           console.log(err);
         });
     },
+    subscribeToGame(gameId) {
+      this.loading = true;
+      gameId = Number(gameId);
+      this.gameStore
+        .requestSubscribe(gameId)
```
Haviles04, line 205:
```suggestion
      this.gameStore
        .requestSubscribe(Number(gameId))
```

Haviles04 (commented):
This is going to be sick! Left a couple nitpick comments :)

src/stores/gameList.js
psinha0: I didn't know there was a function that could do that, thanks a lot! Added your suggestion, much appreciated

psinha0:
> Great job!
> 
> Just an UX suggestion: How about a button, which could say something "Share link for game", which would copy the url to the clipboard? I figure it would make it more obvious that this is a possibility, but it's just a thought

I actually love that suggestion. I'll whip that up soon - not sure if I should add that in this PR or a separate one though

itsalaidbacklife:
> > Great job!
> > Just an UX suggestion: How about a button, which could say something "Share link for game", which would copy the url to the clipboard? I figure it would make it more obvious that this is a possibility, but it's just a thought
> 
> I actually love that suggestion. I'll whip that up soon - not sure if I should add that in this PR or a separate one though

I also love the idea, and I think it can be a separate issue + PR

itsalaidbacklife:
@psinha0 really nice work. We just merged a redesign of the lobby view which has created some merge conflicts. Please reach out if you could use help addressing them. Apologies for the issue, but these two together will make for a hugely improved UX. Thanks for your contribution!

seriouslysean:
@itsalaidbacklife I think a new issue/feature to copy the lobby url from the lobby to "share" it would be really nice, ala stackoverflow share.

image

src/router.js
```diff
@@ -70,8 +71,18 @@ const routes = [
...
-    // TODO: Add logic to redirect if a given game does not exist
-    beforeEnter: mustBeAuthenticated,
+    beforeEnter: async (to, _from, next) => {
+      const authStore = useAuthStore();
+      const gameListStore = useGameListStore();
+      if (!gameListStore.doesGameExist(to.path.split('lobby/')[1])) {
+        next('/');
+      } else if (!authStore.authenticated || authStore.authenticated == null) {
+        authStore.setLoginRedirect(to.path);
+        next('/login');
```
seriouslysean, line 81: This should also be a discrete block with a return next

src/router.js
```diff
@@ -70,8 +71,18 @@ const routes = [
...
+      } else {
+        next();
```
seriouslysean, line 83: And finally return next here if as the final resolution.

src/routes/lobby/LobbyView.vue
```diff
@@ -1,4 +1,23 @@
 <template>
+  <BaseDialog 
+    v-model="errorLoadingDialogVisible"
+    title="Error Joining Lobby"
+  >
+    <template #body>
+      <span>Oops. </span><br>
+      <span>Could not join that game at this time. Go home to find another one</span>
```
seriouslysean, line 8: The text here could use a little work, additionally spans with a break tag really aren't a great way to handle it. I would check other standard usage in the codebase and prefer titles and paragraph tags with the proper vuetify classing.

src/routes/lobby/LobbyView.vue
```diff
@@ -151,6 +174,12 @@ export default {
       }
     },
   },
+  async created() {
+    if (!this.gameStore.id || this.gameStore.id === null) {
+      this.subscribeToGame(window.location.hash.split('lobby/')[1]);
+      this.$router.push(`/lobby/${window.location.hash.split('lobby/')[1]}`);
+    }
+  },
```
seriouslysean, line 182: It feels weird to have a function where the entire thing is wrapped in an if, you should instead use a fast fail pattern here and invert the logic to return.

src/routes/lobby/LobbyView.vue
```diff
@@ -172,6 +201,22 @@ export default {
...
+        .requestSubscribe(Number(gameId))
+        .then(() => {
+          this.loading = false;
+        })
+        .catch((err) => {
+          this.loading = false;
+          this.errorLoadingDialogVisible = true;
+          console.log(err);
+        });
+    },
```
seriouslysean, line 216: Don't we already have a subscribe to game function in the game store? Are you able to reuse it?

src/routes/login/LoginView.vue
```diff
@@ -272,7 +272,12 @@ export default {
       this.username = '';
       this.pw = '';
       this.loading = false;
-      this.$router.push('/');
+      if (this.authStore.loginRedirect) {
+        this.$router.push(this.authStore.loginRedirect);
+        this.authStore.loginRedirect = null;
+      } else {
+        this.$router.push('/');
+      }
```
seriouslysean, line 280: Since both use router push, you can likely consolidate this approach. What's the purpose of nulling out `this.authStore.loginRedirect`? it feels like code smell to me.

```
const redirectPath = this.authStore.loginRedirect ?? '/';
this.authStore.loginRedirect = null; // figure out some way to not need to do this
this.$router.push(redirectPath);
```

src/stores/auth.js
```diff
@@ -31,6 +32,9 @@ export const useAuthStore = defineStore('auth', {
       this.authenticated = false;
       this.username = null;
     },
+    setLoginRedirect(redirectPath) {
+      this.loginRedirect = redirectPath;
+    },
```
seriouslysean, line 37: This is a weird setup for a one time use value as you need to reset it on use. It might make more sense to have a method that also does the redirect and clears the value for you, though I still don't think that feels exactly right.

seriouslysean:
Closing this due to inactivity.

PR #683, opened 2023-10-13
[Fix #666]: set GH_PUSH_TOKEN as an environment variable using env

trishanu-init:
## Issue number

- Resolves #666 

## Please check the following

- [X] Do the tests still pass? (see [Run the Tests](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#run-the-tests))
- [X] Is the code formatted properly? (see [Linting (Formatting)](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#linting-formatting))
- For New Features:
  - [ ] Have tests been added to cover any new features or fixes?
  - [ ] Has the documentation been updated accordingly?

## Please describe additional details for testing this change
Previously the bump-version.yml automation implemented to bump package versions based on PR labels required a PAT to merge to the protected main branch, which is not provided to PR's from forks for security reasons.

Modified workflow:

-set GH_PUSH_TOKEN as an environment variable using env.

-add a step at the beginning and before each version bump step to check for the presence of the GH_PUSH_TOKEN. If it's not provided (empty), the step will no-op, effectively skipping the rest of the job for PRs from forks without the GH_PUSH_TOKEN.

-Now, the workflow will only proceed with the package version bumping and push steps if the GH_PUSH_TOKEN secret is available, ensuring security for the main branch while allowing for PRs from forks to still run the workflow.

Screenshot (50)

PR #700, opened 2023-10-21
feat(bump-version.yml): skip checkout if push token is missing

itsalaidbacklife:
## Issue number

- Resolves #683 

## Please check the following

- For New Features:

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -9,12 +9,13 @@ jobs:
       runs-on: ubuntu-latest
   
       env:
-        GH_PUSH_TOKEN: ${{ secrets.GH_PUSH_TOKEN || github.token }}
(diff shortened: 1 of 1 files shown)


.github/workflows/bump-version.yml
```diff
@@ -9,12 +9,13 @@ jobs:
       runs-on: ubuntu-latest
   
       env:
-        GH_PUSH_TOKEN: ${{ secrets.GH_PUSH_TOKEN || github.token }}
+        GH_PUSH_TOKEN: ${{ secrets.GH_PUSH_TOKEN }}
```
itsalaidbacklife, line 12: The idea of coallescing to `github.token` was to prevent the checkout step from erroring when the token is missing, but then the push step errors and we don't have a way to tell whether the token was provided by the secret or the default. So instead, I'm removing the default value for the `env.GH_PUSH_TOKEN` and adding a conditional for the checkout step

PR #701, opened 2023-10-21
DevEx #621 updateGame() Refactor

### LINKED ISSUE #126: Refactor updateGame to reduce complexity
### Summary

Refactor `updateGame` in two two more more discrete functions

### Detailed Description

`updateGame` in `client/js/store/modules/game.js` currently fires twice when joining a lobby. Once with the full game state which works as expected and updates all the values.

The second call to `updateGame` comes from updating the players and only has the player values. This was noticed when working on the `isRanked` logic. For instance, we currently update values in this way:

```
updateGame(state, newGame) {
  ...
  if (Object.hasOwnProperty.call(newGame, 'isRanked')) state.isRanked = newGame.isRanked;
},
```

We're locking the logic behind the existence of the keys on the object we're passing which means we expect one or more schemas for that `newGame` object. Ideally, this would be refactored to something like this:

- `updateGame` - always takes the full game object so we don't need the `hasOwnProperty` calls and can directly update the state with a set of keys
- `updateGamePlayers` - always takes the players and adds them to the game data
- Other functions may be needed depending on how many ways this function is called and with which payloads

@itsalaidbacklife may have more information here, but this should be enough to do some digging, come up with a plan and then implement an initial pass.

### LINKED ISSUE #621: [DevEx]: Update socket messages to reduce need for type-checks by making all game attributes appear every time
### Improvement Summary

Currently the `gameStore.updateGame()` function checks for the existence of every property on the game before assigning them to the store, because the object is sparsely populated by the server. We could simplify this logic and remove the existence checks if we update the server calls that create games to always include all the properties referenced in the frontend.

### Detailed Description

In particular we should look at all the properties consumed in `gameStore.updateGame()`:

```
    updateGame(newGame) {
      if (Object.hasOwnProperty.call(newGame, 'lastEvent')) {
        if (Object.hasOwnProperty.call(newGame.lastEvent, 'change')) {
          this.lastEventChange = newGame.lastEvent.change;
        } else {
          this.lastEventChange = null;
        }
        if (Object.hasOwnProperty.call(newGame.lastEvent, 'oneOff')) {
          this.lastEventOneOffRank = newGame.lastEvent.oneOff.rank;
        } else {
          this.lastEventOneOffRank = null;
        }
        if (Object.hasOwnProperty.call(newGame.lastEvent, 'oneOffTargetType')) {
          this.lastEventTargetType = newGame.lastEvent.oneOffTargetType;
        } else {
          this.lastEventTargetType = null;
        }
      }
      this.waitingForOpponentToStalemate = false;
      if (Object.hasOwnProperty.call(newGame, 'id')) this.id = newGame.id;
      if (Object.hasOwnProperty.call(newGame, 'turn')) this.turn = newGame.turn;
      if (Object.hasOwnProperty.call(newGame, 'chat')) this.chat = cloneDeep(newGame.chat);
      if (Object.hasOwnProperty.call(newGame, 'deck')) this.deck = cloneDeep(newGame.deck);
      if (Object.hasOwnProperty.call(newGame, 'scrap')) this.scrap = cloneDeep(newGame.scrap);
      if (Object.hasOwnProperty.call(newGame, 'log')) this.log = cloneDeep(newGame.log);
      if (Object.hasOwnProperty.call(newGame, 'name')) this.name = newGame.name;
      if (Object.hasOwnProperty.call(newGame, 'p0Ready')) this.p0Ready = newGame.p0Ready;
      if (Object.hasOwnProperty.call(newGame, 'p1Ready')) this.p1Ready = newGame.p1Ready;
      if (Object.hasOwnProperty.call(newGame, 'passes')) this.passes = newGame.passes;
      if (Object.hasOwnProperty.call(newGame, 'players')) this.players = cloneDeep(newGame.players);
      if (Object.hasOwnProperty.call(newGame, 'spectatingUsers')) {
        this.spectatingUsers = newGame.spectatingUsers;
      }
      if (Object.hasOwnProperty.call(newGame, 'twos')) this.twos = cloneDeep(newGame.twos);

      if (Object.hasOwnProperty.call(newGame, 'topCard')) this.topCard = cloneDeep(newGame.topCard);
      else this.topCard = null;

      if (Object.hasOwnProperty.call(newGame, 'secondCard')) this.secondCard = cloneDeep(newGame.secondCard);
      else this.secondCard = null;

      if (Object.hasOwnProperty.call(newGame, 'oneOff')) this.oneOff = cloneDeep(newGame.oneOff);
      else this.oneOff = null;

      if (Object.hasOwnProperty.call(newGame, 'oneOffTarget'))
        this.oneOffTarget = cloneDeep(newGame.oneOffTarget);
      else this.oneOffTarget = null;

      if (Object.hasOwnProperty.call(newGame, 'isRanked')) this.isRanked = newGame.isRanked;
      if (Object.hasOwnProperty.call(newGame, 'currentMatch')) this.currentMatch = newGame.currentMatch;
    },
```

All of these `Object.hasOwnProperty()` calls make it difficult to read what the function is really doing, and generally make the update more complex than it needs to be. They exist because the various properties on the `Game` sent by the server are sometimes missing. We should rectify this by updating all the server-generated data where a game is sent so that the game includes all of the properties that are consumed in `gameStore.updateGame()`. In particular, we should look at the calls to `updateGame()` that are called outside of the `GameView.vue` e.g. in `gameStore.requestSubscribe()` and `gameStore.requestSpectate()`. 

We should look at the server generated data and ensure that it always has for example an array of players who have hands, points etc, even if the array of players is empty or if the players exist but have no hands (this could be represented with an empty array for the player hands). 

This should then allow us to make the `gameStore.updateGame()` function safely assume that all properties are available and therefore skip all the `Object.hasOwnProperty.call()` checks, simply assigning values to state as passed by the server created game data.

Haviles04:
## Issue number

- Resolves #126  

## Please check the following

- For New Features:

- Refactored `updateGame()` to dynamically update all the variables
- Renamed state variables to better match what comes in from the `game` object


Haviles04:
I'm not 100% sure this is the right way to go with this, but I believe looping through the object keys gets rid of the need for `hasOwnProperty()` and the nullish coalescing is a backup

Haviles04:
@seriouslysean any thoughts on this?

seriouslysean (commented):
I don’t have a lot of time to look at this, I’m in Disney for a couple more days. I can try and take another look tonight but I need to better understand what the purpose of the function is and the requirements and order of operations.

It might help if you documented some of those things and went through the changes you made and why? Don’t feel like you have to, @itsalaidbacklife likely knows more about this than I do, but it would help me give some more targeted feedback.

Haviles04:
> I don’t have a lot of time to look at this, I’m in Disney for a couple more days. I can try and take another look tonight but I need to better understand what the purpose of the function is and the requirements and order of operations.
> 
> It might help if you documented some of those things and went through the changes you made and why? Don’t feel like you have to, @itsalaidbacklife likely knows more about this than I do, but it would help me give some more targeted feedback.

So the thought behind this is to remove calling .hasOwnProperty a bunch of times to check if the key-value pair is in the object and then set it to the corresponding store value, why not just loop through the object and set the store values based on what is in the object, considering almost all of them are the same name as the socket sends over. The if blocks are for the few edge cases like the cards that use a constructor to build them

seriouslysean:
It seems like just checking the props you need directly with optional chaining instead of hasOwnProperty might serve the same purpose without the complexity of requiring a loop and key check? You could also make a map and have some discrete functions for each of the actions, something like this:

```
const actionMap = {
    optionalKey1: optionalKey1Fn,
    optionalKey2: optionalKey2Fn,
};
```

Then you could break up the function and abstract the individual tasks to specific areas like this:

```
Object.keys(game).forEach(k => actionMap[k](game));
```

src/stores/game.js
```diff
@@ -165,33 +164,75 @@ export const useGameStore = defineStore('game', {
...
+      //loop through all keys sent through socket message and run corresponding function
+      Object.keys(newGame).forEach((key) => {
+        if (updateGameMap.key) {
+          updateGameMap[key](newGame);
         } else {
           this[key] = newGame[key] ?? null;
         }
       });
+
+      // Object.keys(newGame).forEach((key) => {
```
seriouslysean, line 212: NoI need to keep the commented out code, you can remove it. How do you feel about this new solution? Feel better or worse? Penny for your thoughts.

Haviles04: I think its a great idea honestly, It makes it clearer where everything is being set where as in my code, if it wasn't a edge case it wasn't clear exactly what was being set

Haviles04: Thanks for the input!!

src/stores/game.js
```diff
@@ -165,106 +164,49 @@ export const useGameStore = defineStore('game', {
...
+          };
+          Object.keys(newGame.lastEvent).forEach((lastEventKey) => {
+            return lastEventMap[lastEventKey] ? lastEventMap[lastEventKey](newGame) : null;
+          });
+        },
+      };
+      //loop through all keys sent through socket message and run corresponding function
+      Object.keys(newGame).forEach((key) => {
+        return updateGameMap[key] ? updateGameMap[key](newGame) : null;
+      });
```
itsalaidbacklife, line 208: Okay so I am _finally_ looking more deeply at this and I'm wondering what's the added value of the loops specifically? Like what if we just replaced the conditionals with optional chaining and skipped looping through keys?

```suggestion
        this.id = newGame.id ?? null;
        this.turn = newGame.turn ?? null;
```
etc? It seems like that would be a little bit shorter and simpler too

seriouslysean: I can see that working. Maybe I didn’t look deep enough initially but it felt like some of them had multiple operations that were encapsulated in the functions. If it’s simple enough your solution makes more sense imo.

Haviles04: There is no added value to the loop using the action map. 

Before I switched it to the action map, the added value of the loop was that it automatically set the store values based on their corresponding newGame values. If you had added a new variable to the store, you would only need to make sure the API was sending it inside newGame and you wouldn't have to mess with updateGame. 

 The if/else blocks were the edge cases where the store value wasn't a direct copy of the newGame value. (running player cards through the card constructor).

That being said I'm not actually sure how often this would be an advantage

Haviles04: The action map was much more readable

Haviles04:
Okay, so I finally got back to this, and I refactored it to use nullish coalescing optional chaining, and the OR operator, I used the OR operator to set stuff back to itself if theres nothing inside the newGame object to match it, this way things like ID are never undefined

Haviles04:
@itsalaidbacklife @seriouslysean any thoughts on this? Should we change this up or maybe wait until we change up the backend to implement this change?

src/stores/game.js
```diff
@@ -174,118 +174,38 @@ export const useGameStore = defineStore('game', {
...
+      this.id = newGame.id || this.id;
+      this.turn = newGame.turn || this.turn;
+      // this.chat = cloneDeep(newGame.chat);
+      this.deck = newGame.deck?.map((card) => createGameCard(card)) || this.deck;
+      this.scrap = newGame.scrap?.map((card) => createGameCard(card)) || this.scrap;
+      this.log = cloneDeep(newGame.log) || this.log;
+      this.name = newGame.name || this.name;
+      this.p0Ready = newGame.p0Ready || this.p0Ready;
+      this.p1Ready = newGame.p1Ready || this.p1Ready;
+      this.passes = newGame.passes || this.passes;
```
itsalaidbacklife, line 193: Any reason to use `||` instead of `??` I don't think we want to exclude false values so I'd think we want ??

itsalaidbacklife (approved):
This looks good to me. One Q re: ||

src/stores/game.js
Haviles04: Good point, I believe you're right it should be ??

Issue #707, opened 2023-10-25
[Bug]: LobbyView is not playing a sound when players join


itsalaidbacklife:
### Bug Summary

The LobbyView is supposed to play a sound when a player joins but this is not happening. 

### Detailed Description

When your opponent joins the lobby you're in, you are supposed to hear a sound effect of some water to indicate that a change has happened and you have an opponent. This was implemented in #364 and fixed up in #387, but it appears no longer to be functional. 

It looks to be the case that the two `<audio>` tags with refs for `enterLobbySound` and `leaveLobbySound` have been removed from the LobbyView. We should re-add them and test that this still works.

### Current Behavior

The LobbyView does not make sounds when players join and leave

### Expected Behavior

The lobby should make one sound effect when a player joins, and another when they leave

### Steps to Reproduce

1. Log in, create a game, and join it
2. Have another player join your game
3. A sound should play but doesn't
4. Have that player leave
5. A different sound should play but doesn't

anudeeps352:
I might try checking this issue out

PR #719, opened 2023-10-28
Bug/707 LobbyView is not playing a sound when players join

JonasThers:
## Issue number

- Resolves https://github.com/cuttle-cards/cuttle/issues/707

## Please check the following

- For New Features:


src/routes/lobby/LobbyView.vue
```diff
@@ -128,6 +130,9 @@ export default {
     gameName() {
       return this.gameStore.name;
     },
+    opponentUsername() {
+      return this.gameStore?.opponentUsername;
+    },
```
itsalaidbacklife, line 135: It looks like you defined this computed prop without actually consuming it in the template code. Also do we need to optionally chain here? The gameStore should always be defined

itsalaidbacklife (commented):
Looks good, one nitpick on a computed prop. Have you confirmed the fix locally?

JonasThers:
@itsalaidbacklife Good to hear! I did the computed property, because the watcher did not seem to react otherwise. I will fix it during this week

src/routes/lobby/LobbyView.vue
itsalaidbacklife: @JonasThers you're right it makes sense why we'd need the computed prop with the way the watcher is written, now I see the problem. Can you please additionally refactor the instance of the opponent's player indicator to use the computed prop on line 26?

seriouslysean: This value feels like it should be mapped in instead of defined this way, though I agree optionally chaining is not required here.

itsalaidbacklife:
@JonasThers are you still interested in completing this issue?

JonasThers:
@itsalaidbacklife Yes, sorry for my silence, but I've been busy. Will get this done this weekend

src/routes/lobby/LobbyView.vue
itsalaidbacklife: @JonasThers not a stupid question. Pinia has `mapState()` and `mapGetters()` helpers that components can use to define computed properties like this that just return a given value from the store. We have several examples in the codebase you can find by searching for mapState and mapGetters

itsalaidbacklife (approved):
Thanks for the fix!

PR #721, opened 2023-10-29
fix: mode change type error

### LINKED ISSUE #705: [Bug]: Switching a game between ranked and casual games in the LobbyView causes a clientside TypeError
### Bug Summary

There is a switch in the LobbyView for switching a game between ranked and casual modes. Hitting it generally works, but causes a TypeError in the gameStore

### Detailed Description

This is the switch:
image

Hitting it shows a snackbar that the game has switched modes, and this functions correctly, but it causes a TypeError in `gameStore.updateGame()`
image

The issue looks to be due to the fact that `gameStore.updateGame()` is being called with `null` or possibly `undefined` instead of an object, which breaks the `Object.hasOwnProperty.call()` function calls.
### Current Behavior

Switching between ranked and casual in the `LobbyView` causes a type error.

### Expected Behavior

Switching between ranked and casual works without causing a type error.

### Steps to Reproduce

1. Log in
2. Create a game
3. Join the game
4. Click the 'casual' switch to switch the game to ranked
5. Look at the devtools console to see the type error

haaarsh619:
## Issue number #705 

- Resolves #705 

## Please check the following

## Summary
- The issue was caused by an unnecessary call to ``` gameStore.resetPNumIfNullThenUpdateGame(evData.game) ``` within the ``` inGameEvents.js ``` method when the ranked was updated. Since changing the game mode was not part of the ``` inGameEvent.js ```, an undefined value was passed, leading to the type error. Removed the call to the ``` game.publish ```  method to resolve the issue.

### DIFF
--- a/api/controllers/game/set-is-ranked.js
+++ b/api/controllers/game/set-is-ranked.js
@@ -10,13 +10,6 @@ module.exports = async function (req, res) {
     
     sails.sockets.blast('setIsRanked', { gameId: game.id, isRanked: isRanked });
     
-    Game.publish([game.id], {
-      userId: req.session.usr,
-      pNum: req.session.pNum,
-      gameId: game.id,
-      isRanked: isRanked,
-    });
-
     return res.ok();
   } catch (err) {
     return res.badRequest(err);


itsalaidbacklife (approved):
Good catch

PR #722, opened 2023-10-29
Feat #33 Initial PWA support

### LINKED ISSUE #33: Add (some) PWA support
1. https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps
2. https://web.dev/progressive-web-apps/

There's a lot to this of course, but there are some minor wins that might be worth doing in the manifest:

### Add a Web App Manifest, https://web.dev/add-manifest/
- More visible than a bookmark
- `short_name` or `name` will show up under the icon as the app name (`Cuttle`?)
- Allows you to set a splash screen (your cuttle logo?)
  - https://medium.com/@applification/progressive-web-app-splash-screens-80340b45d210
  - https://progressier.com/pwa-icons-and-ios-splash-screen-generator
- `start_url`, if someone opens it as a PWA can have them go right to the games screen
- `background_color`/`theme_color` to set your theme
- `display: "fullscreen"` will open your app without any browser url bar, etc when opened as a PWA

Haviles04:
## Issue number

- Resolves #33 

## Please check the following

- For New Features:


tests/e2e/support/commands.js
```diff
@@ -77,7 +77,11 @@ Cypress.Commands.add('requestGameList', () => {
 Cypress.Commands.add('setupGameAsP0', (alreadyAuthenticated = false, isRanked = false) => {
   if (!alreadyAuthenticated) {
     cy.wipeDatabase();
-    cy.visit('/');
+    cy.visit('/', {
+      onBeforeLoad(win) {
+        delete win.navigator.__proto__.serviceWorker;
+      },
+    });
```
Haviles04, line 84: There is an open issue for cypress hanging up on the service worker. It causes a timeout error on the cy.visit(). Currently this is the only way I can bypass that error 

https://github.com/cypress-io/cypress/issues/27501

Haviles04: IMO this feels hacky and will be annoying to do in every `cy.visit()` inside the test

seriouslysean (changes requested):
As cool as I think this is, service workers in general are extremely tricky to get right and with where Cuttle is currently, I would rather sees the other PWA stuff go in first and a service worker be it’s own ticket.

Haviles04:
> As cool as I think this is, service workers in general are extremely tricky to get right and with where Cuttle is currently, I would rather sees the other PWA stuff go in first and a service worker be it’s own ticket.

So the service worker here comes from the vite-pwa plugin. Would you prefer I put just a general manifest and remove the plugin?

seriouslysean:
> > As cool as I think this is, service workers in general are extremely tricky to get right and with where Cuttle is currently, I would rather sees the other PWA stuff go in first and a service worker be it’s own ticket.
> 
> So the service worker here comes from the vite-pwa plugin. Would you prefer I put just a general manifest and remove the plugin?

I think it makes sense to do everything but the service worker right now. Manifest, theme colors, better favicons and icons, splash screens.

https://progressier.com/pwa-icons-and-ios-splash-screen-generator

https://realfavicongenerator.net/

I’m sure there’s more I’m forgetting.

Haviles04:
> > > As cool as I think this is, service workers in general are extremely tricky to get right and with where Cuttle is currently, I would rather sees the other PWA stuff go in first and a service worker be it’s own ticket.
> > 
> > So the service worker here comes from the vite-pwa plugin. Would you prefer I put just a general manifest and remove the plugin?
> 
> I think it makes sense to do everything but the service worker right now. Manifest, theme colors, better favicons and icons, splash screens.
> 
> https://progressier.com/pwa-icons-and-ios-splash-screen-generator
> 
> https://realfavicongenerator.net/
> 
> I’m sure there’s more I’m forgetting.
> 

So I was able to just un-register the service worker by adding `selfDestroying: true` to the vite-pwa config section in `vite.config`. I'm thinking maybe it's good like this way in the future we can just setup the SW the way we want?

seriouslysean:
Honestly I’d just remove it, I know you did the work but it serves no purpose at the moment.

Haviles04:
> Honestly I’d just remove it, I know you did the work but it serves no purpose at the moment.

Okay, and go with a manifest.json file instead?

Haviles04:
I _think_ this should be good, let me know if you want me to change/add anything

seriouslysean (approved):
This looks good to me, final approval from @itsalaidbacklife. It would be worth trying to add this to your Home Screen after merging to verify. If you have iOS simulator in xcode  you could test locally.

Haviles04:
> This looks good to me, final approval from @itsalaidbacklife. It would be worth trying to add this to your Home Screen after merging to verify. If you have iOS simulator in xcode  you could test locally. 

I ran it locally from the dev server through my network on android pixel 7, I was able to add it to the home screen. Not sure about iOS

manifest.json
```diff
@@ -0,0 +1,62 @@
...
+      "sizes": "384x384",
+      "type": "image/png"
+    },
+    {
+      "src": "/pwa-icons/logo-head-512.png",
+      "sizes": "512x512",
+      "type": "image/png"
+    }
+  ]
+}
```
itsalaidbacklife, line 62:
```suggestion
}

```

manifest.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "Cuttle.cards",
+  "short_name": "Cuttle",
+  "description": "The deepest card game under the sea!",
+  "theme_color": "#FFF4D7",
+  "background_color": "#FFF4D7",
+  "display": "fullscreen",
+  "id": "/",
+  "start_url": "/",
```
itsalaidbacklife, line 9: Do these need to be any different for production?

Haviles04: I don't _think_ @seriouslysean any thoughts?

seriouslysean: Nah this seems fine to me, we don't run in any sort of subdirectory situation.

seriouslysean (commented):
Reconfirming this is good with me after the EOL entry is added.

Haviles04:
> Reconfirming this is good with me after the EOL entry is added.

What's an EOL?

Haviles04:
> > Reconfirming this is good with me after the EOL entry is added.
> 
> What's an EOL?

Nevermind, end of line!

itsalaidbacklife:
@Haviles04 and @seriouslysean do y'all have any insight into how this might affect caching? That's my only remaining concern -- keeping folks up to date when we roll out updates

seriouslysean:
@itsalaidbacklife  that’s why I requested the removal of the service worker, the rest of this stuff wrong mess with caching at all.

itsalaidbacklife (approved):
Very nice!

PR #730, opened 2023-10-31
Feature #723 - Animate when players leaves the lobby

### LINKED ISSUE #723: [Feature]: Animate when players leaves the lobby
### Feature Summary

When your opponent leaves your lobby we should see an animation indicating they are going to juice up the interaction

### Detailed Description

The lobby has player indicators for each player (you and your opponent) with three possible states: no player, player who isn't ready, and player who is ready. The ready player and player entry are on the works, this is the last state animation to add. 

I do have the animation done, only thing missing is adding the correct transition trigger which is on the works as well

https://github.com/cuttle-cards/cuttle/assets/36615882/988d1777-f363-4f53-8deb-72b204341bff

### LINKED ISSUE #729: [Bug]: Players joining the lobby animation ends while still overlapping the cards
### Bug Summary

When a player joins the lobby, we animate their player-indicator with some waves and a water effect. The bug is that when this ends, the water still overlaps the card

### Detailed Description

See this video to demonstrate the problem. Notice the blue wave at the bottom of the cards after the animation completes.

https://github.com/cuttle-cards/cuttle/assets/6520704/7e352e79-3736-4ee4-bc17-84c3309637ee

This was on a window about approximately 2400 x 1200 px in size

### Current Behavior

The water animation happens on top of the card and then afterwards still covers a little bit of the card.

### Expected Behavior

After the animation completes, the water does not cover the cards at all

### Steps to Reproduce

1. Sign up
2. Create a game
3. Join it
4. See how the water animation covers the bottom of the card

Rafajrg21:
## Issue number

- Resolves #723 
- Resolves #729  

## Please check the following

- For New Features:

A new animation is added for an opponent leaving the lobby, the animation is a pulse which goes well with the leave-lobby.mp3 available in the project sounds.

https://github.com/cuttle-cards/cuttle/assets/36615882/b979e113-e5e1-4666-b797-45eba0258b0c

This animation triggers for your opponent when it leaves the lobby and when you exit the lobby

<hr />
The other change, in LobbyPlayerIndicator fixes the wave animation and resolves #729. For testing, you can see how the animation react on three different breakpoints 540px - 1240px - 1920px.

### DIFF
--- a/src/routes/lobby/LobbyView.vue
+++ b/src/routes/lobby/LobbyView.vue
@@ -20,11 +20,14 @@
           <img src="/img/logo-stalemate.svg" class="vs-logo" alt="stalemate logo">
         </v-col>
         <v-col md="4" cols="12">
-          <LobbyPlayerIndicator
-            :player-username="gameStore.opponentUsername"
-            :player-ready="gameStore.opponentIsReady"
-            data-cy="opponent-indicator"
-          />
+          <Transition>
+            <LobbyPlayerIndicator
+              v-if="gameStore.opponentUsername"
+              :player-username="gameStore.opponentUsername"
+              :player-ready="gameStore.opponentIsReady"
+              data-cy="opponent-indicator"
+            />
+          </Transition>
         </v-col>
       </v-row>
       <v-row>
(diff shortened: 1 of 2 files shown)


itsalaidbacklife (changes requested):
Actually @Rafajrg21 this fails existing lobby tests because it changes the UX by removing the player indicator that asserted that the 'Invite' version of the indicator appeared (now the whole indicator only appears conditionally).

That said, the 'invite' text is something misleading because we want to have a way of copying your lobby link and inviting another user, but that hasn't been implemented yet. Still this is something we want soon, so how about we move the `Transition` logic into the `LobbyPlayerIndicator` itself and have that handle both entering and leaving of the card so that it's swapped with the "Invite" text.

This would also mean updating the existing water animation to be applied with the same transition tag but for the enter classes while the other would apply on leave, which I think would be more consistent

itsalaidbacklife (approved):
Nice! Looks good. This gels better with the existing animation setup and is a solid improvement to the UX

PR #712, opened 2023-10-27
Card flip

Posoroko:
## #711

Relevant [issue number](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
- Resolves #

## Please check the following

- [ ] Do the tests still pass? I did not write any test for it
- [ ] Is the code formatted properly? yes
- For New Features:
  - [ ] Have tests been added to cover any new features or fixes? no
  - [ ] Has the documentation been updated accordingly? 

## Please describe additional details for testing this change

I added the back and forth flip with a 3d feel.  I didn't use the vue transition component though.  It seemed more straight forward to just use basic css transition.  I hope that's ok.

PR #736, opened 2023-11-01
Fix for Issue/#720.

### LINKED ISSUE #720: [Bug]: Card not displaying the right value
### Bug Summary

After getting ready, exiting and joining again, player is shown as "ready"

### Detailed Description

On Firefox and chrome, the card can show "ready" when the player is not ready.  The variable "playerReady" for the other player shows "true" instead of "false".  

I recreated this bug in local and in production.

### Current Behavior

When a player does "playerReady = true", exit the lobby and come back, the other player sees them as "ready".

### Expected Behavior

A player comming back to the lobby should (I think) always be "not ready"

### Steps to Reproduce

1. create a game
2. player 0 joins the lobby
3. player 1 joins the lobby
4. player 1 get "ready"
5. player 1 exits the lobby
6. player 1 joins the lobby again

(I added an element to display the state of "playerReady" under each card)

Player 0 sees player 1's card as "ready" and the "playerReady" variable for player 0 equals "true".  
bug a - player 0

Player 1 sees cards correctly
bug a - player 1

7. Player 0 gets "ready"

Player 0 can see both card are "ready", when that shouldn't be possible.
showing ready when is not

At the same time, player 1 sees the expected behavior.
seeing cards correctly

Posoroko:
## Issue Number

- Resolves #720 

## Please check the following

- For New Features:

-Refactored the ready/notReady state for the card-flip animation.
-Opponent's ready state is reset in the state.  Client bug, server is ok.
-Fixed a bug where the opponent's indicator would display the player's userName.  The state.players array wasn't being sorted, so p0 was at index 1 and p1 at index 0
-created a test file for testing the lobby join/leave/ready actions.  (unfinished)


src/stores/game.js
```diff
@@ -285,6 +287,13 @@ export const useGameStore = defineStore('game', {
     },
     opponentLeft() {
       this.players = this.players.filter((player) => player.pNum === this.myPNum);
+      
+      //set opponent as not ready (bug repair)
```
itsalaidbacklife, line 291: Let's remove (bug repair) from the comment. Imo which code fixes which thing is useful to indicate in the PR commends (i.e. you can comment on a line in the PR like I'm doing now) to point out to the reviewer what's what, but the code comments themselves should just explain why we're doing what we're doing.
```suggestion
      // set opponent as not ready because they left
```

src/stores/game.js
```diff
@@ -272,6 +272,8 @@ export const useGameStore = defineStore('game', {
     },
     successfullyJoined(player) {
       this.players.push(cloneDeep(player));
+      // sort players so p0 is at index 0 and p1 is at index 1 (bug repair)
```
itsalaidbacklife, line 275: Let's put the comment above the push and explain both lines together
```suggestion
      // Add new player to game and sort based on pNum
      this.players.push(cloneDeep(player));
```

src/stores/game.js
```diff
@@ -387,7 +396,7 @@ export const useGameStore = defineStore('game', {
               this.myPNum = res.pNum;
               this.updateGame(res.game);
               this.successfullyJoined({
-                username: res.playerUsername,
+                username: res.username, // was playerUsername but it doesn't exist
```
itsalaidbacklife, line 399: Same idea with my above comments: comments that explain why you're making a specific change should go into the PR comments rather than the code comments because they are no longer helpful once the code is merged. Like if someone a year from now reads this comment it won't matter or be helpful that it used to be `res.playerUsername` and that this was incorrect. 

But it is helpful to include in the PR comments because it's relevant to the reviewer to have that perspective. So thanks for including the context!
```suggestion
                username: res.username, // was playerUsername but it doesn't exist 
```

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -215,12 +286,26 @@ describe('Lobby - P0 Perspective', () => {
...
+          expect(game.opponentIsReady).to.eq(true);
+        });
+      // opponent : ready class
+      cy.get('[data-cy=opponent-indicator]').find('[data-cy="lobby-card-container"]').should('have.class', 'ready');
       cy.reload();
-      cy.get('[data-cy=opponent-indicator]').find('[data-cy="lobby-ready-card"]').should('exist');
-      cy.get('[data-cy=my-indicator]').find('[data-cy="lobby-back-card"]').should('exist');
-
+      // openent's ready status should be true in the gameStore
+      cy.wait(1000); // fails without this wait
```
itsalaidbacklife, line 299: Generally `cy.wait()` smells that something is wrong with the app or the test. One issue with `cy.wait()` is that it extends the test longer than we want in the case that things run quicker. It's generally preferable instead to use retryable assertions with timeouts so that cypress will keep trying until the assertion passes, within a specified time interval (default is 4 seconds).

That said, it's surprising that this specific `cy.wait()` would be necessary because the assertion before it,
```
cy.get('[data-cy=opponent-indicator]').find('[data-cy="lobby-card-container"]').should('have.class', 'ready');
```

Should only be able to pass once the `opponentIsReady` value is set to true in the store because that's what the component is reading from. So I'd like to try removing this and confirming whether it's still necessary and if so, seeing what we can do to troubleshoot it
```suggestion
```

tests/e2e/specs/out-of-game/lobbyCards.spec.js
itsalaidbacklife, line 1: I don't think that this needs to be a separate test file, as it's all stuff inside the lobby. We should compare these tests against the existing ones in the `lobby.spec` and dedupe them so that we're only testing each flow once. 

That said it could be reasonable to define these helper functions in the lobby spec to make the tests more concise

src/routes/lobby/components/LobbyPlayerIndicator.vue
```diff
@@ -2,18 +2,16 @@
   <Transition name="cards" mode="out-in">
     <div v-if="playerUsername" class="player-card">
       <span class="player-name">{{ playerUsername }}</span>
-      <div class="card-container">
+      <div class="card-container" data-cy="lobby-card-container" :class="[ playerReady ? 'ready' : 'notReady' ]">
```
seriouslysean, line 5: `notReady` feels like a weird class name, I think it should be `not-ready`. Though it might be better to make them more deterministic like `is-ready` or `is-not-ready` if you need to. I'd also argue you may not need the not ready class at all since you would just make whatever is targeting it the default style.

Posoroko: @seriouslysean you're absolutely right. I can modify that

tests/e2e/specs/out-of-game/lobbyCards.spec.js
Posoroko: @itsalaidbacklife OK.  I'll  squish both files into a single one.  The functions really makes it more readable.  I'll keep them and tidy things up.

src/stores/game.js
Posoroko: yeo, ok, got it.  I didn't know about this part or the PR where we can comment every change.  It makes total sense!

Posoroko:
I push my modifications, but forgot it was going to trigger the pr again.  I'm not done yet, I'll finish over the weekend

src/stores/game.js
seriouslysean, line 275: The code is actually very clear and I don’t think we need a comment at all, it’s redundant.

src/stores/game.js
seriouslysean, line 291: Same here

src/stores/game.js
seriouslysean, line 399: No comment needed here either.

Posoroko: I'm confused, I corrected this... I'm git confused I think

src/stores/game.js
Posoroko: I'm confused. Is this comment still there?  I don't see it

seriouslysean: Says outdated now, might have been weird timing on GitHub’s part.

Posoroko:
There is one test ( the very last one ) that I don't know how to set up.  Could someone help me out?

src/stores/game.js
```diff
@@ -285,6 +287,12 @@ export const useGameStore = defineStore('game', {
     },
     opponentLeft() {
       this.players = this.players.filter((player) => player.pNum === this.myPNum);
+
+      if(this.myPNum == 0) {
```
itsalaidbacklife, line 291: We use === to avoid null and undefined goofery
```suggestion
      if(this.myPNum === 0) {
```

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -245,112 +302,263 @@ describe('Lobby - P1 Perspective', () => {
...
+      assertGameStarted();
   });
 
-  it('Game starts when both players are ready - opponent readies first after player joins', () => {
+  it('Game starts when both players are ready - opponent readies first after player joins', function () {
     cy.readyOpponent();
-    cy.get('[data-cy=ready-button]').click();
-    assertGameStarted();
+    playerGetsReady();
+      assertGameStarted();
```
itsalaidbacklife, line 392: whitespace
```suggestion
    assertGameStarted();
```

Posoroko: when writing the file, I indented with 1 white space to create visual separation between the actions and the tests.   Click button, make some verification.  I can see how that might not be a standard practice, but it felt much easier on the eyes and more readable. 
I can take the white space out, no problem.  Would there be a better way to make it easier to read while following good practices?

src/stores/game.js
```diff
@@ -285,6 +287,12 @@ export const useGameStore = defineStore('game', {
     },
     opponentLeft() {
       this.players = this.players.filter((player) => player.pNum === this.myPNum);
+
+      if(this.myPNum === 0) {
+        this.p1Ready = false;
+      } else {
+        this.p0Ready = false;
+      }
```
Haviles04, line 295: We already have `updateReady()` in the store, so I think we can utilize it like this

```suggestion

      updateReady(this.myPnum + 1 % 1)
```

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -245,112 +304,263 @@ describe('Lobby - P1 Perspective', () => {
...
+      expect(game.topCard.rank).to.be.greaterThan(0);
+      expect(game.secondCard.rank).to.be.greaterThan(0);
+      expect(game.scrap.length).to.eq(0);
+      expect(game.twos.length).to.eq(0);
+    });
+}
+
+function clickReadyButton() {
+  cy.get('[data-cy=ready-button]').click();
+}
```
Haviles04, line 497: I'm not sure this is worth putting into a helper function, it's really only one line and reading it in the test makes me think that it is doing something more than just clicking the button.

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -245,112 +304,263 @@ describe('Lobby - P1 Perspective', () => {
...
+
+  cy.get(`[data-cy=${userString}-indicator]`).find('[data-cy="lobby-card-container"]')
+    .should('exist');
+  cy.get(`[data-cy=${userString}-indicator]`).find('[data-cy="lobby-back-card"]').should('exist');
+  cy.get(`[data-cy=${userString}-indicator]`).find('[data-cy="lobby-ready-card"]').should('exist');
+}
+
+function opponentIndicatorShouldBeEmpty() {
+  cy.get('[data-cy=opponent-indicator]').contains('Invite');
+}
```
Haviles04, line 534: This as well, I think we should just use the regular cypress command

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -245,112 +304,263 @@ describe('Lobby - P1 Perspective', () => {
...
+  clickReadyButton();
+}
+function playerGetsNotReady() {
+  clickReadyButton();
+}
+
+
+function playerLeavesLobby() {
+  cy.get('[data-cy=exit-button]').click();
+}
```
Haviles04, line 555: These too

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -1,99 +1,57 @@
...
-        expect(game.players.length).to.eq(2);
-        expect(game.players[0].hand.length).to.eq(5);
-        expect(game.deck.length).to.eq(39);
-      }
-      expect(game.topCard.rank).to.be.greaterThan(0);
-      expect(game.secondCard.rank).to.be.greaterThan(0);
-      expect(game.scrap.length).to.eq(0);
-      expect(game.twos.length).to.eq(0);
-    });
-}
```
Haviles04, line 40: I generally think it's better to have the helper functions at the top of the file, this way I can browse through them first before looking at test, especially in files with hundreds if not thousands of lines

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -1,99 +1,57 @@
...
 
-function checkRanked(ranked) {
-  cy.window()
-  .its('cuttle.gameStore')
-    .then((game) => {
-      expect(game.isRanked).to.eq(ranked);
-    });
-}
+import en from '../../../../src/translations/en.json';
```
Haviles04, line 4: I think it makes sense to just use the words instead of the translations from the file, considering most of them are just one word and we are only pulling english

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -245,112 +304,263 @@ describe('Lobby - P1 Perspective', () => {
...
+    cy.readyOpponent();
+    cy.reload();
+    cy.url().should('include', '/lobby');
+    indicatorShowsNameAndCard('player');
+    userShouldDisplayReady('player', false);
+    indicatorShowsNameAndCard('opponent');
+    userShouldDisplayReady('opponent', true);
+    testReadyValueInState('p1', false);
+    testReadyValueInState('p0', true);
+    checkOpponentIsReadyInGetter(true);
```
Haviles04, line 376: I think these can likely be combined

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -111,124 +69,225 @@ describe('Lobby - P0 Perspective', () => {
...
+    cy.readyOpponent();
+    cy.reload();
+    cy.url().should('include', '/lobby');
+    indicatorShowsNameAndCard('player');
+    userShouldDisplayReady('player', false);
+    indicatorShowsNameAndCard('opponent');
+    userShouldDisplayReady('opponent', true);
+    testReadyValueInState('p0', false);
+    testReadyValueInState('p1', true);
+    checkOpponentIsReadyInGetter(true);
```
Haviles04, line 284: I think we can probably combine these in 1 or 2 test

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -111,124 +69,225 @@ describe('Lobby - P0 Perspective', () => {
...
+    checkOpponentIsReadyInGetter(null);
+    // joins back
+    cy.get('[data-cy=game-list-item]').contains('button.v-btn', 'Join Casual').click();
+    indicatorShowsNameAndCard('player');
+    userShouldDisplayReady('player', false);
+    indicatorShowsNameAndCard('opponent');
+    userShouldDisplayReady('opponent', false);
+    testReadyValueInState('p0', false);
+    testReadyValueInState('p1', false);
+    checkOpponentIsReadyInGetter(false);
```
Haviles04, line 134: I think we can combine these

Haviles04 (changes requested):
A lot of work done here, this will be a nice fix. I suggested a few changes. I think we can likely combine a lot of the new test which will also reduce the need for the helper functions.

I think we can utilize `cy.wrap` to be able to access the game data throughout the test instead of using some of the helpers.
https://docs.cypress.io/api/commands/wrap


I'm around on discord if you need any help :D

tests/e2e/specs/out-of-game/lobby.spec.js
Posoroko: I feel it would be more consistent with the other similar functions.  It seems more clear what the test is. 
But I don't mind swapping it back.
As I was rewriting the tests, it was much easier to read function names, that's why there all over the place.

tests/e2e/specs/out-of-game/lobby.spec.js
Posoroko: yeah I see what you mean.  My idea was that with explicit function names, you don't need to check what they do to understand the test.

seriouslysean: I see the intention you were going for though. A bit of advice that I try and adhere to when working on individual tickets; keep the concerns separated. 

This PR has a specific, targeted fix and while the other changes might be valuable it’s probably better to fork that off to a new issue so we can discuss your reasoning and come up with a better solve to be fixed in a separate update.

tests/e2e/specs/out-of-game/lobby.spec.js
Posoroko: OK.  My idea here was that if we modify/correct the displayed text, we'll need to update all the tests too.  Are we testing the actual text ? or just the fact that the text gets displayed and that it's fetched from the correct source ?

tests/e2e/specs/out-of-game/lobby.spec.js
Posoroko: yup

tests/e2e/specs/out-of-game/lobby.spec.js
Haviles04: While I agree that it's clearer this way, I'm not sure it's worth another function for a single assertion, especially the way cuttle is setup IMO the assertion is pretty clear.

tests/e2e/specs/out-of-game/lobby.spec.js
Haviles04: I agree that if we made changes we'd have to come back to the test and edit it, but I don't think we use the translations in any other test files except the specific localization files and for the sake of consistency we should avoid it here. I think using the translations could possibly be something we can look into, but on a separate PR

tests/e2e/specs/out-of-game/lobby.spec.js
seriouslysean: It’s not worth the abstraction, continue using cypress commands directly until there’s more to encapsulate.

Posoroko:
I realize I got off track with too many changes that were not related to this specific issue.  @seriouslysean you were right.
So I rolled back to the original version of the file and just added the missing tests related to this PR:  when players get ready, leave and join back.
I also modified other assertions that became useless since pr #712.  This PR introduced a new way of handling lobby-cards.

src/stores/game.js
Haviles04: Okay Yea I understand you, maybe check if the opponent is ready, and then use updateReady()

Posoroko: @Haviles04 I push the changes we agreed on.  I had to move it to the first line though, as the code that was already there deletes the opponent from the players array

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -304,6 +339,28 @@ describe('Lobby - P1 Perspective', () => {
...
 
+    it('readying, exit, joining back', () => {
+    cy.window()
+    .its('cuttle.gameStore')
+    .then((gameData) => {
+      //player ready, exit, join back
+      cy.get('[data-cy=ready-button]').click();
+      cy.get('[data-cy=exit-button]').click();
+      cy.get('[data-cy=game-list-item]').contains('button.v-btn', 'Join Casual').click();
+      cy.get(`[data-cy=my-indicator]`).find('[data-cy="lobby-card-container"]').should('not.have.class', 'ready');
```
itsalaidbacklife, line 350: Single quotes please

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -304,6 +339,28 @@ describe('Lobby - P1 Perspective', () => {
...
+      expect(gameData.p1Ready).to.equal(false);
+
+      //opponent ready, exit, join back
+      cy.readyOpponent();
+      cy.leaveLobbyOpponent();
+      cy.get(`[data-cy=opponent-indicator]`).find('[data-cy="lobby-card-container"]').should('not.exist');
+      expect(gameData.p0Ready).to.equal(false);
+      cy.subscribeOpponent(gameData.id);
+      expect(gameData.p0Ready).to.eq(false);
+      cy.get(`[data-cy=opponent-indicator]`).find('[data-cy="lobby-card-container"]').should('exist');
```
itsalaidbacklife, line 360: Single quotes

itsalaidbacklife (commented):
This looks great! There's one random file that was added and then a couple places that should have the quotes replaced, but otherwise this looks good to go

tests/e2e/downloads/iryHRlEC.html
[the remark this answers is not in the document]
Posoroko: I think this file was created when I tried setting up the component testing.  I'll clean it out.

Posoroko:
@Haviles04  
> I see some test were changed to arrow functions, then changed back, now we have a mix, can we switch them all to arrow?

There was a problem with some arrow functions that need to call this.gameSummary.  That's why I changed them back to regular functions.

Haviles04:
> @Haviles04
> 
> > I see some test were changed to arrow functions, then changed back, now we have a mix, can we switch them all to arrow?
> 
> There was a problem with some arrow functions that need to call this.gameSummary. That's why I changed them back to regular functions.

Ah Okay, I understand.. I think we can fix that by using `cy.get('game-summary').then()`, but that's probably for a different PR

Haviles04 (approved):

Haviles04:
LGTM, leaving it to @itsalaidbacklife for final approval

src/stores/game.js
```diff
@@ -225,6 +227,9 @@ export const useGameStore = defineStore('game', {
       }
     },
     opponentLeft() {
+      const opponentPnum = this.opponent.pNum;
```
seriouslysean, line 230: Is this used elsewhere in this better? Might be able to just inline it.

src/stores/game.js
```diff
@@ -334,7 +339,7 @@ export const useGameStore = defineStore('game', {
               this.myPNum = res.pNum;
               this.updateGame(res.game);
               this.successfullyJoined({
-                username: res.playerUsername,
+                username: res.username,
```
seriouslysean, line 342: You could destructure these off red and use the directly to clean up this function some. It’s also allow shorthand when putting keys in this object.
```
{ username, pNum }
```

seriouslysean (approved):
Minor callouts, no blockers, leaving final approval to @itsalaidbacklife 👍🏼

src/stores/game.js
itsalaidbacklife: Given its use in the template literal, I think it's easier to read this way

itsalaidbacklife (approved):

PR #738, opened 2023-11-01
Feat #627 Counter Dialog Localization

### LINKED ISSUE #627: [Feature]: Apply translation/localization to the text in the `CounterDialog`
### Feature Summary

We should use `i18n` to translate all the text in the `CounterDialog` into the user's selected locale

### Detailed Description

We should pull all of the text in the CounterDialog into the translation json files e.g. src/translations/en.json, src/translations/es.json, and src/translations/fr.json in under a key called "game.dialogs.counterDialog"

Look at src/views/LoginView for an example of a page that is correctly using $i18n to localize/translate content. The user-visible strings have been removed from the code and placed into a "login" section of the translation files in src/translations.

The page then imports and consumes the useI18n() composable to grab the t() function used to render/translate text by pulling from the appropriate language's json file. This is done using a setup() block of the page like so:

  setup() {
    const { t } = useI18n();
    return { t };
  },
From there, we can grab a given string by name in the appropriate locale using {{ t('login.joinDiscord') }} since each translation json in this case has a login attribute for all the text in the login page, with a joinDiscord child attribute for the text to show in the 'Join our discord community' button.

We should do the same thing to the CounterDialog by extracting all the text in the page into a rules section of each translation json file (e.g. src/translations/en.json) and then configuring CounterDialog.vue with the useI18n() composable and rendering all its text with t() to translate it to the user's selected locale.

All the text for the CounterDialog should be stored in each language file under the key "game.dialogs.counterDialog" (generally we have one top-level key for each page e.g. "game" and "login", then specific sub-keys for components in that page (additionally grouping all the dialogs under the sub-key "game.dialogs"

The end result of this change should be the ability to see the text of the `CounterDialog` in your preferred locale.

Haviles04:
## Issue number

- Resolves #627 

## Please check the following

- For New Features:

-Combined `counterDialogs` with `cannotCounterDialogs`
-Added localization to Counter Dialogs

### DIFF
--- a/src/routes/game/components/ChooseWhetherToCounterDialog.vue
+++ b/src/routes/game/components/ChooseWhetherToCounterDialog.vue
@@ -2,21 +2,21 @@
   <BaseDialog id="counter-dialog" v-model="show" title="Chance to Counter">
     <template #body>
       <div v-if="!opponentLastTwo" class="my-2">
-        Your opponent has played the 
+        {{ t('game.dialogs.counterDialogs.opponentPlayed') }}
         <GameCardName :card-name="oneOff.name" />
-        as a one-off
+        {{ t('game.dialogs.counterDialogs.oneOff') }}
         <span v-if="target"> 
-          targeting your 
+          {{ t('game.dialogs.counterDialogs.target' + 'global.your') }} 
           <GameCardName :card-name="target.name" />
         </span>
       </div>
       <div v-else class="my-2">
-        Your opponent has played 
+        {{ t('game.dialogs.counterDialogs.opponentPlayed') }}
         <GameCardName :card-name="opponentLastTwo.name" />
 
-        to Counter
+        {{ t('game.dialogs.counterDialogs.toCounter') }} 
         <span v-if="playerLastTwo">
-          your 
+          {{ t('global.your') }}
           <GameCardName :card-name="playerLastTwo.name" />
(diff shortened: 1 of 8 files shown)


src/routes/game/components/ChooseWhetherToCounterDialog.vue
```diff
@@ -1,22 +1,22 @@
...
-        as a one-off
+        {{ t('game.dialogs.counterDialogs.oneOff') }}
         <span v-if="target"> 
-          targeting your 
+          {{ t('game.dialogs.counterDialogs.target' + 'global.your') }} 
           <GameCardName :card-name="target.name" />
         </span>
       </div>
       <div v-else class="my-2">
-        Your opponent has played
```
Haviles04, line 14: I'm not sure if the 'the' was intentionally left out here, but I did add it to only use one variable, if this is going to be an issue I can make another, I also updated the test to now include using 'The'

itsalaidbacklife: I think it was an oversight and I don't see a reason for them to be different. I prefer to keep the 'the' so this is good

itsalaidbacklife (approved):
Nicely done. Good call combining the counter + cannot counter dialog translations, as there is so much overlap.

PR #742, opened 2023-11-05
Bug #702 Sort Top Player

### LINKED ISSUE #702: [Bug]: When spectating, the top player's hand is not sorted
### Bug Summary

When viewing a game as a spectator, you see p0's hand at the bottom correctly sorted, but p1's hand above is not sorted

### Detailed Description

See example image from a spectated game:
image

This is a direct result of the updates in #698 which deliberately does not sort the opponent hand when you are _playing_ the game. We should update the `gameStore.updateGame` function so that the opponent's hand is sorted only if you are viewing as a spectator

### Current Behavior

- Only the bottom player (p0)'s hand is sorted when viewing as a spectator (incorrect)
- Only your own hand is sorted when viewing as a player (correct)

### Expected Behavior

- Both players' hands are sorted when viewing as a spectator (this is the change)
- Only your own hand is sorted when viewing as a player (this should stay the same)

### Steps to Reproduce

1. Join a game as a spectator
2. Look at the top player's hand
3. See it isn't sorted

Haviles04:
## Issue number

- Resolves #702 

## Please check the following

- For New Features:

- Added `isSpectating` variable to state
- Updated `isSpectating` on `case.initialize` and `requestSubscribe`
- Updated sort function to check if `isSpectating === true` or `hasGlassesEight === true`


src/plugins/sockets/inGameEvents.js
```diff
@@ -30,10 +30,11 @@ export async function handleInGameEvents(evData) {
     }
     case SocketEvent.INITIALIZE: {
       // Update state
-      gameStore.resetPNumIfNullThenUpdateGame(evData.game);
       if (isSpectating) {
         gameStore.myPNum = 0; // always spectate as p0
+        gameStore.isSpectating = true;
```
seriouslysean, line 35: I understand why we’re setting this, but it feels weird to me to consider a normal player a spectator. Is this initialize function used anywhere else or just in the main game start up logic?

src/stores/game.js
```diff
@@ -160,7 +166,7 @@ export const useGameStore = defineStore('game', {
       return state.turn % 2 === state.myPNum;
     },
     hasGlassesEight: (state) => {
-      return state.player.faceCards.filter((card) => card.rank === 8).length > 0;
+      return state.player?.faceCards?.filter((card) => card.rank === 8).length > 0 ?? null;
```
seriouslysean, line 169: Shouldn’t player always be set? Is there a case where it’s not set that we should be fixing instead of no-oping here?

seriouslysean (commented):
Couple questions since I’m not as familiar with the code here.

src/plugins/sockets/inGameEvents.js
Haviles04: I'm pretty sure it's called at the main game startup, I kind of put it here as a backup I'm not sure it's 100% needed because I set it in requestSubscribe,

src/stores/game.js
Haviles04: I was getting "can't read .facecards from undefined" on the initial joining of the lobby. I guess because it is checking for it before the players are actually set

seriouslysean: Sounds like we should be fast failing on that case instead then. If there are many cases checking this where we benefit from the chaining I’m not against setting a logical default, I just want to make sure you’re doing your due diligence in validating.

Haviles04: I think fast fail is a good idea, I think it's only checked in two spots

src/stores/game.js
```diff
@@ -13,22 +13,27 @@ function queenCount(player) {
   }
   return player.faceCards.reduce((queenCount, card) => queenCount + (card.rank === 12 ? 1 : 0), 0);
 }
-  
+
 const compareByRankThenSuit = (card1, card2) => {
-  return (card1.rank - card2.rank) || (card1.suit - card2.suit);
+  return card1.rank - card2.rank || card1.suit - card2.suit;
```
itsalaidbacklife, line 18: I know this is a linter thing but imo this reads more clearly with the ().
```suggestion
  return (card1.rank - card2.rank) || (card1.suit - card2.suit);
```

src/stores/game.js
```diff
@@ -13,22 +13,27 @@ function queenCount(player) {
...
 
-const setPlayers = (player, myPnum) => {
-  const sortP1 = (cards) => player.pNum === myPnum ? cards?.sort(compareByRankThenSuit) : cards;
+const setPlayers = (player, myPnum, hasGlassesEight, isSpectating) => {
+  const sortCards = (cards) => {
+    if (isSpectating || hasGlassesEight || player.pNum === myPnum) {
+      return cards?.sort(compareByRankThenSuit);
+    }
+    return cards;
+  };
```
itsalaidbacklife, line 27: This is slick. I think it's an elegant solution. Nicely done

src/stores/game.js
```diff
@@ -160,7 +166,10 @@ export const useGameStore = defineStore('game', {
       return state.turn % 2 === state.myPNum;
     },
     hasGlassesEight: (state) => {
-      return state.player.faceCards.filter((card) => card.rank === 8).length > 0;
+      if (!state.player) {
+        return false;
+      }
+      return state.player.faceCards?.filter((card) => card.rank === 8).length > 0 ?? null;
```
itsalaidbacklife, line 172: This should coalesce to false not null so that the type is always a boolean
```suggestion
      return state.player.faceCards?.filter((card) => card.rank === 8).length > 0 ?? false;
```

src/plugins/sockets/inGameEvents.js
itsalaidbacklife: I think it might be better to read `isSpectating` directly off the currentRoute in the router so that we don't need to set it in multiple different places. That way we have a centralized source of truth for isSpectating (the route) and we don't need to worry about keeping them in sync.

itsalaidbacklife (commented):
Nice work, this is looking good. I just have a couple nitpicks. Well done

src/plugins/sockets/inGameEvents.js
Haviles04: Would it make sense to keep this variable in the store, but set it to true in the `beforeEnter` in the router? And then we can change `isSpectating` in `GameView` to `gameStore.isSpectating`

src/stores/game.js
```diff
@@ -160,7 +166,10 @@ export const useGameStore = defineStore('game', {
       return state.turn % 2 === state.myPNum;
     },
     hasGlassesEight: (state) => {
-      return state.player.faceCards.filter((card) => card.rank === 8).length > 0;
+      if (!state.player) {
+        return false;
+      }
+      return state.player.faceCards?.filter((card) => card.rank === 8).length > 0 ?? false;
```
itsalaidbacklife, line 172: I think these can be collapsed by just nullish coalescing state.players too like
```suggestion
      return state.player?.faceCards?.filter((card) => card.rank === 8).length > 0 ?? false;
```

itsalaidbacklife (approved):
Nice work; looks good! One nitpick to DRY up a function and this is good to go

itsalaidbacklife (approved):
Nice work!

PR #743, opened 2023-11-05
fix for #739

### LINKED ISSUE #739: [Bug]: Can't close the snackbar in the `LobbyView` using its 'X' button
### Bug Summary

The `LobbyView` has a snackbar that appears when you the game is changed from ranked to casual or vice versa. It should be possible to click the 'X' button in this snackbar to close it, but this button doesn't work.

### Detailed Description

See pic:
image

This instance of `BaseSnackbar` in the `LobbyView` closes on a timeout (as it should), but it does not respond to clicking the 'X' button. We should investigate this page to identify why the snackbar doesn't close when you click the X button.

### Current Behavior

Clicking the 'X' button does not close the snackbar.

### Expected Behavior

Clicking the 'X' button does close the snackbar.

### Steps to Reproduce

1. Signup/log in
2. Create a game
3. Join the game
4. Hit the switch to change from ranked to casual or casual to ranked
5. See the snackbar in the bottom of the screen and click its 'X' button
6. Notice that the snackbar does not close

Posoroko:
## Issue Number
- Resolves #739 

## Please check the following

- For New Features:

-handled the emit to clear the snack bar
-added a test file  for the snack bar (I will eventually incorporate the tests in the lobby.spec.js file, but because there is a pending PR about this file, I was afraid to end up git confused )


tests/e2e/specs/out-of-game/snackbar.spec.js
itsalaidbacklife, line 1: Rather than making a new test file for this, let's update the existing test for changing the ranked/casual mode in `tests/e2e/specs/out-of-game/lobby.spec.js`

The existing test is called `it('Changes games to ranked and casual from the lobby')`. We can just add assertions to that the snackbar appears, has the right text, and closes when clicked. 

As a rule of thumb in cypress, it's more efficient to have one longer test that makes multiple assertions, rather than many tests that each make a small number of assertions. So if there's already a test case for a given thing, it's preferable to beef up that test case with additional assertions rather than make a new one from scratch. This is because cypress will show us specifically which assertions failed within a test and because there is more performance overhead to each individual test that runs compared with each assertion (e.g. the beforeEach() hook that runs before each test)

tests/e2e/specs/out-of-game/snackbar.spec.js
```diff
@@ -0,0 +1,47 @@
...
+    cy.get('[data-cy=edit-snackbar]').contains(`${en.lobby.rankedChangedAlert} ${en.global.ranked}`);
+    clickGameModeSwitch();
+    cy.get('[data-cy=edit-snackbar]').contains(`${en.lobby.rankedChangedAlert} ${en.global.casual}`);
+  });
+
+  it('Changes games type. close snackbar', () => {    
+    clickGameModeSwitch();
+    cy.get('[data-cy=close-snackbar]').click();
+    cy.get('[data-cy=edit-snackbar]').should('not.exist');
+  });
```
itsalaidbacklife, line 27: Rather than making each of these a separate test case, we can combine them into the existing lobby test case for switching game modes as additional assertions in a single test

tests/e2e/specs/out-of-game/snackbar.spec.js
```diff
@@ -0,0 +1,47 @@
...
+    clickGameModeSwitch();
+    cy.get('[data-cy=edit-snackbar]').should('exist');
+    
+  });
+
+  it('Changes games type. check appropriate text', () => {
+    clickGameModeSwitch();
+    cy.get('[data-cy=edit-snackbar]').contains(`${en.lobby.rankedChangedAlert} ${en.global.ranked}`);
+    clickGameModeSwitch();
+    cy.get('[data-cy=edit-snackbar]').contains(`${en.lobby.rankedChangedAlert} ${en.global.casual}`);
```
itsalaidbacklife, line 20: Unless we're specifically testing translations I think it's best to use the literal expected text in english since it doesn't have potential gotchas around missed keys in the json
```suggestion
    cy.get('[data-cy=edit-snackbar]').contains(`${en.lobby.rankedChangedAlert} Ranked`);
    clickGameModeSwitch();
    cy.get('[data-cy=edit-snackbar]').contains(`${en.lobby.rankedChangedAlert} Casual`);

```

src/composables/fullScreen.js
itsalaidbacklife, line 1: It looks like this got tangled up with the fullscreen change. These should be kept in separate branches so they can be merged independently. To do so, you can either checkout main + pull main + checkout a new branch and add the specific changes needed for this issue to that then close this PR and open a new one, or you can touch up this branch to remove the extraneous changes

src/routes/lobby/LobbyView.vue
```diff
@@ -182,6 +184,9 @@ export default {
   right: 0;
   top: 20px;
   width: min-content;
+  display: flex;
+  flex-direction: column;
+  align-items: flex-center;
```
itsalaidbacklife, line 189: I think these are also changes for the fullscreen PR
```suggestion
```

src/routes/lobby/LobbyView.vue
```diff
@@ -109,8 +111,8 @@ export default {
   components: {
     LobbyPlayerIndicator,
     BaseSnackbar,
-    TheLanguageSelector,
-  },
+    TheLanguageSelector
```
itsalaidbacklife, line 114: We generally use trailing commas in objects
```suggestion
    TheLanguageSelector,
```

src/routes/game/GameView.vue
itsalaidbacklife, line 1: There's no need for a change in this file

src/components/BaseSnackbar.vue
itsalaidbacklife, line 1: I don't think there are any necessary changes in this file

.gitignore
itsalaidbacklife, line 1: These will hopefully go away after merging a fix for #749. This file does not need to be changed for this update

src/components/TheUserMenu.vue
itsalaidbacklife, line 1: Similarly, this file doesn't need changing for this PR

src/translations/en.json
```diff
@@ -12,7 +12,9 @@
     "cancel": "Cancel",
     "ranked": "Ranked",
     "casual": "Casual",
-    "your": "your"
+    "your": "your",
+    "fullScreen": "full screen",
+    "exitFullScreen": "Exit Full Screen"
```
itsalaidbacklife, line 17: For the fullscreen pr
```suggestion
    "your": "your"
```

itsalaidbacklife (changes requested):
Nice work. The fix itself looks right, but there are some administrative changes necessary for this to merge.

The main issue is that this code got mixed up with #744. In the future, to avoid that, checkout `main` and pull before checking out your new feature branch when submitting multiple PR's. The new branch always starts off a copy of the one you'd checked out when making the new branch so this way you'll always copy main for the new branch and avoid cross-contamination.

Other than that I'm requesting consolidating the tests as assertions in the existing test case for changing game modes, and to remove unnecessary changes

tests/e2e/specs/out-of-game/snackbar.spec.js
Posoroko: Yep ok . Got it. I was wandering about the length of the tests. I'll incorporate the snack bar in the lobby file and make sure we run as fewer tests as possible.

src/composables/fullScreen.js
Posoroko: Yes I got git confused a few times and mixed my branches.. I'm sorry about this.  I'm still getting used to work with multiple branches.

src/routes/lobby/LobbyView.vue
Posoroko: Got it

PR #747, opened 2023-11-06
Snackbar errors translated in login view #584

### LINKED ISSUE #584: [Feature]: Enable translating snackbar errors in LoginView
### Feature Summary

Error messages appearing in snackbars cannot yet be translated because they are generated serverside and the server doesn't know language preference

### Detailed Description

In order to be able to translate the snackbar errors, we need to change the way they are rendered so that the server sends a key to be entered into a language file, and the frontend uses that key to lookup and translate the respective message.
image

We could organize the snack messages for each page in the language files like so:

```
en.json
{
  "global": {}, // text for global/app wide text
  "home": {
    // snackbar errors for the home page
    "snackbar": {
        'gameIsFull': "Cannot join that game because it's already full"
        // other home page specific errors
    },
  },
  // login page text
  "login": {
    "noAccount": "Don't have an account?",
    "haveAccount": "Already have an account?",
    //   other login page text etc
    // ....
    "snackbar": {
       "usernameIsTaken": "That username is already registered to another user; try logging in!"
       "wrongPassword": "Username and password do not match",
       // other username specific errors
     },
  },
  // other pages' 
}
```

Then the error message from the server can send the key to the snackbar message. So instead of the current error sent when username doesn't match password:
```
return reject({ message: 'Username and password do not match' });
```

The server would send 
```
return reject({ message: 'wrongPassword' });
```

And when consuming the message, the client would use `i18n`'s `t()` function to render the translated message:
```
    handleError(messageName) {
      this.showSnackBar = true;
      this.snackBarMessage = t(`login.${messageName}`);
      this.loading = false;
    },
```

This way each language file e.g. `en.json` and `fr.json` can define its own translations of the error messages so they can be translated in the UI.

Posoroko:
## #584

- Resolves #584

## Please check the following

- For New Features:

-modified userController.js to return the json key instead of hard coded message 
-modified loginView.vue to handle the message key and pass the proper error message to the snack bar
-added the error message to en.json


.gitignore
```diff
@@ -1,145 +1,145 @@
...
+# committing settings which will smash the local
+# settings of  other developers on your team.
+#
+# Some reasonable defaults are included below,
+# but, of course, you should modify/extend/prune
+# to fit your needs!
+################################################
+
+
+.husky/pre-push
```
itsalaidbacklife, line 21: Why have you added this? This was intended to be in the source code. Has the prepush message been annoying? I see another PR from a different user who did the same so I'm wondering what's motivating the change

.gitignore
itsalaidbacklife, line 1: Something looks to be off with the git config if it's saying this entire file was updated just to edit seemingly the one line about `.husky/prepush`. My suspicion is the line endings are coming in with crlf isntead cr like unix systems has. We should figure this out to reduce noise and keep the line endings consistent

api/controllers/UserController.js
itsalaidbacklife, line 1: Same idea as above; git is saying this entire file was rewritten even though there are only two lines changing. We should sort out why

.gitignore
Posoroko: I've had this issue a few times.  the pre-push file was annoying, so I tried to add it to the gitignore, but it didn't work.  It was a week ago, I reinstalled everything since then because I was git confused.  I didn't think this line was still in there though.  I didn't mean to keep it.
I don't know what this pre-push is and what it does...

.gitignore
Posoroko: I saw that the whole file was updated.  I don't know anything about crlf/cr.  I would gladly learn though.

api/controllers/UserController.js
itsalaidbacklife: I'm hopeful that fixing #749 will resolve the issue

src/routes/login/LoginView.vue
```diff
@@ -1,393 +1,393 @@
...
+                target="_blank"
+              >
+                <img class="discord" src="/img/loginView/logo-discord.svg">
+                <span class="discordButton">{{ t('login.joinDiscord') }}</span>
+              </v-btn>
+            </v-form>
+
+            <BaseSnackbar
+              v-model="showSnackBar"
+              :message="t(snackBarMessage)"
```
itsalaidbacklife, line 134: atm the variable `snackBarMessage` isn't actually the snackbar message, it's the key to that message. Let's move the `t()` call into the function that _sets_ `snackBarMessage` so that this can be simpler here
```suggestion
              :message="snackBarMessage"

```

src/routes/login/LoginView.vue
```diff
@@ -1,393 +1,393 @@
...
+    },
+    handleLogin() {
+      this.username = '';
+      this.pw = '';
+      this.loading = false;
+      this.$router.push('/');
+    },
+    handleError(messageKey) {
+      this.showSnackBar = true;
+      this.snackBarMessage = messageKey;
```
itsalaidbacklife, line 279: Let's move the translation here
```suggestion
      this.snackBarMessage = this.t(messageKey);
```

src/routes/login/LoginView.vue
```diff
@@ -1,393 +1,393 @@
...
+    BaseSnackbar,
+    BaseVideo,
+    MarkdownContent,
+    TheLanguageSelector
+  },
+  setup() {
+    // Vuetify has its own translation layer that isn't very good
+    // It seems to conflict with the namespace of vue-i18n so we need to import it at the component
+    // level and utilize it this way with a composable. There may be another more global way but
+    // I haven't found anything just yet
```
itsalaidbacklife, line 201: While we're at it, let's remove this comment
```suggestion
```

.gitignore
itsalaidbacklife: @Posoroko this file is still being updated erroneously in this PR. Can you revise this so that `.gitignore` is unchanged?

PR #748, opened 2023-11-06
fixed bug #745

### LINKED ISSUE #745: [Bug]: Player counts on games in the home view can be negative or greater than two
### Bug Summary

The HomeView shows games available to join. They should show 0/2, 1/2, or 2/2 players. Sometimes they can show -1/2 or 3/2

### Detailed Description

See pic:
Screen_Shot_2023-11-02_at_12 27 26_PM

This happens sometimes when socket messages are missed intermittently e.g. to players join a lobby and then leave it, but the your client misses the second player joined. So the event stream looks to your client like: join, leave, leave. This sets the player count to -1 instead of zero.

To fix this easily we can update the gameList store so that the `joinGame()` and `otherLeftGame()` actions which increment and decrement player counts use `Math.min()` and `Math.max()` respectively to ensure that the player count never goes above 2 or below 0. So instead of 
```
updatedGame.numPlayers++;
```

we’d have
```
updatedGame.numPlayers = Math.min(2, updatedGame.numPlayers + 1)
```

and the reciprocal for otherLeftGame 

### Current Behavior

The player counts of games in the gameList can sometimes be < 0 or > 2.

### Expected Behavior

The player counts are always either 0, 1, or 2

### Steps to Reproduce

1.
2.
3.

pradnyabhukan:
## Issue number : #745 

- Resolves #745 

## Please check the following

- For New Features:

### DIFF
--- a/src/stores/gameList.js
+++ b/src/stores/gameList.js
@@ -52,16 +52,14 @@ export const useGameListStore = defineStore('gameList', {
     joinGame(data) {
       const updatedGame = this.openGames.find((game) => game.id === data.gameId);
       if (updatedGame) {
-        // updatedGame.numPlayers++;
-        updatedGame.numPlayers = Math.min(2, updatedGame.numPlayers + 1)
+        updatedGame.numPlayers = Math.min(2, updatedGame.numPlayers + 1);
         updatedGame.status = data.newStatus;
       }
     },
     otherLeftGame(gameId) {
       const updatedGame = this.openGames.find((game) => game.id === gameId);
       if (updatedGame) {
-        // updatedGame.numPlayers--;
-        updatedGame.numPlayers = Math.min(2, updatedGame.numPlayers - 1)
+        updatedGame.numPlayers = Math.min(2, updatedGame.numPlayers - 1);
       }
     },
     setIsRanked({gameId, isRanked}) {
--- a/src/stores/gameList.js
+++ b/src/stores/gameList.js
@@ -59,7 +59,7 @@ export const useGameListStore = defineStore('gameList', {
     otherLeftGame(gameId) {
       const updatedGame = this.openGames.find((game) => game.id === gameId);
       if (updatedGame) {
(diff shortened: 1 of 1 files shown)


src/stores/gameList.js
```diff
@@ -52,14 +52,16 @@ export const useGameListStore = defineStore('gameList', {
     joinGame(data) {
       const updatedGame = this.openGames.find((game) => game.id === data.gameId);
       if (updatedGame) {
-        updatedGame.numPlayers++;
+        // updatedGame.numPlayers++;
+        updatedGame.numPlayers = Math.min(2, updatedGame.numPlayers + 1)
```
itsalaidbacklife, line 56: Let's remove the old line and add a ; at the end of this line
```suggestion
        updatedGame.numPlayers = Math.min(2, updatedGame.numPlayers + 1);
```

src/stores/gameList.js
```diff
@@ -52,14 +52,16 @@ export const useGameListStore = defineStore('gameList', {
...
         updatedGame.status = data.newStatus;
       }
     },
     otherLeftGame(gameId) {
       const updatedGame = this.openGames.find((game) => game.id === gameId);
       if (updatedGame) {
-        updatedGame.numPlayers--;
+        // updatedGame.numPlayers--;
+        updatedGame.numPlayers = Math.min(2, updatedGame.numPlayers - 1)
```
itsalaidbacklife, line 64: Right idea but the problem for leaving games is that the count is sometimes negative. So instead of Math.min with 2, we should use Math.max with 0
```suggestion
        updatedGame.numPlayers = Math.max(0, updatedGame.numPlayers - 1);
```

itsalaidbacklife (changes requested):
Thanks for your contribution! You have the right idea, but a couple changes are needed to merge. Please reply in the comments if you have any questions

pradnyabhukan:
I have made the changes. Thank you for the suggestions! :)

itsalaidbacklife (approved):
Nice work. Thank you for your contribution! Feel free to pick up any other issues that catch your interest.

PR #754, opened 2023-11-09
fix: replace Math.min() with Math.max() to prevent player count from …

### LINKED ISSUE #745: [Bug]: Player counts on games in the home view can be negative or greater than two

itsalaidbacklife:
…becoming negative

## Issue number

- Resolves #745 (again)

## Please check the following

- For New Features:

### DIFF
--- a/src/stores/gameList.js
+++ b/src/stores/gameList.js
@@ -59,7 +59,7 @@ export const useGameListStore = defineStore('gameList', {
     otherLeftGame(gameId) {
       const updatedGame = this.openGames.find((game) => game.id === gameId);
       if (updatedGame) {
-        updatedGame.numPlayers = Math.min(0, updatedGame.numPlayers - 1);
+        updatedGame.numPlayers = Math.max(0, updatedGame.numPlayers - 1);
       }
     },
     setIsRanked({gameId, isRanked}) {


src/stores/gameList.js
```diff
@@ -59,7 +59,7 @@ export const useGameListStore = defineStore('gameList', {
     otherLeftGame(gameId) {
       const updatedGame = this.openGames.find((game) => game.id === gameId);
       if (updatedGame) {
-        updatedGame.numPlayers = Math.min(0, updatedGame.numPlayers - 1);
+        updatedGame.numPlayers = Math.max(0, updatedGame.numPlayers - 1);
```
seriouslysean, line 62: Is this the best way to do that? What’s the range this can feasibly be? The typical game will be 0 with no one, 1 (lobby) or 2 (lobby/game) which would be:

- min: -1, 0, 1?
- max: 0, 1, 2?

Does it make more sense to count this a different way and just prevent it from going below 0?

Alternatively if you want to be extra safe pull it out to a utility and test it via vitest.

seriouslysean (approved):
Solution is fine but I have some questions.

src/stores/gameList.js
itsalaidbacklife: @seriouslysean yes this is a kludge or rather a fix to a previous kludge.
1. The player count is expected to be exactly 0, 1, or 2
2. I agree that it makes sense to store it a different way. Specifically instead of a number, we should store the users of the objects in the games so that we could later display their usernames if we wanted to, and so that the count is just the length of the server-provided array of players, which should reduce the chance for desync due to connection issues

I think it's overkill to pull this line into a utility b/c it's an intermediary solution anyway. I've written #755 to capture a better solution

PR #756, opened 2023-11-10
put info button to the left of switch and fixed alignment

### LINKED ISSUE #517: [Feature]: Left-align info icon in the CreateGameDialog
### Feature Summary

The CreateGameDialog has an info icon next to the 'Ranked' switch that explains ranked scoring, but the icon is right-justified which keeps it too far away from the ranked switch that it explains. We should left-align this icon so that it's clear that it is info about ranked.

### Detailed Description

Here is the CreateGameDialog (Seen from the HomeView)
image

We should move this info icon (which is the activator for the `StatsScoringDialog`) over to the left so that it appears next to the ranked switch that it explains

AlexVCS:
## Issue number
- Resolves #517 

## Please check the following

- [✅] Do the tests still pass? (see [Run the Tests](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#run-the-tests))
- [✅] Is the code formatted properly? (see [Linting (Formatting)](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#linting-formatting))


Haviles04:
Nice solution 👍 glad you figured it out, that switch is setup really weird

src/routes/home/components/CreateGameDialog.vue
```diff
@@ -30,13 +30,17 @@
...
+        <StatsScoringDialog
+          activator-color="surface-2"
+          :show-button-text="false"
+        />
         <v-switch
           v-model="isRanked"
           :label="isRanked ? t('global.ranked') : t('global.casual')"
           data-cy="create-game-ranked-switch"
           color="surface-2"
+          class="d-flex align-center"
```
itsalaidbacklife, line 42: align-center is for the flexbox container, not the child element, so this shouldn't be needed. The problem with the vertical alignment before was the bottom margin applied by the `class="mb-2"` on the other element. Removing that as you have should fix the vertical alignment without this line here

```suggestion
```

itsalaidbacklife (approved):
Nice fix, thank you

PR #763, opened 2023-11-12
closes #755

### LINKED ISSUE #755: [DevEx]: Track player counts in the HomeView/gameList using an array of the player objects, rather than an integer
### Improvement Summary

The homepage displays how many players are in the games available for you to join e.g. 1/2. This is stored as a `playerCount` number on the games in the `gameList` store on each game, but it would be more robust for the player count to be the actual length of the players array.

### Detailed Description

We should update the `GameSummary` class in the `gameList` store so that it directly consumes the `players` array on each game instead of transforming it into the `playerCount` integer, then update `HomeView` so that the `numPlayers` prop it sends to its instance of `GameListItem` is set to be the length of the `players` array. This will allow us to display the specific usernames of the players in games and also to keep the games displayed in the homeview in better sync with the server state (preventing bugs like #745).

```
// Current state of src/stores/gameList.js
class GameSummary {
  constructor(obj) {
    this.numPlayers = Object.prototype.hasOwnProperty.call(obj, 'players') ? obj.players.length : 0;    // replace this line
   // instead we should have game.players = object.players ?? [];
    this.status = Object.prototype.hasOwnProperty.call(obj, 'status') ? obj.status : GameStatus.ARCHIVED;
    this.isRanked = Object.prototype.hasOwnProperty.call(obj, 'isRanked') ? obj.isRanked : false;
    this.isOver = false;
  }
}
```

```
                 <!-- HomeView.vue -->
                  <GameListItem
                    :name="game.name"
                    :p0ready="game.p0Ready ? 1 : 0"
                    :p1ready="game.p1Ready ? 1 : 0"
                    :game-id="game.id"
                    :status="game.status"
                    :num-players="game.numPlayers" <!-- change this -->
                    :num-players="game.players.length" <!-- to this -->
                    :is-ranked="game.isRanked"
                    @error="handleSubscribeError(game.id, $event)"
                  />
```

We should then update the `gameList` store's `otherLeftGame()` and `joinGame()` actions to directly use player objects instead of incrementing a decrementing counts.

For security, we should update `api/hooks/customGameHook.index.js` so that it filters the player objects sent from the backend to only include `username` and `pNum` like so:

```
            // api/hooks/customGameHook.index.js 
       findOpenGames() {
            // after querying for the games
            const openGames = games.filter(({ players }) => players.length < 2).map(({username, pNum}) => {
              return { username, pNum };
            });
        }
```

We should do a similar move in `api/helpers/find-spectatable-games.js` so that it queries for the players themselves then maps them to just username and pNum:
```
      const games = await Game.find({
        status:  gameService.GameStatus.STARTED,
        updatedAt: { '>=': recentUpdateThreshhold },
      })
      .populate('players') // add this to get the players
       // and this to remove all the attributes except username and pNum
      .map(({username, pNum} => {
        return { username, pNum };
      });
```

pradnyabhukan:
## Issue number

- Resolves #

## Please check the following

- For New Features:

### DIFF
--- a/api/helpers/ find-spectatable-games.js
+++ b/api/helpers/ find-spectatable-games.js
@@ -12,7 +12,11 @@ module.exports = {
       const games = await Game.find({
         status:  gameService.GameStatus.STARTED,
         updatedAt: { '>=': recentUpdateThreshhold },
-      });
+      })
+          .populate('players')
+          .map(({username, pNum}) =>{
+            return { username, pNum};
+          });
       return exits.success(games);
     } catch (err) {
       return exits.error(err);
--- a/src/stores/gameList.js
+++ b/src/stores/gameList.js
@@ -7,7 +7,7 @@ class GameSummary {
   constructor(obj) {
     this.id = obj.id ? obj.id : null;
     this.name = obj.name ? obj.name : null;
-    this.numPlayers = Object.prototype.hasOwnProperty.call(obj, 'players') ? obj.players.length : 0;
+    this.players = obj.players ?? [];
     this.status = Object.prototype.hasOwnProperty.call(obj, 'status') ? obj.status : GameStatus.ARCHIVED;
     this.isRanked = Object.prototype.hasOwnProperty.call(obj, 'isRanked') ? obj.isRanked : false;
     this.isOver = false;
@@ -52,14 +52,14 @@ export const useGameListStore = defineStore('gameList', {
     joinGame(data) {
       const updatedGame = this.openGames.find((game) => game.id === data.gameId);
       if (updatedGame) {
-        updatedGame.numPlayers = Math.min(2, updatedGame.numPlayers + 1);
+        updatedGame.players.push(data.newPlayer);
         updatedGame.status = data.newStatus;
       }
     },
     otherLeftGame(gameId) {
       const updatedGame = this.openGames.find((game) => game.id === gameId);
-      if (updatedGame) {
-        updatedGame.numPlayers = Math.max(0, updatedGame.numPlayers - 1);
+      if (updatedGame && updatedGame.players.length > 0) {
+        updatedGame.players.pop();
       }
     },
     setIsRanked({gameId, isRanked}) {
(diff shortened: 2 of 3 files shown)


api/helpers/ find-spectatable-games.js
```diff
@@ -12,7 +12,11 @@ module.exports = {
       const games = await Game.find({
         status:  gameService.GameStatus.STARTED,
         updatedAt: { '>=': recentUpdateThreshhold },
-      });
+      })
+          .populate('players')
+          .map(({username, pNum}) =>{
+            return { username, pNum};
```
Haviles04, line 18: I dont think we need the return keyword here, I believe you can just `.map(({username, pNum}) => ({username, pNum}))`

src/stores/gameList.js
```diff
@@ -52,14 +52,14 @@ export const useGameListStore = defineStore('gameList', {
...
+        updatedGame.players.push(data.newPlayer);
         updatedGame.status = data.newStatus;
       }
     },
     otherLeftGame(gameId) {
       const updatedGame = this.openGames.find((game) => game.id === gameId);
-      if (updatedGame) {
-        updatedGame.numPlayers = Math.max(0, updatedGame.numPlayers - 1);
+      if (updatedGame && updatedGame.players.length > 0) {
+        updatedGame.players.pop();
```
Haviles04, line 62: I'm not 100% on how this whole array works, but will using `.pop()` be an issue it was coming from the second person to join? should we filter out based on username?

api/helpers/ find-spectatable-games.js
pradnyabhukan: Yes. I added that because I needed to put some consoles. I am going to remove it

src/stores/gameList.js
pradnyabhukan: Yes, I am trying to figure how to remove the specific player. Maybe using pNum

Haviles04: You could probably filter out by using the userName on the auth store

itsalaidbacklife:
Closing as inactive

PR #773, opened 2023-11-13
Issue/#584

### LINKED ISSUE #584: [Feature]: Enable translating snackbar errors in LoginView

Posoroko:
## #584

- Resolves #584

## Please check the following

- For New Features:

Closed the problematic pr and pushed new changes without the problems

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -19,7 +19,7 @@ module.exports = {
       const foundUser = await User.findOne({ username: username });
       if (foundUser) {
         throw {
-          message: 'That username is already registered to another user; try logging in!',
+          message: 'login.snackbar.usernameIsTaken',
         };
       }
       // Encrypt pw and create new user
@@ -40,7 +40,7 @@ module.exports = {
       const user = await User.findOne({ username: username });
       if (!user) {
         throw {
-          message: 'Could not find that user with that username. Try signing up!',
+          message: 'login.snackbar.userNotFound',
         };
       }
       await passwordAPI.checkPass(password, user.encryptedPassword);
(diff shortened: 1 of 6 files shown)


itsalaidbacklife (approved):
Excellent

PR #775, opened 2023-11-14
flex and align-center classes added to the v-witch

### LINKED ISSUE #772: [Bug]: Create Game Dialog ranked switch + button vertical alignment is off-center
### Bug Summary

The switch in the `CreateGameDialog` is vertically top-aligned instead of centered

### Detailed Description

See pic
image

We should adjust the container with the button to vertically center the contents. This should have been achieved already using `class="d-flex align-center"` on the parent element of the `v-switch`, so we'll need to investigate and debug

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Steps to Reproduce

1.
2.
3.

Posoroko:
## #772

- Resolves #772

## Please check the following

- For New Features:

I added
```
class="d-flex align-center"
```
on the v-switch.  The same classes were already applied to the form parent element, but the v-switch has a container div that needs to be vertically align as well.
There might be a more 'vuetify-ish' way of handleing this ?

the v-swith component has a center-affix prop:
image

It's true by default, so I thaought it would be centering its content, but maybe I misunderstand this...

### DIFF
--- a/src/routes/home/components/CreateGameDialog.vue
+++ b/src/routes/home/components/CreateGameDialog.vue
@@ -36,6 +36,7 @@
         />
         <v-switch
           v-model="isRanked"
+          class="d-flex align-center"
           :label="isRanked ? t('global.ranked') : t('global.casual')"
           data-cy="create-game-ranked-switch"
           color="surface-2"


src/routes/home/components/CreateGameDialog.vue
```diff
@@ -36,6 +36,7 @@
         />
         <v-switch
           v-model="isRanked"
+          class="d-flex align-center"
```
itsalaidbacklife, line 39: Does this actually fix the issue? Can I get a before/after screenshot?

Posoroko: yep :
with flex
without flex

Posoroko: yep, it seems to be fixing it :
with flex
without flex

itsalaidbacklife (approved):

PR #777, opened 2023-11-14
Bug #753 May counter dialog

### LINKED ISSUE #753: [Bug]: Drawing extra card during resolve/counter phase
### Bug Summary

Being able to draw a card while the opponent decides whether to counter or resolve your one-off

### Detailed Description

After you've played a one-off, when the opponent is deciding if he's going to counter it, the game lets you draw a card even though you shouldn't be able to.
It works with all one-offs. When it's a 4 it also works while the opponent is choosing his discards, and a very similar thing happens with the 3 when he's choosing the card he's going to recover.

Quick example:

Playing a 6 as a one-off

imagen

Clicking anywhere in the highlighted area makes the "Opponent may counter" label go away, as if it was your turn again

imagen

imagen

And the game lets you draw

imagen

All of this while the opponent hasn't decided yet

imagen

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Steps to Reproduce

1.
2.
3.

Haviles04:
## Issue number

- Resolves #753 

## Please check the following

- For New Features:

- Added persistent prop so clicking outside dialog will not close it
- Added error handling inside `draw.js` api endpoint

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -19,7 +19,7 @@ module.exports = {
       const foundUser = await User.findOne({ username: username });
       if (foundUser) {
         throw {
-          message: 'That username is already registered to another user; try logging in!',
+          message: 'login.snackbar.usernameIsTaken',
         };
       }
       // Encrypt pw and create new user
@@ -40,7 +40,7 @@ module.exports = {
       const user = await User.findOne({ username: username });
       if (!user) {
         throw {
-          message: 'Could not find that user with that username. Try signing up!',
+          message: 'login.snackbar.userNotFound',
         };
       }
       await passwordAPI.checkPass(password, user.encryptedPassword);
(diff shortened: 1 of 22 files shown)


itsalaidbacklife (commented):
This looks great. @Haviles04 could you add a test case where we set up this situation and then hook into the `gameStore`  using the window in order to directly request to draw a card as the player while ignoring the overlay in order to assert that the snackbar appears?

Haviles04:
> This looks great. @Haviles04 could you add a test case where we set up this situation and then hook into the `gameStore`  using the window in order to directly request to draw a card as the player while ignoring the overlay in order to assert that the snackbar appears?

Absolutely can, should I add it to the current test that checks for the opponent may counter overlay, or should I make a new test

itsalaidbacklife:
> > This looks great. @Haviles04 could you add a test case where we set up this situation and then hook into the `gameStore`  using the window in order to directly request to draw a card as the player while ignoring the overlay in order to assert that the snackbar appears?
> 
> Absolutely can, should I add it to the current test that checks for the opponent may counter overlay, or should I make a new test

Let's add it to the existing test since it's already about the same situation

itsalaidbacklife (approved):
Great work

PR #783, opened 2023-11-17
Fix/application locking

### LINKED ISSUE #750: [Bug]: Players joining games or readying/unreadying simultaneously causes unexpected behavior
### Bug Summary

If both players hit the ready button at the same times, the game sometimes fails to start or behaves strangely. Same goes for joining a game at the same time.

### Detailed Description

Some requests e.g. joining games or hitting ready/unready are interdepenent e.g. when the 2nd player hits ready, the game should start. So when two requests to ready happen at the same time, the server sometimes fails to recognize that the game should start because the other player isn't ready yet (even though the other player's request is midflight). 

We should address this using [application locking](https://stackoverflow.com/questions/5436021/implementing-application-level-locking) where we can essentially create a lock on a game that other requests can await before proceeding. This will force interrelated requests to be processed in sequence instead of in parallel, allowing for consistent behavior. The request to join a game or to ready up should lock the game while the request is processed.

The basic approach is to define two helper methods: 
```
lockGame(gameId): Promise<string>{}
unlockGame(lockId: string): Promise<void>{}
```
Where lockGame will await the current lock on the specified game if there is one, then apply a new lock and resolve with a uuid representing the new lock which can be used later to unlock the game.

Here is a snippet outlining the approach
```
const crypto = require('crypto');
async lockGame(timeoutSeconds = 60, priorAttemps = 0) {
  const lock = crypto.randomUUID();
  // update record set lock = lock, locked_at = NOW() where lock is null OR where locked_at <= now - (timeoutSeconds * '1 SECOND'::INTERVAL);
  const updatedRecord = query for new record just updated;
  if (updatedRecord.lock === lock) {
    this.lock = updatedRecord.lock;
    this.lockedAt = updatedRecord.lockedAt;
    return lock
  }

const waitTime = Math.min( 2 ** priorAttempts * LOCK_WAIT_TIME_MS, MAX_LOCK_WAIT_TIME_MS);
await sleep(Math.random() * 1000 + waitTime);
return this.obtainLock(timeoutSeconds, priorAttempts + 1);
}

async unlockGame(lock) {
  const updatedRecord = // update record set lock = null, locked_at = null where lock = lock;
  if (updatedRecord) { // if above updated
    this.lock = undefined;
    this.lockedAt = undefined;
  }
}
```

The key here is that we will add `lock` and `lockedAt` columns to the `game` table/model and that locking a game will only write if the specified game's `lock` is currently null or if the game's `lockedAt` timestamamp is sufficiently old. Then when locking we read the updated game record's lock back to confirm it is the uuid we just attempted to write. If not, then we retry until we succeed or timeout. And if we succeed then we return the uuid of the new lock which can later be used to `unlockGame()`.

### Current Behavior

Simultaneous interdependent requests e.g. readying up or joining a game can bypass or fail checks on the game that expect updates to be processed in sequence e.g. a game doesn't start when both players hit ready at the same time.

### Expected Behavior

Interdependent requests are processed atomically in sequence to prevent race conditions. Hitting ready at the same time will still start the game.

### Steps to Reproduce

1. Signup, create a game, join it as two players
2. Both players hit ready simultaneously
3. The game may not start or may start incorrectly

itsalaidbacklife:
Applies Application Locking strategy to force interdependent requests to await each other. Specifically applied to the 'Ready' action to fix a bug where two players hitting ready at the same time causes the game to fail to start.

## Issue number

- Resolves #750 

## Please check the following

- For New Features:


api/helpers/lock-game.js
```diff
@@ -0,0 +1,63 @@
...
+      descritiption: 'The database id of game to lock',
+      example: 44,
+      required: true,
+    },
+  },
+
+  fn: async ({ gameId }, exits) => {
+    const LOCK_RETRY_TIME_MS = 250;
+    const LOCK_MAX_WAIT_TIME_MS = 5000;
+    const MAX_ATTEMPTS = 20;
```
itsalaidbacklife, line 28: It's kindof redundant to have both a max attempts count and a max wait time when we have a fixed interval retry time. As it stands, we'll retry every 250ms for up to 5s or 20 tries. Theoretically those are equivalent, but each attempt will take nonzero time so in reality only the 5s timeout would apply. 

So we could remove either the max wait time or the attempt count to simplify this because they can't both be used at once meaningfully. 

Alternatively, we could adjust the wait time to be non-fixed, perhaps exponentially increasing up to a maximum. That's what ButterBothSides (who recommended this overall approach) does at his work. This decreases the load on the database and increases the chance for successfully obtaining a lock when multiple simultaneous requests are processed. Having both max attempts and max wait time would allow for fine tuning of these params as we see fit.

In practice though for a given game, we're likely looking at a maximum of 3 simultaneous requests at once, so I suspect that exponential decay is overkill. That means we're likely best off using either max wait time, or max attempts. In that I'd say that max attempts is easier to read and understand b/c it's an incremented counter, and max wait time would arguably have the UX benefit of having a hard cap on the user wait time before their request is fully processed. @seriouslysean penny for your thoughts.

My inkling is to go for max attempts, in which case we could ditch the while loop for a for loop which will probably be easier to read

api/helpers/lock-game.js
```diff
@@ -0,0 +1,63 @@
...
+        const updatedGame = Game.updateOne({
+          id: gameId,
+          or: [
+            {lock: null},
+            {lockedAt: {'<=': lockIsStaleTimeout}}
+          ],
+        }).set({lock: uuId, lockedAt: now});
+
+        // If we successfully wrote our uuid, resolve
+        const newLock = updatedGame?.lock;
```
itsalaidbacklife, line 50: It's notable that sails' `.updateOne()` method automatically queries for the updated record and returns undefined if it's unavailable (as opposed to erroring). That's why we need to optionally chain `updatedGame?.lock`. I considered putting a comment there to explain that behavior, but it seemed like it would clutter the flow of comments about the logic of the helper itself which is why I omitted it. I'm hoping that the update call followed by optionally chaining is relatively clear by itself.

api/helpers/lock-game.js
```diff
@@ -0,0 +1,63 @@
+const { randomUUID } = require('crypto'); // Added in: node v14.17.0
+const dayjs = require('dayjs');
+
+async function sleep(durationInMillis) {
+  return new Promise((resolve) => {
+    setTimeout(resolve, durationInMillis);
+  });
+}
```
itsalaidbacklife, line 8: A suggestion I got from ButterBothSides was to pull `sleep()` into its own utility function. I'm open to doing so, but my initial instinct is to wait for a second usecase before relocating this so that until then, `sleep()` is colocated with its only caller. Curious for your thoughts @seriouslysean

api/helpers/lock-game.js
```diff
@@ -0,0 +1,63 @@
...
+        // If we successfully wrote our uuid, resolve
+        const newLock = updatedGame?.lock;
+        if (newLock === uuId) {
+          return exits.success(uuId);
+        }
+
+        // Otherwise wait and try agin
+        await sleep(LOCK_RETRY_TIME_MS);
+      } catch (err) {
+        return exits.error(err);
```
itsalaidbacklife, line 58: This error should only happen if we have a database error or something unexpected when writing to the db.

api/helpers/unlock-game.js
```diff
@@ -0,0 +1,20 @@
...
+      type: 'string',
+      descritiption: 'The lock UUID to unset/release ',
+      example: '921a780a-514e-4086-8c9c-fccc2a8d26d8',
+      required: true,
+    },
+  },
+
+  fn: async ({ lock }, exits) => {
+    await Game.updateOne({ lock }).set({lock: null, lockedAt: null});
+    return exits.success();
```
itsalaidbacklife, line 18: Interestingly, I don't think we care about whether the unlock write operation successfully writes to the db because if someone else has locked the game, we just let them have it until the lock times out. That's why this returns void. That might be worth a comment

api/models/Game.js
```diff
@@ -119,5 +119,14 @@ module.exports = {
...
       model: 'user',
     },
+    lock: {
+      type: 'string',
+      allowNull: true,
+    },
+    // Time game was last locked in millis since epoch
+    lockedAt: {
+      type: 'number',
+      columnType: 'integer',
```
itsalaidbacklife, line 129: It has been a mistake to fail to set the `columnType` of our other numbers because the default to `float` at the physical db layer even though most of our number columns are database ids. I'm hopeful that this practice will clean up new columns going forward. It's worth writing an issue to clean that up for our other columns, app-wide and to write a db migration to make it so.

api/controllers/game/ready.js
```diff
@@ -5,10 +5,11 @@ module.exports = function (req, res) {
   if (req.session.game && req.session.usr) {
     const promiseGame = gameAPI.findGame(req.session.game);
     const promiseUser = userAPI.findUser(req.session.usr);
-    Promise.all([promiseGame, promiseUser])
+    const promiseLock = sails.helpers.lockGame(req.session.game);
+    Promise.all([promiseGame, promiseUser, promiseLock])
```
itsalaidbacklife, line 9: This is where the rubber meets the road for actually using the lock. It will be much cleaner once #448 is merged because the control flow is simpler and we can use async/await here, but I like how simple and minimally invasive it is to apply a lock/unlock pattern in an existing endpoint

api/controllers/game/ready.js
```diff
@@ -101,7 +103,9 @@ module.exports = function (req, res) {
           pNum: user.pNum,
           gameId: game.id,
         });
-        return Game.updateOne({ id: game.id }).set(gameUpdates);
+        const updateGamePromise = Game.updateOne({ id: game.id }).set(gameUpdates);
+        const unlockGamePromise = sails.helpers.unlockGame(lock);
+        return Promise.all([updateGamePromise, unlockGamePromise]);
```
itsalaidbacklife, line 108: It occurred to me that it might make sense to eschew the `unlockGame()` call here and instead just add to the `gameUpdates` to set the lock to null. The only reason not to is so as not to overwrite someone elses' lock if ours has timed out since generally you only want a request to release its own lock (not someone else's). But it's a weird use case, so I'm not positive what's best.

Semantically, especially once we have async/await for this action, I like the simplicity of
```
const lock = await sails.helpers.lockGame(gameId);
// do stuff

await sails.helpers.unlockGame(lock);
```

though this does add an extra query

api/helpers/lock-game.js
seriouslysean: Typically plan for the now, I wouldn’t move it unless you need to use it elsewhere.

api/helpers/lock-game.js
```diff
@@ -0,0 +1,63 @@
+const { randomUUID } = require('crypto'); // Added in: node v14.17.0
+const dayjs = require('dayjs');
+
+async function sleep(durationInMillis) {
```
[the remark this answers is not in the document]
itsalaidbacklife: As an idiot, I love me some explicit clarity

api/helpers/lock-game.js
```diff
@@ -0,0 +1,63 @@
...
+        now = dayjs().valueOf();
+        const lockIsStaleTimeout = dayjs().subtract(30, 'second').valueOf();
+
+        // Lock & re-fetch game if unlocked or lock is expired
+        const updatedGame = Game.updateOne({
+          id: gameId,
+          or: [
+            {lock: null},
+            {lockedAt: {'<=': lockIsStaleTimeout}}
+          ],
```
itsalaidbacklife, line 46: This `or` predicate on the update's where clause is the secret sauce that forces the locking mechanism to be atomic and to prevent race conditions. We only write the lock for one specified game if its lock is already null 
(meaning its unlocked) or if the lock is expired based on the expiration interval.

Then we read the latest value of the lock back and if it matches our uuid, we're good an can resolve. Otherwise we sleep and retry next loop

api/controllers/game/ready.js
seriouslysean: I think it’s best to do the unlock and close the loop. We can investigate not needing one but it feels cleaner to actually remove it instead of unsetting it.

api/helpers/unlock-game.js
seriouslysean: Does it matter with a distinction between when it times out vs. when the operation succeeds from the user who locked it? It sounds like:

- a user who locks it can unlocks it
- a user who did not lock it can not unlock it
- It can timeout and unlock on its own (is this part of the check of a new request or does a separate job run to unlock?)

seriouslysean (approved):
I’m not as close to it but I do have a couple questions.

api/controllers/game/ready.js
itsalaidbacklife: Interesting question. I think really the lock should come _before_ getting the game and user so that whatever game data you query for is up to date by the time you get to business logic. That said, I think it likely makes sense to revise `lock-game` so that it returns the game itself, which will contain the lock uuid.. Then we don't to wait for that before querying for the game because we'll get the game directly from the lock. It'll save us a query.

Then we'll wait for the game/lock before querying for the user and doing the guts of it

api/helpers/unlock-game.js
itsalaidbacklife: Correct. The timeout is only folded into the check for a new attempt to lock; there is no separate job to clear it later. 
When you try to unlock a game, either:
1. You succeed and the game is now unlocked
2. You fail because your lock expired and someone else locked the game before you unlocked
3. An unexpected error e.g. server loses connection to the database causes a failure.

3) should be an error and already should cause the method to error when the `Game.updateOne()` call errors. I don't _think_ the caller of `unlock-game` needs to care about the difference between 1) and 2) because either way, the request that has called `unlock-game` is no longer locking that game and now it's up to whatever other requests interact with the game later on to sort out who's locking the game (if anyone). So I think it's still reasonable for this method to resolve void as long as it doesn't hit an adapter error, whether the unlock write operation is successful or not

api/controllers/game/ready.js
itsalaidbacklife: @seriouslysean I've updated the `lock-game` helper to return the game object itself since the method already requires querying for it and this way consumers (like ready.js) who were going to lock the game and immediately query for it before doing other stuff (the most likely use case) can just use the lock game helper to get the game they need.

Further, I've refactored `ready.js` to use async/await so that it's easy to simply `await` the `lockGame()` call as the first line and then ensure that the rest of the data is only queried and considered valid after the lock is obtained. This should prevent weird cases around other data being queried before the lock is obtained then updated while this request is midflight

api/controllers/game/ready.js
```diff
@@ -1,55 +1,52 @@
-const gameAPI = sails.hooks['customgamehook'];
 const userAPI = sails.hooks['customuserhook'];
 
-module.exports = function (req, res) {
-  if (req.session.game && req.session.usr) {
```
itsalaidbacklife, line 5: I've removed this check from the function body to replace it with the `isInGame` and `isLoggedIn` policies, which we generally do for other endpoints and which lets us have this endpoint focus on the business logic

api/controllers/game/ready.js
```diff
@@ -1,55 +1,52 @@
...
-            break;
-          case 1:
-            gameUpdates.p1Ready = !game.p1Ready;
-            if (game.p0Ready) {
-              bothReady = true;
-            }
-            break;
+module.exports = async function (req, res) {
+  try {
+    const game =  await sails.helpers.lockGame(req.session.game);
```
itsalaidbacklife, line 5: I refactored this to use `async/await` largely so that this line is easy to use and work with without needing an extra layer of `.then()` to pass the game through the promise chain. Now consumers can lock a game and query for it by simply `awaiting` the `lockGame()` helper

config/policies.js
```diff
@@ -37,7 +37,7 @@ module.exports.policies = {
   'game/get-list': 'isLoggedIn',
   'game/subscribe': ['isLoggedIn', 'hasGameId'],
   'game/spectate': ['isLoggedIn', 'hasGameId'],
-  'game/ready': 'isLoggedIn',
+  'game/ready': ['isLoggedIn', 'isInGame'],
```
itsalaidbacklife, line 40: This checks that `req.session` has values for `usr` and `game`, which removes the need to check that in the ready action itself

api/controllers/game/ready.js
```diff
@@ -1,55 +1,54 @@
...
+      gameUpdates.status = gameService.GameStatus.STARTED;
+      // Deal cards
+      await gameService.dealCards(game, gameUpdates);
+    } else {
+      // Send socket message
+      Game.publish([game.id], {
+        change: 'ready',
+        userId: user.id,
+        pNum: user.pNum,
+        gameId: game.id,
```
itsalaidbacklife: I added this else block because sending the socket message for 'ready' after we initialize the deck and deal the cards was causing a bug where the client would update to say that one of the players wasn't ready and therefore triggering an overlay that said "Waiting for game to start" even when the game was clearly initialized.

A bit of digging revealed that generally our setup for when to show an overlay that the game isn't ready is pretty spotty and disorganized. We have one such overlay in the `GameOverlays`, one in `RematchView`, and one in the `GameUnavailableView`, all of which have their own conditional logic. We should write up an issue to reorganize those so they are easier to troubleshoot and reason about.

For now though this prevents the bug

itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/806

itsalaidbacklife:
Migration to add `lock` and `lockedAt` columns
```
alter table "game"
add column "lock" text,
add column "lockedAt" bigint;
```

PR #809, opened 2023-11-25
Bug #794 Flaky spectate test

### LINKED ISSUE #794: [Bug]: `it('spectates a game')` test fails flakily sometimes
### Bug Summary

One of the tests in `tests/e2e/in-game/spectating.spec.js` fails randomly sometimes

### Detailed Description

The first test in the `spectating.spec` has been failing randomly:
```
it('Spectates a game', () => {
```

Here is an example test run: https://cloud.cypress.io/projects/i8bxr8/runs/2d121cf6-4a2c-439f-b463-215cb3d6483e/test-results/969394b4-b389-494f-a781-31ebe89f906b/replay?ts=1700497544242.2686&att=1

It appears to be the case that after the `cy.reload()` the DOM shows the `GameUnavailableView`

We should investigate why and fix the test and/or app so that it behaves correctly

### Current Behavior

The `'spectates a game'` test sometimes fails randomly

### Expected Behavior

The test should consistently pass

### Steps to Reproduce

It can't be consistently reproduced, but

Haviles04:
## Issue number

- Resolves #794 

## Please check the following

- For New Features:

Added router check to `isSpectating` computed value in `GameView.vue` to check after reload

### DIFF
--- a/src/routes/game/GameView.vue
+++ b/src/routes/game/GameView.vue
@@ -420,7 +420,7 @@ import { mapStores } from 'pinia';
 import { useI18n } from 'vue-i18n';
 import { useGameStore } from '@/stores/game';
 import { useAuthStore } from '@/stores/auth';
-import { ROUTE_NAME_HOME } from '@/router';
+import { ROUTE_NAME_HOME, ROUTE_NAME_SPECTATE } from '@/router';
 import BaseSnackbar from '@/components/BaseSnackbar.vue';
 import UsernameToolTip from '@/routes/game/components/UsernameToolTip.vue';
 import GameCard from '@/routes/game/components/GameCard.vue';
@@ -472,7 +472,7 @@ export default {
     ...mapStores(useGameStore),
     ...mapStores(useAuthStore),
     isSpectating() {
-      return this.gameStore.isSpectating;
+      return this.gameStore.isSpectating || this.$router.currentRoute.value.name === ROUTE_NAME_SPECTATE;
     },
     showOpponentHand() {
       return this.gameStore.hasGlassesEight || this.isSpectating;
--- a/src/routes/game/GameView.vue
+++ b/src/routes/game/GameView.vue
@@ -420,7 +420,6 @@ import { mapStores } from 'pinia';
 import { useI18n } from 'vue-i18n';
 import { useGameStore } from '@/stores/game';
 import { useAuthStore } from '@/stores/auth';
-import { ROUTE_NAME_HOME, ROUTE_NAME_SPECTATE } from '@/router';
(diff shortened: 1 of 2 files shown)


src/routes/game/GameView.vue
```diff
@@ -472,7 +472,7 @@ export default {
     ...mapStores(useGameStore),
     ...mapStores(useAuthStore),
     isSpectating() {
-      return this.gameStore.isSpectating;
+      return this.gameStore.isSpectating || this.$router.currentRoute.value.name === ROUTE_NAME_SPECTATE;
```
itsalaidbacklife, line 475: Do we want both the store value and the route or should we update the consumers of the store value to use the router instead?

itsalaidbacklife (commented):
Good catch. One question: should we still use the store value at all or replace it altogether with the route check?

Haviles04:
> Good catch. One question: should we still use the store value at all or replace it altogether with the route check?

We still need the store value inside the store, this way updateGame knows we are spectating.

Without it, when you join a game from the lobby it doesn't sort the top cards because the name is home


Inside the `GameView.vue` we can just use the route

src/routes/game/GameView.vue
seriouslysean: If the store value can be derived from the route it might make sense to keep both so you have one setter and all consumers use the store since that’s typically easier to check in the code.

seriouslysean:
@Haviles04 @itsalaidbacklife i don’t think a getter has access to route, but you can add some logic when entering or exiting the route to dispatch an action. Might take some tinkering, there’s some interesting cases between resetting before each, etc.

Give it a crack and see where you land. I’m busy most of today, but can take a look tomorrow morning.

src/stores/auth.js
```diff
@@ -110,6 +111,13 @@ export const useAuthStore = defineStore('auth', {
         //     - `gameService.populateGame` is called
         //     - `Game.subscribe` is called
         //     - `Game.publish` is called
+
+        const gameStore = useGameStore();
+        if (!gameId && isSpectating) {
+          let { gameId } = route.params;
```
itsalaidbacklife, line 117: should be const if we're not modifying this
```suggestion
          const { gameId } = route.params;
```

src/stores/auth.js
```diff
@@ -110,6 +111,13 @@ export const useAuthStore = defineStore('auth', {
...
+          gameStore.requestSpectate(Number(gameId));
+        }
+
         if (gameId && (isGame || isLobby)) {
           const gameStore = useGameStore();
```
itsalaidbacklife, line 122: Let's remove this since we're now declaring gameStore above
```suggestion
```

PR #818, opened 2023-11-29
Feature: Automatic release based on version bump

seriouslysean:
## Issue number

- Resolves #

## Please check the following

- For New Features:

See https://github.com/seriouslysean/demo--auto-bump-forked-prs/pull/24/files for an explanation.

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -1,83 +1,106 @@
+name: Bump Version
+
 on:
   pull_request:
-    branches:
-      - main
     types: [closed]
+    branches: [main]
 
 jobs:
-  bump-version:
+  check-token:
     runs-on: ubuntu-latest
-
-    env:
-      GH_PUSH_TOKEN: ${{ secrets.GH_PUSH_TOKEN }}
-
+    outputs:
+      GH_PUSH_TOKEN: ${{ steps.check.outputs.GH_PUSH_TOKEN }}
     steps:
+      - name: Check GH_PUSH_TOKEN
+        id: check
+        run: |
+          if [[ -z "${{ secrets.GH_PUSH_TOKEN }}" ]]; then
+            echo "GH_PUSH_TOKEN is not set. Exiting."
+            exit 1
+          fi
+          echo "::set-output name=GH_PUSH_TOKEN::${{ secrets.GH_PUSH_TOKEN }}"
+        shell: bash
 
-      - name: Checkout
-        uses: actions/checkout@v4
-        if: env.GH_PUSH_TOKEN != ''
+  check-merged:
+    needs: check-token
+    runs-on: ubuntu-latest
+    outputs:
+      merged: ${{ steps.check.outputs.merged }}
+    steps:
+      - name: Check if PR is merged
+        id: check
+        run: |
+          if [[ "${{ github.event.pull_request.merged }}" != "true" ]]; then
+            echo "Pull request is not merged. Exiting."
+            exit 1
+          fi
+          echo "::set-output name=merged::${{ github.event.pull_request.merged }}"
+        shell: bash
+
+  get-labels:
+    needs: check-merged
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.get.outputs.version }}
+    steps:
+      - name: Get labels
+        id: get
+        run: |
+          LABELS=$(jq -r '.pull_request.labels[].name' <<< '${{ toJson(github.event) }}')
+          VERSION_LABELS=$(echo "$LABELS" | grep -E '^version-(patch|minor|major)$' || true)
+          if [[ $(echo "$VERSION_LABELS" | wc -l) -ne 1 ]]; then
+            echo "Exactly one version label is required. Exiting."
+            exit 1
+          fi
+          echo "::set-output name=version::$(echo "$VERSION_LABELS" | sed 's/^version-//')"
+        shell: bash
+
+  bump-and-push-version:
+    needs: [check-token, get-labels]
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.bump.outputs.version }}
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
         with:
-          token: ${{ env.GH_PUSH_TOKEN }}
+          node-version: 18
 
-      - name: configure git
-        if: env.GH_PUSH_TOKEN != ''
-        run: |
-          git remote set-url origin https://itsalaidbacklife@github.com/cuttle-cards/cuttle
-          git config --global user.name "Cuttle Continuous Integration"
-          git config --global user.email "itsalaidbacklife@gmail.com"
+      - name: Install Dependencies
+        run: npm ci
 
-      # Bump version
-      - name: bump patch version
-        id: bump_patch
+      - name: Bump Version
+        id: bump
         run: |
-          if contains(github.event.pull_request.labels.*.name, 'version-patch') && env.GH_PUSH_TOKEN != ''
-          then
-            npm run version:patch
-            echo "::set-output name=bumped::true"
-          fi
+          npm version ${{ needs.get-labels.outputs.version }} --no-git-tag-version
+          VERSION=$(node -p "require('./package.json').version")
+          git config --global user.email "actions@github.com"
+          git config --global user.name "GitHub Actions"
+          git commit -am "Bump version to $VERSION"
+          echo "::set-output name=version::$VERSION"
         shell: bash
 
-      - name: bump minor version
-        id: bump_minor
-        run: |
-          if contains(github.event.pull_request.labels.*.name, 'version-minor') && env.GH_PUSH_TOKEN != ''
-          then
-            npm run version:minor
-            echo "::set-output name=bumped::true"
-          fi
-        shell: bash
-
-      - name: bump major version
-        id: bump_major
-        run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/bump-version.yml
```diff
@@ -1,44 +1,83 @@
...
+
+      # Bump version
+      - name: bump patch version
+        id: bump_patch
+        run: |
+          if contains(github.event.pull_request.labels.*.name, 'version-patch') && env.GH_PUSH_TOKEN != ''
+          then
+            npm run version:patch
+            echo "::set-output name=bumped::true"
+          fi
```
seriouslysean, line 37: This is new per step, if any of the bumps occur, we set a new variable called `bumped` to be used later in the workflow.

.github/workflows/bump-version.yml
```diff
@@ -1,44 +1,83 @@
...
+      - name: Get version
+        id: package
+        run: |
+          echo "::set-output name=version::$(node -p "require('./package.json').version")"
+        shell: bash
+
+      # Push changes to main if a version bump occurred
+      - name: Push changes
+        run: git push origin main
+        if: steps.bump_patch.outputs.bumped == 'true' || steps.bump_minor.outputs.bumped == 'true' || steps.bump_major.outputs.bumped == 'true'
```
seriouslysean, line 69: If any of the bumps occurred, we can push changes to main. This means it won't run unless a bump did actually happen which is an improvement.

.github/workflows/bump-version.yml
```diff
@@ -1,44 +1,83 @@
...
+        shell: bash
+
+      # Create a new GitHub release if a version bump occurred
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        if: steps.bump_patch.outputs.bumped == 'true' || steps.bump_minor.outputs.bumped == 'true' || steps.bump_major.outputs.bumped == 'true'
```
seriouslysean, line 76: Same thing here, if, any only if a bump took place, will a release be published.

.github/workflows/bump-version.yml
```diff
@@ -1,44 +1,106 @@
+name: Bump Version
+
 on:
-    pull_request:
-      branches:
-        - main
-      types: [closed]
-  
+  pull_request:
+    types: [closed]
```
seriouslysean, line 5: I noticed a bug here.

1. **Trigger:** The workflow is initiated when a pull request is closed on the `main` branch.

This means that any closed PR would run this logic, potentially.

.github/workflows/bump-version.yml
```diff
@@ -1,44 +1,106 @@
...
-          run: npm run version:minor
-  
-        - name: bump major version
-          if: contains(github.event.pull_request.labels.*.name, 'version-major') && env.GH_PUSH_TOKEN != ''
-          run: npm run version:major
-  
-        - name: push
-          if: env.GH_PUSH_TOKEN != ''
-          run: git push origin main
+  check-token:
```
seriouslysean, line 9: 2. **Check Token:** The `check-token` job runs first. It verifies if the `GH_PUSH_TOKEN` secret is set. If it's not, the job fails and the workflow terminates. If the secret is set, the job outputs the token for subsequent jobs.

.github/workflows/bump-version.yml
```diff
@@ -1,44 +1,106 @@
...
+        id: check
+        run: |
+          if [[ -z "${{ secrets.GH_PUSH_TOKEN }}" ]]; then
+            echo "GH_PUSH_TOKEN is not set. Exiting."
+            exit 1
+          fi
+          echo "::set-output name=GH_PUSH_TOKEN::${{ secrets.GH_PUSH_TOKEN }}"
+        shell: bash
+
+  check-merged:
```
seriouslysean, line 24: 3. **Check Merge Status:** The `check-merged` job runs next, contingent on the successful completion of `check-token`. It verifies if the pull request was merged. If it wasn't, the job fails and the workflow terminates. If the pull request was merged, the job outputs the merge status for subsequent jobs.

This is the fix for the above callout-- if the branch was closed AND ALSO MERGED, then we can continue.

.github/workflows/bump-version.yml
```diff
@@ -1,44 +1,106 @@
...
+        id: check
+        run: |
+          if [[ "${{ github.event.pull_request.merged }}" != "true" ]]; then
+            echo "Pull request is not merged. Exiting."
+            exit 1
+          fi
+          echo "::set-output name=merged::${{ github.event.pull_request.merged }}"
+        shell: bash
+
+  get-labels:
```
seriouslysean, line 40: 4. **Get Labels:** The `get-labels` job runs next, contingent on the successful completion of `check-merged`. It retrieves the labels from the pull request and verifies if exactly one of them matches the pattern `version-(patch|minor|major)`. If not, the job fails and the workflow terminates. If it does match, the job outputs the version label for subsequent jobs.

.github/workflows/bump-version.yml
```diff
@@ -1,44 +1,106 @@
...
+          LABELS=$(jq -r '.pull_request.labels[].name' <<< '${{ toJson(github.event) }}')
+          VERSION_LABELS=$(echo "$LABELS" | grep -E '^version-(patch|minor|major)$' || true)
+          if [[ $(echo "$VERSION_LABELS" | wc -l) -ne 1 ]]; then
+            echo "Exactly one version label is required. Exiting."
+            exit 1
+          fi
+          echo "::set-output name=version::$(echo "$VERSION_LABELS" | sed 's/^version-//')"
+        shell: bash
+
+  bump-and-push-version:
```
seriouslysean, line 58: 5. **Bump and Push Version:** The `bump-and-push-version` job runs next, contingent on the successful completion of `check-token` and `get-labels`. It checks out the repository, sets up Node.js, installs dependencies, bumps the version in `package.json` according to the version label, commits the change, and pushes the commit and tags to the `main` branch.

I originally had this broken in to two jobs, but it means duplicating the node setup, checkout and git setup logic so I opted to consolidate them.

.github/workflows/bump-version.yml
```diff
@@ -1,44 +1,106 @@
...
+
+      - name: Push changes
+        run: |
+          git push origin main
+          git push origin --tags
+        env:
+          GH_PUSH_TOKEN: ${{ needs.check-token.outputs.GH_PUSH_TOKEN }}
+        shell: bash
+
+  create-release:
```
seriouslysean, line 94: 6. **Create Release:** The `create-release` job runs last, contingent on the successful completion of `get-labels` and `bump-and-push-version`. It creates a new release on GitHub with the version label as the tag name and release name.

This will create the actual tag in the repo AND the release, and is why we use `--no-git-tag-version` when bumping the package version.

.github/workflows/bump-version.yml
itsalaidbacklife: Similarly to above, why is erroring when the PR was manually closed better than no-op succeeding?

.github/workflows/bump-version.yml
itsalaidbacklife: Good catch. I think I had the 
```    if: github.event_name == 'pull_request' && github.event.action == 'closed'
```
line before and it seems to have been nixed somewhere in getting the 'happy flow' to work

.github/workflows/bump-version.yml
```diff
@@ -1,44 +1,106 @@
...
+      GH_PUSH_TOKEN: ${{ steps.check.outputs.GH_PUSH_TOKEN }}
+    steps:
+      - name: Check GH_PUSH_TOKEN
+        id: check
+        run: |
+          if [[ -z "${{ secrets.GH_PUSH_TOKEN }}" ]]; then
+            echo "GH_PUSH_TOKEN is not set. Exiting."
+            exit 1
+          fi
+          echo "::set-output name=GH_PUSH_TOKEN::${{ secrets.GH_PUSH_TOKEN }}"
```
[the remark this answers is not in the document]
seriouslysean: For 1. I was copying older code I had around, so I've updated that to the new method.

2. Ideally a fail would be better, but I guess each fork does need to run this so in that case a noop makes more sense.

3. That was also a miss on my part, so I've switched to the ENV.

.github/workflows/bump-version.yml
seriouslysean: I've switched this to noop each consecutive job is the token isn't set.

package.json
```diff
@@ -92,9 +92,6 @@
     "storybook:build": "storybook build",
     "test:unit": "vitest run",
     "prepare": "husky install",
-    "version:patch": "npm version patch",
-    "version:minor": "npm version minor",
-    "version:major": "npm version major",
```
seriouslysean, line 97: We don't need these, they're only simple aliases to the npm command.

.github/workflows/bump-version.yml
```diff
@@ -1,44 +1,151 @@
...
+              echo "Artifact $artifact_name is not available yet"
+              if [ "$attempt_number" -eq "$max_attempts" ]; then
+                echo "### ❌ Artifact $artifact_name is not available after $max_attempts attempts" >> $GITHUB_STEP_SUMMARY
+                exit 1
+              fi
+              attempt_number=$((attempt_number+1))
+              echo "Waiting before retrying..."
+              sleep 10
+            fi
+          done
```
itsalaidbacklife, line 79: Could this be simplified using the official [download artifact action](https://github.com/actions/download-artifact)? We're currently using it to cache the app build during e2e testing. I imagine this would be similar and might allow us to rely on the action for retry logic?

seriouslysean: As far as I can tell the action doesn't actually do any retry logic, but I need to double check.

PR #821, opened 2023-11-30
fix: set game overlays to be persistent

### LINKED ISSUE #820: [Bug]: Players can illegally draw while waiting for their opponent to discard
### Bug Summary

Players are able to dismiss various game overlays by clicking, and this allows them to make a move out of turn.

### Detailed Description

Follow up to #753. None of the other game overlays besides the one for waiting for opponent to counter are `persistent`, meaning clicking will close them out and then make moves when their opponent is doing something (out of turn).

See this game history:
image

We should make the overlays in `GameOverlays` all `persistent`, but moreover we should update all the in game moves to validate that we are not waiting for something else to happen. 

### Current Behavior

When waiting for an opponent to discard a card, you can click to close out the overlay and then click the deck to draw a card.

The same applies for waiting for opponent to resolve a three, waiting for an opponent to play from the deck, waiting for an opponent to discard from a jack.

### Expected Behavior

Clicking should not close the above overlays. Making the api request to make a move (even if you forcibly remove the overlay clientside) should reject these moves with errors.

### Steps to Reproduce

1. Log in two accounts 
2. Create, join, and start a game
3. Have player 1 play a 4
4. Have player 2 resolve it
5. Have player 1 click to close out the 'Opponent Discarding' overlay
6. Have player 1 click the deck to draw a card. They are able to draw a card even though their move was playing the 4 and the opponent should be the only one allowed to act

itsalaidbacklife:
Makes the overlays in the `GameOverlays` `persistent` so that you can't click to close them. We made the same move for the overlay when waiting for opponent to counter, but we haven't applied this to the other overlays.
## Issue number

- Resolves #820 

## Please check the following

- For New Features:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cuttle",
-  "version": "6.4.2",
+  "version": "6.4.3",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "cuttle",
-      "version": "6.4.2",
+      "version": "6.4.3",
       "dependencies": {
         "chart.js": "4.4",
         "connect-redis": "3.0.2",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "6.4.2",
+  "version": "6.4.3",
   "dependencies": {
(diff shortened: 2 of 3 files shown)


Haviles04 (approved):
I think we should look into possibly making the game menu available while the overlay is up this way you can't get "trapped" inside a game.

PR #824, opened 2023-12-01
DevEx #811 Postgres migration to timestamptz

### LINKED ISSUE #811: [DevEx]: Update the timestamp columns across the codebase to use postgres’ `timestampz` data type
### Improvement Summary

Currently, our time-related columns are stored as simple numbers, represented as milliseconds since Epoch. We should update these in both the code and database to use a proper timestamp type so that the values are human readable and so that we avoid limitations on the size of numbers in both postgres int8 and JavaScript types.

### Detailed Description

This will require updating frontend, the backend, and migrating the existing database rows to convert the existing values to the new format.

The timezone should be set to EST for all timestamp rows.

At the code level, the three types of changes to look for:
1. Updating the models in api/models so that the time-related attributes have a `columnType` of  `timestampz`
2. Updating code that consumes or writes to these columns so that it correctly parses them to and from `dayjs()` instances (where most of the processing logic happens)
3. Checking and revising any queries that compare times to ensure that they still correctly do what they are supposed to do.

In particular, we should check the Season and Match models’ `startTime` and `endTime` attributes, and the `lockedAt` attribute of the `Game` model. We should update the consumers of those columns, including the `StatsController` and its associated helpers, the helpers used when the game ends to add the game to its appropriate match, and the `lock-game` and `unlock-game` helpers.

Generally we should look for usages of `dayjs` which are the areas that involve time calculations and ensure those features still work correctly.

For the database migration, we will need to alter the column types and convert the existing values to match the new types, being careful not to corrupt the existing data. For this reason, the migrations will require significant testing.

Note that we should not need to change the existing `createdAt` and `updatedAt` columns which are automatically managed by sails

Haviles04:
## Issue number

- Resolves #811

## Please check the following

- For New Features:


api/controllers/StatsController.js
```diff
@@ -148,7 +148,7 @@ module.exports = {
       });
       const currentSeasonGames = Game.find({
         status: gameService.GameStatus.FINISHED,
-        updatedAt: { '>': currentSeason.startTime, '<': currentSeason.endTime },
+        updatedAt: { '>': dayjs(currentSeason.startTime).valueOf(), '<': dayjs(currentSeason.endTime).valueOf() },
```
Haviles04, line 151: I wonder if It makes sense to change `updatedAt` and `createdAt` to the same timestamptz just to keep everything consistent. This works but I'm not sure it's the best way

Haviles04:
This still needs a bunch of testing and a database migration, but I got a good start on the backend changes.

psqls coerces all the timestamptz columns to be in UTC timezone. So as a safety I pushed all our dates to be UTC BEFORE they hit they backend just for an extra safety factor.

Haviles04:
```
alter table "match"
add column "new_startTime" timestamptz,
add column "new_endTime" timestamptz;

update "match"
set "new_startTime" = to_timestamp("startTime"/1000.0),
	"new_endTime" = to_timestamp("endTime"/1000.0);

alter table "match"
drop column "startTime",
drop column "endTime";

alter table "match"
rename column "new_startTime" to "startTime",
rename column "new_endTime" to "endTime";
```

tests/e2e/fixtures/statsFixtures.js
```diff
@@ -51,67 +53,67 @@ const seasonOneMatches = [
...
+    endTime: dayjs('2022-01-20').format(),
   },
   // Player 2 week 1
   {
     player1: 'player2',
     player2: 'player3',
     winner: 'player2',
-    startTime: dayjs('2022-01-20', 'yyyy-mm-dd').valueOf(),
-    endTime: dayjs('2022-01-20', 'yyyy-mm-dd').valueOf(),
+    startTime: dayjs('2022-01-20').format(),
```
Haviles04, line 85: Put back mmyydd

api/controllers/StatsController.js
itsalaidbacklife: Yes I think we want all date formates to become `timestampz`

tests/e2e/specs/in-game/game-over/rematch.spec.js
```diff
@@ -70,16 +72,16 @@ describe('Creating And Updating Ranked Matches With Rematch', () => {
           player1: this.playerOneId,
           player2: this.playerTwoId,
           winner: this.playerOneId,
-          startTime: dayjs().subtract(1, 'week').subtract(1, 'day').valueOf(),
-          endTime: dayjs().subtract(1, 'week').subtract(1, 'day').valueOf(),
+          startTime: dayjs().utc().subtract(1, 'week').subtract(1, 'day').format(),
+          endTime: dayjs().utc().subtract(1, 'week').subtract(1, 'day').format(),
```
itsalaidbacklife, line 76: Why are we using `.utc()` in some places and not others? I'm not clear on whether we want to be using utc() or not.  What's your understanding, @Haviles04 ?

Haviles04: I don't believe we want UTC

api/controllers/StatsController.js
```diff
@@ -145,14 +145,17 @@ module.exports = {
         select: ['id', 'username'],
       });
       const currentSeasonMatches = Match.find({
-        startTime: { '>': currentSeason.startTime },
-        endTime: { '<': currentSeason.endTime },
+        startTime: { '>': dayjs(currentSeason.startTime).format('YYYY-MM-DD HH:mm:ss.SSS Z') },
```
itsalaidbacklife, line 148: Silly question, but how does `currentSeason.startTime` compare to 

```
dayjs(currentSeason.startTime).format('YYYY-MM-DD HH:mm:ss.SSS Z');
```

Is `currentSeason.endTime` already a string in that format (if that's how postgres is storing timestampz columns? Is this equivalent to the follwing?
```suggestion
        startTime: { '>': currentSeason.startTime},
```

Haviles04: So it should already be in that format, this is the format that postgres puts it in.

api/models/Card.js
```diff
@@ -57,4 +59,12 @@ module.exports = {
...
     },
   },
+  beforeCreate(record, proceed) {
+    record.createdAt = dayjs(record.createdAt).format('YYYY-MM-DD HH:mm:ss.SSS Z');
+    return proceed();
+  },
+  beforeUpdate(record, proceed) {
+    record.updatedAt = dayjs(record.createdAt).format('YYYY-MM-DD HH:mm:ss.SSS Z');
+    return proceed();
+  },
```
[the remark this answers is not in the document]
itsalaidbacklife: Upon further reading, it looks like we can't use the `models` config to set lifecycle hooks so this may not work. Worth a shot imo

tests/e2e/specs/in-game/game-over/rankedMatches.spec.js
```diff
@@ -5,15 +5,17 @@ import { playerOne, playerTwo, playerThree } from '../../../fixtures/userFixture
...
 
 const dayjs = require('dayjs');
+const utc = require('dayjs/plugin/utc');
+dayjs.extend(utc);
 
 function validateMatchResult(match, length, p1, p2, winnerId) {
   expect(match.player1.id).to.eq(p1);
   expect(match.player2.id).to.eq(p2);
-  expect(match.startTime).to.be.greaterThan(0);
+  expect(match.startTime).to.not.eq(null);
```
itsalaidbacklife, line 14: Have we confirmed that the match will have a `null` `startTime` if none is set?

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -1,4 +1,6 @@
 import dayjs from 'dayjs';
+import utc from 'dayjs/plugin/utc';
+dayjs.extend(utc);
```
itsalaidbacklife, line 3: I'd prefer to place this below the imports so all the imports go together

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -112,7 +114,7 @@ describe('Home - Game List', () => {
     it('Does not show games older than 24 hours', () => {
       cy.loadFinishedGameFixtures([
         { name: 'New Game', status: GameStatus.CREATED },
-        { name: 'Old Game', status: GameStatus.CREATED, createdAt: dayjs().subtract(1, 'day').valueOf() },
+        { name: 'Old Game', status: GameStatus.CREATED, createdAt: dayjs.utc().subtract(1, 'day').valueOf() },
```
itsalaidbacklife, line 117: Should this be `.toDate()`?
```suggestion
        { name: 'Old Game', status: GameStatus.CREATED, createdAt: dayjs.utc().subtract(1, 'day').toDate() },
```

api/controllers/TestController.js
```diff
@@ -30,7 +34,14 @@ module.exports = {
 
   loadSeasonFixture: async function (req, res) {
     try {
-      const seasons = await Season.createEach(req.body).fetch();
+      const editedSeasons = req.body.map((season) => {
```
itsalaidbacklife, line 37: This feels like it warrants a comment to explain the need to transform the data. Ideally if we can write up an issue in sails we could link it here. In the meantime maybe something like this?
```suggestion
      // transform timestamps to `Date` objects, as sails-disk doesn't support ISO timestamps
      const editedSeasons = req.body.map((season) => {
```

tests/e2e/specs/in-game/game-over/rankedMatches.spec.js
Haviles04: Not yet, this could probably use some looking at

itsalaidbacklife: Confirmed this will be null when unset.

itsalaidbacklife (approved):
Excellent work. Final nitpicks in the comments

api/controllers/StatsController.js
```diff
@@ -121,18 +121,30 @@ function addMatchToRankings(season, match, player, opponent) {
...
   });
+
+  // Remove zero-valued gameCounts + uniquePlayersPerWeek from the end
+  if (gameCounts[gameCounts.length - 1] === 0) {
+    gameCounts.pop();
+    uniquePlayersPerWeek.pop();
+  }
+
+  // Convert uniquePlayersPerWeek from Set[] to int[] number of players
+  uniquePlayersPerWeek = uniquePlayersPerWeek.map((playerSet) => playerSet.size);
```
itsalaidbacklife, line 141: Changing to using `let` instead of `const` and mutating the season in-place has the disadvantage of no longer being a pure function that returns new data w/o side effects, because the season you input here gets mutated in-place, in addition to the data being returned. But it's also more concise, so I think that it's reasonable, especially given that the only intended use of this function is right before returning to the client.

api/controllers/StatsController.js
```diff
@@ -121,18 +121,30 @@ function addMatchToRankings(season, match, player, opponent) {
...
+
   return {
     ...rest,
-    rankings: rankingsAsArray,
-    uniquePlayersPerWeek: season.uniquePlayersPerWeek.map((playerSet) => playerSet.size),
+    rankings,
+    gameCounts,
+    uniquePlayersPerWeek,
   };
```
itsalaidbacklife, line 148: One note is I think we could alternatively skip `...rest` and simply `return season` since we change data in place now. That might be cleanest

itsalaidbacklife: Update: I tried this and it pisses off the linter because the mutated variables are "unused". That's not really true in practice but I'm not looking to fight the linter on such a small issue

api/controllers/StatsController.js
seriouslysean: I still think the const makes more sense. Changing the shape of what you send in is kinda weird and feels like an anti pattern. Making a new named const and returning that instead is a little less brittle imo. Consider if this fails, you’ll have a half modified season since it works by ref, whereas if you used a new const it wouldn’t have any side effects outside of this function as far as I can tell.

itsalaidbacklife (approved):
Excellent work. This has been a long time in the making and it will make it so much easier to maintain, especially with respect to creating seasons in the db. Very well done.

PR #829, opened 2023-12-04
fix: Prevent scrap from revealing the last-played card unless the last move was a 3

### LINKED ISSUE #828: [Bug]: Card will appear in the scrap and stay there sometimes
### Bug Summary

Some of the time a card appears revealed in the scrap pile, revealed randomly after someone counters or does another move that *isn't* picking a card with a three.

### Detailed Description

See image:
image

One player played a three and it was countered, but the two appears as revealed.

Looking at the `GameView`, the `scrapChoice` computed property doesn't look to check that the `lastEvent` is specifically a three resolution. This is likely because the property that it is checking `lastEventCardChosen` used to be the more specific `cardSelectedFromScrap` before #126. 

We should update this computed prop to additionally check the lastEvent name. Because the behavior is inconsistent, the cause may be a race condition.

### Current Behavior

Sometimes a card (usually a two just used to counter) appears in the scrap revealed and sticks around until the player refreshes the page. 

### Expected Behavior

The card in the scrap should only be revealed during the three animation, and then only temporarily.

### Steps to Reproduce

1. Sign in/up, join + start a game
2. Put at least one card in the scrap
3. Have player 1 play a 3
4. Have player 2 counter it
5. Have player 1 resolve so the three is countered
6. See that the two used to counter appears and freezes in the scrap.

itsalaidbacklife:
Now that the `lastEvent` object has been flattened into its individual properties, it's important that various consumers of `lastEvent` data be able to tell which type of move (change) the `lastEvent` was immediately rather than waiting for the change-specific logic to apply.

Additionally, consumers of `lastEventCardChosen` etc should specifically check if the `lastEvent` is a particular change i.e. was a three played vs a four in order to only display their content for the appropriate moves.

## Issue number

- Resolves #828 

## Please check the following

- For New Features:

### DIFF
--- a/src/plugins/sockets/inGameEvents.js
+++ b/src/plugins/sockets/inGameEvents.js
@@ -28,6 +28,7 @@ export async function handleInGameEvents(evData) {
       gameStore.setGameOver(evData.victory);
     }, 1000);
   }
+  gameStore.lastEventChange = evData.change;
   switch (evData.change) {
     case SocketEvent.READY: {
       gameStore.updateReady(evData.pNum);
--- a/src/routes/game/GameView.vue
+++ b/src/routes/game/GameView.vue
@@ -420,6 +420,7 @@ import { mapStores } from 'pinia';
 import { useI18n } from 'vue-i18n';
 import { useGameStore } from '@/stores/game';
 import { useAuthStore } from '@/stores/auth';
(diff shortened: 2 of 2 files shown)


itsalaidbacklife:
Closing as dupe of #832

PR #834, opened 2023-12-06
Optimized the bg-image file

### LINKED ISSUE #158: [Feature]: Add loading state to background image in `GameView`
### Feature Summary

Currently the `GameView` will sometimes load a blank white page for the background while the background svg loads. We should update the `GameView` to display some nicer loading state until the image loads

### Detailed Description

Open discussion on what that loading state should entail

Posoroko:
## Issue Number

- Resolves #158

## Please check the following

- For New Features:

I created and added a backgroud gradient for the #app element so it will show while the bg image loads.
I deleted the white bg color in LoginView

The board-background.svg file was over 2mb, so I turned it into a webp image.
I optimized it and lowered the file size to 64k, but the image quality is still good.

I changed the bg image in LoginView and RematchView to use the .webp file


src/App.vue
```diff
@@ -53,6 +53,10 @@ export default {
 <style lang="scss">
 @import url('https://fonts.googleapis.com/css2?family=Cormorant+Infant:wght@400;600;700&family=Luckiest+Guy&family=Changa:wght@400;500;700&family=Libre+Baskerville:wght@400;700&family=PT+Serif:wght@400;700&display=swap');
 
+#app {
+  background: radial-gradient(circle, rgba(209,154,78,1) 0%, rgba(158,105,64,1) 100%);
+}
```
itsalaidbacklife, line 58: Does this really belong as a global gradient for every page? Even the ones that don't have this background image?

Posoroko: I tried applying the gradient to the #game-view-wrapper, where the board-background.webp is applied, but it wasn't as smooth.  
The #app seems to load faster, showing it's gradient earlier.  The login view has the board-background, so loading the gradient on this page is fine.  I thought we might as well keep it alive in the lobby, even if it hidden by the lobby brown bg-color, because the user will eventually load the gameView , where this gradient is needed again.

itsalaidbacklife: What about for the home view?

itsalaidbacklife: @Posoroko the problem with this is that it changes the design in pages across the entire app, including when the app is not loaded. For example, it changes the colors in the LoginView behind the logo. I think we need a solution that is more limited in scope so that areas of the application not using the game board aren't affected. Putting it in the GameView itself makes the most intuitive sense to me, though I'm open to discussing alternatives

itsalaidbacklife (approved):
Thank you! This should improve loading times for the biggest asset in the app.

PR #835, opened 2023-12-06
Testing the language files

### LINKED ISSUE #808: [DevEx]: Testing the language files
### Improvement Summary

Make sure that all language files have all required keys

### Detailed Description

As we add new features, more entries will be added to the language files.
It will be important that all text content gets added to every language files.  So, we need to write a test to verify that all files are populated correctly.

the test should follow a few rules :

1 - The en.json files must be used as a reference.
2 - every key from en.json must exist in every other file
3 - other files shouldn't have extra keys that don't exist in en.json
4 - none of the keys should be an empty string
5 - the test should be added to tests/e2e/specs/out-of-game/localization.spec.js

Posoroko:
## Issue number

- Resolves #808 

## Please check the following

- For New Features:

Tests that check that all files have the same keys as the English file; none missing, none extra.
Tests that check that no keys have a value of an empty string.

I corrected the Spanish file that was missing a key.


seriouslysean (changes requested):
I think these are interesting, specifically the no empty string tests. I don't think that the same number of keys is a good test to fail the build as we have a fallback translation of English intentionally. It is useful information though.

Curious what @itsalaidbacklife has to say.

tests/e2e/specs/out-of-game/localization.spec.js
```diff
@@ -72,3 +72,50 @@ describe('Localization', () => {
...
+
+describe('language files', () => {
+  function extractKeys(obj, keyList = []) {
+    for (let key in obj) {
+      if (typeof obj[key] === 'object' && obj[key] !== null) {
+        extractKeys(obj[key], keyList);
+      } else {
+        keyList.push(key);
+      }
+    }
```
Haviles04, line 84: I think it may be more readable and faster to use Object.keys and forEach here, but I'm not 100% sure on that

```suggestion
    Object.keys(obj).forEach(key => {
      if (typeof obj[key] === 'object' && obj[key] !== null) {
        extractKeys(obj[key], keyList);
      } else {
        keyList.push(key);
      }
    });
```

tests/e2e/specs/out-of-game/localization.spec.js
```diff
@@ -72,3 +72,50 @@ describe('Localization', () => {
...
+  }
+
+  let enKeys = extractKeys(en).sort();
+  let frKeys = extractKeys(fr).sort();
+  let esKeys = extractKeys(es).sort();
+
+  it('English should have no empty strings', () => {
+    enKeys.forEach((key) => {
+      expect(en[key]).to.not.eql('');
+    });
```
itsalaidbacklife, line 95: We can spruce up the assertion to be more specific and legible in case of failures, using the 2nd argument of `.eq()` to write an assertion message. Also we generally use the `.eq()` assertion rather than `.eql()`. There's no difference, but it's best to have a singular convention.

```suggestion
    enKeys.forEach((key) => {
      const assertionMsg = `en.json value at key ${key} should be non-empty, but it was empty`;
      expect(en[key]).to.not.eq('', assertionMsg);
    });
```

tests/e2e/specs/out-of-game/localization.spec.js
```diff
@@ -72,3 +72,50 @@ describe('Localization', () => {
...
+  });
+
+  it('French should have the same keys', () => {
+    for(let i = 0; i < enKeys.length; i++) {
+      expect(enKeys[i]).to.eql(frKeys[i]);
+    }
```
itsalaidbacklife, line 101: Same idea as above; let's use a descriptive error message so it's easier to identify which key failed, why
```suggestion
    for (let i = 0; i < enKeys.length; i++) {
      const assertionMsg = `fr.json should have the key ${ enKeys[i] } for key number ${ i }, but instead it had key ${ frKeys[i] }`;
      expect(enKeys[i]).to.eq(frKeys[i], assertionMsg);
    }
```

tests/e2e/specs/out-of-game/localization.spec.js
```diff
@@ -72,3 +72,50 @@ describe('Localization', () => {
...
+  });
+
+  it('French should have no empty strings', () => {
+    frKeys.forEach((key) => {
+      expect(fr[key]).to.not.eql('');
+    });
```
itsalaidbacklife, line 107:
```suggestion
    frKeys.forEach((key) => {
      const assertionMsg = `fr.json should have nonempty value for key ${key}, but it value was empty`;
      expect(fr[key]).to.not.eq('', assertionMsg);
    });
```

tests/e2e/specs/out-of-game/localization.spec.js
```diff
@@ -72,3 +72,50 @@ describe('Localization', () => {
...
+  }); 
+
+  it('Spanish should have the same keys', () => {
+    for(let i = 0; i < enKeys.length; i++) {
+      expect(enKeys[i]).to.eql(esKeys[i]);
+    }
```
itsalaidbacklife, line 113:
```suggestion
    for(let i = 0; i < enKeys.length; i++) {
      const assertionMsg = `es.json should have the key ${ enKeys[i] } for key number ${ i }, but instead it had key ${ esKeys[i] }`
      expect(enKeys[i]).to.eq(esKeys[i], assertionMsg);
    }
```

tests/e2e/specs/out-of-game/localization.spec.js
```diff
@@ -72,3 +72,50 @@ describe('Localization', () => {
...
+  });
+
+  it('Spanish should have no empty strings', () => {
+    frKeys.forEach((key) => {
+      expect(es[key]).to.not.eql('');
```
itsalaidbacklife, line 118:
```suggestion
      const assertionMsg = `es.json should have nonempty value for key ${key}, but it value was empty`;
      expect(es[key]).to.not.eql('', assertionMsg);
```

itsalaidbacklife (approved):
Nice work! This is a cool solution and per the discord conversation, I think it will be better to enforce that the language files have all expected text as updates are made. I'm going to commit some changes to clarify the error messages when there is a failure.

PR #841, opened 2023-12-08
error handling refactored, key added to language files

### LINKED ISSUE #790: [DevEx]: Refactor `GameOverDialog` so that it emits errors up to the GameView instead of instantiating its own `BaseSnackbar`
### Improvement Summary

The `GameOverDialog` has a `BaseSnackbar` to display errors from its requests for rematches. We should refactor this component so that instead of displaying the errors itself, it emits events with the errors up to the `GameDialogs` and then up to the `GameView` so that we can leverage the existing `handleError` logic in the `GameView` without needing an extra snackbar

### Detailed Description

This should follow the same pattern seen in how the other dialogs in the `GameOverDialogs` emit errors e.g. `SevenDoubleJacksDialog` emits a `'handle-error'` event to the `GameDialog`. 

The `GameDialogs` should listen for a `'handle-error'` event on the `GameOverDialog` and bind the existing `handleError()` handler to it so that this causes the `GameDialogs` to emit an event to the `GameView` which it uses to render the existing error in its existing `BaseSnackbar`. We should then remove the `BaseSnackbar` from the `GameOverDialog` along with all unnecessary variables and logic

Posoroko:
## Issue Number

- Resolves #790 

## Please check the following

- For New Features:

deleted the snackbar from gameOverView
created a "handle-error" emit that passes the error message key all the way up to gameView
added new keys to the language files

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cuttle",
-  "version": "7.0.0",
+  "version": "7.0.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "cuttle",
-      "version": "7.0.0",
+      "version": "7.0.1",
       "dependencies": {
         "chart.js": "4.4",
         "connect-redis": "3.0.2",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "7.0.0",
+  "version": "7.0.1",
   "dependencies": {
     "chart.js": "4.4",
     "connect-redis": "3.0.2",


Posoroko:
When reviewing, could you have a look at lines 160 to 169 in GameDialogs.vue ?  handleError is called a few times, but with error message as argument ... Is that normal ? Am I missing something ?

Haviles04:
> When reviewing, could you have a look at lines 160 to 169 in GameDialogs.vue ? handleError is called a few times, but with error message as argument ... Is that normal ? Am I missing something ?

I believe those functions should return an error message, so you could probably pass it along to the handleError

src/routes/game/GameView.vue
```diff
@@ -410,7 +410,7 @@
         @points="playPoints"
         @target="beginTargeting"
       />
-      <GameDialogs @clear-selection="clearSelection" @handle-error="handleError" />
+      <GameDialogs @clear-selection="clearSelection" @handle-error="handleError($event)" />
```
Haviles04, line 413: I don't think you need to pass it the event, I believe it will automatically be passed the event object

Posoroko:
I updated my question.  I realized it was not clear.  When emitting 'handle-error', don't we need to pass the error messageKey along ?

Haviles04:
> I updated my question.  I realized it was not clear.  When emitting 'handle-error', don't we need to pass the error messageKey along ? 

I believe it will automatically receive the message when the backend throws the error .

 @itsalaidbacklife thoughts?

src/routes/game/GameView.vue
itsalaidbacklife: Correct, the second argument passed to `$emit(eventName, eventData)` is passed to the corresponding event handler w/o needing to specifing $event as an argument of the handler ([docs](https://vuejs.org/guide/components/events.html#event-arguments))
```suggestion
      <GameDialogs @clear-selection="clearSelection" @handle-error="handleError" />
```

src/routes/game/components/GameOverDialog.vue
```diff
@@ -256,15 +245,10 @@ export default {
       try {
         await this.gameStore.requestRematch({ gameId:this.gameStore.id, rematch: true});
       } catch (e) {
-        this.showSnackbar = true;
-        this.snackBarMessage = 'Error requesting rematch';
+        this.$emit('handle-error', 'game.snackbar.rematch.errorRequesting');
```
Haviles04, line 248: This feels weird because usually the backend is the one sending back the error, but I noticed the `Rematch` endpoint doesn't actually have any errors to throw. 

@itsalaidbacklife is this something your PR addresses?

PR #846, opened 2023-12-17
Feat #845 Auto-generate rematch game names with score

### LINKED ISSUE #845: [Feature]: Auto-generate game name based on score.
### Feature Summary

Auto-generate score based names inside Rematch

### Detailed Description

Currently the Game name inside for rematch games does not indicate the score. We should update the logic inside `rematch.js` to count the number of games wins in that current rematch chain and name the game accordingly. 

`Player1 VS Player 2 1-1-1`

Haviles04:
## Issue number

- Resolves #845 

## Please check the following

- For New Features:

- Added functions to make an array of all previous rematch games
- Got the count of all wins and stalemates (p0 is based off p0 of game 1 in the series)
- Added assertions to check for correct game names

### DIFF
--- a/api/controllers/game/rematch.js
+++ b/api/controllers/game/rematch.js
@@ -65,9 +65,9 @@ module.exports = async function (req, res) {
     const seriesP0 = [p0, p1].find(({ id }) => id === rematchGames[0].p0);
     const seriesP1 = [p0, p1].find(({ id }) => id === rematchGames[0].p1);
     //Get rematchGame win counts
-    let player0Wins = rematchGames.filter(({winner}) => winner === seriesP0.id).length;
-    let player1wins = rematchGames.filter(({winner}) => winner === seriesP1.id).length;
-    let stalemates = rematchGames.filter(({winner}) => winner === null).length;
+    const player0Wins = rematchGames.filter(({winner}) => winner === seriesP0.id).length;
+    const player1wins = rematchGames.filter(({winner}) => winner === seriesP1.id).length;
+    const stalemates = rematchGames.filter(({winner}) => winner === null).length;
     
     const newName = `${seriesP0.username} VS ${seriesP1.username} ${player0Wins}-${player1wins}-${stalemates}`;
     await Game.updateOne({ id: newGame.id }).set({ name: newName });
--- a/api/controllers/game/rematch.js
+++ b/api/controllers/game/rematch.js
@@ -49,24 +49,24 @@ module.exports = async function (req, res) {
     updatedGame.rematchGame = newGame.id;
 
     //Get all exisiting rematchGames
-    let pastGames = [];
-    const getGames = async (gameId) => {
+    let rematchGames = [];
+    const getRematchGames = async (gameId) => {
       const [gameToAdd] = await Game.find({ rematchGame: gameId });
       if (gameToAdd) {
-        pastGames.unshift(gameToAdd);
+        rematchGames.unshift(gameToAdd);
         if (gameToAdd.rematchGame) {
-          await getGames(gameToAdd.id);
+          await getRematchGames(gameToAdd.id);
         }
       }
       return;
     };
-    await getGames(newGame.id);
-    const seriesP0 = [p0, p1].find(({ id }) => id === pastGames[0].p0);
-    const seriesP1 = [p0, p1].find(({ id }) => id === pastGames[0].p1);
+    await getRematchGames(newGame.id);
+    const seriesP0 = [p0, p1].find(({ id }) => id === rematchGames[0].p0);
+    const seriesP1 = [p0, p1].find(({ id }) => id === rematchGames[0].p1);
     //Get rematchGame win counts
-    let player0Wins = pastGames.filter(({winner}) => winner === seriesP0.id).length;
-    let player1wins = pastGames.filter(({winner}) => winner === seriesP1.id).length;
-    let stalemates = pastGames.filter(({winner}) => winner === null).length;
+    let player0Wins = rematchGames.filter(({winner}) => winner === seriesP0.id).length;
+    let player1wins = rematchGames.filter(({winner}) => winner === seriesP1.id).length;
+    let stalemates = rematchGames.filter(({winner}) => winner === null).length;
     
     const newName = `${seriesP0.username} VS ${seriesP1.username} ${player0Wins}-${player1wins}-${stalemates}`;
     await Game.updateOne({ id: newGame.id }).set({ name: newName });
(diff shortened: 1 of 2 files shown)


api/controllers/game/rematch.js
```diff
@@ -40,14 +40,37 @@ module.exports = async function (req, res) {
...
       Game.replaceCollection(newGame.id, 'players').members([newP0Id, newP1Id]),
       User.updateOne({ id: newP0Id }).set({ pNum: 0 }),
       User.updateOne({ id: newP1Id }).set({ pNum: 1 }),
     ]);
     updatedGame.rematchGame = newGame.id;
 
+    //Get all exisiting rematchGames
+    let rematchGames = [];
+    const getRematchGames = async (gameId) => {
+      const [gameToAdd] = await Game.find({ rematchGame: gameId });
```
itsalaidbacklife, line 54: You can use `.findOne()` to avoid the need for destructuring here
```suggestion
      const gameToAdd = await Game.findOne({ rematchGame: gameId });
```

api/controllers/game/rematch.js
```diff
@@ -40,14 +40,37 @@ module.exports = async function (req, res) {
...
+    const [ , , p0, p1] = await Promise.all([
       Game.updateOne({ id: updatedGame.id }).set({ rematchGame: newGame.id }),
       Game.replaceCollection(newGame.id, 'players').members([newP0Id, newP1Id]),
       User.updateOne({ id: newP0Id }).set({ pNum: 0 }),
       User.updateOne({ id: newP1Id }).set({ pNum: 1 }),
     ]);
     updatedGame.rematchGame = newGame.id;
 
+    //Get all exisiting rematchGames
+    let rematchGames = [];
```
itsalaidbacklife, line 52: This is never redeclared, so we can use const
```suggestion
    const rematchGames = [];
```

api/controllers/game/rematch.js
```diff
@@ -40,14 +40,37 @@ module.exports = async function (req, res) {
...
+    const getRematchGames = async (gameId) => {
+      const [gameToAdd] = await Game.find({ rematchGame: gameId });
+      if (gameToAdd) {
+        rematchGames.unshift(gameToAdd);
```
itsalaidbacklife, line 56: My initial reaction was that it would be tidier if this function returned the array instead of mutating it as a side effect like
```
const getRematchGames = async (gameId, previousGames) () => {
  const gameToAdd = await Game.findOne({ rematchGame: gameId });
  if (gameToAdd) {
    previousGames.unshift(gameToAdd);
    if (gameToAdd.rematchGame) {
      previousGames = await getRematchGames(gameToAdd.id, previousGames);
    }
  }
  return previousGames;
}
```

But now I wonder about the performance implications of each recursive function having its own separate reference to the array being mutated. I suspect your implementation is more performant, and it still reads reasonably clearly. 

@seriouslysean what do you think?

seriouslysean: Honestly, I'm a fan of less loops and this is a clean implementation of a recursive function. If I were writing it, I'd prefer fast fails to what we have here, but I wouldn't really change much else -- this is really nice!

```
// Get all existing rematchGames
const rematchGames = [];
const getRematchGames = async (gameId) => {
  const gameToAdd = await Game.findOne({ rematchGame: gameId });
  if (!gameToAdd) {
    return;
  }
  rematchGames.unshift(gameToAdd);
  if (!gameToAdd.rematchGame) {
    return;
  }
  await getRematchGames(gameToAdd.id);
};
```

I would just be very clear about any pitfalls with this approach as I don't have the full context. Make sure it's not able to run forever, test it well and look for any other areas you may need to exit cleanly. Do you need any sort of tray watch if what await fails inside of the rematchGame check, etc.

Haviles04: I started with a fast fail but it didn't look as clean visually imo

seriouslysean: I’d say switch it, it’s logically more coherent and generally more performant.

Haviles04: Cool, okay will do this when I get home from work later 🙂

api/controllers/game/rematch.js
```diff
@@ -40,14 +40,38 @@ module.exports = async function (req, res) {
...
+    };
+    await getRematchGames(newGame.id);
+    const seriesP0 = [p0, p1].find(({ id }) => id === rematchGames[0].p0);
+    const seriesP1 = [p0, p1].find(({ id }) => id === rematchGames[0].p1);
+    //Get rematchGame win counts
+    let player0Wins = rematchGames.filter(({winner}) => winner === seriesP0.id).length;
+    let player1wins = rematchGames.filter(({winner}) => winner === seriesP1.id).length;
+    let stalemates = rematchGames.filter(({winner}) => winner === null).length;
+    
+    const newName = `${seriesP0.username} VS ${seriesP1.username} ${player0Wins}-${player1wins}-${stalemates}`;
```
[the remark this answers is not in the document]
itsalaidbacklife: @Haviles04 @seriouslysean what do you think is clearer?

itsalaidbacklife (commented):
One wondering

api/controllers/game/rematch.js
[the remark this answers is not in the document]
Haviles04: I kinda like how after the .find P0, and P1 are no longer used and the rest of it uses the same objects instead of jumping back and forth, but I think either way is pretty clear, and am happy with either

api/controllers/game/rematch.js
```diff
@@ -40,14 +40,38 @@ module.exports = async function (req, res) {
...
+      }
+      await getRematchGames(gameToAdd.id);
+    };
+    await getRematchGames(newGame.id);
+    const seriesP0 = [p0, p1].find(({ id }) => id === rematchGames[0].p0);
+    const seriesP1 = [p0, p1].find(({ id }) => id === rematchGames[0].p1);
+    //Get rematchGame win counts
+    let player0Wins = rematchGames.filter(({winner}) => winner === seriesP0.id).length;
+    let player1wins = rematchGames.filter(({winner}) => winner === seriesP1.id).length;
+    let stalemates = rematchGames.filter(({winner}) => winner === null).length;
```
itsalaidbacklife, line 70:
```suggestion
    const player0Wins = rematchGames.filter(({winner}) => winner === seriesP0.id).length;
    const player1wins = rematchGames.filter(({winner}) => winner === seriesP1.id).length;
    const stalemates = rematchGames.filter(({winner}) => winner === null).length;
```

itsalaidbacklife (approved):
Excellent work. Really cool QoL improvement

Issue #503, opened 2023-08-12
[Bug]: Spectate endpoint server 500 response


Haviles04:
### Bug Summary

The spectate endpoint currently throws a 500 server error while running test in home.spec.js, this should be changed to a  400 error.

### Detailed Description

Currently if you run test in `home.spec.js` the test starting on line 215  `it('Does not show open or completed games in spectate tab', () => {}` will pass, but your server console will log a 500 error 
```
error: Sending 500 ("Server Error") response:
 TypeError: Cannot read properties of undefined (reading 'id')
    at Object.populateGame (cuttle/api/services/gameService.js:10054)
    at async Promise.all (index 0)
    at async module.exports [as game/spectate] 
```
This is happening after we force disconnect the socket and try to rejoin the game. We should handle this case and throw a `res.badRequest()` resulting in a 400s error code.

### Current Behavior

Error: 500 server error

### Expected Behavior

400 error

### Steps to Reproduce

1.
2.
3.

arnaut08:
![image](https://github.com/cuttle-cards/cuttle/assets/52157140/361d1f0d-c1c6-4915-8d9b-c6883b6667b8)

I noticed this today while trying to test my changes. Please correct me if I'm wrong, but from what I can tell is that we already have a condition to handle this and apparently the players property of game is an empty array in such case and not any false-y value. We can probably fix the condition to check for array length instead.

Haviles04:
> 
> ![image](https://github.com/cuttle-cards/cuttle/assets/52157140/361d1f0d-c1c6-4915-8d9b-c6883b6667b8)
> 
> I noticed this today while trying to test my changes. Please correct me if I'm wrong, but from what I can tell is that we already have a condition to handle this and apparently the players property of game is an empty array in such case and not any false-y value. We can probably fix the condition to check for array length instead.
> 

This check happens after we fetched a game, given the Id was provided, this issue should address the API not receiving an ID, which would happen at the start of the file.

arnaut08:
I tried to add logs to debug this further. The gameId is there in this test case and that's how the game data is also being fetched successfully right before the condition which I mentioned earlier.
Also, I can see that we already have a check for gameId in the beginning of this method. 
![image](https://github.com/cuttle-cards/cuttle/assets/52157140/6a40d252-ae78-4f3f-86a1-2f31e202ac6c)

Haviles04:
[...]

PR #849, opened 2023-12-19
Bug/503

fuveluck:
#503 
Relevant [503](https://github.com/cuttle-cards/cuttle/issues/503)

--Do the tests still pass? (cypress tests were passed)

I hope its ok if I just return status 400 endpoint but I ve not got 500 Error anymore.

### DIFF
--- a/api/services/gameService.js
+++ b/api/services/gameService.js
@@ -73,15 +73,8 @@ module.exports = {
 
     if (game.players) {
       if (game.players.length < 2) {
-        /*no game was found*/
-        if(game.players.length == 0){
-          return { status: 400, message: 'Bad Request: Cannot find a game' };
-        }
-        else{
-          throw new Error({ message: 'Cannot populate game without two players' });
-        }
+        throw new Error({ message: 'Cannot populate game without two players' });
       } 
-
     } else {
       throw new Error({ message: 'Cannot populate game, because it does not have players collection' });
     }
--- a/api/services/gameService.js
+++ b/api/services/gameService.js
@@ -74,7 +74,7 @@ module.exports = {
     if (game.players) {
       if (game.players.length < 2) {
         throw new Error({ message: 'Cannot populate game without two players' });
-      } 
+      }
     } else {
       throw new Error({ message: 'Cannot populate game, because it does not have players collection' });
     }
--- a/api/services/gameService.js
+++ b/api/services/gameService.js
@@ -76,10 +76,14 @@ module.exports = {
       if (game.players.length < 2) {
         /*no game was found*/
         if(game.players.length == 0){
-          return res.status(400).send('Bad Request: Specific condition encountered');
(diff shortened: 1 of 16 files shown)


fuveluck:
I hope I did everything right )

api/services/gameService.js
```diff
@@ -73,8 +73,15 @@ module.exports = {
...
-      }
+        /*no game was found*/
+        if(game.players.length == 0){
+          return { status: 400, message: 'Bad Request: Cannot find a game' };
+        }
+        else{
+          throw new Error({ message: 'Cannot populate game without two players' });
+        }
+      } 
+
```
itsalaidbacklife, line 84: This is the right idea, but I think we should instead handle this in the consuming endpoint itself. The error from this issue is specifically triggered by a request to spectate using the /game/spectate endpoint which is managed by the `api/controllers/game/spectate.js` file. I think what we want to do is basically update that file so that its initial call to `gameService.populateGame()` and essentially the entire method is wrapped in the existing `try/catch`, that way this error is caught and transformed to a badRequest response, which that endpoint already does.

api/services/gameService.js
```diff
@@ -73,8 +73,15 @@ module.exports = {
 
     if (game.players) {
       if (game.players.length < 2) {
-        throw new Error({ message: 'Cannot populate game without two players' });
-      }
+        /*no game was found*/
+        if(game.players.length == 0){
+          return { status: 400, message: 'Bad Request: Cannot find a game' };
```
itsalaidbacklife, line 78: The reason that handling the issue in this specific way here is awkward is that this a helper method that isn't directly in controller code and doesn't have access to the `response` object, meaning whatever this returns doesn't inherently correspond to a specific response that ultimately gets sent to the client. So returning a differently shaped object here (rather than the Game that callers of this helper are expecting) will potentially goof the other places that use this method and it won't force them to return a `res.badRequest()`. That's why (per the comment below) we want to handle this change in the `spectate` endpoint file, rather than in this `gameService` helper.

itsalaidbacklife (changes requested):
Thanks for your contribution! This gets at the right error, but to really solve the problem we should instead make the fix in `api/controllers/game/spectate.js`. See my comments for details.

fuveluck:
@itsalaidbacklife
If I understood it right I just need start _try_ in moment when a spectator is being added to the game and then if it catch an error it just return badRequest instead of 500. I rollback gameServices as well.

If you can please run tests because I still have some problems with cypress it sometimes return socket disconnected error (there was few times when tests were passed) and I not see 500 endpoint error in terminal anymore.

BTW Wish you more luck and happiness in New Year 🎉

itsalaidbacklife (approved):
Nice work: remove the unnecessary whitespace and this will be good to go

itsalaidbacklife (approved):
Nicely done!

PR #852, opened 2023-12-31
Feat/rematch redesign

### LINKED ISSUE #784: [Feature]: Add a dialog in RematchView when opponent declines rematch
### Feature Summary

Add a dialog in RematchView when opponent declines rematch

### Detailed Description

Depends on https://github.com/cuttle-cards/cuttle/pull/448

After game over, the players have the option to jump into a new game immediately by clicking the "Rematch" button on the gameover dialog. Players can also decline a rematch by clicking the "go home" button.

Once the button is clicked, the page redirects to a "rematch view", and waits for response from the opponent.

If the player is on the game over dialog, and the opponent declines the rematch, the dialog shows a text saying "opponent has declined rematch".

This doesn't exist on rematch view yet, and is what this feature request is for.

### LINKED ISSUE #786: [Bug]: Two players hitting 'rematch' at the same time can cause a race condition where the new game might fail to initialize
### Bug Summary

When one player requests a rematch we record their decision. When the second player requests a rematch, we start the new game. This causes a race condition if both players request a rematch at once.

### Detailed Description

When a game ends, both players are presented the option to play another game (a rematch).

As a follow up to #783 we should apply the `lock-game` and `unlock-game` helpers to the `rematch` and `join-rematch` endpoints so that simultaneous requests are forcibly processed in series rather than in parallel. This will ensure that only one request to rematch is processed at a time, ensuring that the game properly initializes.

### Current Behavior

Both players hitting rematch at the same time can cause the game to fail to start

### Expected Behavior

The game starts after both players hit rematch, even if both hit the button at the same time

### Steps to Reproduce

1. Play a game as two players
2. Have both players hit 'rematch' at the same time
3. New game may fail to start

### LINKED ISSUE #810: [Bug]: The first player to request a rematch sees the old gamestate flicker when the rematch game starts.
### Bug Summary

Players can request a rematch from their opponent from the GameOverDialog. The first player to do so is brought to the RematchView while they wait for their opponent to decide whether to play the rematch. When the game starts, the player who went first sees the state of their previous game before the game state updates to the new game.

### Detailed Description

We should examine the control flow of what happens when the first user hits the Rematch button. In particular we should consider whether to reset the game store after entering the RematchView (either via the created()` hook of the `RematchView` or via the function that navigates the user *into* the RematchView. 

If it turns out that data in the game store is actually necessary in order for the user to be correctly directed to the GameView when the game starts, we could instead look to ensure that the order of events makes it so that the game store is updated with the new game’s data before the user is redirected to the GameView from the RematchView.

### Current Behavior

The first player to join a rematch sees the old game flicker before the game starts.

### Expected Behavior

The first player to join the rematch smoothly sees the new game start without traces of the old one flickering.

### Steps to Reproduce

1. Sign up as two players in two browsers 
2. Create a game
3. Join it and ready up as both players (starting the game)
4. End the game e.g. by conceding
5. Hit rematch as one player (let’s call this user player1)
6. Hit rematch as the other player (player2)
7. Look at player1’s screen and see that the cards initially appear in the previous positions from the game that just finished before showing the newly dealt cards for the rematch game

### LINKED ISSUE #850: [Feature]: Implement redesigned rematch flow
### Feature Summary

We have designed an improved UX for the GameOver dialog and the 'rematch' flow

### Detailed Description

Here is the [design file](https://www.figma.com/file/5QrqqPdN8VwRExxl5K5HRt/Cuttle_Design_System-EDITABLE?type=design&node-id=458-5247&mode=design&t=Onx6bxDdxEGwoSoc-0). 

We should revise the `GameOverDialog` as designed so that it functions as a mini-lobby, where both players remain until they both agree to continue their match (or they go home).

itsalaidbacklife:
Reworks the "Rematch" flow from the GameOver screen by implementing the designs shown in #850.

## Issue number

- Resolves #850 
- Resolves #810 
- Resolves #786 
- Closes #784 by making it irrelevant 

## Please check the following

- For New Features:


tests/e2e/specs/in-game/gameOver.spec.js
itsalaidbacklife, line 1: This file was over 1k lines so I broke it up into the new `/tests/e2e/specs/in-game/game-over` folder and split it into 3 files

types/SocketEvent.js
```diff
@@ -29,5 +29,4 @@ export default {
   UPDATED: 'updated',
   REMATCH: 'rematch',
   NEW_GAME_FOR_REMATCH: 'newGameForRematch',
-  JOIN_REMATCH: 'joinRematch',
```
itsalaidbacklife, line 32: This isn't necessary as a socket event. It represents when a player joins a rematch game after it's created i.e. subscribes to it for updates, but there is nothing that other users need to do to respond to the event, so we don't need it.

It was previously used to handle spectators subscribing to the game, but instead they now respond to the `NEW_GAME_FOR_REMATCH` event just like the players do

src/routes/game/components/dialogs/components/gameOver/GameOverDialog.vue
```diff
@@ -1,66 +1,47 @@
...
+            :won-match="opponentWinsMatch"
+            data-cy="opponent-match-result"
+          />
+        </template>
+        <template v-else>
+          <LobbyPlayerIndicator
+            :player-username="gameStore.player.username"
+            :player-ready="gameStore.iWantRematch"
+            small
+            data-cy="my-rematch-indicator"
```
itsalaidbacklife, line 34: Todo: rename and relocate this component to be generic, now that it's used outside the lobby

src/plugins/sockets/inGameEvents.js
```diff
@@ -125,57 +125,33 @@ export async function handleInGameEvents(evData) {
...
+        return;
       }
-      return;
-    case SocketEvent.NEW_GAME_FOR_REMATCH: {
       gameStore.setRematchGameId(evData.gameId);
+      const { gameId: oldGameId } = currentRoute.params;
 
       if (currentRoute.name === ROUTE_NAME_SPECTATE) {
+        await gameStore.requestSpectate(evData.gameId);
         gameStore.updateGame(evData.newGame);
```
Haviles04, line 142: `requestSpectate()` calls `updateGame()` inside of its response handlers so this is probably uneeded

src/routes/game/components/dialogs/components/gameOver/components/MatchStatusBanner.vue
```diff
@@ -0,0 +1,110 @@
...
+</template>
+
+<script setup>
+import { computed } from 'vue';
+import { useGameStore } from '@/stores/game.js';
+
+const gameStore = useGameStore();
+
+const isRanked = computed(() => gameStore.isRanked);
+const isSpectating = computed(() => gameStore.isSpectating);
```
Haviles04, line 30: I wonder if it makes sense to use pinias built in storeToRefs()

Haviles04: https://pinia.vuejs.org/core-concepts/

src/plugins/sockets/inGameEvents.js
```diff
@@ -125,57 +126,47 @@ export async function handleInGameEvents(evData) {
...
+      gameStore.p0Rematch = true;
+      gameStore.p1Rematch = true;
+      gameStore.rematchGameId = evData.gameId;
+      // stop early if spectating and haven't yet chosen to continue spectating
+      if (gameStore.isSpectating && !gameStore.iWantToContinueSpectating) {
+        return;
+      }
+
+      // wait for card flip animations
+      await sleep(500);
```
seriouslysean, line 147: This is only used here, I'd argue you don't need a utility and instead should inline the promise check. It would be better to file a follow up to use the vue transition component and the hooks it provides instead of a manual wait though, see https://stackblitz.com/edit/vue-3-card-flip-animation?file=src%2FApp.vue,src%2Fcomponents%2FCard.vue.

src/plugins/sockets/inGameEvents.js
```diff
@@ -125,57 +126,47 @@ export async function handleInGameEvents(evData) {
...
-      }
+
+      router.push({
+          name: currentRoute.name,
+          params: {
+            gameId: evData.gameId,
+          },
+        });
+      
+      gameStore.iWantToContinueSpectating = false;
```
seriouslysean, line 167: This name is rough.

src/plugins/sockets/inGameEvents.js
```diff
@@ -184,7 +175,10 @@ export async function handleInGameEvents(evData) {
       gameStore.resetPNumIfNullThenUpdateGame(evData.game);
       break;
     case SocketEvent.REQUEST_STALEMATE:
-      if (evData.requestedByPNum !== gameStore.myPNum && !evData.victory.gameOver) {
+      if (evData.victory.gameOver) {
+        gameStore.consideringOpponentStalemateRequest = false;
+      }
+      else if (evData.requestedByPNum !== gameStore.myPNum) {
         gameStore.consideringOpponentStalemateRequest = true;
       }
```
seriouslysean, line 183: Seems like this might be able to be simplified to something like this, but I don't understand the full chain to say for sure.
```suggestion
      gameStore.consideringOpponentStalemateRequest = evData.requestedByPNum !== gameStore.myPNum || !evData.victory.gameOver;
```
It might be better to organize how we decide that to make sure it makes sense. Little hard for me to follow.

api/controllers/game/rematch.js
```diff
@@ -21,17 +26,16 @@ module.exports = async function (req, res) {
 
     const { p0: newP1Id, p1: newP0Id } = updatedGame;
 
-    Game.publish([game.id], {
```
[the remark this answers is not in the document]
itsalaidbacklife: The endpoint should only emit a single socket event. Previously it would emit one to say that a user had opted for/against a rematch, then another one when the game started (if both users said yes). But that creates a race condition between the two socket events triggered by the same endpoint. 

This way there is just one socket event that fires for any given code path through this endpoint, and it's either `'rematch'` which means we're just updating a rematch card, or it's `'newGameForRematch'`, which means the game started. When it's the new game, the client handles the animation for flipping the card before processing the new game and navigating to the new game url

api/controllers/game/rematch.js
```diff
@@ -79,7 +83,8 @@ module.exports = async function (req, res) {
     Game.publish([game.id], {
       change: 'newGameForRematch',
       game: updatedGame,
-      pNum,
+      pNum: oldPNum,
+      oldGameId,
```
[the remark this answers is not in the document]
itsalaidbacklife: It's used to compare if the user's current url is pointing towards the old game in which case we should navigate to the new one. Currently in prod we have but #867 which is essentially the client failing to filter out socket events for games you aren't watching.

Now in `inGameEvents` we compare the `oldGameId` to the current url for this event and only navigate to the new game id + subscribe if the current url has the id of the oldGame sent in this event

api/controllers/game/stalemate.js
```diff
@@ -6,7 +6,8 @@ module.exports = async function (req, res) {
   try {
     const { game: gameId, pNum, usr: userId } = req.session;
     // Track which turn each player most recently requested stalemate
-    const game = await Game.findOne({ id: gameId }).populate('players');
+    const game = await Game.findOne({ id: gameId })
+      .populate('players', { sort: 'pNum' });
```
itsalaidbacklife, line 10: Adding this sort fixes #864

api/helpers/add-game-to-match.js
```diff
@@ -16,7 +16,8 @@ module.exports = {
   fn: async ({ game }, exits) => {
     try {
       if (!game.isRanked) {
-        return exits.success();
+        const match = sails.helpers.mockCasualMatch(game);
```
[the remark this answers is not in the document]
itsalaidbacklife: Ultimately I'd like to rework the db so that casual games have a real match as well. But the idea here is to keep score during rematch games even for casual games like this
image

Which for consistency's sake we use the same data structure that the ranked matches do: an object w/ a list of the games along w/ some metadata about who's playing etc

api/helpers/mock-casual-match.js
```diff
@@ -0,0 +1,40 @@
+async function getRematchGames (game, priorRematchGames = []) {
+  const rematchGames = [game, ...priorRematchGames];
+  const gameToAdd = await Game.findOne({ rematchGame: game.id });
+  if (!gameToAdd) {
+    return rematchGames;
+  }
+  if (!gameToAdd.rematchGame) {
+    return rematchGames;
+  }
```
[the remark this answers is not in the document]
itsalaidbacklife: Yeah I can't see how those wouldn't be equivalent. More concise; definitely better

src/routes/game/components/dialogs/components/CannotCounterDialog.vue
itsalaidbacklife, line 1: Created a new /src/routes/game/components/dialogs folder to group the dialogs, largely because I wanted to organize subcomponents. So now we have a /components subdirectory whenever there are child components. It's analogous to what we have with /routes where there is the parent component file and then a /components directory with all its direct child components
It looks like this
```
src
  routes
    game
      components
        GameView.vue
        dialogs
          GameDialogs.vue
          components
            CannotCounterDialog.vue
            ThreeDialog
            // etc the other in-game dialogs (children of GameDialogs)
            // GameOverDialog gets its own directory b/c it has sub components
            gameOver
              GameOverDialog.vue
              // child components for GameOverDialog
              components
                MatchScoreCounter.vue
                MatchStatusBanner.vue
                MatchWonOrLostIndicator.vue
```

itsalaidbacklife (commented):
Initial comment responses

src/routes/game/components/dialogs/GameDialogs.vue
```diff
@@ -46,14 +46,14 @@
...
 import { useAuthStore } from '@/stores/auth';
-import CannotCounterDialog from '@/routes/game/components/CannotCounterDialog.vue';
-import CounterDialog from '@/routes/game/components/CounterDialog.vue';
-import FourDialog from '@/routes/game/components/FourDialog.vue';
-import GameOverDialog from '@/routes/game/components/GameOverDialog.vue';
-import ReauthenticateDialog from '@/routes/game/components/ReauthenticateDialog.vue';
-import SevenDoubleJacksDialog from '@/routes/game/components/SevenDoubleJacksDialog.vue';
-import ThreeDialog from '@/routes/game/components/ThreeDialog.vue';
-import OpponentRequestedStalemateDialog from '@/routes/game/components/OpponentRequestedStalemateDialog.vue';
+import CannotCounterDialog from './components/CannotCounterDialog.vue';
```
seriouslysean, line 49: We really should be using the `@` syntax as much as we can. If you need them shorter you could make an `gameComponents` alias or something.

src/routes/game/components/dialogs/components/gameOver/GameOverDialog.vue
```diff
@@ -0,0 +1,346 @@
...
+      return this.$vuetify.display.xs;
+    },
+    currentMatch() {
+      return this.gameStore.currentMatch;
+    },
+    matchGameStats() {
+      const currentMatchGames = this.gameStore.currentMatch?.games ?? [];
+      return currentMatchGames.map((game) => {
+        if (game.status === GameStatus.FINISHED){
+          return game.winner === null ? 'D' : game.winner === this.gameStore.opponent.id ? 'L' : 'W';
```
seriouslysean, line 234: I don't really like nested ternaries, I would switch this up.

```
return currentMatchGames.map((game) => {
  if (game.status !== GameStatus.FINISHED){
    return 'I';
  }
  if (game.winner === null) {
    return 'D';
  }

  if (game.winner === this.gameStore.opponent.id) {
    return 'L'
  }
  
  return 'W';
});
```

If you do this in multiple places convert it to a utility or something and reuse it.

src/routes/game/components/dialogs/components/gameOver/GameOverDialog.vue
```diff
@@ -0,0 +1,346 @@
...
+        }
+        return 'I';
+      });
+    },
+    stalemates() {
+      return this.gameStore.currentMatch?.games
+        .filter((game) => game.status === GameStatus.FINISHED && game.winner === null).length;
+    },
```
seriouslysean, line 242: Does this need to be a getter in the store?

src/stores/game.js
```diff
@@ -11,7 +11,7 @@ function queenCount(player) {
   if (!player) {
     return null;
   }
-  return player.faceCards.reduce((queenCount, card) => queenCount + (card.rank === 12 ? 1 : 0), 0);
+  return player?.faceCards.reduce((queenCount, card) => queenCount + (card.rank === 12 ? 1 : 0), 0) ?? 0;
```
seriouslysean, line 14: It might be worth doing the player check first instead of combining it all for clarity. Easy to lose when trailing the reducer.

```
if (!player) {
    return 0;
}

return player.faceCards.reduce((queenCount, card) => queenCount + (card.rank === 12 ? 1 : 0), 0);
```

src/stores/game.js
```diff
@@ -731,13 +750,15 @@ export const useGameStore = defineStore('game', {
     async requestStalemate() {
       return new Promise((resolve, reject) => {
         io.socket.get('/game/stalemate', (res, jwres) => {
+          this.consideringOpponentStalemateRequest = false;
```
seriouslysean, line 753: I know why you're setting this in both cases, feels like an easy thing to lose sight of though.

seriouslysean (approved):
I didn't test the functionality due to time, so make sure to do additional regression on your end. No real blockers, but some questions and suggestions.

src/routes/game/components/dialogs/components/gameOver/GameOverDialog.vue
itsalaidbacklife: 1. I think you're right that it reads better if we remove the nested ternary. 
2. I think a switch statement would be nice since we have multiple comparisons to the same value
3. Most importantly, looking again, this is unused lol. I'm ditching it.

But if we _did_ want this, I'd think style wise this would be the sweet spot:

```
return currentMatchGames.map((game) => {
  if (game.status !== GameStatus.FINISHED){
    return 'I';
  }

  switch(game.winner) {
    case this.gameStore.player.id:
      return 'W';
    case this.gameStore.opponent.id:
      return 'L';
    default:
      return 'D';
  }
});
```

src/routes/game/components/dialogs/components/gameOver/GameOverDialog.vue
itsalaidbacklife: It's in a funky spot b/c it feels like something that does belong in the store (because it a value computed from the store state that seems logically like it could be used elsewhere), but it's actually only consumed in the GameOverDialog.

My inclination is to leave it here because the store is already a problematically large file so if something isn't needed as a re-used utility, then we can keep it in the consuming component (especially if it's anything smaller than GameView)

src/routes/game/components/dialogs/GameDialogs.vue
itsalaidbacklife: So counterpoint: I've reorganized the components directory to better represent parent/child relationships. I see this as an extension of how we already colocated page files with their constituent components. This basically means that for any given component, its children will always be in `./components`. 

I think that actually makes it more intuitive to find what you're looking for, because everything is organized from the page level down and any given thing should always be able to find it's direct children in `./components`. I don't have strong feelings about it, but I do like the look of everything getting to use ./components as its succinct, logical, and gels with the possibility of having ./composables, etc

api/controllers/game/rematch.js
```diff
@@ -1,89 +1,94 @@
...
+    // Determine who was p0 and p1 in first game in the series
+    const [ firstGame ] = rematchGames;
+    const seriesP0Id = firstGame.p0;
+    const seriesP1Id = firstGame.p1;
+    const seriesP0Username = players.find((player) => player.id === seriesP0Id);
+    const seriesP1Username = players.find((player) => player.id === seriesP1Id);
+    // Get rematchGame win counts
+    const player0Wins = rematchGames.filter(({winner}) => winner === seriesP0Id).length;
+    const player1wins = rematchGames.filter(({winner}) => winner === seriesP1Id).length;
+    const stalemates = rematchGames.filter(({winner}) => winner === null).length;
```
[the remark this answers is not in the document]
itsalaidbacklife: In practice I think `!winner` should be fine. Like technically we are _only_ expecting `null`, there should never be a user with id 0, so in one sense it's more explicitly careful to only count things that have `null`, but if it reads easier then I think it's fine to go the shorter route

api/controllers/game/rematch.js
```diff
@@ -1,89 +1,94 @@
...
+
+    // Set game name and isRanked
+    const newName = `${seriesP0Username} VS ${seriesP1Username} ${player0Wins}-${player1wins}-${stalemates}`;
```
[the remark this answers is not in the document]
itsalaidbacklife: I don't think translations necessarily play a part in this issue. There's funkary around creating the string perhaps. It's an interesting challenge when it comes to anything that lives in the db. There's only one game name, so there's no room to translate it differently for different users. 

Curious to unpack this later

src/stores/game.js
seriouslysean: lol yep

api/controllers/game/rematch.js
```diff
@@ -1,89 +1,94 @@
...
-    await Game.updateOne({ id: game.id }).set(gameUpdates);
-
-    // In case other player updated the game at the same time
-    const updatedGame = await Game.findOne({ id: game.id }).populate('players');
-
-    const currentMatch = await Match.findOne({ id: updatedGame.match });
-    const shouldNewGameBeRanked = currentMatch?.winner ? false : updatedGame.isRanked;
-
-    const { p0: newP1Id, p1: newP0Id } = updatedGame;
+    let game =  await sails.helpers.lockGame(req.session.game);
```
itsalaidbacklife, line 15: We know lock and unlock the requested game to prevent race conditions around which user is the 2nd one to request a rematch (and therefore to start the new game)

src/routes/game/components/dialogs/components/gameOver/GameOverDialog.vue
seriouslysean: Makes sense, I agree. I didn’t have full context so that tracks.

src/routes/game/components/dialogs/GameDialogs.vue
seriouslysean: I’m certainly not going to push the point, I just never really like mixing those styles but I agree that it’s pretty nice for parent/sibling refs.

api/controllers/game/rematch.js
```diff
@@ -4,10 +4,15 @@
...
+    const { usr: userId } = req.session;
+    const { gameId: oldGameId, rematch } = req.body;
 
-    const game = await Game.findOne({ id: gameId }).populate('players');
+    const game = await Game.findOne({ id: oldGameId }).populate('players');
+
+    // Early return if requesting user was not in the game
+    if (game.p0 !== userId && game.p1 !== userId) {
+      return;
+    }
```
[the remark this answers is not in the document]
itsalaidbacklife: done

src/plugins/sockets/inGameEvents.js
```diff
@@ -125,57 +126,47 @@ export async function handleInGameEvents(evData) {
...
+        Number(urlGameId) !== evData.oldGameId || 
+        ![ROUTE_NAME_GAME, ROUTE_NAME_SPECTATE].includes(currentRoute.name)
       ) {
-        router.push({
-          name: ROUTE_NAME_REMATCH,
-          params: {
-            gameId: gameStore.id,
-          },
-        });
+        return;
```
itsalaidbacklife, line 136: This new early return fixes #867

src/plugins/sockets/inGameEvents.js
itsalaidbacklife: Yeah let's see. The way I was thinking about it is "I'm not considering my opponent's request anymore if the game is over b/c that means we both answered already. And otherwise, I'm it's my turn to accept/reject the stalemate if it the request was triggered by my opponent.

itsalaidbacklife: Oh actually we can swap the || for && I think and it should work
```suggestion
      // Show OpponentRequestedStalemateDialog if opponent requested stalemate
      //   and game is not yet over
      gameStore.consideringOpponentStalemateRequest = !evData.victory.gameOver && evData.requestedByPNum !== gameStore.myPNum;
```

src/plugins/sockets/inGameEvents.js
itsalaidbacklife: Yeahhhh. Open to suggestions 😅

api/helpers/mock-casual-match.js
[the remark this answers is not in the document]
itsalaidbacklife: Actually it turns out we don't care if the new game has a rematchGame attribute because we're recursing _backward_ through time. We care if there is a game that _this game_ is a rematch _of._ So we only need the `if(!gameToAdd)` without checking `rematchGame` at all

api/helpers/mock-casual-match.js
```diff
@@ -0,0 +1,40 @@
...
+  return getRematchGames(gameToAdd, rematchGames);
+}
+
+module.exports = {
+  friendlyName: 'Mock Casual Match',
+
+  description:
```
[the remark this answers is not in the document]
itsalaidbacklife: As I think on it, I think we can sidestep calling it `mock` or `adapted` or `from rematch`, and just say `getCasualMatch()`. The fact that it's not a Match in the db is kindof an implementation detail that callers of this method don't necessarily need to contend with.

api/helpers/add-game-to-match.js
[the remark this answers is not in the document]
itsalaidbacklife: Yeah exactly. Upon further thought and based on your suggestion, I think we can just pretend it's a real match and use the name `getCasualMatch()`

src/routes/game/components/dialogs/components/gameOver/components/MatchStatusBanner.vue
itsalaidbacklife: This is intriguing and something for us to look at globally

src/plugins/sockets/inGameEvents.js
itsalaidbacklife: Per our discord convo, we'll keep this for the moment to avoid indenting these. We should roll out `sleep()` across the existing setTimeout use cases then rework transitions across the board to properly utilize `Transition` components

src/routes/game/components/dialogs/GameDialogs.vue
itsalaidbacklife: good enough for now

src/routes/game/components/dialogs/GameDialogs.vue
```diff
@@ -120,13 +120,8 @@ export default {
...
     },
-    consideringOpponentStalemateRequest: {
-      get() {
-        return this.gameStore.consideringOpponentStalemateRequest;
-      },
-      set(val) {
-        this.gameStore.setConsideringOpponentStalemateRequest(val);
-      },
+    consideringOpponentStalemateRequest() {
+      return this.gameStore.consideringOpponentStalemateRequest;
```
itsalaidbacklife, line 124: Removed setter because the store now fully controls this variable like for the other dialogs, so we only read here. Store sets it based on socket events

api/helpers/get-rematch-games.js
itsalaidbacklife, line 1: @Haviles04 I'm pleased with how this helper turned out. Great work laying the foundations of this logic. I think the end result is clear and reusable

PR #857, opened 2024-01-14
Only Make text selectable only in History pane & Rules dialog

### LINKED ISSUE #856: [Feature]: Allow user to not click deck after down-click (Firefox)
### Feature Summary

Allow user to not click deck after down-click 

### Detailed Description

Currently on desktop, if you down click the deck to draw a card, you cannot move the cursor away from the deck to stop drawing the card. 

This does not seem to be an issue when playing on mobile or on Safari, but was observed on Firefox for desktop 
[Screencast from 2024-01-13 22-46-30.webm](https://github.com/cuttle-cards/cuttle/assets/10717998/7326f6ea-b66a-4692-89af-423b4caf8d96)

I'm not sure if this applies to any other browsers since most users said it was working fine for them.

meichthys:
## Issue number

- Resolves #856

## Please check the following

- For New Features:

### DIFF
--- a/src/routes/game/GameView.vue
+++ b/src/routes/game/GameView.vue
@@ -1176,8 +1176,12 @@ export default {
 // Styles //
 ////////////
 
-#body {
+body {
   user-select: none;
+}
+
+#rulesDialog {
+  user-select: text;
 }
 
 #game-view-wrapper {


itsalaidbacklife:
Is this about #856? It looks like this is a separate issue, but I don't fully follow the specific problem that's addressed by preventing users from selecting other text outside the history.

What is this aiming to solve?

itsalaidbacklife:
> @itsalaidbacklife Yes, this is for #856 (https://github.com/cuttle-cards/cuttle/assets/10717998/7326f6ea-b66a-4692-89af-423b4caf8d96) The title was misleading on this pull request. I've updated it to be more exact. This issue is aiming to solve the issue described in #856 (Dragging off the deck after down-clicking should not draw a card). My solution was to make no text selectable by preventing the user from selecting text in the body element `user-select: none`, then only allowing selectable text in the history area and the rules dialog with `user-select: text`

Ah okay, understood. I didn't initially see the connection between whether the click was cancellable and the selectability of text on the page; thanks for explaining. If this is specifically an issue with the clicking the deck to draw, I think it would make more sense if we limited the scope of the style to the deck itself. 

Does it work if you put `user-select: none` in the existing `& #deck {`  style block around line 1270?

seriouslysean (changes requested):
While this may fix the issue, this is not the solution and is an accessibility nightmare. Look in to reproducing with the vcard element and seeing if anything there can help (props, config settings, etc).  I added some additional info here, too, https://github.com/cuttle-cards/cuttle/issues/856#issuecomment-1900569783.

PR #858, opened 2024-01-14
[Feature]: Add reverse water effect in the lobby cards #792

### LINKED ISSUE #792: [Feature]: Add reverse water effect in the lobby cards
### Feature Summary

When both players are ready we should see an animation for the cards to indicate the players are leaving the lobby

### Detailed Description

Right now, when both players are ready, we jump to the game, and there is no indication that they are leaving the lobby.

This is just one part of the solution; we should add an exit animation to the cards, complementing the entry lobby animation.

The entry animation is a water-draining effect; we could do a water-draining effect for the exit animation. This animation would play when both players are ready; after it is finished, we should redirect to the game.

This is the second step in a four step upgrade to the "start game" animation bundle. First step:  #791

Rafa-romero-dev:
## Issue number

- Resolves #792

## Please check the following

- For New Features:

I tested with desktop and mobile resolutions and it looks ok, if someone finds a problem with responsiveness, let me know.

This is the exit animation for the cards on lobby, it lasts 3s but the route transition happens in 2s. 

https://github.com/cuttle-cards/cuttle/assets/152430863/b99a805b-142b-4dbf-8212-6b55e78897af

This animation runs when both players are ready and want to move to the /game route. It shouldn't play in any other case.


src/routes/lobby/components/LobbyPlayerIndicator.vue
```diff
@@ -154,13 +158,32 @@ export default {
...
+.reverseWater {
+  width: 3600px;
+  height: 800px;
+  background-image: url('/img/waves.svg');
+  background-repeat: no-repeat;
+  background-size: cover;
+  position: relative;
+  animation: waving-reverse 3s;
+  animation-fill-mode: forwards;
+  z-index: 1;
```
itsalaidbacklife, line 170: In order to DRY this up, how about pulling everything except the animation itself into a style block shared by `.water` and `.reverseWater` since the rest is the same. We could either do that by introducing a new class shared in both cases, or by adding a style block with `.water, .reverseWater`

itsalaidbacklife (approved):
This looks awesome. Can you combine the style blocks of water and reverse water for the shared styles?

src/routes/lobby/components/LobbyPlayerIndicator.vue
Rafa-romero-dev: That was actually my mistake, the `reverseWater` class is only there to trigger the animation! Because is the same component that has the `water` class properties. I will upload the fix right away

src/routes/lobby/components/LobbyPlayerIndicator.vue
```diff
@@ -149,9 +153,12 @@ export default {
   background-repeat: no-repeat;
   background-size: cover;
   position: relative;
-  animation: waving 5s;
-  animation-fill-mode: forwards;
   z-index: 1;
+  animation: waving 5s forwards;
+}
+
+.reverseWater {
```
itsalaidbacklife, line 160: Does this work? I'm a little surprised that the specificity overrides the .water style block because they're both single class selectors. Maybe for explicitness' sake we should apply both here?
```suggestion
.water.reverseWater {
```

itsalaidbacklife (commented):
Looks good. One question about specificity of the .reverseWater selector

src/routes/lobby/components/LobbyPlayerIndicator.vue
Rafa-romero-dev: I think is just preference, you might see it like, "this class belongs to the water component" and the suggestion works as well, but both options are doing the exact same thing. 

They are removing the existing animation `waving` for the new one `waving-reverse`. Personally I feel is more confusing to re-apply the `water` styles with `reverseWater`, because what matter to us is to apply the new animation.  But that's how I see it based on my personal interpretation.

itsalaidbacklife: The thing that I don't like about the current way is that it is order-dependent. If `.water` style were moved below `.reverseWater` then `.water` would take precedence. At the cost of muddying the hierarchy, I'd rather ensure that the overriding style block has a higher-specificity selector to guarantee that it will apply.

Since it's a simple change, I'll commit it myself then merge

itsalaidbacklife (approved):
Well done; this looks great!

src/routes/lobby/components/LobbyPlayerIndicator.vue
Rafa-romero-dev: I totally understand your point and I'll take that into consideration moving forward. Thanks for taking the time on doing the change

PR #860, opened 2024-01-16
prevent ranked-switch text to wrap, media query and padding adjustment

### LINKED ISSUE #859: [Bug]: Rank-switch display bug
### Bug Summary

In the lobby, text in .rank-switch is displayed on two lines. 

### Detailed Description

In the lobby, the game type wraps between 960px and 979px of viewport width

Capture d'écran 2024-01-15 213901

### Current Behavior

Game type text wraps

### Expected Behavior

Game type text shouldn't wrap

### Steps to Reproduce

1. fix viewport with of window between 960 and 979 px.
2. game type is on two lines

Posoroko:
## 

- Resolves #859

- modified the media query so the text doesn't wrap between 966 and 980px
- added a media query to handle the same problem on  xxs screens ( < 350px)
- adjusted the padding on small screen so the element are displayed more evenly in the ranked-switch component.


Posoroko:
I feel he whole component could be styled in a much simpler way. Some simple flex-box would be enough to prevent the use of media queries.  But I'm not sure how customizable vuetify components are.

Posoroko:
@Haviles04 Probably.  I'm not used to styling vuetify components and how you guys handle it for cuttle.

Haviles04:
> @Haviles04 Probably.  I'm not used to styling vuetify components and how you guys handle it for cuttle.  

Just put `text-wrap: nowrap` in the .rank-switch main style

src/routes/lobby/LobbyView.vue
```diff
@@ -217,7 +217,9 @@ export default {
   padding: 0 64px;
   width: 50%;
 }
-
+.rank-switch * {
+  flex-shrink: 0;
+}
```
Haviles04, line 222: I think it's better to put `flex-shrink-0` as a vuetify utility class directly on the v-swtich

Haviles04 (changes requested):
Nice, this is a weird problem caused by Vuetify's styling. Nice catch though and good fix! The added media query for 350px definitely looks better to me! One small change requested to utilize Vuetifys utility classes

Haviles04 (approved):
Nice work @Posoroko!!

PR #872, opened 2024-01-25
Feat #674 join lobby by URL

### LINKED ISSUE #674: [Feature]: Enable joining a lobby by sharing a link to your current game

Haviles04:
## Issue number

- Resolves #674 

## Please check the following

- For New Features:

-Added logic to `router.js` to handle redirections or game subscribing


Haviles04:
I kinda of went off the mark of how the issue states a fix, let me know if I went too far off the deep end :sweat_smile:

src/router.js
```diff
@@ -37,6 +38,32 @@ const logoutAndRedirect = async (to, from, next) => {
   return next('/login');
 };
 
+const checkAndSubscribeToLobby = async (to, from, next) => {
+  const gameStore = useGameStore();
+  const authStore = useAuthStore();
+  const { gameId } = to.params;
+  
+  if (!authStore.authenticated) {
+    authStore.redirectGameId = +gameId;
```
seriouslysean, line 47: This doesn't feel like a great way to coerce this value imo as you may end up with `NaN` as a value with nonstandard input.
image

Haviles04: Should we use parseInt() then? I believe Number() will also give you NaN in that situation

seriouslysean: You might need to use something like this:
```
const gameId = parseInt(game.gameId);
if (!Number.isNaN(gameId) && Number.isFinite(gameId)) {
          // valid
}
```
Otherwise throw the error. Paraphrasing a bit here but use your best judgement. 👍🏼

src/components/LobbyErrorDialog.vue
```diff
@@ -0,0 +1,40 @@
...
+    data-cy="create-game-dialog"
+  >
+    <template #body>
+      <div>
+        <h4>Cannot Join Lobby {{ errorMessage }}</h4>
+      </div>
+    </template>
+    <template #actions>
+      <v-btn
+        data-cy="how-it-works-okay"
```
itsalaidbacklife, line 16: This data-cy looks inaccurate
```suggestion
        data-cy="lobby-error-go-home"
```

src/components/LobbyErrorDialog.vue
```diff
@@ -0,0 +1,40 @@
...
+const { showLobbyError } = defineProps({ showLobbyError: {type: Boolean, default: false } });
+const router = useRouter();
+const route = useRoute();
+
+const errorMessage = route.meta.error ? route.meta.error : 'Unknown Error';
+
+const close = () => {
+ router.push('/');
+};
+</script>
```
itsalaidbacklife, line 40:
```suggestion
</script>

```

src/components/LobbyErrorDialog.vue
```diff
@@ -0,0 +1,40 @@
...
+        Go Home
+      </v-btn>
+    </template>
+  </BaseDialog>
+</template>
+
+<script setup>
+import BaseDialog from '@/components/BaseDialog.vue';
+import { useRouter, useRoute } from 'vue-router';
+const { showLobbyError } = defineProps({ showLobbyError: {type: Boolean, default: false } });
```
itsalaidbacklife, line 31: This is interesting. I believe this is the deprecated [reactive props destructuring](https://vuejs.org/guide/extras/reactivity-transform#reactive-props-destructure) feature of vue, correct? This has been removed from vue as of 3.4 so I think we should replace it so that we refrain from destructuring. It looks like we don't need the props in js atm (just the template) so I think we'd want this
```suggestion
defineProps<{
  showLobbyError: {
    type: Boolean,
    default: false
  }
}>();
```

src/components/LobbyErrorDialog.vue
```diff
@@ -0,0 +1,40 @@
...
+  <BaseDialog
+    id="ErrorDialog"
+    :model-value="showLobbyError"
+    title="Oops!"
+    :opacity="1"
+    data-cy="create-game-dialog"
+  >
+    <template #body>
+      <div>
+        <h4>Cannot Join Lobby {{ errorMessage }}</h4>
```
itsalaidbacklife, line 11: Semantically this is more of a paragraph than a header, so I think it makes sense to put this in a `<p>` tag and style it however necessary to make it look right

src/router.js
```diff
@@ -35,6 +36,37 @@ const logoutAndRedirect = async (to, from, next) => {
...
+  
+  try {
+    await gameStore.requestSubscribe(+gameId);
+    next();
+  }
+  catch (err) {
+    to.meta.error = err.message;
+    next();
+  }
+};
```
itsalaidbacklife, line 67: I'm wondering how this gels with/conflicts with #874

src/components/LobbyErrorDialog.vue
itsalaidbacklife, line 11: Need to translate this

src/components/LobbyErrorDialog.vue
```diff
@@ -0,0 +1,40 @@
...
+    </template>
+    <template #actions>
+      <v-btn
+        data-cy="how-it-works-okay"
+        color="surface-1"
+        variant="flat"
+        class="text-surface-2"
+        @click="close"
+      >
+        Go Home
```
itsalaidbacklife, line 22: needs translation

src/routes/lobby/LobbyView.vue
```diff
@@ -142,6 +147,9 @@ export default {
   },
   computed: {
     ...mapStores(useGameStore, useAuthStore),
+    showLobbyError() {
+      return this.$route.meta.error ? true : false;
```
itsalaidbacklife, line 151: Thoughts on ? vs !!   ?
```suggestion
      return !!this.$route.meta.error; 
```

src/router.js
```diff
@@ -35,6 +36,37 @@ const logoutAndRedirect = async (to, from, next) => {
...
+  
+  if (Number.isNaN(gameId) || !Number.isFinite(gameId)) {
+    to.meta.error = 'Invalid lobby number';
+    return next();
+  }
+
+  if (!authStore.authenticated) {
+    authStore.redirectGameId = gameId;
+    return next('/login');
+  }
```
itsalaidbacklife, line 52: I think it might be better to use a url param instead of a store variable for the `redirectGameId` because that way we can refresh the page on the `LoginView` and still get back to your game after logging in

src/stores/auth.js
```diff
@@ -17,6 +17,7 @@ export const useAuthStore = defineStore('auth', {
     username: null,
     mustReauthenticate: false,
     isReturningUser: null,
+    redirectGameId: null,
```
itsalaidbacklife, line 20: As I said above, I think this data will be better served as a url param so that we can recover it on page refresh

itsalaidbacklife (commented):
Nice! This is exciting stuff. Questions and feedback are in the comments

src/routes/login/LoginView.vue
```diff
@@ -272,7 +272,11 @@ export default {
       this.username = '';
       this.pw = '';
       this.loading = false;
-      this.$router.push('/');
+      const {redirectId} = this.$route.params;
+      if (!redirectId) {
+        return this.$router.push('/');
+      }
+      return this.$router.push(`/lobby/${redirectId}`);
```
itsalaidbacklife, line 279: I think this is fine for now, but it's notably limiting in that it assumes we only go to the lobby. We may later want to generalize the redirect but until we have a clear vision for the UX across pages I think this is solid

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -413,3 +413,43 @@ describe('Lobby - P1 Perspective', () => {
...
+    cy.createGamePlayer({ gameName: 'Test Game', isRanked: false }).then((gameSummary) => {
+      cy.wrap(gameSummary).as('gameSummary');
+      // Sign up new (other) user and subscribe them to game
+      cy.signupOpponent(opponentOne);
+      cy.subscribeOpponent(gameSummary.gameId);
+    });
+  });
+  
+  
+  it('Joins Lobby from url', function () {
```
itsalaidbacklife, line 432: Let's check that the so let subscriptions work here by reading both players up and confirming the game starts

itsalaidbacklife: We can do that in both tests and in one of them have the opponent ready up before the player joins via lobby url (so we confirm we initialize the existing ready state) and then in the other one both players can ready up after the player joins via url so we can see responses to the ready action

tests/e2e/specs/out-of-game/lobby.spec.js
itsalaidbacklife, line 1: How about adding a test case for refreshing the lobby page to confirm the game still starts as normal after hitting the initialization logic twice?

src/router.js
```diff
@@ -35,6 +36,35 @@ const logoutAndRedirect = async (to, from, next) => {
...
+  
+  try {
+    await gameStore.requestSubscribe(+gameId);
+    next();
+  }
+  catch (err) {
+    to.meta.error = err.message;
+    next();
+  }
```
itsalaidbacklife, line 64: How does this affect the flow when a user hits the join game button on a lobby and the request fails because the game is full? Does the user still get a snackbar on the home view or are they now redirected to the lobby error view?

src/router.js
```diff
@@ -35,6 +36,35 @@ const logoutAndRedirect = async (to, from, next) => {
...
+  if (!authStore.authenticated) {
+    return next(`/login/${gameId}`);
+  }
+
+  if (gameStore.players.some(({username}) => username === authStore.username)) {
+    return next();
+  }
+  
+  try {
+    await gameStore.requestSubscribe(+gameId);
```
itsalaidbacklife, line 58: Do we still need to cast this if we already called parseInt() above?

src/router.js
```diff
@@ -69,7 +99,7 @@ const routes = [
     path: '/lobby/:gameId',
     component: LobbyView,
     // TODO: Add logic to redirect if a given game does not exist
-    beforeEnter: mustBeAuthenticated,
+    beforeEnter: checkAndSubscribeToLobby,
```
itsalaidbacklife, line 102: This means that we double request to join the game, right? Once from the home page before the router push and then right here before the page is loaded by the router. I'm wondering how we can nicely hit it once instead. @seriouslysean penny for your thoughts:

I think we want the home page to be able to show a loading state when you hit the join button and keep you on the home page until the you successfully subscribe to the game so that if the subscribe call fails, you can stay on the home page and get an error snackbar without being redirected to an error view and then having to come back.

But if you navigate directly to the lobby url, how do we subscribe you to the game in a way that isn't redundant to the flow where you're clicking the join button from the home page?

itsalaidbacklife: I think the move may be to set a piece of store state when coming from the home page that this nav guard checks from and uses to skip the subscription call in that case

itsalaidbacklife (commented):
Great work! I'm asking for additional translations, a new test case + a test case tweak, and a re-look at de-duping requests for the flows where the user comes from the home page vs where they navigate directly to a lobby url

src/router.js
Haviles04: I don't believe it double requests, After the initial request it triggers `gameStore.updateGame()` which should update the players in gameStore, which this function checks against the username, so if the user is already apart of the game we give the go ahead, if not we request subscribe

src/router.js
Haviles04: The snackbar still shows up, the router does not push the lobby url unless the `requestSubscribe` in successful, otherwise it emits the error which displays the snackbar. I confirmed this with cypress just to be sure

src/router.js
Haviles04: nope, good catch

src/router.js
seriouslysean: I talked to @itsalaidbacklife about this a little bit. It feels like all (or most) of the logic should be occurring here so it fires for both direct loads, and client side loads.

In my opinion, the store should handle the state of the incoming user and whether here or in the component you’ll be able to map the data you need without duplication.

This will also allow you to utilize a bit of state for the toast message if the game can’t be found or it otherwise invalid.

seriouslysean (commented):
Left some thoughts on the implementation.

src/routes/home/HomeView.vue
```diff
@@ -198,6 +198,11 @@ export default {
     await this.gameListStore.requestGameList();
     this.loadingData = false;    
   },
+  mounted() {
+    if (history?.state?.error) {
```
itsalaidbacklife, line 202: Slick

src/router.js
```diff
@@ -35,6 +36,38 @@ const logoutAndRedirect = async (to, from, next) => {
...
+    if (gameStore.players.some(({username}) => username === authStore.username)) {
+      return true;
+    }
+      
+    await gameStore.requestSubscribe(gameId);
+    return true;
+  }
+  catch (err) {
+    if (from.name === 'Home') {
+      throw new Error(err);
```
itsalaidbacklife, line 62: Does throwing here get caught in the component that calls .push()?

itsalaidbacklife (approved):
Very well done. This came together cleaner than I'd previously hoped, and it'll be a solid QoL improvement

src/router.js
Haviles04: Yup

src/routes/home/HomeView.vue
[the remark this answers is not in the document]
Haviles04: @seriousySean I used the history state to pass the error message on redirect in `router.js`

`return {name: 'Home', state: {error: err.message} };`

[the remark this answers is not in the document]
seriouslysean: I don't see the `state` in their documentation, only `query` or `params`. I'm certainly not against it if it's a standard usage of the router api, just want to make sure as I haven't seen this particular use case before. It seems like a query param might serve the same purpose without needing to use state at all, then you'd be able to remove the query param on mounted of the component if it exists.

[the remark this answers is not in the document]
Haviles04: Yea, I'm not 100% sure to be honest which is the best way, I do feel like adding it as a query or param is a little weird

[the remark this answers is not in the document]
seriouslysean: Yeah, I get what you’re saying but query param is at least a standard usage and pretty normal in my experience. If you can find me documentation on that being a standard or expected use of Vue router I’m totally down, otherwise I’d like you to switch it.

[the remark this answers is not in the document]
seriouslysean: What about a conditional route param? Would that work in this case instead? Trying to find a decent alternative while sticking with some semblance of standards is tricky :)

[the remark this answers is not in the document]
seriouslysean: Last note, I promise! It's also ok to ship it and resolution later if you just want it out there. I would make sure to test in all browsers-- at least Chrome, Firefox and Safari (desktop + mobile) as they may behave differently with the api as called out in the RFC.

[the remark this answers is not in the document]
itsalaidbacklife: @Haviles04 is there a reason that a query param would be more difficult to use?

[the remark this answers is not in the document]
itsalaidbacklife: @seriouslysean after talking to @Haviles04 about it, one thing that I think is nice about using history state over a route param is that refreshing the page on the home view after getting the error will not re-trigger the error, without needing extra cleanup logic to clear the route param.

Given that the approaches are otherwise comparable, I think that clinches it for state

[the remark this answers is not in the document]
Haviles04: So I ended up pushing it to be a query param, I was having an issue on displaying the error message if you were already on '/' and the router was redirecting you back to '/' because the component wouldn't reload which wouldn't fire the mounted hook. The only way around this was to force a refresh of the page before redirecting it which felt clunky to me. 

I tried to `watch` the history but it didn't fire, so instead I watched the `$route`, and show the snackbar anytime there is an error property in the `router.query`

src/routes/home/HomeView.vue
```diff
@@ -204,6 +212,7 @@ export default {
       this.showSnackBar = false;
     },
     handleSubscribeError(gameId, message) {
+      debugger;
```
seriouslysean, line 215: Errant debugger!

Haviles04: Ooooops

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -413,3 +415,85 @@ describe('Lobby - P1 Perspective', () => {
...
+  it('Navigates Home and shows error snackbar when user visits invite link of full game', function () {
+    cy.get(`[data-cy-join-game=${this.gameSummary.gameId}]`).should('be.enabled');
+    cy.signupOpponent(opponentTwo);
+    cy.subscribeOpponent(this.gameSummary.gameId);
+    cy.get(`[data-cy-join-game=${this.gameSummary.gameId}]`).should('be.disabled');
+    cy.visit(`#/lobby/${this.gameSummary.gameId}`);
+    assertSnackbarError(SnackBarError.GAME_IS_FULL, 'newgame');
+  });
+
+
```
itsalaidbacklife, line 498:
```suggestion
```

src/router.js
```diff
@@ -35,6 +36,32 @@ const logoutAndRedirect = async (to, from, next) => {
   return next('/login');
 };
 
+const checkAndSubscribeToLobby = async (to) => {
+  const gameStore = useGameStore();
+  const authStore = useAuthStore();
+  const gameId = parseInt(to.params.gameId);
+  try {
+    if (Number.isNaN(gameId) || !Number.isFinite(gameId)) {
+      throw new Error('Invalid Lobby Number');
```
itsalaidbacklife, line 45: Does this need to be translated too?

itsalaidbacklife (approved):
Excellent work! There's one error left to translate and then this is ready

PR #874, opened 2024-01-28
Optional gameId/seasonId in routes and redirection

### LINKED ISSUE #871: [Bug]: /Game and /Spectate bring you to blank screen
### Bug Summary

Navigating to /game or /spectate without params brings you to a blank screen

### Detailed Description

Currently if you navigate to `/game` or `/spectate` without passing any `gameId` as a param you are navigate to a blank white page with the cuttle navbar above it. This should also be handled for `/lobby` and any other routes that might not have explicit paths

We should redirect these routes to the home view `/` 

### Current Behavior

Navigates to a blank white screen with navbar

### Expected Behavior

Navigates back to the home view

Posoroko:
## Issue number

- Resolves #871

## Please check the following

- For New Features:

-made the gameId and seasonId optional in the routes that should contain it. (lobby, game, spectate and rematch, stats)
-redirection to home if no gameId/seasonId is found in the route path


src/router.js
```diff
@@ -37,6 +37,14 @@ const logoutAndRedirect = async (to, from, next) => {
   return next('/login');
 };
 
+const redirectIfNoIdParam = (to, from, next) => {
+  if (to.params.gameId || to.params.seasonId) {
+    next();
+  } else {
+    next('/');
```
itsalaidbacklife, line 44: This is the right idea, but the reason this won't work is that the first time you navigate to the /stats page, we don't have the ids of any of the seasons yet (the stats view downloads them on created).

itsalaidbacklife (commented):
Right idea, but this prevents initial navigation to the stats page (before season ids are downloaded). 

This issue was just closed, coincidentally, so I'll close this PR. Sorry for the miscommunication!

itsalaidbacklife:
Upon a second look, I think this is appropriate for the /game, /lobby and /spectate routes, just not for /stats, because stats downloads the season ids on created().

In that case I think it makes sense to update the helper function to specifically look for the gameId param (and to rename it accordingly)

Posoroko:
@itsalaidbacklife yep, ok. I'll do the changes.  👍

Haviles04:
So, after looking at the your code and noticing you added the `?` to make the param optional it made me think that maybe we shouldn't redirect to the home page. If this will still load the `GameView.vue` but it will show the `GameUnavailableDialog` it might be better that way. This way the end user has an explicit error message stating what's going on, instead of just a redirect which could potentially confuse them. 

@Posoroko @itsalaidbacklife what are your thoughts?

Posoroko:
>This way the end user has an explicit error message stating what's going on, instead of just a redirect which could potentially confuse them.

@Haviles04 

About the possible confusion...   If I have this right, this redirection would only occur if the user manually enter a url with no param.  Wich is not really the expected behaviour.  So weird results when trying to "hack" the router is probably not so problematic.  

Could we redirect home and get the router to throw an error to be displayed if the gameId is missing ?  something like :  "Please provide a game id or select a game in the list"

Haviles04:
> >This way the end user has an explicit error message stating what's going on, instead of just a redirect which could potentially confuse them.
> 
> @Haviles04 
> 
> About the possible confusion...   If I have this right, this redirection would only occur if the user manually enter a url with no param.  Wich is not really the expected behaviour.  So weird results when trying to "hack" the router is probably not so problematic.  
> 
> Could we redirect home and get the router to throw an error to be displayed if the gameId is missing ?  something like :  "Please provide a game id or select a game in the list"
>  
> 

So by just adding the optional param to `/game` will still load the game view(p reviously it was just a white screen) but without the `gameId` it should display the `GameUnavailable` overlay that displays an error message with a "Go home" button. It's technically less logic for a mistake that likely won't occur often while still giving an message

itsalaidbacklife:
@Posoroko sorry it's taken me so long to get back to a review here! I've been hellbent on wrapping up some changes and then coordinating the logistics of the Cuttle World Championship last Saturday and now that we're through that, I have bandwidth to dig back into QoL improvements like this.

I agree with @Haviles04 that it would be ideal to show the `GameUnavailableView` when the game id is missing so you can see that something is wrong and hit the button to go home. I think we could do that by adding a route mapping in the router for when you have /game without a /:gameId that points to the `GameUnavailableView`. Does that make sense?

Haviles04:
@itsalaidbacklife @Posoroko I think the ? On the gameID is enough actually.. it'll load gameView but the store won't have an id which will trigger the dialog

src/router.js
```diff
@@ -104,7 +104,7 @@ const routes = [
     },
   },
   {
-    path: '/stats/:seasonId?',
+    path: '/stats/:seasonId',
```
Haviles04, line 107: this should have the `?` this way you can see the stats page with or without a seasonId

itsalaidbacklife:
Closing as stale

PR #879, opened 2024-02-15
Chane opponent card border radius to match primary player's

meichthys:
Simple fix to make opponent card border radius match primary player's card border radius


itsalaidbacklife (approved):
Thanks for your contribution! Could you comment a before and after pic for the patch note?

PR #891, opened 2024-03-12
feat: switch from hash history to web history

### LINKED ISSUE #580: [Feature]: Convert page routing to use history mode
### Feature Summary

We should update the routing strategy to use vue's history mode

### Detailed Description

Currently the Cuttle vue SPA uses the [hash mode](https://router.vuejs.org/guide/essentials/history-mode.html#Hash-Mode) which separates all the individual pages with a # e.g. https://cuttle.cards/#/login. This has negative SEO impact because web crawlers do not correctly index pages using #.

To address this, we should update the router configuration to [utilize history mode](https://router.vuejs.org/guide/essentials/history-mode.html#Hash-Mode), which would change the routes to https://cuttle.cards/login etc. This will require updating tests that refer to routes using #/, so that they correctly look for the #. 

Additionally, per the Vue router docs for history mode, we will need to update the server `config/routes.js` file in order to define a catch-all route that directs all unrecognized urls to `index.html` so that vue can handle them. That way if a user navigates directly in their browser to https://cuttle.cards/about they will load the vue app and vue can render the about (rules) page. Without this, the server will not recognize the https://cuttle.cards/about url and direct navigation to pages of the app will 404.

seriouslysean:
## Issue number

- Resolves #580 

## Please check the following

- For New Features:

Dev build

1. Spin up the build normally with `npm run start:dev` and then go to http://localhost:8080/
2. Log in normally
3. Verify the paths all still work without the hash (e.g. http://localhost:8080/rules, etc)

Prod build

1. Spin up a prod build with `npm run start:preview` and then go to http://localhost:1337/
2. Log in normally
3. Verify the paths all still work without the hash (e.g. http://localhost:1337/rules, etc)

Tests

1. Make sure they all still work 🙃


config/routes.js
```diff
@@ -104,4 +104,7 @@ module.exports.routes = {
   '/test/loadSeasonFixture': 'TestController.loadSeasonFixture',
   '/test/loadMatchFixtures': 'TestController.loadMatchFixtures',
   '/test/loadFinishedGameFixtures': 'TestController.loadFinishedGameFixtures',
+
+  // Catch all for Vue paths
+  '/*': { controller: 'ViewController', action: 'serveIndex', skipAssets: true },
```
seriouslysean, line 109: This is required for sails to properly server the vue index, including any and all paths that aren't listed above functioning as a catch all.

src/router.js
```diff
@@ -102,7 +102,7 @@ const routes = [
 ];
 
 const router = createRouter({
-  history: createWebHashHistory(),
+  history: createWebHistory(),
```
seriouslysean, line 105: This is the major change, the route will be inferred from the url now, and not the hash suffix.

tests/e2e/specs/in-game/game-over/rematch.spec.js
```diff
@@ -121,6 +121,7 @@ describe('Creating And Updating Ranked Matches With Rematch', () => {
 
     // Match should be incomplete with one game
     cy.request('http://localhost:1337/match').then((res) => {
+      debugger;
       expect(res.body.length).to.eq(3);
```
seriouslysean, line 125: @itsalaidbacklife this line is failing now in this branch. Where is `/match` defined as a route? I can't find it anywhere.

src/router.js
itsalaidbacklife, line 1: Can we safely add route bindings to redirect the old hash routes to the new urls without the hash? Like redirect /#/rules to /rules. That way any references outside the application remain usable.

itsalaidbacklife (commented):
This is looking great. Awesome work. This should let us set up a proper robots.txt and also make enable our jump link setup to properly leverage heading anchors

src/router.js
seriouslysean: Probably, those might need to be a navigation guard or hook in the router as static redirects.

config/routes.js
```diff
@@ -104,4 +104,10 @@ module.exports.routes = {
   '/test/loadSeasonFixture': 'TestController.loadSeasonFixture',
   '/test/loadMatchFixtures': 'TestController.loadMatchFixtures',
   '/test/loadFinishedGameFixtures': 'TestController.loadFinishedGameFixtures',
+  '/test/game': 'TestController.getGames',
+  '/test/match': 'TestController.getMatches',
+
+  // Catch all for Vue paths only require get for http request so we don't interfer
+  // with any blueprints routes
+  '/*': { controller: 'ViewController', action: 'serveIndex', skipAssets: true },
```
seriouslysean, line 112: I considered importing the routes from vue router in to this file and making it only proxy vue routes, but then we lose the benefit of vue handling any non app route where we'd want a true Vue powered 404. Speaking of which, @itsalaidbacklife -- we'll need a todo for a proper 404 now once this goes in.

Maybe some sort of animated Cuttleworth, floating upside down with x eyes, hahahaha.

config/env/development.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
   //   connection: 'someMongodbServer'
   // }
   blueprints: {
-    rest: true,
+    rest: false,
+    shortcuts: false,
```
seriouslysean, line 23: This is sort of annoying but the catch all wreaks havoc on the api calls from blueprints, overriding all of them. I'm sure there's got to be a better way of doing this (converting to post instead of get, which the http requests use), but this works for now.

All previous logic has moved to test controller functions here, https://github.com/cuttle-cards/cuttle/pull/891/files#diff-ff18ca6c9bd29eacdb89029e8f309522bd67e49d21b18e18f42de044a48c20e4R58-R78.

tests/e2e/specs/in-game/spectating.spec.js
```diff
@@ -358,7 +358,7 @@ describe('Spectating Games', () => {
         .should('not.contain', 'Request Stalemate')
         .should('not.contain', 'Concede');
       cy.get('[data-cy="stop-spectating"]').click();
-      cy.hash().should('eq', '#/');
+      cy.location('pathname').should('eq', '/');
```
seriouslysean, line 361: We no longer can use the hash, we need to use the actual pathname of the route in these tests.

api/controllers/ViewController.js
```diff
@@ -0,0 +1,13 @@
+const path = require('path');
+
+const indexHtmlFile = path.resolve(__dirname, '../../assets/index.html');
+
+module.exports = {
+  serveIndex: function (req, res) {
+    return res.sendFile(indexHtmlFile, (err) => {
+      if (err) {
+        return res.status(500).send('Internal Server Error');
```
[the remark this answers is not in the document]
seriouslysean: I actually opted not to because of the information it puts out there but I can see the need. I don’t mind it either way.

src/router.js
seriouslysean: Added in the latest commit, and added in a way where it only affects the initial route. I don’t think we should leave that long term though, it’s not ideal. Maybe for a few months.

src/router.js
```diff
@@ -101,8 +101,22 @@ const routes = [
...
+    const path = window.location.hash.replace('#/', '');
+    window.location.hash = '';
+    return path;
+  }
+  return null;
+};
+
+const initialPath = getInitialPath();
+if (initialPath) {
+  window.history.replaceState({}, '', initialPath);
```
seriouslysean, line 115: This ensures the initial path is set before the router is initialized. It limits the initial flash of one page then another, and avoids the need for any sort of router push. It will just read the right URL the first time.

tests/e2e/specs/out-of-game/other.spec.js
```diff
@@ -0,0 +1,13 @@
+describe('Hash Router Redirects', () => {
```
seriouslysean, line 1: New test for the hash routing redirects.

src/routes/error/NotFoundView.vue
```diff
@@ -0,0 +1,42 @@
+<script setup>
```
seriouslysean, line 1: Super barebones, just need someone to pretty it up.

image

tests/e2e/specs/out-of-game/error.spec.js
```diff
@@ -0,0 +1,13 @@
+describe('Error Page', () => {
```
seriouslysean, line 1: Added tests for it, too. These will all obviously change, but bones are there.

src/routes/error/NotFoundView.vue
```diff
@@ -0,0 +1,42 @@
...
+const { logoSrc } = useThemedLogo();
+</script>
+
+<template>
+  <div class="pa-4">
+    <v-container>
+      <v-row>
+        <v-col cols="12" class="text-center">
+          <router-link to="/">
+            <img
```
seriouslysean, line 14: You need a task to abstract this in to a shared header and footer cross app, with a wrapper component functioning as the base layout so we can add header/footer easily to any page.

src/translations/es.json
```diff
@@ -434,5 +434,8 @@
     "pointsAndWins": "Puntos y victorias",
     "pointsOnly": "Solo puntos",
     "winsOnly": "Solo gana"
+  },
+  "error": {
+    "title": "404 No Encontrado"
```
itsalaidbacklife, line 439: Since 404 is a specific error, I think it makes sense to nest this under a "404" key like:
```suggestion
  "error": {
    "404": {
      "title": "404 No Encontrado"
    }
  }
```

That way we can have whole error pages for other errors and they can have separate titles etc

tests/e2e/specs/out-of-game/error.spec.js
itsalaidbacklife: Let's call this NotFoundPage
```suggestion
describe('Not Found Page', () => {
```

config/env/development.js
```diff
@@ -18,7 +18,11 @@ module.exports = {
   // models: {
   //   connection: 'someMongodbServer'
   // }
+  log: {
+    level: 'debug'
+  },
```
seriouslysean, line 23: More debug logs for local development

config/routes.js
```diff
@@ -46,62 +46,73 @@ module.exports.routes = {
    ***************************************************************************/
 
   // '/'                             :   'userController.homepage',
-  '/user/signup': 'userController.signup',
-  '/user/login': 'userController.login',
-  '/user/reLogin': 'userController.reLogin',
-  '/user/logout': 'userController.logout',
-  '/user/status': 'userController.status',
+  '/api/user/signup': 'userController.signup',
```
seriouslysean, line 49: When I moved to web history from hash history, it created a conflict in the route namespace. We use `/#/game/*` in vue but already used `/game/*` on sails
I had to move all of sails to a /api/* namespace

config/routes.js
```diff
@@ -46,62 +46,73 @@ module.exports.routes = {
...
+  '/api/test/match': 'TestController.getMatches',
+
+  // Catch all for Vue paths only require get for http request so we don't interfer
+  // with any blueprints routes
+  '/*': {
+    controller: 'ViewController',
+    action: 'serveIndex',
+    skipAssets: true,
+    skipRegex: /^\/api\/.*$/ // ignore /api routes
+  },
```
seriouslysean, line 117: We skip the `/api/*` namespace here so it isn't part of the catch all logic that goes to the Vue `index.html`.

vite.config.js
```diff
@@ -43,29 +44,17 @@ export default defineConfig(({ mode }) => {
...
         },
       },
       // Watching doesn't work on windows, so we need to use polling -- this does lead to high CPU
-      // usage though, which is a bit of a bummer. Should probably make this conditional at some
+      // usage though, which is a bit of a bummer.
       // point, see https://v3.vitejs.dev/config/server-options.html#server-watch
       watch: {
-        usePolling: true,
+        // returns 'win32' for Windows, regardless of the architecture
+        usePolling: os.platform() === 'win32',
```
seriouslysean, line 57: Windows only fix, we should get a mac/linux speed boost from this.

vite.config.js
```diff
@@ -43,29 +44,17 @@ export default defineConfig(({ mode }) => {
...
-        '/health': {
-          target: HOST_SERVER_URL,
-          changeOrigin: true,
-        },
-        '/user': {
-          target: HOST_SERVER_URL,
-          changeOrigin: true,
-        },
-        '/test': {
+        '/api': {
```
seriouslysean, line 47: Only need to proxy `/api/*` now in Vite since it's all grouped under the same namespace.

Haviles04 (approved):
This looks great!

src/router.js
```diff
@@ -126,10 +126,33 @@ const routes = [
     component: StatsView,
     beforeEnter: mustBeAuthenticated,
   },
+  // Catch every other unsupported route
+  {
+    path: '/:pathMatch(.*)*',
+    name: 'Not Found',
+    component: () => import(
+      /* webpackChunkName: "not-found-view" */
```
itsalaidbacklife, line 134: Is this applicable to vite since it's rollup based? Looks like webpack config

itsalaidbacklife (approved):
Very well done

src/router.js
seriouslysean: Nope, we’d need a plugin. Old habit :) — kind removing it? I have limited access today.

itsalaidbacklife:
```suggestion
```

PR #895, opened 2024-03-18
added a conditional statement in statsUsageChart template for the issue #885

### LINKED ISSUE #885: [Feature]: Hide the "Site Usage" chart for seasons that have no games played
### Feature Summary

The `StatsUsageChart` should not be shown for seasons that have no data to display.

### Detailed Description

Some seasons are placeholders for the World Championship tournaments, and these don't have any games played in them. The Usage Stats chart shows an empty chart for these seasons:
image

Instead, we should update the `StatsView` so that its instance of the `StatsUsageChart` only shows if there are any games in the current season i.e. if the `selectedSeason.gameCounts` is > 0.

karan3267:
template

## #885 
Relevant [#885 ](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
- Resolves #885

## Please check the following

- For New Features:

after the changes the stats chart will only render if there are games played in the selected tournament

### DIFF
--- a/src/routes/stats/StatsView.vue
+++ b/src/routes/stats/StatsView.vue
@@ -84,7 +84,7 @@
           <h2 class="text-h2 mt-8 mb-4">
             {{ t('stats.siteUsage') }}
           </h2>
-          <StatsUsageChart :season="selectedSeason" />
+          <StatsUsageChart v-if="selectedSeason.gameCounts>0" :season="selectedSeason" />
         </div>
         <!-- Error display -->
         <div v-if="error" class="d-flex flex-column align-center text-center">
--- a/src/routes/stats/StatsView.vue
+++ b/src/routes/stats/StatsView.vue
@@ -84,7 +84,7 @@
           <h2 class="text-h2 mt-8 mb-4">
             {{ t('stats.siteUsage') }}
           </h2>
-          <StatsUsageChart v-if="selectedSeason.gameCounts>0" :season="selectedSeason" />
+          <StatsUsageChart v-if="showStatsUsageChart" :season="selectedSeason" />
         </div>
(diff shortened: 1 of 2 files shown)


karan3267:
The above mentioned changes have been added but i have a doubt regarding the values of gameCounts and uniquePlayersPerWeek, how to get these values? or should i just populate some values in the arrays so that the Stats chart can be rebdered?

src/routes/stats/StatsView.vue
```diff
@@ -84,7 +84,7 @@
           <h2 class="text-h2 mt-8 mb-4">
             {{ t('stats.siteUsage') }}
           </h2>
-          <StatsUsageChart :season="selectedSeason" />
+          <StatsUsageChart v-if="selectedSeason.gameCounts>0" :season="selectedSeason" />
```
itsalaidbacklife, line 87: Looking more closely at the shape of the season data, this conditional is incorrect, because the `gameCounts` is an array of integers, not a single integer. Because this makes it more complicated, let's pull this into its own computed property called `showStatsUsageChart`, which can essentially return false if either the array is empty, or every element in the array is zero. It could look like this:

```
computed: {
  showStatsUsageChart() {
    const gameCounts = this.selectedSeason?.gameCounts ?? [];
    // Show unless gameCounts is empty or all zeros
    return gameCounts.length > 0 && !gameCounts.every(gameCount => gameCount === 0)
  }
}
```

karan3267: I have tried this and the tests still fail

itsalaidbacklife (approved):
Great work! Thank you for your contribution

PR #900, opened 2024-03-30
DevEx #866 Utilize Sleep

### LINKED ISSUE #866: [DevEx]: Define clientside `sleep()` utility to replace use of `setTimeout()` calls for animation processing
### Improvement Summary

We should define and consume a new `sleep()` utility to streamline the use of delays in clientside code.

### Detailed Description

The client uses various delays in js to manage multi-part transitions by changing some variables in state then doing the rest after a delay using `setTimeout()`. This requires nesting the delayed code in a callback. To simplify it we could use a `sleep()` utility that returns a promise which resolves after a specified delay, allowing us to 

```
await sleep(1000); // wait 1s
doSomething();
```

instead of

```
setTimeout(() => {
doSomething()}, 1000);
```

Haviles04:
## Issue number

- Resolves #866 

## Please check the following

- For New Features:

-Imported sleep and replaced `setTimeout`


itsalaidbacklife (approved):
Nicely done

itsalaidbacklife (commented):
Not sure what the heck happened to cause these test failures though

src/plugins/sockets/inGameEvents.js
```diff
@@ -25,9 +25,9 @@ export async function handleInGameEvents(evData) {
   }
   // Handle GameOver
   if (evData.victory && evData.victory.gameOver) {
-    setTimeout(() => {
+    sleep(1000).then(() => {
       gameStore.setGameOver(evData.victory);
-    }, 1000);
+    });
```
Haviles04, line 30: I tried to use async await here like I did in the other parts of the PR but for some reason, it does not work unless it's promise chained

itsalaidbacklife: @seriouslysean any idea why tf `.then()` works but `await` doesn't? It looks to be the case that the subsequent lines fire at the appropriate time either way, but when using `await` the DOM fails to update.

itsalaidbacklife (approved):

src/plugins/sockets/inGameEvents.js
itsalaidbacklife: Created #923 to follow up

itsalaidbacklife (approved):

src/plugins/sockets/inGameEvents.js
seriouslysean: Lemme dig in to this a bit today and see if I can suss anything out.

seriouslysean: Just to be sure, you were doing something like this?

```
// Handle GameOver
if (evData.victory?.gameOver) {
  await sleep(1000);
  gameStore.setGameOver(evData.victory);
}
```

Haviles04: Yea

seriouslysean: @Haviles04 any chance you can get a minimal reproduction of it not working?

Haviles04: @seriouslysean I'm at work ATM but can do it either tonight or tomorrow

PR #904, opened 2024-05-01
Fix/heap memory crashes

itsalaidbacklife:
This fixes a critical bug where the amount of data processed by the `StatsController` endpoints was far more than needed and therefore in the production environment the app would crash whenever a single user visited the /stats page.

This partial fix is to use `select` statements to project the queries for games and users down to just the ones necessary for the computation. That is to say we no longer load all the fields for all users and the relevant games into memory, just the users' `id` + `usernames`, and the games `updatedAt` + `p0` and `p1` fields.

Performance wise, besides the crash, this improves performance of the stats page by an order of magnitude. That said, it is still an incomplete fix because if enough users play enough games, we will hit the same limit memory issue, even with the omitted fields. We should write up an issue to use the `.steam()` api for sails' waterline so that the records can be processed in manageable batch sizes.

- For New Features:

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-worker: node app.js
+worker: node --max-old-space-size=460 app.js
--- a/api/controllers/StatsController.js
+++ b/api/controllers/StatsController.js
@@ -141,14 +141,19 @@ module.exports = {
     try {
       const seasons = await sails.helpers.getSeasonsWithoutRankings();
       const [currentSeason] = seasons;
-      const allUsers = User.find({});
+      const allUsers = User.find({
+        select: ['id', 'username'],
+      });
       const currentSeasonMatches = Match.find({
         startTime: { '>': currentSeason.startTime },
         endTime: { '<': currentSeason.endTime },
       });
       const currentSeasonGames = Game.find({
-        status: gameService.GameStatus.FINISHED,
-        updatedAt: { '>': currentSeason.startTime, '<': currentSeason.endTime },
+        select: ['updatedAt', 'p0', 'p1'],
+        where: {
+          status: gameService.GameStatus.FINISHED,
+          updatedAt: { '>': currentSeason.startTime, '<': currentSeason.endTime },
+        },
       });
       const [users, matches, games] = await Promise.all([allUsers, currentSeasonMatches, currentSeasonGames]);
       updateRankingsFromMatches(users, matches, currentSeason);
@@ -163,14 +168,19 @@ module.exports = {
     const seasonId = parseInt(req.params.seasonId);
     try {
       const [requestedSeason] = await sails.helpers.getSeasonsWithoutRankings(seasonId);
-      const allUsers = User.find({});
+      const allUsers = User.find({
+        select: ['id', 'username'],
+      });
       const requestedSeasonMatches = Match.find({
         startTime: { '>': requestedSeason.startTime },
         endTime: { '<': requestedSeason.endTime },
       });
       const requestedSeasonGames = Game.find({
-        status: gameService.GameStatus.FINISHED,
-        updatedAt: { '>': requestedSeason.startTime, '<': requestedSeason.endTime },
+        select: ['updatedAt', 'p0', 'p1'],
+        where: {
+          status: gameService.GameStatus.FINISHED,
+          updatedAt: { '>': requestedSeason.startTime, '<': requestedSeason.endTime },
+        },
       });
       const [users, matches, games] = await Promise.all([
         allUsers,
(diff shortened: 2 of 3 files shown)


Procfile
```diff
@@ -1 +1 @@
-worker: node app.js
+worker: node --max-old-space-size=460 app.js
```
itsalaidbacklife, line 1: This doesn't look like it's actually required for the fix, but it's a [recommended best practice](https://devcenter.heroku.com/articles/node-best-practices#avoid-garbage) for the 512mb memory containers

api/controllers/StatsController.js
```diff
@@ -141,14 +141,19 @@ module.exports = {
     try {
       const seasons = await sails.helpers.getSeasonsWithoutRankings();
       const [currentSeason] = seasons;
-      const allUsers = User.find({});
+      const allUsers = User.find({
+        select: ['id', 'username'],
+      });
```
itsalaidbacklife, line 146: Previously, this grabbed all data from every user record. Now we're only taking their username + id. This is the crux of the fix, and it's rolled out in a few places that queried for users and for games

Haviles04: Does it make sense to query for only users where there `updatedAt` is between the season start and end times?

seriouslysean: That’s a good idea, best case you limit the data as low as humanly possible. When @itsalaidbacklife and I talked about this originally I wanted to cache it in some way and statically serve it, only updating after a game is played with some sort of other delay to stop slamming the server.

I think it would make sense to use a custom endpoint in sails and then server the cached version until the delay is up then re-request it. If you wanted to collect all the data like we used to do it would be even more
Important.

itsalaidbacklife: Reasonable ideal. The user table itself doesn't update when games are completed so that doesn't actually get us the users who have played this season. Alternatively, we could wait for the query for games first and then get the list of user ids from the games, then query for the list of users based on those ids. That should theoretically use less memory, but take more time (because we need to query for games before we can get the users.

From what I can tell, updating the select statement to use less data per user record drastically reduces the memory footprint such that we can afford to query both simultaneously in order to load the page faster

Haviles04: Does the user table get updated on `clearGame` ?

Procfile
seriouslysean, line 1: We do this with a higher number at URBN and in my experience has been common practice, so worth it imo 👍🏼

seriouslysean (approved):

api/controllers/StatsController.js
itsalaidbacklife: That's an interesting point. I think that would be more efficient for the current setup, actually. My two concerns are that it would only work for the current season data, not an old one (since `updatedAt` will be overwritten compared the next time you play), and that this wouldn't work once we rework things so that the `user` table doesn't have a fk to game any more (just using the fk from game to user).

All in all I think that would be more efficient for right now, but maybe not as clean in the longer term

itsalaidbacklife: @seriouslysean re: caching. I think the optimal way to manage caching the data, performance-wise, would be to update it incrementally as games are completed. We could store the leaderboard response data in its own table, update that when games and matches complete, and then the endpoints would just dump table data. 

I think that makes sense to look to as we continue to scale, but it's a larger lift so I think we hold off for the moment. Just adding the `select` clauses has given us a 10x performance boost and kept the endpoint's computational intensity well within the server's limits.

config/env/development.js
```diff
@@ -11,18 +11,16 @@
...
   },
+  // Disable default endpoints for each model e.g. GET /game/:id
   blueprints: {
     rest: false,
     shortcuts: false,
   },
+  // Needed to use custom 'select' statements using sails-disk
+  models: {
+    schema: true,
+  }
```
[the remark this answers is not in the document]
itsalaidbacklife: Here are the [docs](https://sailsjs.com/documentation/reference/configuration/sails-config-models)

And here's the error that occurs when you query with `select` in development mode without this config:
```
"The provided criteria contains a custom `select` clause, but since this model (`user`) is `schema: false`, this cannot be relied upon... yet.  In the mean time, if you'd like to use a custom `select`, configure this model to `schema: true`. Or, better yet, since this is usually an app-wide setting,configure all of your models to have `schema: true` -- e.g. in `config/models.js`.  (Note that this WILL be supported in a future, minor version release of Sails/Waterline.  Want to lend a hand?  http://sailsjs.com/contribute)"
```

PR #913, opened 2024-05-18
Fix/rules navigation highlight

itsalaidbacklife:
Fixes bug in RulesView where clicking on the 'Royals' jump link would highlight the 'One-Offs' section. The issue was due to the IntersectionObservers triggering as soon as any portion of the element came into view. Now using a combination of `rootMargin` and `thresshold` to trigger them only when in the middle-ish of the screen.

Bug:

https://github.com/cuttle-cards/cuttle/assets/6520704/577fc4fd-55a2-4669-97a7-1fc05f75a6bd

## Issue number

- Resolves #

## Please check the following

- For New Features:

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -153,7 +153,7 @@ if (initialPath) {
 const router = createRouter({
   history: createWebHistory(),
   routes,
-  scrollBehavior(to, from, savedPosition) {
+  scrollBehavior(to, _from, savedPosition) {
     if (to.hash) {
       return { el: to.hash, behavior: 'smooth' };
     } else if (savedPosition) {
--- a/src/routes/rules/RulesView.vue
+++ b/src/routes/rules/RulesView.vue
@@ -392,7 +392,7 @@ export default {
         handler: this.onIntersect,
         options: {
           threshhold: 0,
-          rootMargin: '0px 0px -300px 0px',
+          rootMargin: '-150px 0px -300px 0px',
         }
       };
     }
--- a/src/routes/rules/RulesView.vue
+++ b/src/routes/rules/RulesView.vue
@@ -362,11 +362,6 @@ export default {
   },
     data() {
     return {
-      scrollOptions:{
-        duration: 1000,
-        offset: -100,
-        easing: 'easeInOutCubic',
-      },
       activeTitle: 'introduction',
       previewDialog: false,
       imageUrl: '',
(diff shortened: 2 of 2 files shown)


src/routes/rules/RulesView.vue
```diff
@@ -123,13 +119,8 @@
...
 
           <!-- Royals -->
-          <section class="section">
-            <div
-              id="royals"
-              v-intersect="{
-                handler: onIntersect,
-              }"
-            >
+          <section id="royals" v-intersect="intersectConfig" class="section">
```
itsalaidbacklife, line 122: I moved the intersection observers onto the sections instead of the headers so that the threshold can trigger based on a % of the entire section being visible

src/routes/rules/RulesView.vue
```diff
@@ -380,11 +362,6 @@ export default {
   },
     data() {
     return {
-      scorllOptions:{
-        duration: 1000,
-      offset: -100,
-      easing: 'easeInOutCubic',
-      },
```
itsalaidbacklife, line 387: Pulled this out of data since it doesn't need to be reactive

src/routes/rules/RulesView.vue
```diff
@@ -411,18 +388,35 @@ export default {
...
+
+    // Scrolling
+    this.scrollOptions = {
+      duration: 1000,
+      offset: -100,
+      easing: 'easeInOutCubic',
+    };
+
+    // Intersection
+    const onIntersect = (_isIntersecting, entries) => {
```
itsalaidbacklife, line 400: Defining this inline inside created as it's only to be used in the context of the `intersectConfig`

src/routes/rules/RulesView.vue
```diff
@@ -411,18 +388,35 @@ export default {
...
       });
+    };
+
+    this.intersectConfig = {
+      handler: onIntersect,
+      options: {
+          threshhold: 0,
+          rootMargin: '-150px 0px -300px 0px',
+        }
+    };
```
itsalaidbacklife, line 414: Put this here to make it nonreactive

src/router.js
```diff
@@ -153,7 +153,7 @@ if (initialPath) {
 const router = createRouter({
   history: createWebHistory(),
   routes,
-  scrollBehavior(to, from, savedPosition) {
+  scrollBehavior(to, _from, savedPosition) {
```
itsalaidbacklife, line 156: Just a chore I bumped into since from is unused

itsalaidbacklife (commented):
It would likely be a further improvement to pull the entire left nav into its own component, but that can be a subsequent update

src/routes/rules/RulesView.vue
```diff
@@ -411,18 +388,35 @@ export default {
...
           this.activeTitle = entry.target.id;
         }
       });
+    };
+
+    this.intersectConfig = {
+      handler: onIntersect,
+      options: {
+          threshhold: 0,
+          rootMargin: '-150px 0px -300px 0px',
```
itsalaidbacklife, line 412: This is the crux of the fix. It makes it so the element only counts as intersecting when at least 150px below the top of the viewport and 300 above the bottom

Haviles04: Will this still work correctly on big screens? Should it use a percentage?

itsalaidbacklife: Negative %'s aren't supported, so I think we are best off sticking with px. My hope is that this ensures it's at least fairly far onto the screen. I'll bump the threshold to .2 so that it also requires at least 20% of the section to be visible, which I think will further mitigate the possibility of it triggering prematurely on large screens (though it may not be perfect).

PR #917, opened 2024-05-19
Feat/unauthenticated nav bar

### LINKED ISSUE #916: [Feature]: RulesView - display TheHeader even when unauthenticated
### Feature Summary

We should display TheHeader component (the top toolbar) in the RulesView, HomeView, and StatsView, even when unauthenticated 

### Detailed Description

Currently the router.js file hides TheHeader component whenever you are unauthenticated. Because the RulesView can be accessed while authenticated or not, currently it will show TheHeader while authenticated and not show it otherwise. 

**Unauthenticated** 
IMG_5455

**Authenticated**
IMG_5456

This makes it difficult to return to the Signup Page when you are viewing the RulesView while unauthenticated. We should display TheHeader in either case , but when unauthenticated show a login button instead of TheUserMenu

itsalaidbacklife:
This updates the RulesView to show the top toolbar when unauthenticated. Shows a signup/login button in place of TheUserMenu (signup for first time users and login for returning users)
image

## Issue number

- Resolves #916 

## Please check the following

- For New Features:

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -4,11 +4,12 @@
     <v-main>
       <RouterView />
     </v-main>
-    <TheFooter v-if="showNav && isSmallDevice" :variant="variant" />
+    <TheFooter v-if="showNav && isSmallDevice && isAuthenticated" :variant="variant" />
   </v-app>
 </template>
 
 <script>
+import { mapStores } from 'pinia';
 import TheHeader from '@/components/TheHeader.vue';
 import TheFooter from '@/components/TheFooter.vue';
 import { useGameStore } from '@/stores/game';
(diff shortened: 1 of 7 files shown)


src/App.vue
```diff
@@ -4,11 +4,12 @@
     <v-main>
       <RouterView />
     </v-main>
-    <TheFooter v-if="showNav && isSmallDevice" :variant="variant" />
+    <TheFooter v-if="showNav && isSmallDevice && isAuthenticated" :variant="variant" />
```
Haviles04, line 7: This conditional is kinda long, does it make sense to make this a computed value?

Haviles04 (approved):
Definitely a user friendly improvement, Nice job!

src/routes/home/components/FiveChangeDialog.vue
```diff
@@ -6,7 +6,7 @@
     :opacity="1"
     data-cy="five-change-dialog"
     scrollable
-    max-width="750"
+    :max-width="750"
```
itsalaidbacklife, line 9: Fixes prop type warning b/c BaseDialog expects a number

PR #928, opened 2024-05-28
Bug/897

### LINKED ISSUE #897: [Bug]: Rule Formatting seems to have regressed
### Bug Summary

Rule formatting isn't great

### Detailed Description

I think the formatting in the rules dialog used to be better. It is currently:
image

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Steps to Reproduce

_No response_

kawaljen:
## 897

Relevant 897
- Resolves #897

## Please check the following

- For New Features:

With this change, the rulesdialog vue uses the rulesview as component

Some content is  hidden on the rulesdialog modal => the ready to play section
the backtotop button needs to show on scroll on both vues
the nav is positionned at the top on the modal and hidden on small devices


src/routes/rules/components/RulesNav.vue
```diff
@@ -82,14 +97,24 @@ watch(() => props.activeTitle,
 .active-title {
   background-color: rgba(var(--v-theme-newPrimary));
 }
-
+.inModal.sidebar-container{
+     background: rgba(var(--v-theme-surface-1));
+    position: sticky;
+    top: 0;
+    z-index:18;
```
itsalaidbacklife, line 104: Whitespace
```suggestion
.inModal.sidebar-container {
  background: rgba(var(--v-theme-surface-1));
  position: sticky;
  top: 0;
  z-index:18;
```

itsalaidbacklife (changes requested):
Good start. One general thing to look for is that we are moving over to using the `<script setup>` syntax, as it's the preferred option for vue 3. New components should all be written using script setup, and we will be changing old components over to it, over time. Please use `<script setup>` syntax in any components that already do so. Here are [the docs](https://vuejs.org/api/sfc-script-setup.html) on this syntax. 

We can discuss here or in discord if you have any questions. Another point is that it may be better to conditionally remove the `RulesNav` altogether in the `isInModal` case, rather than trying to conditionally style it. Can you show some pics how this looks across device sizes? Then we can decide whether to keep it

kawaljen:
I have made the changed you ask for, and made some additional onces :)

-> I made the targetid name for the modal comes from the parent, where that id is actually used, it looks more readible to me, what do you think ? in the process I am creating an additional prop, parentModalId


   -> I change the code to use the <script setup> you mentionned in the rulesdialogue view, and in the process, I have found an unused piece of code.
   Should I made it work : IE change the buttons size depending on viewport ?
   computed: {
     buttonSize() {
       return this.$vuetify.display.mdAndDown ? 'small' : 'medium';
     },
   },

-> I made some small styling changes on rulesview mainly: 
- centering the titles on small devices, 
- make the cards images go full length on small devices 
- changing some small bits so it will fit on all size even in a modal

src/routes/game/components/dialogs/components/RulesDialog.vue
```diff
@@ -1,217 +1,50 @@
 <template>
   <BaseDialog id="rulesDialog" max-width="1250">
```
itsalaidbacklife, line 2: Let's add 
```suggestion
  <BaseDialog id="rulesDialog" max-width="1250" :persistent="false">
```

So that this closes when clicking outside of it

src/routes/game/components/dialogs/components/RulesDialog.vue
```diff
@@ -1,217 +1,50 @@
...
   </BaseDialog>
 </template>
 
-<script>
-import BaseDialog from '@/components/BaseDialog.vue';
+<script setup>
+  import BaseDialog from '@/components/BaseDialog.vue';
+  import RulesView from '@/routes/rules/RulesView.vue';
+
+  const emit = defineEmits(['close']);
```
itsalaidbacklife, line 32: We normally don't indent js an extra layer, just align it with the script tag
```suggestion
import BaseDialog from '@/components/BaseDialog.vue';
import RulesView from '@/routes/rules/RulesView.vue';

const emit = defineEmits(['close']);
```
```suggestion
  import BaseDialog from '@/components/BaseDialog.vue';
  import RulesView from '@/routes/rules/RulesView.vue';

  const emit = defineEmits(['close']);
```

src/routes/game/components/dialogs/components/RulesDialog.vue
```diff
@@ -1,217 +1,50 @@
...
 </script>
+
 <style lang="scss" scoped>
 .rules-button {
   width: 100%;
 }
+.closeBtn {
+  right: 30px;
+  z-index: 19;
+  top: 15px;
```
itsalaidbacklife, line 43: Prefer multiples of 4
```suggestion
  top: 16px;
```

src/routes/rules/components/RulePreview.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <v-row class="mb-10">
-    <v-col lg="4" sm="5">
+    <v-col lg="4" sm="5" cols="12">
```
itsalaidbacklife, line 3: What's the reason for this change? It looks to make the card images too big on small devices:

this branch:
image


Vs `main`:
image

itsalaidbacklife: Here it is in the dialog:
image

src/routes/rules/components/RulesNav.vue
```diff
@@ -112,4 +115,4 @@ watch(() => props.activeTitle,
     font-size: 2.5rem !important;
   }
 }
-</style>
+</style>
```
itsalaidbacklife, line 118: Keep the whitespace at EOF
```suggestion
</style>

```

src/components/BackToTop.vue
```diff
@@ -35,5 +44,12 @@ function onClick() {
 .fab {
   margin: auto 8px 64px 8px;
   z-index: 1;
+  right: 23px;
+  bottom: 0;
+}
+@media (max-width: 600px) {
+  .fab {
+    right: 13px;
```
itsalaidbacklife, line 52: Pref multiples of 4 to keep the grid easier to manage. Either 12px or 16 please, at your discretion

src/components/BackToTop.vue
```diff
@@ -35,5 +44,12 @@ function onClick() {
 .fab {
   margin: auto 8px 64px 8px;
   z-index: 1;
+  right: 23px;
```
itsalaidbacklife, line 47:
```suggestion
  right: 24px;
```

src/components/BackToTop.vue
```diff
@@ -19,12 +18,22 @@ import { ref } from 'vue';
...
-function onScroll() {
-  show.value = window.scrollY > 200;
+const props = defineProps({
+  parentModalId: {
+        type: String,
+        default: '',
+        required: false
+      },
+});
+console.log( 'gfhdjs'+props.parentModalId);
```
itsalaidbacklife, line 28:
```suggestion
```

src/routes/rules/components/RulePreview.vue
kawaljen: yes I changed that, I though it looked better on small devices, especially in the modal.. 
is it worth passing a prop to rulepreview for that styling issue ?

itsalaidbacklife: In this case I think it's preferable to leave it as it was with the cards inline with the text, because then you can read more than one at a time without scrolling. Otherwise it will take twice as much scrolling to read the same amount of text.
```suggestion
    <v-col lg="4" sm="5">
```

src/routes/rules/RulesView.vue
```diff
@@ -393,6 +412,7 @@ export default {
   background-color: rgba(var(--v-theme-newPrimary));
   width: 100%;
   padding-top: 16px;
+  padding-bottom: 16px;
```
itsalaidbacklife, line 415: Because this font does not nicely calculate its line heights, applying equal padding to the top and bottom causes the magenta background to appear off-centered:
image


Vs 16px top padding with no bottom padding:
image
```suggestion
```

itsalaidbacklife: This is because, for some annoying reason, "Luckiest Guy" adds 'height' to the text on the bottom, but not the top. you can see this when you highlight the text:
image

src/routes/rules/RulesView.vue
```diff
@@ -405,10 +425,15 @@ export default {
 .text-label-lg {
   margin-bottom: 8px;
 }
-
+@media (max-width: 600px) {
+  .section-title {
+    padding-left: 0px;
+    text-align: center;
+  }
+}
```
itsalaidbacklife, line 433: I like the look of this change, but I fear it makes the headers look to much like big buttons, since we use the same color and center-align the text in our buttons
image


```suggestion
```

itsalaidbacklife (approved):
Great work. Thank you for your contribution!

PR #931, opened 2024-06-04
Fix for issue 930

### LINKED ISSUE #930: [Bug]: AwardCard.vue style regression
### Bug Summary

The AwardCard.vue component is styled incorrectly, showing incorrect colors and sizes.

### Detailed Description

Previously, the Award Cards (showing the players who won tournaments), looked like this:
image

#888 introduced a bug where they now look like this:
image

* The overall card is too small, and is missing padding
* The text is white and therefore not legible on the grey background

The issue should be fixed in `src/components/AwardCard.vue`, where we should appropriately style the component when given the `isCard` prop so that the colors and styles are appropriate.

In general, because we are phasing out white backgrounds in favors of the chocolate (surface-1) and cream (surface-2) colors, we should experiment with a different version that utilizes these colors. Maybe the card could be cream (surface-1), with the section surrounding the award trophy being white, and the text being chocolate (surface-1)? 

If that color scheme doesn't work, we should use the original version with white, grey, and black text. Then a revision can be done when the entire StatsView is redesigned.

### Current Behavior

Award Cards have incorrect color, size, and padding

### Expected Behavior

Award cards have the correct size, and either the original colors, or revised colors utilizing surface-1 and surface-2.

### Steps to Reproduce

1. Log in/sign up
2. Go to the stats view
3. Navigate to a season that has season champions on it e.g. Clubs 2022
4.Look at the award cards and their incorrect styles

Plubedam:
## 930
- Resolves  #930

## Please check the following

- For New Features:

## Additional info
Tried it with surface-1 and surface-2, it looks like this:
image


itsalaidbacklife:
Definitely better! How does it look if the left section (with the medals) is cream, and the right section (with the text) is chocolate (surface-1) with cream text (surface-2)?

src/components/AwardCard.vue
```diff
@@ -17,7 +17,7 @@
     </div>
     <div
       class="text d-flex flex-column align-center px-10 text-center justify-center"
-      :class="[isCard ? 'white' : 'cream']"
+      :class="[isCard ? 'chocolate background-cream py-5' : 'cream']"
```
Haviles04, line 20: I think we should be able to use Vuetifys utility classes here `text-surface1`

Plubedam:
> Definitely better! How does it look if the left section (with the medals) is cream, and the right section (with the text) is chocolate (surface-1) with cream text (surface-2)?

Here it is, I like how this one looks:
Captura de pantalla 2024-06-04 232656

itsalaidbacklife:
> > Definitely better! How does it look if the left section (with the medals) is cream, and the right section (with the text) is chocolate (surface-1) with cream text (surface-2)?
> 
> 
> 
> Here it is, I like how this one looks:
> 
> Captura de pantalla 2024-06-04 232656

Excellent, thank you. The second one shown here looks the best imo. Let's go with that. Nicely done!

src/components/AwardCard.vue
itsalaidbacklife: Specifically, Vuetify exposes `bg-{colorName}` and `text-{colorName}` classes out of the box, so we can do this:
```suggestion
      :class="[isCard ? 'bg-surface-2 text-surface-1 py-5' : 'cream']"
```

src/components/AwardCard.vue
```diff
@@ -5,7 +5,7 @@
     :data-tournament="`${place}${suffix}`"
     :variant="isCard ? 'elevated' : 'text'"
   >
-    <div class="medal-icon d-flex align-center" :class="[isCard ? 'white' : 'cream']">
+    <div class="medal-icon d-flex align-center" :class="[isCard ? 'white px-5 py-5' : 'cream']">
```
itsalaidbacklife, line 8: With the design change I think this container always has a surface-2 background. Similarly, I think the padding is probably fine to keep the same in the card + not-card contexts. Also 'pa' (short for padding-all) is a shorthand for px + py. So then this would simplify to 
```suggestion
    <div class="medal-icon d-flex align-center bg-surface-2 pa-5">
```

itsalaidbacklife (changes requested):
Nice work. The second iteration, using the cream background on the left looks best:
image


All that's left to do is to clean up & simplify the class usage, as the overlap in styles means we need fewer dynamic classes, and then to remove the css setting color, since that will be handled by the vuetify classes. 

Ultimately I think this will look even better than it did before the regression. Thanks for your contribution!

itsalaidbacklife (approved):
Nicely done. Thanks for exploring style alternatives. I think this landed in a good place

PR #971, opened 2024-06-18
Store API Call Refactor

Haviles04:
## Issue number

Relevant [issue number](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
- Resolves #972 

## Please check the following

- [x] Do the tests still pass? (see [Run the Tests](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#run-the-tests))
- [x] Is the code formatted properly? (see [Linting (Formatting)](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#linting-formatting))
- For New Features:
  - [ ] Have tests been added to cover any new features or fixes?
  - [ ] Has the documentation been updated accordingly?

## Please describe additional details for testing this change
- Created new reusable function
- Refactored old functions to utilize new reusable generic api call function

PR #973, opened 2024-06-20
Issue 938: Create feature flag for using new API

### LINKED ISSUE #938: [Feature]: Create feature flag for using new API
### Feature Summary

We need to define environment variables that serve as feature flags for enabling and disabling use of the new move-making API

### Detailed Description

This is a key piece of the [Move + GameStateHistory MVP](https://github.com/orgs/cuttle-cards/projects/8/views/1) project, as we are adopting the strategy of merging everything into main without utilizing them in production, so we need the ability to turn this entire feature set off for production deployments and CI.

Specifically, we need:
1) A new backend `gameStateAPIEnabled` policy which will be applied to all the new endpoints in order to disable them unless an environment variable e.g. `VITE_USE_GAMESTATE_API` is set to true
2) Frontend updates in the `src/stores/game.js` store to read from the `VITE_USE_GAMESTATE_API` env var to differentiate which api they call for each api-calling action. Initially, if the variable is set to true, every one of the in-scope actions (listed [here in the spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.axyx90fblcat)) should throw an error saying that this action is not yet supported in the GameState api. In subsequent issues, we will build these out, one at a time, in step with the backend implementation of each of these
3) Test code in `tests/e2e/support/commands.js` that differentially calls the new gamestate api when an env var `CYPRESS_USE_GAMESTATE_API` is set to true. All of the move-making api commands, plus `loadGameFixture` should be updated to throw an error saying they are unsupported in the GameState API when this variable is true (and otherwise to call the existing code)
4) Added commands to `package.json` that set the `VITE_USE_GAMESTATE_API` and `CYPRESS_USE_GAMESTATE_API` environement variables when booting, building, and opening cypress so that we can run with the new api enabled (which initially should throw errors in all the abovementioned places.

Plubedam:
## 938

- Resolves #938

## Please check the following

- For New Features:

## Notes
Maybe just `loadGameFixture` check is enough in commands.js?


src/stores/game.js
```diff
@@ -4,6 +4,15 @@ import { cloneDeep } from 'lodash';
...
 
+/**
+ * Throws an error if VITE_USE_GAMESTATE_API is enabled
+ */
+const checkGameStateApi = () => {
+  if (import.meta.env.VITE_USE_GAMESTATE_API === 'true') {
+    throw 'This action is not supported yet in GameState API';
+  }
+};
+
```
Haviles04, line 15: I don't believe we need to check for an error here anymore after updating the way the API is consumed in `makeSocketRequest`.

tests/e2e/support/commands.js
```diff
@@ -412,6 +424,9 @@ Cypress.Commands.add('playPointsSpectator', (card, pNum) => {
 });
 
 Cypress.Commands.add('playPointsById', (cardId) => {
+  if (Cypress.env('CYPRESS_USE_GAMESTATE_API')) {
+    throw new Error(gameStateApiError);
+  }
```
Haviles04, line 429: I think it may make sense to make a function similar to `makeSocketRequest` in the gamestore for these API calls instead of checking for an error like this.

It may be out of scope on this specific PR though

Curious on your thoughts @itsalaidbacklife , @seriouslysean

Haviles04 (commented):
Nice work on this. A couple of architectural thoughts I have but otherwise it's great to see this project moving!

src/stores/game.js
```diff
@@ -4,6 +4,15 @@ import { cloneDeep } from 'lodash';
...
 import { sleep } from '../util/sleep';
 
+/**
+ * Throws an error if VITE_USE_GAMESTATE_API is enabled
+ */
+const checkGameStateApi = () => {
+  if (import.meta.env.VITE_USE_GAMESTATE_API === 'true') {
+    throw 'This action is not supported yet in GameState API';
+  }
+};
```
itsalaidbacklife, line 14: While this is essentially what the issue states, with the refactoring introduced in #971, we have the opportunity to centralize the weirdness of maintaining the two apis into just the `makeSocketRequest()` function.

The unsolved nuance is how we manage api calls that have different forms to their urls in the GameState API.

All the move-making requests i.e. `draw`, `points` etc, are changing to use `POST /api/game/:gameId/move/<the_move_name>`

whereas the other endpoints related to games that are not move-making e.g. creating a game, joining, readying up, are not currently changing their urls. 

The end result of this PR should be that we can boot, build and run the tests in `VITE_USE_GAMESTATE_API` mode in that mode we can log in, create a game, join it, ready up and start the game, at which point the cards should deal using the old method of creating `card` records, but then we can't make any of the moves e.g. drawing etc because they error. The question now is how to manage making the non-move requests work as normal while making the move-making ones error in the correct fashion.

I think the mechanism that we want to use for causing the move-making requests to error is basically what's already in `makeSocketRequest()`: to conditionally request an updated URL and then the backend will simply 404 if that url isn't supported yet. Then the question is how to leave the other requests alone. The three ways I can think to do that are:

1. Add an argument to `makeSocketRequest()` that enables/disables the url transform and setting it appropriately in each caller
2. Updating `makeSocketRequest()` itself to check the requested `slug` against a known set of ones to mutate (all the move making slugs) and have that function itself determine whether to change the request based on whether the slug is found in its set of known ones to change
3. Splitting off a separate function for the transforming version of `makeSocketRequest()` e.g. `makeGameStateAPIRequest()`, putting the conditional logic in there, and calling `makeGameStateAPIRequest()` in the move-making functions and `makeSocketRequest()` in the others.


@Haviles04 @Plubedam penny for your thoughts. Once this is settled I can go through and update the issues to remove the section on fixing these frontend errors because it should come for 'free' with the backend implementing the appropriate endpoint.

itsalaidbacklife (commented):
Nice work. This is what was written up, but I think we can actually better handle the `game` store changes by centralizing the conditional logic for handling the env var (making different requests) using the `makeSocketRequest()` function. Details are in a comment.

tests/e2e/support/commands.js
seriouslysean: If I’d just for testing a specific action wit would be kinda weird. You could have a conditional action that’s only there for testing maybe?

Either way I agree with you though, to keep things clean. Definitely out of scope :) — separate your concerns.

src/stores/game.js
Plubedam: After giving it some thought, I decided to go with number 2.
```suggestion
const nonGameStateAPIMoves = new Set(['concede', 'stalemate', 'reject-stalemate', 'rematch']);
```
Instead of creating a Set of the move-making slugs, I used the non move-making ones so the list is much shorter.
```suggestion
      const usingGameStateAPI = import.meta.env.VITE_USE_GAMESTATE_API === 'true';
      let url = `/api/game/${slug}`;

      if (usingGameStateAPI && !nonGameStateAPIMoves.has(slug)) {
        url = `/api/gamestate/${this.id}/move/${slug}`;
        method = 'post';
      }
```
Then we check if the slug is not part of the Set and the env variable is true. Since all move-making slugs are POST we always set the `method`. After we implement all new GameState moves and it's time to normalize the rest of the routes, we probably should make the `method` parameter default to 'post', and define the minority of non POST endpoints `method` individually

seriouslysean: Doesn’t make sense to have a function per url, as getters essentially, that either return the new url or the old one? It’s a little more overhead to start but would allow you to opt in per route.

src/stores/game.js
```diff
@@ -366,10 +368,13 @@ export const useGameStore = defineStore('game', {
       }
     },
     makeSocketRequest(slug, data, method = 'get') {
-      const url =
-        import.meta.env.VITE_USE_GAMESTATE_API === 'true'
-          ? `/api/gamestate/${this.id}/move/${slug}`
-          : `/api/game/${slug}`;
+      const usingGameStateAPI = import.meta.env.VITE_USE_GAMESTATE_API === 'true';
```
seriouslysean, line 371: It's typically better to use the check by itself instead of assigning it to a variable. It aids in static analysis and tree shaking. I know we don't really do that in Cuttle but it's just good to consider longer term.

src/stores/game.js
```diff
@@ -366,10 +368,13 @@ export const useGameStore = defineStore('game', {
...
+      let url = `/api/game/${slug}`;
+  
+      if (usingGameStateAPI && !nonGameStateAPIMoves.has(slug)) {
```
seriouslysean, line 374: This block of logic feels like it should be pulled out in to a separate utility and tested to validate the proper call goes out for the specific APIs when the env is set.

src/stores/game.js
```diff
@@ -378,6 +383,9 @@ export const useGameStore = defineStore('game', {
             data,
           },
           (res, jwres) => {
+            if (usingGameStateAPI && res === 'Not Found') {
```
seriouslysean, line 386: It's pretty weird to statically check the raw string. There's a better way to do this, probably by either allow listing the api paths than have been converted and checking that first, or making functions to generate each api path based on the env to get the right path the first time.

seriouslysean (commented):
Left a few more comments while I was thinking about it this morning.

tests/e2e/support/commands.js
itsalaidbacklife: #974

itsalaidbacklife (approved):
Nicely done!

src/stores/game.js
seriouslysean: Any sort of tool that statically analyses your code, if we ever do this in the client code it would also tree shake the variables so it’s best to standardize between them. Additionally I’m working on transpiling the server, or at least moving it to ES6 syntax so it’s beneficial to understand these things now even it’s not explicitly necessary yet.

You’re correct in that it’s fine as written now, but I would prefer we pick a style and stick to it between both client and server.

src/stores/game.js
itsalaidbacklife: @seriouslysean interesting point. An issue with one function per url is that each caller needs to call their correct function. I think what might instead be ideal is to create a url-transforming function that inputs a slug and outputs the correct url with a big switch statement for managing the different cases. 

At the moment, that should work equivalently to what is written here, but then we could add urls to the switch case to handle different urls differently when we get there. Something like this.

```
function transformGameUrl() {
  if (import.meta.env.VITE_USE_GAMESTATE_API !== 'true') {
    return  `/api/game/${slug}`;
  }

  switch (slug) {
    case 'pass':
    case 'draw':
    // add all the move-making ones here
      return `api/game/${this.gameId}/move/${slug};
    default: 
      return  `/api/game/${slug}`;
  }
}

```

package.json
```diff
@@ -75,14 +75,20 @@
...
+    "start:server:gamestate": "cross-env VITE_USE_GAMESTATE_API=true node app.js",
     "start:preview": "npm run build;npm run start:server",
     "start:devtools": "npx vue-devtools",
     "start:dev": "concurrently \"npm run start:client\" \"npm run start:server\" \"npm run start:devtools\"",
+    "start:dev:gamestate": "concurrently \"npm run start:client:gamestate\" \"npm run start:server:gamestate\" \"npm run start:devtools\"",
     "start": "node app.js",
     "build": "vite build",
+    "build:gamestate": "cross-env VITE_USE_GAMESTATE_API=true vite build",
     "e2e:client": "cypress run",
+    "e2e:client:gamestate": "cypress run --env \"CYPRESS_USE_GAMESTATE_API=true\"",
```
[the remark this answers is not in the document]
Plubedam: Not really, I added those double quotes after reading the Cypress docs about env variables, but they are actually only advised when adding multiple variables and only for Windows PowerShell, so I'll remove them.

itsalaidbacklife (commented):
Beautiful:
image

PR #978, opened 2024-06-26
DevEx #974 Refactor Cypress Commands

### LINKED ISSUE #974: [DevEx]: Refactor cypress commands file to use a a centralized helper function for making socket requests
### Improvement Summary

We should consolidate the cypress commands so that the socket api calls are made with a utility function like the `game` store's `makeSocketRequest()` function. This should DRY up the commands file and afford the flexibility to more centrally manage dynamically utilizing the GameState API vs the legacy API during development.

### Detailed Description

We should employ the same maneuver used in #971 to DRY up all the `io.socket.get()` calls from the various cypress commands. We can then additionally use the technique laid out in #973 where the `nonGameStateAPIMoves` are defined in a `Set` that is used to programatically alter the gamestate request urls when the `VITE_USE_GAMESTATE_API` env var is set. 

This will allow us to more concisely control which endpoints hit which urls in both the GameState and legacy APIs, while shortening up the commands file.

Haviles04:
## Issue number

- Resolves #974

## Please check the following

- For New Features:

Added helper cypress command to re-use code and use dynamic URLs

### DIFF
--- a/tests/e2e/support/commands.js
+++ b/tests/e2e/support/commands.js
@@ -92,7 +92,7 @@ Cypress.Commands.add('loadSeasonFixture', (season) => {
 });
 
 Cypress.Commands.add('loadMatchFixtures', (matches) => {
-  cy.makeSocketRequest('test', 'loadMatchFixture', matches);
+  cy.makeSocketRequest('test', 'loadMatchFixtures', matches);
 });
 
 Cypress.Commands.add('loadFinishedGameFixtures', (games) => {
--- a/tests/e2e/support/commands.js
+++ b/tests/e2e/support/commands.js
@@ -1,15 +1,70 @@
 import { getCardIds, hasValidSuitAndRank, cardsMatch, printCard } from './helpers';
 import { myUser, opponentOne, playerOne, playerTwo } from '../fixtures/userFixtures';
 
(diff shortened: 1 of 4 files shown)


tests/e2e/support/commands.js
```diff
@@ -1,14 +1,70 @@
...
+        case 'seven/scuttle':
+        case 'seven/faceCard':
+        case 'seven/jack':
+        case 'seven/untargetedOneOff':
+        case 'seven/targetedOneOff':
+        case 'pass':
+          // add all the move-making ones here
+          return `api/game/${gameId}/move/${slug}`;
+        default:
+          return `/api/${api}/${slug}`;
```
itsalaidbacklife, line 42: Why does one have a preceding slash, but not the other?

Haviles04: I think we probably want the preceding slash.

This is a copy paste from the game store so we should probably change it there too

itsalaidbacklife (approved):
Looks good. Can we add the request methods from the Spec? All the move-making reqs should be POST. It should be okay to switch over to POST requests even for the non-gamestate api because the current backend matches urls regardless of the req method, so they should still match.

We might want to make POST the default req method in `makeSocketRequest` then update the consumers for non-POST requests

PR #984, opened 2024-07-23
Feat/deal cards gamestate

### LINKED ISSUE #939: [Feature]: Update `gameService.dealCards()` to use new GameState model
### Feature Summary

We should update the function that deals the cards/initializes the game state so that it utilizes the new GameState model.

### Detailed Description

`gameService.dealCards()` is the utility that initializes the game by creating and placing all the cards. We should update it to read from the `process.env.SAILS_USE_GAME_STATE_API` env var and if that is set to “true”, use a new sails helper to instantiate and save a new `GameState` associated with the specified game. This should make it so that we can start a game via ready and rematch.

itsalaidbacklife:
Implements initializing/dealing the cards using the GameState API

## Issue number

- Resolves #939 

## Please check the following

- For New Features:


api/helpers/gameStates/deal-cards.js
```diff
@@ -0,0 +1,73 @@
...
+        return exits.error({ message: 'Game has already started' });
+      }
+
+      const previousGameStates = await GameStateRow.find({ gameId: game.id});
+      if (previousGameStates?.length) {
+        return exits.error({ message: 'Cards are already dealt' });
+      }
+
+      const deck = _.shuffle(
+        DeckIds.map(sails.helpers.gameStates.convertCardIdToCard)
```
itsalaidbacklife, line 32: Helper name changed. Also need the .map() callback to pass the arg because sails helpers input a single object arg
```suggestion
        DeckIds.map((str) => sails.helpers.gameStates.convertStrToCard({ str }))
```

api/helpers/gameStates/deal-cards.js
```diff
@@ -0,0 +1,73 @@
...
+
+  fn: async ({ game }, exits) => {
+    
+    try {
+      if (game.status !== GameStatus.CREATED) {
+        return exits.error({ message: 'Game has already started' });
+      }
+
+      const previousGameStates = await GameStateRow.find({ gameId: game.id});
+      if (previousGameStates?.length) {
```
itsalaidbacklife, line 27: If we require/expect `game` to have populated its gameStates. But then what do we do if it's missing? Error? Or query for it?
```suggestion
      const previousGameStates = game.gameStates ?? await GameStateRow.find({ gameId: game.id});
      if (previousGameStates?.length) {
```

api/helpers/gameStates/deal-cards.js
```diff
@@ -0,0 +1,73 @@
...
+        return exits.error({ message: 'Cards are already dealt' });
+      }
+
+      const deck = _.shuffle(
+        DeckIds.map(sails.helpers.gameStates.convertCardIdToCard)
+      );
+
+      const p0 = {
+        hand: deck.splice(0, 5),
```
itsalaidbacklife, line 36: Pretty nifty to be able to call .splice() synchronously here to remove 5 cards from the deck and put them into the player hand in one line. This kind of move (and the ability to write it all to the db with a single insert and  function call) was a core motivator for the arch of the GameState API

api/helpers/gameStates/deal-cards.js
```diff
@@ -0,0 +1,73 @@
...
+        points: [],
+        faceCards: [],
+      };
+
+      const p1 = {
+        hand: deck.splice(0, 6),
+        points: [],
+        faceCards: [],
+      };
```
itsalaidbacklife, line 45:
```suggestion
      };

```

utils/DeckIds.json
```diff
@@ -0,0 +1,6 @@
+[
+  "AC", "2C", "3C", "4C", "5C", "6C", "7C", "8C", "9C", "TC", "JC", "QC", "KC",
+  "AD", "2D", "3D", "4D", "5D", "6D", "7D", "8D", "9D", "TD", "JD", "QD", "KD",
+  "AH", "2H", "3H", "4H", "5H", "6H", "7H", "8H", "9H", "TH", "JH", "QH", "KH",
+  "AS", "2S", "3S", "4S", "5S", "6S", "7S", "8S", "9S", "TS", "JS", "QS", "KS"
+]
```
itsalaidbacklife, line 6:
```suggestion
]

```

utils/MoveType.json
itsalaidbacklife, line 1: This should be reset to match main without changes

itsalaidbacklife (commented):
Writing a round of todos.

Also todo;
1) call dealCards() conditionally based on env var for gameState API in both ready and rematch
2) update lockGame() to populate gameStates

api/controllers/game/ready.js
```diff
@@ -31,7 +31,9 @@ module.exports = async function (req, res) {
       sails.sockets.blast('gameStarted', { gameId: game.id });
       gameUpdates.status = gameService.GameStatus.STARTED;
       // Deal cards (also emits socket event)
-      await gameService.dealCards(game, gameUpdates);
+      await Game.updateOne({ id: game.id }).set(gameUpdates);
+
+      await gameService.dealCards({...game, ...gameUpdates}, {});
```
itsalaidbacklife, line 36: Because dealing the cards and  other logic for starting the game require updating the Game row, the legacy utility for dealing the card's inputs a dictionary of changes to apply during the dealCards() function, so that the Game can be updated once instead of twice. 

Since the GameState API version of `dealCards()` just creates a new (initial) GameState, it doesn't require updating the Game, so it makes sense to update the Game's status and the p0Ready/p1Ready columns *before* calling `dealCards()`.

This has the awkward side effect, at least in the way I'm doing it now, of making two game updates in the legacy (actual production) environment, since now we're updating the game here and then again. It feels awkward, and I'm not sure what's the best way to resolve it. 

It might be cleaner if we managed the update in gsmeService.dealCards() in the section where we bifurcate on the env var

api/controllers/game/subscribe.js
```diff
@@ -40,7 +40,14 @@ module.exports = function (req, res) {
       const addPlayerToGame = Game.addToCollection(game.id, 'players').members([user.id]);
       const updatePlayer = User.updateOne({ id: user.id }).set({ pNum });
 
-      return Promise.all([game, updatePlayer, addPlayerToGame]);
+      // Set game.p0 or game.p1 as requesting user's id
+      const pNumUpdate = {};
+      const pNumKey = `p${pNum}`;
+      pNumUpdate[pNumKey] = user.id;
+      game[pNumKey] = user.id;
+      const updateGame = Game.updateOne({id: game.id}).set(pNumUpdate);
```
itsalaidbacklife, line 48: Same idea with leave-lobby: we're now setting p0 and p1 upon joining the lobby instead of when the game starts in the ready action. Eventually this will deprecate the need for the players collection

api/helpers/gamestate/deal-cards.js
```diff
@@ -0,0 +1,80 @@
...
+      required: true,
+    },
+  },
+
+  fn: async ({ game }, exits) => {
+    
+    try {
+
+      if (game.status !== GameStatus.STARTED) {
+        return exits.error({ message: 'Game has not yet started or is over' });
```
itsalaidbacklife, line 25: Todo: check the UX of these errors. They shouldn't generally occur through the client, but if they do, a snackbar would be appropriate

api/helpers/gamestate/deal-cards.js
```diff
@@ -0,0 +1,80 @@
...
+        hand: deck.splice(0, 5),
+        points: [],
+        faceCards: [],
+      };
+
+      const p1 = {
+        hand: deck.splice(0, 6),
+        points: [],
+        faceCards: [],
+      };
```
itsalaidbacklife, line 50: this is the beauty of the GameState API. Just manipulating arrays in memory before saving and publishing

api/helpers/gamestate/deal-cards.js
```diff
@@ -0,0 +1,80 @@
...
+        playedBy: null,
+        playedCard: null,
+        targetCardId: null,
+        discardedCards: [],
+        oneOff: null,
+        oneOffTarget: null,
+        resolving: null,
+      };
+
+      await sails.helpers.gamestate.saveGamestate(newGameState);
```
itsalaidbacklife, line 71: Todo: I'd like to rename the `gamestate` folder to `gameStates`

api/helpers/gamestate/deal-cards.js
```diff
@@ -0,0 +1,80 @@
...
+
+      await sails.helpers.gamestate.publishGameState(game, newGameState);
```
Haviles04, line 73: Kinda preference, but we can destructure these so we can just call `saveGameState()`... `publishGameState`...`convertStrToCard`

api/helpers/gamestate/validate-gamestate.js
```diff
@@ -33,8 +33,8 @@ function validatePhase(param) {
 }
 
 function validatePlayedBy(param) {
-  if (param === null || param === undefined || Number.isNaN(param) || param < 0 || param > 1) {
-    throw new Error('The playedBy attribute cannot be null and must be a number in [0,1]');
+  if (![null, 0, 1].includes(param)) {
```
itsalaidbacklife, line 36: It makes the most sense for `playedBy` to be null for the first gamestate (initialize) since neither player "did" anything. That said alternatively we could say it's p1 because they are the "dealer"

api/helpers/gamestate/deal-cards.js
```diff
@@ -0,0 +1,80 @@
...
+        return exits.error({ message: 'Game was not populated with .gameStates collection' });
+      }
+
+      if (game.gameStates.length) {
+        return exits.error({ message: 'Cards are already dealt' });
+      }
+
+      const deck = _.shuffle(
+        DeckIds.map((cardId) => sails.helpers.gamestate.convertStrToCard(cardId))
+      );
```
Haviles04, line 38: This is really cool, it's a one line function, but it does sooooooo much

Haviles04 (commented):
Awesome stuff, Can't wait to see this merged, it's a big one for creating games!!

api/helpers/lock-game.js
```diff
@@ -34,15 +34,22 @@ module.exports = {
...
 
         // If we successfully wrote our uuid, resolve
         const newLock = updatedGame?.lock;
         if (newLock === uuId) {
+
+          // Re-fetch game and populate players and gamestates
+          updatedGame = await Game.findOne({ id: gameId })
+            .populate('p0')
+            .populate('p1')
+            .populate('gameStates', { sort: 'createdAt ASC' });
```
itsalaidbacklife, line 51: Updating `lockGame()` to include these associations so that consumers can exclusively rely on it for game queries instead of having to requery for associations after locking the game. Unfortunately it still requires an extra find query (in addition the `Game.updateOne()` above which also fetches after update), but at least that's abstracted into this helper.

The upshot is that endpoints that need to grab an individual game can lock it and have all the data they need to complete the request

api/models/GameStateRow.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
     playedBy: {
       type: 'number',
       isIn: [0, 1],
-      required: true,
+      allowNull: true,
```
itsalaidbacklife, line 32: Same as above. Need to decide if this should be null or 1 for the first gamestate where the cards are dealt

utils/DeckIds.json
itsalaidbacklife, line 6: It seemed like it might be helpful to have a representation of the full deck somewhere, and it's useful for creating the deck when we deal.

the other use I have in mind is that we could update validateGameState() to verify that all these cards exist exactly once with no extras

api/services/gameService.js
```diff
@@ -232,6 +232,11 @@ module.exports = {
   },
 
   dealCards: function (game, gameUpdates) {
+    // Use gamestate api if feature flag is set, otherwise legacy
+    if (process.env.VITE_USE_GAMESTATE_API) {
+      return sails.helpers.gamestate.dealCards(game);
+    }
+
```
itsalaidbacklife, line 239: Dynamically use either gamestate api version or legacy based on feature flag. When we finalize the project we can/should delete this function altogether and directly call the gamestate helper in the consumers

itsalaidbacklife (commented):
First round of explanatory comments and todos

api/helpers/gamestate/validate-gamestate.js
itsalaidbacklife: Update: opting to make it p1 the player who 'deals' and to rename the 'INITIALIZE` moveType to 'DEAL'

api/models/GameStateRow.js
itsalaidbacklife: Keeping it required and setting it to 1 for the DEAL move as p1 deals

api/helpers/gameStates/deal-cards.js
itsalaidbacklife: Opted to throw because the intention is that it should always be an empty array when hitting this function and if not that indicates a problem

api/helpers/gamestate/deal-cards.js
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/1002

utils/MoveType.json
```diff
@@ -1,5 +1,5 @@
 {
-  "INITIALIZE": "initialize",
+  "DEAL": "deal",
```
itsalaidbacklife, line 2: Renamed `initialize` to `deal` to be more thematically consistent with the idea that p1 is the dealer and the first gamestate happens when p1 deals.

tests/unit/fixtures/gameStates/points.js
```diff
@@ -137,10 +137,10 @@ export const points = {
             lockedAt: null,
             match: null,
             name: 'Test Game',
-            p0: gameFixture.p0,
+            p0: gameFixture.p0.id,
```
itsalaidbacklife, line 140: updated the way matches come throught to the client so `p0` and `p1` on all the individual games are just the player ids, not the objects

api/helpers/gamestate/deal-cards.js
itsalaidbacklife: I've updated the errors so they flow better through the endpoints, but the lobby doesn't have an error snackbar. Let's write a separate issue to add one

api/controllers/game/ready.js
itsalaidbacklife: I explored reworking this to save game updates in the same helper, but either way we'll need to rework rematch to make this work, so let's set that aside until we get to rematch

api/helpers/get-rematch-games.js
```diff
@@ -1,6 +1,13 @@
 // Pull the logic into func outside fn to allow recursion w/o bumping into weirdness w/ sails' exits
 async function getRematchGames (game, priorRematchGames = []) {
-  const rematchGames = [game, ...priorRematchGames];
+  const rematchGames = [game, ...priorRematchGames].map((matchGame) => {
+    return {
+      ...matchGame,
+      // Get p0 and p1 id, whether populated or not
+      p0: matchGame.p0?.id ?? matchGame.p0,
```
seriouslysean, line 7: Should these be added as keys if they don’t exist? I don’t how or when that case would occur though.

api/helpers/lock-game.js
```diff
@@ -34,15 +34,22 @@ module.exports = {
         const now = dayjs.utc().toDate();
         const lockIsStaleTimeout = dayjs.utc().subtract(30, 'second').toDate();
 
-        // Lock & re-fetch game if unlocked or lock is expired
-        const updatedGame = await Game.updateOne({
+        // Lock game if unlocked or lock is expired
+        let updatedGame = await Game.updateOne({
```
seriouslysean, line 38: Do you find the same game here? I don’t see why you need a let, it should be updated game and refreshed updated game or skipping the variable assignment for the first one altogether since you just need to await the call.

src/plugins/sockets/inGameEvents.js
```diff
@@ -34,7 +34,7 @@ export async function handleInGameEvents(evData) {
       gameStore.updateReady(evData.pNum);
       return;
     }
-    case SocketEvent.INITIALIZE: {
+    case SocketEvent.DEAL: {
```
seriouslysean, line 37: These should also be constants that we can reuse.

seriouslysean (approved):
Looks good, but I didn’t run it. Few questions before this goes in, though.

api/controllers/game/rematch.js
```diff
@@ -15,12 +15,12 @@ module.exports = async function (req, res) {
     let game =  await sails.helpers.lockGame(req.session.game);
 
     // Early return if requesting user was not in the game
-    if (![game.p0, game.p1].includes(userId)) {
+    if (!userId || ![game.p0?.id, game.p1?.id].includes(userId)) {
```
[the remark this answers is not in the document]
itsalaidbacklife: Agreed. Seeing more use cases will help to clarify whether how it should return, whether to throw, etc. but it will be a common check

api/helpers/gamestate/deal-cards.js
```diff
@@ -0,0 +1,80 @@
+const GameStatus = require('../../../utils/GameStatus.json');
```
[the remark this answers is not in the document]
itsalaidbacklife: Like aliasing directories similar to what we have in vite? That could be convenient. We'd have to wrangle a separate config/setup from vite since this is backend, but it could be done

api/helpers/gamestate/deal-cards.js
```diff
@@ -0,0 +1,80 @@
...
+        return exits.error({ message: 'Game has not yet started or is over' });
+      }
+
+      if (!game.gameStates) {
+        return exits.error({ message: 'Game was not populated with .gameStates collection' });
+      }
+
+      if (game.gameStates.length) {
+        return exits.error({ message: 'Cards are already dealt' });
+      }
```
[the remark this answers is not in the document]
itsalaidbacklife: Particularly because these are not yet consumed on the frontend. There is no error UX in place for dealing the cards, so that deserves its own issue

api/helpers/get-rematch-games.js
itsalaidbacklife: I think we always want to p0 and p1 properties to exist, but to be nullable. We shouldn't hit this part and have them missing because they are set when users join the game and this is going through finished games, but just in case I think it's better that the keys exist so games can be assumed to have a p0 and p1 attribute even if null (meaning there is no player)

src/plugins/sockets/inGameEvents.js
itsalaidbacklife: Isn't it already? This is imported as a reusable enum. 

That said, the weirdness imo is the heavy overlap with MoveType. They aren't exactly the same because SocketEvent includes some event names that aren't actually GameStates eg conceding, a new rematch game being created. So it feels weird to have them as two separate imports, but MoveType doesn't allow all the values used here, so I'm not sure how to align them. One move would be to have SocketEvent extend MoveType by adding extra properties

seriouslysean: I don’t even think I noticed it was imported lol, but I agree with the extend idea if it truly is a subset.

api/controllers/game/ready.js
```diff
@@ -31,7 +31,9 @@ module.exports = async function (req, res) {
       sails.sockets.blast('gameStarted', { gameId: game.id });
       gameUpdates.status = gameService.GameStatus.STARTED;
       // Deal cards (also emits socket event)
-      await gameService.dealCards(game, gameUpdates);
+      await Game.updateOne({ id: game.id }).set(gameUpdates);
```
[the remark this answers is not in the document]
itsalaidbacklife: Yeah, this is the [sails syntax for updating a record](https://sailsjs.com/documentation/reference/waterline-orm/models/update-one). `Model.updateOne()` inputs a dictionary describing which record to update (the one with an `id` matching `game.id`), and then it has a `.set()` takes a dictionary describing what changes to make to the one that matches

api/controllers/game/get-list.js
```diff
@@ -28,7 +28,7 @@ module.exports = async function (req, res) {
         Game.publish(
           [req.session.game],
           {
-            change: 'initialize',
+            change: 'deal',
```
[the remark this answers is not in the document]
itsalaidbacklife: They should be MoveTypes. This will take some more sweeping backend cleanup since atm the constants are only used clientside and in the GameState API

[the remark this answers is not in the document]
itsalaidbacklife: Most of these use cases will be removed altogether when we roll out the GameState API. I think we can clean up the stragglers downstream

src/plugins/sockets/inGameEvents.js
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/1004

api/helpers/gamestate/deal-cards.js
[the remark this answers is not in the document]
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/1008

api/helpers/lock-game.js
itsalaidbacklife: The reason I had it this way is because `Model.updateOne()` returns the updated record, but not its associations (in this case `p0` and `p1`). We need the record itself to verify that the lock is set, and what I had was checking that first, and if the lock matches, _then_ querying again to populate the other associations.

What I could do instead would be skip capturing it in the `updateOne()` call itself and then immediately re-query without checking the lock to grab the associations. I think it might read more clearly that way. In the case where we fail to acquire the lock (basically because we hit multiple competing concurrent requests), this means we would be running an extra query for the Game + associations even though we were going to throw it away to try again. I think that's probably okay for clarity's sake

PR #985, opened 2024-07-26
Fix vittest document error

seriouslysean:
## Issue number

- Resolves #

## Please check the following

- For New Features:


itsalaidbacklife (approved):
Looks like this fixes the issue of document being undefined. Peculiar how it cropped up, and that it is so inconsistent, but I will take a semi magical fix for the moment, especially given that the test in question is something that we'll see pretty clearly in development if the devtools really break

PR #986, opened 2024-07-27
Unit test Devtools Bugfix

Haviles04:
## Issue number

- Resolves #

## Please check the following

- For New Features:

### DIFF
--- a/utils/config-utils.js
+++ b/utils/config-utils.js
@@ -1,5 +1,5 @@
-import { version } from '_/package.json';
 import devtools from '@vue/devtools';
+import { version } from '_/package.json';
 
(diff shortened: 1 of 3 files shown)


utils/config-utils.js
```diff
@@ -1,7 +1,6 @@
-import devtools from '@vue/devtools';
 import { version } from '_/package.json';
 
-export function initCuttleGlobals(app) {
+export async function initCuttleGlobals(app) {
```
itsalaidbacklife, line 3: I think if we do this we should also look to callers of this function and make them `await` it so that we avoid race condition shenanigans

itsalaidbacklife (approved):
Weird problem; nice fix

PR #990, opened 2024-08-12
Feat #946 Playing Points

### LINKED ISSUE #946: [Feature]: Build Playing Points in the new GameState API
### Feature Summary

We need to implement playing points in the GameState API in order to support the full game logic in this API. See the [spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.wpeuh0qc6ps2) for a general overview of the API, its underlying motivations, and the development approach.

### Detailed Description

Implementation should take the general approach shown in the [Backend Control Flow section of the GameState API Spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.vgf103gono5r) and follow the example of issue #942.

The endpoint should be:

POST /api/game/:gameId/move/points
```
{ playedCardId: '4H'}; // playing the 4H for points. Should be a valid card id per the spec doc
```

In addition to building the endpoint, the frontend consumption of it in the src/stores/game.js file's `requestPlayPoints()` function should be updated to make a request to this endpoint (instead of throwing an error) when the `process.env.VITE_USE_GAME_STATE_API` var equals `"true"`.

The same should be applied to the `playPointsOpponent()` and `playPointsSpectator()` test commands in tests/e2e/support/commands.js when the `CYPRESS_USE_GAMESTATE_API` env var is set to `"true"`. If so, (instead of erroring), they should send a request to the new api endpoint.

Haviles04:
## Issue number

- Resolves #946 

## Please check the following

- For New Features:

-Creating re-usabled endpoint for all moves
-Created helpers for playing points


api/controllers/gamestate/points.js
```diff
@@ -0,0 +1,25 @@
...
+  try {
+    const game = await sails.helpers.lockGame(req.params.gameId);
+    const currentState = sails.helpers.unpackGameState(game.gameStates[game.gameStates.length - 1]);
+    const requestedMove = sails.helpers.moves.points.parseRequest(req);
+
+    // Validate move -- Throws if move is illegal e.g. not your turn
+    sails.helpers.moves.points.validate(currentState, requestedMove);
+
+    // Move cards around, increment turn, etc
+    const updatedState = sails.helpers.moves.points.execute(currentState, requestedMove);
```
itsalaidbacklife, line 13: As I look more closely, I see that these are the only three lines that are specific to the particular move made, and only because of the prefix on the helper names. We could generalize this into a single endpoint for all moves and DRY up the code (while being more RESTful, too)
```suggestion
    const { parseRequest, validate, execute } = sails.helpers.moves[req.body.moveType];
    const requestedMove = parseRequest(req);

    // Validate move -- Throws if move is illegal e.g. not your turn
    validate(currentState, requestedMove);

    // Move cards around, increment turn, etc
    const updatedState = execute(currentState, requestedMove);
```

This will require two other changes outside this file:
1) we need to apply a policy to this endpoint that verifies that req.body has a moveType and that it's valid (ie one of the values in MoveType)
2) we need to update the route for this endpoint and all consumers to instead be simply POST /api/game/:gameId/move (removing /points etc) and expect that the req body include the move type

itsalaidbacklife (commented):
Great work. This makes me want to generalize the move endpoint into a single endpoint for all moves

api/helpers/gamestate/moves/points/execute.js
```diff
@@ -0,0 +1,45 @@
...
+     */
+    requestedMove: {
+      type: 'ref',
+      description: 'The move being requested. Specifies which player is asking to play a card for points',
+    },
+  },
+  sync: true, // synchronous helper
+  fn: ({ currentState, requestedMove }, exits) => {
+    const { playedBy, cardPlayed } = requestedMove;
+    let result = _.cloneDeep(currentState);
```
Haviles04, line 29: We could alternatively use `structuredClone` here instead of having to import lodash

seriouslysean: Just make sure it exists in the node version we're using.

Haviles04: Oohhh, good call I didn't even think to check that.

We're running 18, and structuredClone was added in 17

api/helpers/gamestate/moves/points/validate.js
```diff
@@ -0,0 +1,52 @@
+// api/helpers/gameState/moves/points/validate.js
+const GamePhase = require('../../../../../utils/GamePhase.json');
+
+module.exports = {
+  friendlyName: 'Parse points request',
+
+  description: 'Returns needed values to make points requests',
+
+  inputs: {
+    requestedMove: {
```
itsalaidbacklife, line 10: Can we comment its shape?

api/helpers/gamestate/moves/points/validate.js
```diff
@@ -0,0 +1,52 @@
...
+      descriptions: 'Object containing the current game state',
+      required: true,
+    },
+  },
+  sync: true, // synchronous helper
+  fn: ({ requestedMove, gameState }, exits) => {
+    const cardPlayed = gameState[`${requestedMove.playedBy}Hand`].find(requestedMove.cardPlayed);
+
+    if (gameState.phase !== GamePhase.MAIN) {
+      throw { message: 'Cannot play points during this phase' };
```
itsalaidbacklife, line 32: Todo: move into i18n

api/helpers/gamestate/moves/points/validate.js
```diff
@@ -0,0 +1,52 @@
...
+     * @param { MoveType.POINTS } requestedMove.moveType - Specifies that this a Points move
+     */
+    currentState: {
+      type: 'ref',
+      descriptions: 'Object containing the current game state',
+      required: true,
+    },
+  },
+  sync: true, // synchronous helper
+  fn: ({ requestedMove, gameState }, exits) => {
```
itsalaidbacklife, line 28: We should add a check that the requesting user is in the game. Should that go here, in parserequest, or in post move?

itsalaidbacklife: I think it should go in the move endpoint so that we only need it in one place, since it's a requirement of every move

api/policies/hasMoveType.js
```diff
@@ -0,0 +1,9 @@
+const moveType = require('../../utils/MoveType.json');
+
+module.exports = function (req, res, next) {
+  if (Object.keys(moveType).some((key) => moveType[key] === req.body.moveType)) {
```
itsalaidbacklife, line 4: This creates an array from an object then iterates through the keys, when looking up the key in an object is already a constant time operation.

Three constant time alternatives:

```suggestion
  if (req.body.moveType in MoveType)) {
```


```suggestion
  if (Object.hasOwn(MoveType, req.body.moveType) {
```


```suggestion
  if (MoveType[req.body.moveType]) {
```

Honestly I'm not sure which is best, but I think any of these could be better. I think `in` is the clearest, though it has the theoretical disadvantage of "counting" any properties in the prototype chain as opposed to ones defined in the obj itself, but honestly I think that's fine

itsalaidbacklife: @seriouslysean curious if you have thoughts on the best way to check an obj for a key

config/policies.js
```diff
@@ -75,6 +75,9 @@ module.exports.policies = {
   ],
   'game/seven/untargeted-one-off': ['isLoggedIn', 'isInGame', 'hasCardId'],
 
+  //GameStateApi
+  'game/:gameid/move': ['isLoggedIn', 'isInGame', 'hasMoveType'],
```
itsalaidbacklife, line 79: I think we should remove the isInGame policy here in favor of a check in the move making endpoint.  


The main reason is because I think the check in being in the game should be a db read on the game obj matching your user id in p0 or p1. (Currently it's just a check if your redis session, which is easier to desync and is less conducive to being in multiple games). 

We already query the game in the endpoint so it is inefficient to query for it in the policy as well. An alternative could be to do the check in the policy but make that the place where we actually query for the game, appending the locked game into the `req` object so the endpoint can access it as `req.game`. That way we still only query once but the logic for verifying access is kept in the policy which is arguably more reusable.

It would likely be more concise but also arguably more magical since the endpoint would assume a query had been made in the policy that you don't see directly in the endpoint code. That makes me lean towards just checking in the endpoint itself

itsalaidbacklife (changes requested):
Great work; this is really coming along.

The testing PR should be merged first. Then we should execute the points test and include that in the scope of the gamestate api CI tests.

Do you think that should be a separate issue? It requires separating basicMoves specs into its own directory and and making a basicMoves/points.spec file that we include in the CI run, and implementing the loadGameFixture endpoint for the GameState api

api/helpers/gamestate/moves/points/execute.js
```diff
@@ -0,0 +1,45 @@
...
+    const player = playedBy ? result.p1 : result.p0;
+    const cardIndex = player.hand.findIndex((id) => id === cardPlayed);
+
+    player.points.push(...player.hand.splice(cardIndex, 1));
+    result.turn++;
+
+    result = {
+      ...result,
+      ...requestedMove,
+      cardPlayed: [...player.points].pop(),
```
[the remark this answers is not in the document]
Haviles04: My thought was that using the spread operator inside a new array wont mutate the original

api/controllers/game/move.js
```diff
@@ -0,0 +1,19 @@
+// api/controller/game/:gameId/move
+// Request to make a move
+module.exports = async function (req, res) {
+  try {
+    const { validate, execute, parseRequest } = sails.helpers[req.body.moveType];
+    const game = await sails.helpers.lockGame(req.params.gameId);
+    const gameState = sails.helpers.unpackGameState({ gameStateRow: [...game.gameStates.pop()] });
```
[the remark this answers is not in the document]
Haviles04: this should be `[...game.gameStates].pop()`

api/policies/hasMoveType.js
seriouslysean: Out of those options, I personally prefer `MoveType[req.body.moveType]`. It’s O(1) complexity which means it’s a constant no matter how many items are in the collection. Even if it’s not going to be a large collection we want things to be as efficient as possible.

Additionally, with the loop the way you have it you’re looking at O(n) complexity, which even though isn’t bad it’s less efficient the more items in the array.

That said, the in operator is the fastest since it’s only checking the existence and not actually accessing the property directly. See https://jsbench.me/rim04kjqvx/1.

Either way, always prefer direct key access/validation if possible.

Haviles04: So the problem here is that the value will be in the gamestaterow, not the key

seriouslysean: Is there any way to better align them? Short of that you can make a mapping to lookup the key name, or send it along with the body. Any of those would be better than a loop in my opinion.

itsalaidbacklife: I didn't realize initially that the issue was in needing a reverse-mapping. We could update the json values to equal the keys so they could be used interchangeably and then update the socket event handlers + all legacy endpoints to match. Then this becomes a dictionary lookup. Honestly it's such a small juice to squeeze ratio that I'm fine with iterating as long as it's more succinct. 

It should be `.values().includes()`

itsalaidbacklife: I think creating an inverse mapping variable feels a little awkward because we have to keep them in sync. And programmatically generating an inverse mapping variable requires looping anyway. So imo this is okay

itsalaidbacklife (commented):
Getting there. Another round of comments

api/controllers/game/move.js
```diff
@@ -0,0 +1,22 @@
+// Request to make a move
+module.exports = async function (req, res) {
+  try {
+    const { saveGamestate, publishGameState, unpackGamestate } = sails.helpers.gamestate;
+    const { execute } = sails.helpers.gamestate.moves[req.body.moveType];
+    const game = await sails.helpers.lockGame(req.params.gameId);
+    const gameState = unpackGamestate(game.gameStates.at(-1));
+
+    if (gameState.turn % 2 !== req.body.playedBy) {
```
Haviles04, line 9: This will be replaced by the validate function

api/controllers/game/move.js
```diff
@@ -0,0 +1,18 @@
+// Request to make a move
+module.exports = async function (req, res) {
+  try {
+    const { saveGamestate, publishGameState, unpackGamestate } = sails.helpers.gamestate;
+    const { execute, validate } = sails.helpers.gamestate.moves[req.body.moveType];
```
itsalaidbacklife, line 5: We need to verify that the requesting player is in the game. Instead of it sending `playedBy` in the req.body, this needs to be verified in the db. Essentially if 
```
req.session.usr === game.p0
```

Then `playedBy` should be p0, same goes for p1. But if the requesting user is neither, then we error with you are not a player in this game (there's an existing error for it in the legacy endpoints)

api/helpers/gamestate/moves/points/validate.js
```diff
@@ -0,0 +1,62 @@
...
+      type: 'ref',
+      description: 'Object containing data needed for current move',
+      required: true,
+    },
+    /**
+     * @param {Object} requestedMove - Object describing the request to play points
+     * @param {1 | 0} requestedMove.playedBy - Which player is playing
+     * @param { Card } requestedMove.cardPlayed - Card Played for points
+     * @param { MoveType.POINTS } requestedMove.moveType - Specifies that this a Points move
+     */
```
itsalaidbacklife, line 19: Move this above the requestedMove input

api/helpers/gamestate/moves/points/validate.js
```diff
@@ -0,0 +1,62 @@
...
+    currentState: {
+      type: 'ref',
+      descriptions: 'Object containing the current game state',
+      required: true,
+    },
+    lock: {
+      type: 'string',
+      descriptions: 'UUID Lock',
+      required: true,
+    },
```
itsalaidbacklife, line 29: We can remove this once we set the  unlock operation to happen in the endpoint 
```suggestion
```

api/helpers/gamestate/moves/points/validate.js
```diff
@@ -0,0 +1,62 @@
...
+  },
+  fn: async ({ requestedMove, currentState, lock }, exits) => {
+    try {
+      const playedBy = requestedMove.playedBy ? 'p1' : 'p0';
+
+      const cardPlayed = currentState[playedBy].hand.find(({ id }) => id === requestedMove.cardId);
+
+      if (currentState.phase !== GamePhase.MAIN) {
+        throw new Error('Cannot play points during this phase');
```
itsalaidbacklife, line 38: Let's be more specific
```suggestion
        throw new Error(`Can only play points in main phase, not ${gameState.phase});
```

api/helpers/gamestate/moves/points/validate.js
```diff
@@ -0,0 +1,62 @@
...
+        throw new Error('game.snackbar.global.cardFrozen');
+      }
+
+      if (currentState.turn % 2 !== requestedMove.playedBy) {
+        throw new Error('game.snackbar.global.notYourTurn');
+      }
+
+      if (cardPlayed.rank > 10) {
+        throw new Error('game.snackbar.points.numberOnlyForPoints');
+      }
```
itsalaidbacklife, line 55:
```suggestion
      }

```

api/policies/hasMoveType.js
itsalaidbacklife, line 1: I think we should update this into a super policy for validating that the req body has a valid moveType and corresponding inputs. We can switch on the MoveType, error on `default` to detect illegal values, then for each explicitly allowed case, check `req.body` for the required inputs eg playedCard for points

itsalaidbacklife: Correspondingly we can call it something to the effect of hasValidMoveBody

config/policies.js
```diff
@@ -75,6 +75,9 @@ module.exports.policies = {
   ],
   'game/seven/untargeted-one-off': ['isLoggedIn', 'isInGame', 'hasCardId'],
 
+  //GameStateApi
+  'game/move': ['isLoggedIn'],
```
itsalaidbacklife, line 79: We should also apply the req body validation policy

src/stores/game.js
```diff
@@ -515,7 +515,7 @@ export const useGameStore = defineStore('game', {
       await this.makeSocketRequest('draw');
     },
     async requestPlayPoints(cardId) {
-      await this.makeSocketRequest('points', { cardId });
+      await this.makeSocketRequest('points', { cardId, moveType: 'points', playedBy: this.myPNum });
```
itsalaidbacklife, line 518: playedBy will be computed based on the requesting user's session and won't be included in the req body
```suggestion
      await this.makeSocketRequest('points', { cardId, moveType: 'points' });
```

Also nitpick but I like the convention of putting the moveType first since that will always be present and dictates what else goes in

tests/e2e/support/commands.js
```diff
@@ -305,7 +305,7 @@ Cypress.Commands.add('playPointsOpponent', (card) => {
         );
       }
       const cardId = foundCard.id;
-      cy.makeSocketRequest('game', 'points', { cardId });
+      cy.makeSocketRequest('game', 'points', { cardId, playedBy: opponent.pNum, moveType: 'points' });
```
itsalaidbacklife, line 308: Here as well: moveType first and remove playedBy

itsalaidbacklife (changes requested):
Excellent work. A few important pieces left, but none of them complex or requiring backpedaling. This has come together beautifully

api/helpers/gamestate/moves/points/validate.js
```diff
@@ -0,0 +1,62 @@
...
+      if (!cardPlayed) {
+        throw new Error('game.snackbar.global.playFromHand');
+      }
+
+      if (cardPlayed.isFrozen) {
+        throw new Error('game.snackbar.global.cardFrozen');
+      }
+
+      if (currentState.turn % 2 !== requestedMove.playedBy) {
+        throw new Error('game.snackbar.global.notYourTurn');
```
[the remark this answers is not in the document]
Haviles04: This current works

api/helpers/gamestate/moves/points/validate.js
Haviles04: do we still want this comment? it would just be the req.body now

api/controllers/game/move.js
```diff
@@ -0,0 +1,36 @@
...
+    const gameStateRow = await saveGamestate(updatedState);
+    game.gameStates.push(gameStateRow);
+    await publishGameState(game, updatedState);
+    await sails.helpers.unlockGame(game.lock);
+
+    return res.ok();
+  } catch (err) {
+    //unlock game if failing due to validation
+    if (game?.lock) {
+      await sails.helpers.unlockGame(game.lock);
```
itsalaidbacklife, line 32: We should probably handle possible errors here, just for good measure. Maybe try/catch it and do nothing in the catch so we fall through to the existing response?

api/helpers/gamestate/moves/points/validate.js
itsalaidbacklife: Yes because req.body is the value but it's still helpful to know the shape of data expected in order to understand what you're looking at

itsalaidbacklife (approved):
Beautifully done

api/controllers/game/move.js
```diff
@@ -0,0 +1,44 @@
...
+  try {
+    const { saveGamestate, publishGameState, unpackGamestate } = sails.helpers.gamestate;
+    const { execute, validate } = sails.helpers.gamestate.moves[req.body.moveType];
+    game = await sails.helpers.lockGame(req.params.gameId);
+    const gameState = unpackGamestate(game.gameStates.at(-1));
+    if (!game.gameStates.length || !gameState) {
+      throw new Error({ message: 'Game has not yet started' });
+    }
+
+    //Check player exist, and is in game
```
itsalaidbacklife, line 13:
```suggestion
    // Verify whether user is in requested game and as which player
```

api/controllers/game/move.js
```diff
@@ -0,0 +1,44 @@
...
+    switch (req.session.usr) {
+      case game.p0.id:
+        playedBy = 0;
+        break;
+      case game.p1.id:
+        playedBy = 1;
+        break;
+      default:
+        throw new Error('You are not a player in this game!');
+    }
```
itsalaidbacklife, line 24:
```suggestion
    }

```

itsalaidbacklife (approved):

PR #992, opened 2024-08-19
GH Actions GameState Config

### LINKED ISSUE #963: [Feature]: Update CI runs to test twice: with and without the `USE_GAMESTATE_API` feature flags
### Feature Summary

Once we have completed the feature set for the GameState API, we should update CI so that we test the application twice: one build and test config that uses the GameState API, and one that uses the Legacy API.

### Detailed Description

This way we can determine if we have successfully completed all the features such that the app is fully usable in both modes, before we clean up and switch over to the GameState API in production.

One cypress run should build the frontend with `VITE_USE_GAMESTATE_API=true`, then boot the backend with `SAILS_USE_GAMESTATE_API=true` and run the tests with `CYPRESS_USE_GAMESTATE_API=true`.

The other should build, boot, and run the tests without these environment variables set i.e. using the regular commands.

**Production config should not be affected and should still OMIT these env vars in order to continue to use the Legacy API**

In order to keep CI green, this update should fix all the tests that are broken in both boot modes before it can be merged.

Haviles04:
## Issue number

- Resolves #963 

## Please check the following

- For New Features:


itsalaidbacklife (approved):
Great work. Can you benchmark the array .includes against a Set() and against a dictionary where the slugs are the keys?

Otherwise the only thing left I'd like to check is how the cypress cloud results are reported

itsalaidbacklife (approved):
Nicely done

PR #996, opened 2024-08-23
Fixes Bug of CreateGameDialog

### LINKED ISSUE #518: [Bug]: CreateGameDialog has two separate forms inside it
### Bug Summary

Currently the `CreateGameDialog` instantiates a `<form>` element that wraps the switch for controlling ranked, and a separate `<v-form>` element that wraps the text field for the game name. Instead, the `<v-form>` should wrap both the ranked-play switch and the game-name text field. 

### Detailed Description

The `<form>` tag that is used to wrap the switch also includes the `<stats-scoring-dialog>` in order to make them appear on the same line. We should replace this with a regular `<div>` that puts the switch + stats scoring dialog on one line, and move the `<v-form>` wrapper to include the entire contents of the dialog #body

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Steps to Reproduce

1.
2.
3.

hammad-521:
## Issue number
- Resolves #518 

## Please check the following

- [x] Do the tests still pass? 
- [x] Is the code formatted properly? 

## Description:

> Replaced the form around the switch and scoring dialog with a  div, ensuring they appear on the same line.
> Moved the v-form wrapper to include both the switch and the text field within the dialog body.
> The submit button will trigger the form submission using the submitNewGame method.


src/routes/home/components/CreateGameDialog.vue
```diff
@@ -29,20 +29,20 @@
...
-        />
-        <v-switch
-          v-model="isRanked"
-          class="d-flex align-center"
-          :label="isRanked ? t('global.ranked') : t('global.casual')"
-          data-cy="create-game-ranked-switch"
-          color="surface-2"
-        />
-      </form>
       <v-form @submit.prevent="submitNewGame">
```
Haviles04, line 32: I believe we wanted to wrap the whole #body with the v-form, so from lines 20 - 56

Haviles04 (commented):
Thanks for contributing! This looks good, one small comment about wrapping the whole contents as per the PR description 👍

src/routes/home/components/CreateGameDialog.vue
hammad-521: I've made the changes you requested, wrapping the entire #body with the <v-form> tag as per the PR description. The commit has been pushed, and everything should now be in line with the requirements.

Please let me know if you need any further adjustments.

Thanks,
Hammad

src/routes/home/components/CreateGameDialog.vue
```diff
@@ -18,31 +18,32 @@
...
+        <!-- V form now wraps the whole #body -->
```
Haviles04, line 22: This comment is unnecessary

Haviles04 (commented):
This is looking good, good job!

Curious if we should include the #actions template in the form as well

@itsalaidbacklife thoughts?

Haviles04:
> This is looking good, good job!
> 
> Curious if we should include the #actions template in the form as well
> 
> @itsalaidbacklife thoughts?

After some discussion I think the move here is to add an `id` to the `form` and add a `form=id` attribute to the submit button, we could then remove the `@click` on the button

hammad-521:
@Haviles04 Thank you for the feedback and discussion. I have implemented the suggested changes:

- Added an id attribute to the v-form element: id="create-game-form".
- Added a form attribute to the submit button: form="create-game-form".
- Removed the @click handler from the submit button.

I've also removed the unnecessary comment about wrapping the whole #body with v-form.  

Please review the latest commit and let me know if any further adjustments are needed.

src/routes/home/components/CreateGameDialog.vue
hammad-521: I have removed this comment in latest commit.

Haviles04 (approved):
Great Work! Thank you for your contribution!

PR #1013, opened 2024-09-01
Revert "Feat/deal cards gamestate"

itsalaidbacklife:
Reverts cuttle-cards/cuttle#984

Bug was introduced that breaks readying. Doesn't appear in CI. This revert is a bit of a crapshoot.


itsalaidbacklife:
Closing as I believe I've isolated the problem and hack-fixed it in #1014 

Need a migration and rollback of that PR to fix fully

PR #1020, opened 2024-09-04
Update en.json

AntonBeYt:
Removed sentence about setting up from paragraph about goal.


itsalaidbacklife (commented):
Good catch. The same mistake exists in the es.json and fr.json files as well. Please fix them there as well so that Spanish and French speaking users have the same experience, and then this is good to merge

AntonBeYt:
I don't really understand Spanish or French but hopefully I understood enough to remove the correct sentences

itsalaidbacklife (commented):
Close, but you removed one to many sentences from the es and fr files. The part to remove is what is duplicated in the next section

itsalaidbacklife (approved):
Good catch. Thank you for your contribution!

PR #1021, opened 2024-09-05
Feature issue #1009: Add text with rules for 3 and 4 player Cuttle

### LINKED ISSUE #1009: [Feature]: Add text with rules for 3 and 4 player Cuttle
### Feature Summary

The rules for 3 and 4 player Cuttle were accidentally removed in the redesign and implementation #888. We should add them back after the one-offs section and before faq.

### Detailed Description

The text already exists in the translation files, it's simply not displayed. We should add it to `src/routes/rules/RulesView.vue` in its own `<section>` with the same `v-intersect` config as the other sections and its own `id="multiplayer"`. A corresponding entry should be made to `src/routes/rules/rulesData`'s `sectionTitles` to ensure the section appears in the `RulesNav`. The text should be rendered using `i18n` by passing the translation keys to the already-imported `t()` function. See the FAQ section of the `RulesView` for examples.

The heading should be the text at the key `rules.multiplayerText2` followed by a `<p>` that combines the text from the `rules.multiplayerText1` and `rules.multiplayerText2` strings which include a link to cuttle.cards in the middle. It should read:

> Presently only 2-player cuttle is available at [cuttle.cards](https://cuttle.cards), however Cuttle can be played with 2-4 players. 3 and 4-player Cuttle each have slight variations on the rules to optimize the game balance while changing the pace and the feel. Give them a try next time you're playing Cuttle in a group!

There should then be sub-headings and divs for the `rules.3Players` and `rules.4Players` blocks found in the translation files e.g. `en.json`.  These should have the heading, a paragraph of text, and then a bulleted list of all their individual rules. It is likely a good idea to create a component to represent these sub-sections since they will be almost identical in the 3Player and 4Player sub-sections.

livoszlak:
## Issue number

- Resolves #1009 

## Please check the following

- For New Features:

Added a new component for displaying multiplayer rule sets (src/routes/rules/components/MultiplayerRuleBlock.vue) as suggested in the issue. Also added a new variable (multiplayer) to rulesData.js to include multiplayer rule set keys to the rulesData export. The text translates as it should. 

As for tests added and documentation updated - I'm not sure what to do here (whether it needs a new test or if it's already covered with the rest of the RulesView,  and whether the multiplayer options need to be added anywhere else than specified in the issue.


src/routes/rules/data/rulesData.js
itsalaidbacklife, line 1: Also add an entry to the `sectionTitles` variable below to add a jump link to the nav

src/routes/rules/RulesView.vue
itsalaidbacklife, line 1: Let's add some additional vertical margin between the above 'ready to play?' section (with the button) and the new section for multiplayer so that it better matches the margin above the ready to play section (between that and the cards with their effects)

src/routes/rules/components/MultiplayerRuleBlock.vue
```diff
@@ -0,0 +1,58 @@
...
+  padding-left: 2rem;
+}
+
+li {
+  padding: 0 0 1rem 0;
+}
+
+ul > li:first-child {
+  margin-top: 16px !important;
+}
```
itsalaidbacklife, line 57: tag-based selectors are generally less performant and specific than attribute based ones. Let's put some classes on the relevant elements and use those for selectors instead. Otherwise, looks good

src/routes/rules/components/MultiplayerRuleBlock.vue
```diff
@@ -0,0 +1,58 @@
...
+ul {
+  padding-left: 2rem;
+}
+
+li {
+  padding: 0 0 1rem 0;
+}
+
+ul > li:first-child {
+  margin-top: 16px !important;
```
itsalaidbacklife, line 56: The need for `!important` smells here. Is it conflicting with another style? Can we tell from where? Does switching over to a class-based selector sufficiently increase specificity to apply the style without the need for `!important`?

itsalaidbacklife (changes requested):
Nice work. This is well organized and it's working nicely. I have a few minor changes requested in the comments. The biggest of them is adding the section to the rules navigation by creating an entry in the rulesData section titles (see comment). The rest is small potatoes.

As for testing, there's no need for static content like this. We don't generally test that the static text appears, and mostly reserve tests for stateful/conditionally rendered elements.

src/routes/rules/components/MultiplayerRuleBlock.vue
livoszlak: Agreed, looking at it now I don't even know why I used `!important` there in the first place - no other styles interfered with it whatsoever. Removing it now!

src/routes/rules/components/MultiplayerRuleBlock.vue
livoszlak: True! Great suggestion. Adding "rules-ul" & "rules-li" as classes for the elements.

src/routes/rules/RulesView.vue
livoszlak: I've added top margin to `.section#multiplayer`, however, I can see that on smaller screens the margin needs to adapt and shrink by a few pixels by each breakpoint. I set one breakpoint in the already existing `@media (max-width: 960px)`, but I bet there's a better way to go about it than putting a bunch of breakpoints for just one section in this file - would love input! 🙏 (CSS isn't my strong suit, and SCSS is new to me)

src/routes/rules/RulesView.vue
```diff
@@ -421,12 +448,19 @@ export default {
...
 .text-label-lg {
   margin-bottom: 8px;
 }
 @media (max-width: 960px) {
   .section-title {
     font-size: 2.4rem !important;
   }
+  .section#multiplayer {
+    margin-top: 118px;
+  }
```
itsalaidbacklife, line 464: No need to sweat this level of precision
```suggestion
```

src/routes/rules/RulesView.vue
```diff
@@ -421,12 +448,19 @@ export default {
   margin-top: 64px;
 }
 
+.section#multiplayer {
+  margin-top: 121px;
```
itsalaidbacklife, line 452: Prefer multiples of 8 for consistency and a regular grid size
```suggestion
  margin-top: 120px;
```

itsalaidbacklife (approved):
Nice work. Thank you for your contribution!

PR #1025, opened 2024-09-07
Feat/draw gamestate

### LINKED ISSUE #942: [Feature]: Implement `draw` helpers in the GameState API
### Prerequisites
This cannot be completed until the following issues are resolved: #946 

### Feature Summary

We need to implement the endpoint for drawing a card in the new GameState API in order to allow users to draw cards. Drawing a card means moving the first card of the deck into the specified player's hand ( and out of the deck).

### Detailed Description

This will likely be one of the first full move-making utilities built using the GameState API. It should take the general approach shown in the [Backend Control Flow section of the GameState API Spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.vgf103gono5r). 

```
POST /api/game/:gameId/move
```
Which move gets made is determined by the request body, which should specify a `moveType` field with a value from the `MoveType` enum. Drawing a card will be triggered by sending a request body like: 
```
{ moveType: MoveType.DRAW};
```

The endpoint will already exist once #964 is completed. When a `moveType: MoveType.DRAW` is sent in the req body, the endpoint will attempt to utilize the `draw` helpers: `sails.helpers.gamestate.draw.validate()` and `sails.helpers.gamestate.execute()`. Each of these should follow the paradigm outlined in the [docs](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.vgf103gono5r):

1. update the `api/policies/hasValidMoveBody` policy to allow and validate requests with a moveType of `MoveType.DRAW`
2. Implement the `api/helpers/gamestate/moves/draw/validate` helper to verify if the move is legal
3. Implement the `api/helpers/gamestate/moves/draw/execute` helper to create a new GameState that draws a card for the requested player

We must also update the policy `api/policies/hasValidMoveBody.js` to accommodate the `DRAW` moveType. This policy is already applied to the endpoint for making a move and it uses a `switch()` statement on `req.body.moveType` to validate that a given req body has all the data necessary for the given `moveType`. We should add a `case MoveType.DRAW` to this policy.

A `moveType`  of `MoveType.DRAW` does not require any additional data (because no cards are played or targeted) and so the policy can simply `return res.next()` in this switch case.

### Implementing `validate()`
Create a new `api/helpers/gamestate/moves/draw/validate.js` helper file that matches the general setup of the existing `api/helpers/gamestate/moves/points/validate.js` helper. It should throw errors if the requested move is illegal. It should have three inputs: 
`validate()` should throw if:
* The player is at the hand limit

### Implementing `execute()`
Create a new `api/helpers/gamestate/moves/draw/execute.js` helper file that matches the general approach of the existing execute helper `api/helpers/gamestate/moves/points/execute.js`. It should create and return a new `GameState` that is the result of making the `draw` move for the requested player i.e. **it should remove the top card from the deck and add it to the requested player's points** array. It should have 3 inputs:

### Unskipping Tests

**Essential tests**
* `basicMoves.spec.js` - `it('Draws from deck')`

itsalaidbacklife:
Implements drawing a card in the GameState API
## Issue number

- Resolves #942 

## Please check the following

- For New Features:

### DIFF
--- a/src/stores/game.js
+++ b/src/stores/game.js
@@ -512,7 +512,7 @@ export const useGameStore = defineStore('game', {
     // In-Game Moves //
     ///////////////////
     async requestDrawCard() {
-      await this.makeSocketRequest('draw');
+      await this.makeSocketRequest('draw', { moveType: 'draw' });
     },
     async requestPlayPoints(cardId) {
       await this.makeSocketRequest('points', { moveType: 'points', cardId });
--- a/tests/e2e/specs/in-game/basicMoves.spec.js
+++ b/tests/e2e/specs/in-game/basicMoves.spec.js
@@ -393,7 +393,6 @@ describe('Game Basic Moves - P1 Perspective', () => {
   });
 
   it('Draws from deck', () => {
-    cy.skipOnGameStateApi();
     // Opponent draws card
     cy.drawCardOpponent();
     // Opponent now has 6 cards in hand
--- a/api/helpers/gamestate/moves/draw/execute.js
+++ b/api/helpers/gamestate/moves/draw/execute.js
@@ -11,7 +11,7 @@ module.exports = {
     },
     /**
      * @param {Object} requestedMove - Object describing the request to play points (req.body)
-     * @param { MoveType.DRAW } requestedMove.moveType - Specifies that this a Points move
+     * @param { MoveType.DRAW } requestedMove.moveType - Specifies that this a Draw move
      */
     requestedMove: {
       type: 'ref',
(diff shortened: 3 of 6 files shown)


src/stores/game.js
```diff
@@ -512,7 +512,7 @@ export const useGameStore = defineStore('game', {
     // In-Game Moves //
     ///////////////////
     async requestDrawCard() {
-      await this.makeSocketRequest('draw');
+      await this.makeSocketRequest('draw', { moveType: 'draw' });
```
Haviles04, line 515: I think maybe we should import the enum, this should probably have been done in points too

Haviles04 (commented):
Looking good, this will help a lot of tests when we start building the one-offs

tests/e2e/specs/in-game/basicMoves.spec.js
```diff
@@ -393,7 +393,6 @@ describe('Game Basic Moves - P1 Perspective', () => {
   });
 
   it('Draws from deck', () => {
-    cy.skipOnGameStateApi();
```
itsalaidbacklife, line 396: Just one test so far but it's working

src/stores/game.js
itsalaidbacklife: Yes I think it deserves its own PR so that this can be a lean example of just what's needed for a move (once that's fixed)

itsalaidbacklife: Resolved by #1026

PR #1038, opened 2024-09-12
Feat/one off gamestate

itsalaidbacklife:
* Implements and combines playing targeted and untargeted one-offs into one `MoveType.ONE_OFF`.  Does not enable countering or resolving
* Fixes a bug in `publishGameState` where it incorrectly set the `oneOff` attribute of the game
* Adds `targetType` to `GameState` and `GameStateRow`
## Issue number

Relevant [issue number](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
- Resolves #950, #951 

## Please check the following

- [ ] Do the tests still pass? (see [Run the Tests](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#run-the-tests))
- [ ] Is the code formatted properly? (see [Linting (Formatting)](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#linting-formatting))
- For New Features:
  - [ ] Have tests been added to cover any new features or fixes?
  - [ ] Has the documentation been updated accordingly?

PR #1041, opened 2024-09-16
Feat/one off error reconciliation

### LINKED ISSUE #1040: [Feature]: Reconcile error messages for untargetedOneOff / sevenUntargetedOneOff & targetedOneOff / sevenTargetedOneOff
### Feature Summary

The untargetedOneOff/sevenUntargetedOneOff & targetedOneOff/sevenTargetedOneOff endpoints have near-identical error messages for the same things. We should make them the same.

### Detailed Description

`api/controllers/game/seven/untargetedOneOff` and `api/controllers/game/seven/targetedOneOff` correctly use translated error messages by throwing errors as translation keys e.g. `'game.snackbar.oneOffs.sevenWithEmptyDeck'` which i18n can look up in the translation files e.g. `src/translations/en.json`.

In contrast, `api/controllers/game/untargetedOneOff` and `api/controllers/game/targetedOneOff` throw almost identical errors for some of the same issues e.g. you can't play a 5 one-off with an empty deck with hard-coded strings that aren't found in the translation files. We should update the errors thrown in both `api/controllers/game/untargetedOneOff` and `api/controllers/game/targetedOneOff` to throw errors matching their counterparts in `api/controllers/game/seven/untargetedOneOff` and `api/controllers/game/seven/targetedOneOff`.

Correspondingly, we should update the tests so that the `assertSnackBarErorr()` calls for the hard coded strings previously found in `api/controllers/game/untargetedOneOff` and `api/controllers/game/targetedOneOff` are updated to check for the exact text in the en.json translation file for the given error.

itsalaidbacklife:
## Issue number
Applies i18n to error messages from the legacy api's `untargetedOneOff` and `targetedOneOff` endpoints, and reconciles them with their /seven counterpoints. Applies the translated errors to the GameStateAPI's `oneOff/validate` and `oneOff/execute` helpers.

- Resolves #1040 

## Please check the following

- For New Features:

### DIFF
--- a/api/controllers/game/resolve-five.js
+++ b/api/controllers/game/resolve-five.js
@@ -7,16 +7,16 @@ module.exports = async function (req, res) {
     const [game, player, cardToDiscard] = await Promise.all([promiseGame, promisePlayer, promiseCard]);
 
     if (!game.topCard) {
-      throw { message: 'game.snackbar.five.fiveDeckIsEmpty' };
+      throw { message: 'game.snackbar.oneOffs.five.fiveDeckIsEmpty' };
     }
     if (player.hand?.length && !cardToDiscard) {
-      throw { message: 'game.snackbar.five.selectCardToDiscard' };
+      throw { message: 'game.snackbar.oneOffs.five.selectCardToDiscard' };
     }
     if (game.turn % 2 !== player.pNum) {
       throw { message: 'game.snackbar.global.notYourTurn' };
     }
     if (!game.oneOff || game.oneOff?.rank !== 5) {
-      throw { message: 'game.snackbar.five.incorrectCard' };
+      throw { message: 'game.snackbar.oneOffs.five.incorrectCard' };
     }
 
     const gameUpdates = {
--- a/api/controllers/game/untargeted-one-off.js
+++ b/api/controllers/game/untargeted-one-off.js
@@ -28,7 +28,7 @@ module.exports = function (req, res) {
                     if (game.scrap.length < 1)
                       {
                         return Promise.reject({
-                          message: 'You can only play a 3 as a one-off, if there are cards in the scrap pile',
+                          message: 'game.snackbar.oneOffs.scrapIsEmpty',
                         });
                       }
                     break;
--- a/api/controllers/game/untargeted-one-off.js
+++ b/api/controllers/game/untargeted-one-off.js
@@ -45,7 +45,7 @@ module.exports = function (req, res) {
                     if (!game.topCard)
                       {
                         return Promise.reject({
-                          message: "You can't play that card as a one-off, unless there are cards in the deck",
+                          message: 'game.snackbar.oneOffs.emptyDeck',
                         });
                       }
                     break;
--- a/api/controllers/game/untargeted-one-off.js
+++ b/api/controllers/game/untargeted-one-off.js
@@ -28,7 +28,7 @@ module.exports = function (req, res) {
                     if (game.scrap.length < 1)
                       {
                         return Promise.reject({
-                          message: 'game.snackbar.oneOffs.scrapIsEmpty',
+                          message: 'game.snackbar.oneOffs.three.scrapIsEmpty',
                         });
                       }
                     break;
--- a/api/policies/hasCardId.js
+++ b/api/policies/hasCardId.js
@@ -13,11 +13,11 @@ module.exports = function (req, res, next) {
     if (typeof req.body.cardId === 'number') {
       return next();
     }
-    return res.forbidden({
+    return res.badRequest({
       message: 'Error: You must provide a valid card for your move (ID was non-integer)',
     });
   }
-  return res.forbidden({
+  return res.badRequest({
     message: 'Error You must provide a valid card to make that move (No ID given)',
   });
 };
--- a/src/translations/en.json
+++ b/src/translations/en.json
@@ -282,7 +282,6 @@
         "scrapIsEmpty": "You can only play a 3 as a one-off if there are cards in the scrap pile",
         "opponentHasNoCards": "You cannot play a 4 as a one-off while your opponent has no cards in hand",
         "emptyDeck": "You can't play that one-off unless there are cards in the deck",
-        "sevenWithEmptyDeck": "You can only play a 7 as a ONE-OFF if there are cards in the deck",
         "sevenWithOneCardInDeck": "You cannot play a 7 as a ONE-OFF if it is the last card in the deck",
         "cantPlayWithoutATarget": "You cannot play that card as a ONE-OFF without a target",
         "onlyTwoOrNine": "You can only play a 2, or a 9 as targeted one-offs.",
--- a/src/translations/en.json
+++ b/src/translations/en.json
@@ -280,7 +280,7 @@
       },
       "oneOffs": {
         "scrapIsEmpty": "You can only play a 3 as a one-off if there are cards in the scrap pile",
-        "opponentHasNoCards": "You cannot play a 4 as a ONE-OFF while your opponent has no cards in hand",
+        "opponentHasNoCards": "You cannot play a 4 as a one-off while your opponent has no cards in hand",
         "sevenWithEmptyDeck": "You can only play a 7 as a ONE-OFF if there are cards in the deck",
         "sevenWithOneCardInDeck": "You cannot play a 7 as a ONE-OFF if it is the last card in the deck",
         "cantPlayWithoutATarget": "You cannot play that card as a ONE-OFF without a target",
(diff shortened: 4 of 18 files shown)


itsalaidbacklife:
Todo: 
* ensure all errors in the gamestate one-off helper are translated.
* translate all errors in /untargetedOneOff
* translate all errors in /targetedOneOff
* Reconcile keys so the one-off errors are all colocated

api/controllers/game/resolve-five.js
```diff
@@ -7,16 +7,16 @@ module.exports = async function (req, res) {
     const [game, player, cardToDiscard] = await Promise.all([promiseGame, promisePlayer, promiseCard]);
 
     if (!game.topCard) {
-      throw { message: 'game.snackbar.five.fiveDeckIsEmpty' };
+      throw { message: 'game.snackbar.oneOffs.five.fiveDeckIsEmpty' };
```
itsalaidbacklife, line 10: Reorganized translation keys for one-off errors so they all live under `game.snackbar.oneOffs` with the general ones that are used in multiple places there, and move specific ones under their one-off rank e.g. `game.snackbar.oneOffs.five`

api/controllers/game/untargeted-one-off.js
```diff
@@ -28,7 +28,7 @@ module.exports = function (req, res) {
                     if (game.scrap.length < 1)
                       {
                         return Promise.reject({
-                          message: 'You can only play a 3 as a one-off, if there are cards in the scrap pile',
+                          message: 'game.snackbar.oneOffs.three.scrapIsEmpty',
```
itsalaidbacklife, line 31: untargeted oneoff was not using i18n keys for its errors and had duplicate, hard-coded english errors for things that the /seven/untargetedOneOff endpoint already had in i18n. I consolidated them into the i18n version

api/policies/hasCardId.js
```diff
@@ -13,11 +13,11 @@ module.exports = function (req, res, next) {
     if (typeof req.body.cardId === 'number') {
       return next();
     }
-    return res.forbidden({
+    return res.badRequest({
```
itsalaidbacklife, line 16: This was sending a 403 instead of 400 because it's one of the oldest policies and frankly I didn't know the difference back when I was writing this. But it's not an authorization issue it's a matter of missing/incorrect request data so it should be 400 instead.

src/translations/en.json
itsalaidbacklife, line 1: Reorganized these but the text is mostly the same

tests/e2e/fixtures/snackbarError.js
```diff
@@ -5,7 +5,12 @@ const SnackBarError = {
   FROZEN_CARD: 'That card is frozen! You must wait a turn to play it',
   NOT_IN_GAME: "You can't do that because you're not a player in this game",
   GAME_IS_FULL: `Cannot join that game because it's already full`,
-  CANT_FIND_GAME: "Can't find game"
+  CANT_FIND_GAME: "Can't find game",
+  ONE_OFF: {
+    THREE_EMPTY_SCRAP: 'You can only play a 3 as a one-off if there are cards in the scrap pile',
+    FOUR_EMPTY_HAND: 'You cannot play a 4 as a one-off while your opponent has no cards in hand',
+    EMPTY_DECK: "You can't play that one-off unless there are cards in the deck",
+  }
```
itsalaidbacklife, line 13: Updated to add additional errors that the tests can look for via fixtures instead of hard coding in the tests. It would be nice to reconcile this more concretely with the language files themselves, but I think that's an issue for another update

tests/e2e/specs/in-game/one-offs/3_threes.spec.js
```diff
@@ -8,7 +9,6 @@ describe('Playing THREEs', () => {
   });
 
   it('Plays 3s with no cards in scrap', () => {
-    cy.skipOnGameStateApi();
```
itsalaidbacklife, line 11: The only reason this wasn't in scope as of #1038 was the error message was inconsistent across legacy/gamestate api version of this move. Now that they all use the same i18n key, we can run this test in gamestate mode. Notably it doesn't require resolving the one-off or _doing_ it's effect, just requesting to make the move and validating that it's illegal because the scrap is empty

tests/e2e/specs/in-game/one-offs/4_fours.spec.js
```diff
@@ -98,7 +99,6 @@ describe('FOURS', () => {
     });
 
     it('Prevents playing a 4 when opponent has no cards in hand', () => {
-      cy.skipOnGameStateApi();
```
itsalaidbacklife, line 101: Same idea as with threes. Can now run this in the gamestate api, now that the error messages are normalized.

tests/e2e/specs/in-game/one-offs/5_fives.spec.js
```diff
@@ -350,8 +350,11 @@ describe('FIVES', () => {
           deck: [],
         });
       });
+    });
+
+    describe('Illegal FIVES', () => {
 
-      it('Attempts to play 5 with an empty deck', () => {
+      it('Cannot to play 5 with an empty deck', () => {
         cy.skipOnGameStateApi();
```
itsalaidbacklife, line 358: Still can't run this test in gamestate api because it requires passing. I'll add this to the list of specs to unskip in that issue

itsalaidbacklife: Actually it looks like we might be missing an issue to enable passing? I'll investigate

tests/e2e/specs/in-game/one-offs/5_fives.spec.js
```diff
@@ -367,15 +370,34 @@ describe('FIVES', () => {
...
+              // If the promise resolves, this assertion will fail the test (as expected in your case)
+              expect(true).to.eq(
+                false,
+                `Expected request to resolve five without discarding to error, but instead came back 200: ${res}`,
+              );
+            } catch (err) {
+              // If the promise rejects, this will handle the error
+              expect(err).to.eq('game.snackbar.oneOffs.emptyDeck');
+            }
+          });
```
itsalaidbacklife, line 398: This took some fancy footwork to get cypress to play nice with the promise returned by the pinia store action `gameStore.requestResolveFive()`. We had a way of doing it originally that looked correct and would show the result of the assertion as passing, but it didn't actually fail the test when the assertion failed, due to a race condition.

I think there is an issue where `cy.window().its().then()` doesn't actually create a promise in such a way that the inner callback would bubble up. By using `async/await` here we can ensure that cypress will actually wait for the result of the async call and fail the test if the request comes back 200 or if the server error doesn't match the expected string.

tests/e2e/specs/in-game/one-offs/5_fives.spec.js
```diff
@@ -393,19 +415,18 @@ describe('FIVES', () => {
...
-              .requestResolveFive(undefined)
-              .then((res) => {
-                expect(true).to.eq(
-                  false,
-                  `Expected request to resolve five without discarding to error, but instead came back 200: ${res}`,
-                );
-              })
-              .catch((err) => {
-                expect(err).to.eq('game.snackbar.five.selectCardToDiscard');
-              });
```
itsalaidbacklife, line 408: @seriouslysean and @Haviles04 I think you might get a kick out of this

This looked right but didn't actually work. Cypress would execute the `catch()` callback and show the result of the assertion, but it would pass the test before the `gameStore.requestResolveFive(undefined)` promise resolved because cypress' use of `.then()` is not a true promise. The rewrite below uses an actual promise via `async/await` so as to force cypress to await resolution/rejection of the promise from the app code before proceeding with the test.

This ensures that the test will fail if the results of the request are not what we expect. We should take a look through the test code to see if there are other examples of this pattern being used incorrectly so we can ensure that our tests of api requests like this correctly fail their respective tests if the result is not what's expected.

itsalaidbacklife:
Okay I think this is ready

tests/e2e/specs/in-game/one-offs/5_fives.spec.js
seriouslysean: I absolutely love stuff like this and it’s why I still have a problem with Cypress using a “promise-like” format. Still, awesome find!

tests/e2e/specs/in-game/one-offs/5_fives.spec.js
seriouslysean: This is great. It feels like an excellent time to do a write up on the problem and cement your knowledge on the topic, and potentially dive in to the nuances around it. I’d also be curious if other testing frameworks, like playwright, have this same problem.

Really nice work, @itsalaidbacklife.

PR #1045, opened 2024-09-18
Implement playing Jacks in the GameState API #949

### LINKED ISSUE #949: [Feature]: Implement playing Jacks in the GameState API
### Feature Summary

We need to implement playing Jacks in the GameState API in order to support the full game logic in this API. See the [spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.wpeuh0qc6ps2) for a general overview of the API, its underlying motivations, and the development approach.

Playing a jack should result in the `playedCard` moving from the requesting player's `hand` to the `attachments` on the `targetCard`, and the `targetCard` moving from the opponent's `points` to the requesting player's `points`. It should also increment the turn and set the `phase` to `GamePhase.MAIN`.

### Detailed Description

Implementation should take the general approach shown in the [Backend Control Flow section of the GameState API Spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.vgf103gono5r) and follow the example of #1025 

```
POST /api/game/:gameId/move
```
Which move gets made is determined by the request body, which should specify a moveType field with a value from the MoveType enum. Playing a jack will be triggered by sending a request body like:

```
{ moveType: MoveType.JACK, playedCard: 'JH', targetCard: '7D'};
```
When a `moveType: MoveType.JACK` is sent in the req body, the endpoint will attempt to utilize the `jack` helpers: `sails.helpers.gamestate.moves.jack.validate()` and `sails.helpers.gamestate.moves.jack.execute()`. Each of these should follow the paradigm outlined in the [docs](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.vgf103gono5r):

1. update the `api/policies/hasValidMoveBody` policy to allow and validate requests with a moveType of `MoveType.JACK`
2. Implement the `api/helpers/gamestate/moves/jack/validate` helper to verify if the move is legal
3. Implement the `api/helpers/gamestate/moves/jack/execute` helper to create a new GameState that plays a jack for the requested player to steal their opponent's point card

We must also update the policy `api/policies/hasValidMoveBody.js` to accommodate the `MoveType.JACK` moveType. This policy is already applied to the endpoint for making a move and it uses a `switch()` statement on `req.body.moveType` to validate that a given req body has all the data necessary for the given `moveType`. We should add a `MoveType.JACK` case to this policy.

A `moveType`  of `MoveType.JACK` requires a `playedCardId` and a `targetCardId` which are both valid card ids. If both `playedCardId` and `targetCardId` present in req.body and it can be found in `utils/DeckIds.json`, then we can `return res.next()` in this switch case. Otherwise we should return `res.badRequest()` with an error that explains which `req.body` attribute is missing or invalid.

### Implementing `validate()`
Create a new `api/helpers/gamestate/moves/jack/validate.js` helper file that matches the general setup of the existing `api/helpers/gamestate/moves/points/validate.js` helper. It should throw errors if the requested move is illegal. It should have three inputs: 

`validate()` should throw if:
* The `playedCard` is not in the player's hand (`throw new Error("game.snackbar.global.playFromHand")`)
* The `playedCard` is frozen (`throw new Error('game.snackbar.global.cardFrozen')`)
* The `playedCard` is not a jack (`throw new Error('game.snackbar.jack.stealOnlyPointCards')`)
* The `targetCard` is not in the opponent's `points` (`throw new Error('game.snackbar.jack.stealOnlyPointCards')`)
* The opponent has a queen in their `faceCards` (`throw new Error('game.snackbar.jack.noJackWithQueen')`)

### Implementing `execute()`
Create a new `api/helpers/gamestate/moves/jack/execute.js` helper file that matches the general approach of the existing execute helper `api/helpers/gamestate/moves/points/execute.js`. It should create and return a new `GameState` that is the result of making the `jack` move for the requested player i.e. **it should remove `playedCard` from the player's hand, and add it the `attachments` of the `targetCard`. It should then move the `targetCard` from the opponent's `points` to the requesting player's `points`. It should also increment the turn and set the phase to `GamePhase.MAIN`. The `playedBy`, `playedCard`, and `targetCard` attributes should be included in the returned `GameState`.

### Unskipping Tests

**Essential tests**
* `basicMoves.spec.js` - `it('Player and Opponent plays Jacks on different cards')`
* `basicMoves.spec.js` - `it('Quadruple jacks successfully')`

kawaljen:
## 949

- Resolves #949

## Please check the following

- For New Features:

### DIFF
--- a/api/helpers/gamestate/moves/jack/execute.js
+++ b/api/helpers/gamestate/moves/jack/execute.js
@@ -1,0 +1,71 @@
+const GamePhase = require('../../../../../utils/GamePhase.json');
+
+module.exports = {
+  friendlyName: 'Play a Jack -> steal an opponent points card',
+
+  description: 'Returns new GameState resulting from requested move',
+
+  inputs: {
+    currentState: {
+      type: 'ref',
+      description: 'The latest gamestate before playing Jack',
+      required: true,
+    },
+    /**
+     * @param { Object } requestedMove - The move being requested.
+     * @param { String } requestedMove.cardId - Card Played (Jack))
+     * @param { String } [ requestedMove.targetId ] - Card targeted by Jack
+     */
+    requestedMove: {
+      type: 'ref',
+      description: 'The move being requested',
+      required: true,
+    },
+    playedBy: {
+      type: 'number',
+      description: '0 or 1 for whether p0 or p1 is making the move',
+      required: true,
+    },
+  },
+  sync: true,
+
+  fn: ({ currentState, requestedMove, playedBy }, exits) => {
+    try {
+      let result = _.cloneDeep(currentState);
+
+      const { cardId, targetId } = requestedMove;
+
+      const player = playedBy ? result.p1 : result.p0;
+      const opponent = playedBy ? result.p0 : result.p1;
+
+      const playedCard = player.hand.find(({ id }) => id === cardId);
+      const targetCard = opponent.points.find(({ id }) => id === targetId);
+
+      //add jack(playedCard) to targetcard attachment
+      targetCard.attachments.push(playedCard);
+      //add targetCard to player points
+      player.points.push(targetCard);
+      // remove jack(playedCard) from player hand
+      const cardIndex = player.hand.findIndex(({ id }) => id === cardId);
+      player.hand.splice(cardIndex, 1);
+      //remove target card from oppponent points
+      const targetIndex = opponent.points.findIndex(({ id }) => id === targetId);
+      opponent.points.splice(targetIndex, 1);
+
+      result.turn++;
+
+      result = {
+        ...result,
+        ...requestedMove,
+        playedBy,
+        playedCard,
+        targetCard,
+        phase: GamePhase.MAIN,
+      };
+
+      return exits.success(result);
+    } catch (err) {
+      return exits.error({ message: 'Cannot play Jack ' + err });
+    }
+  },
+};
--- a/api/helpers/gamestate/moves/jack/validate.js
+++ b/api/helpers/gamestate/moves/jack/validate.js
@@ -1,0 +1,75 @@
+const GamePhase = require('../../../../../utils/GamePhase.json');
+
+module.exports = {
+  friendlyName: 'Validate request to play Jack',
+
+  description: 'Verifies whether a request to play a Jack is legal, throwing explanatory error if not.',
+
+  inputs: {
+    currentState: {
+      type: 'ref',
+      description: 'Object containing the current game state',
+      required: true,
+    },
+    /**
+     * @param { Object } requestedMove - Object describing the request to play Jack
+     * @param { String } requestedMove.cardId - Card Played (Jack)
+     * @param { String } [ requestedMove.targetId ] - Card targeted by Jack
+     */
+    requestedMove: {
+      type: 'ref',
+      description: 'Object containing data needed for current move',
+      required: true,
+    },
+    playedBy: {
+      type: 'number',
+      description: 'Player number of player requesting move',
+      required: true,
+    },
+  },
+  sync: true,
+
+  fn: ({ currentState, requestedMove, playedBy }, exits) => {
+    try {
+      const player = playedBy ? currentState.p1 : currentState.p0;
+      const opponent = playedBy ? currentState.p0 : currentState.p1;
+
+      const cardPlayed = player.hand.find(({ id }) => id === requestedMove.cardId);
+      const targetCard = opponent.points.find(({ id }) => id === requestedMove.targetId);
+
+      //gameState phase should be MAIN
+      if (currentState.phase !== GamePhase.MAIN) {
+        throw new Error(`Can only play a jack in main phase, not ${currentState.phase}`);
+      }
+      //is it player turn
+      if (currentState.turn % 2 !== playedBy) {
+        throw new Error('game.snackbar.global.notYourTurn');
+      }
+      //is cardPlayed in player hand
+      if (!cardPlayed) {
+        throw new Error('game.snackbar.global.playFromHand');
+      }
+      //is cardPlayed a jack
+      if (!cardPlayed.rank === '11') {
+        throw new Error('game.snackbar.jack.stealOnlyPointCards');
+      }
+      //is targetcard in the opponent points
+      if (!targetCard) {
+        throw new Error('game.snackbar.jack.stealOnlyPointCards');
+      }
+      //is any card frozen
+      if (cardPlayed.isFrozen || targetCard.isFrozen) {
+        throw new Error('game.snackbar.global.cardFrozen');
+      }
+      //does opponent have a queen in his faceCards to block the move
+      const queenCount = opponent.faceCards.filter((faceCard) => faceCard.rank === 12).length;
+      if (queenCount > 0) {
+        throw new Error('game.snackbar.jack.noJackWithQueen');
+      }
+
+      return exits.success();
+    } catch (err) {
+      return exits.error(err);
+    }
+  },
+};
--- a/api/policies/hasValidMoveBody.js
+++ b/api/policies/hasValidMoveBody.js
@@ -18,24 +18,46 @@ module.exports = function (req, res, next) {
       return next();
 
     case MoveType.POINTS:
-    case MoveType.ONE_OFF: {
-      if (!cardId) {
-        return res.badRequest({ message: 'Must specify a card' });
-      }
+    case MoveType.ONE_OFF:
+      {
+        if (!cardId) {
+          return res.badRequest({ message: 'Must specify a card' });
+        }
 
-      if (!DeckIds.includes(cardId)) {
-        return res.badRequest({ message: `${cardId} is not a valid cardId` });
-      }
+        if (!DeckIds.includes(cardId)) {
+          return res.badRequest({ message: `${cardId} is not a valid cardId` });
+        }
 
-      const [ rankAsStr ] = cardId;
-      const isTwoOrNine = ['2', '9'].includes(rankAsStr);
-      const missingTarget = !targetId || !DeckIds.includes(targetId) || !['point', 'jack', 'faceCard'].includes(targetType);
-      if (isTwoOrNine && missingTarget) {
-        return res.badRequest({ message: 'You cannot play that one-off without a target' });
+        const [rankAsStr] = cardId;
+        const isTwoOrNine = ['2', '9'].includes(rankAsStr);
+        const missingTarget =
+          !targetId || !DeckIds.includes(targetId) || !['point', 'jack', 'faceCard'].includes(targetType);
+        if (isTwoOrNine && missingTarget) {
+          return res.badRequest({ message: 'You cannot play that one-off without a target' });
+        }
       }
-    }
 
       return next();
+
+    case MoveType.JACK:
+      {
+        if (!cardId) {
+          return res.badRequest({ message: 'Must specify a card' });
+        }
+        if (!targetId) {
+          return res.badRequest({ message: 'Must have a card targeted' });
+        }
+
+        if (!DeckIds.includes(cardId)) {
+          return res.badRequest({ message: `${cardId} is not a valid cardId` });
+        }
+
+        if (!DeckIds.includes(targetId)) {
+          return res.badRequest({ message: `${targetId} is not a valid cardId` });
+        }
+      }
+      return next();
+
     default:
       return res.badRequest({ message: `Invalid moveType of ${req.body.moveType}` });
   }
--- a/tests/e2e/specs/in-game/basicMoves.spec.js
+++ b/tests/e2e/specs/in-game/basicMoves.spec.js
@@ -510,7 +510,6 @@ describe('Play Jacks', () => {
   });
 
   it('Player and Opponent plays Jacks on different cards', () => {
-    cy.skipOnGameStateApi();
     // Set Up
     cy.loadGameFixture(0, {
       p0Hand: [Card.ACE_OF_SPADES, Card.JACK_OF_CLUBS, Card.KING_OF_SPADES],
@@ -554,7 +553,6 @@ describe('Play Jacks', () => {
   });
 
   it('Quadruple jacks successfully', () => {
-    cy.skipOnGameStateApi();
     // Set Up
     cy.loadGameFixture(0, {
       p0Hand: [Card.ACE_OF_SPADES, Card.JACK_OF_CLUBS, Card.KING_OF_SPADES, Card.JACK_OF_HEARTS],


api/helpers/gamestate/moves/jack/validate.js
```diff
@@ -0,0 +1,75 @@
...
+      //is targetcard in the opponent points
+      if (!targetCard) {
+        throw new Error('game.snackbar.jack.stealOnlyPointCards');
+      }
+      //is any card frozen
+      if (cardPlayed.isFrozen || targetCard.isFrozen) {
+        throw new Error('game.snackbar.global.cardFrozen');
+      }
+      //does opponent have a queen in his faceCards to block the move
+      const queenCount = opponent.faceCards.filter((faceCard) => faceCard.rank === 12).length;
```
Haviles04, line 65:
```suggestion
      const queenCount = opponent.faceCards.filter(({ rank }) => rank === 12).length;
```

Haviles04 (commented):
This is looking really good, I've only really got a few nitpicks about spellings and one destrucuring. Great work!

api/helpers/gamestate/moves/jack/execute.js
```diff
@@ -0,0 +1,71 @@
...
+
+      const { cardId, targetId } = requestedMove;
+
+      const player = playedBy ? result.p1 : result.p0;
+      const opponent = playedBy ? result.p0 : result.p1;
+
+      const playedCard = player.hand.find(({ id }) => id === cardId);
+      const targetCard = opponent.points.find(({ id }) => id === targetId);
+
+      //add jack(playedCard) to targetcard attachment
```
itsalaidbacklife, line 44:
```suggestion
      // add jack (playedCard) to targetCard's attachments
```

api/helpers/gamestate/moves/jack/execute.js
```diff
@@ -0,0 +1,71 @@
...
+      targetCard.attachments.push(playedCard);
+      //add targetCard to player points
+      player.points.push(targetCard);
+      // remove jack(playedCard) from player hand
+      const cardIndex = player.hand.findIndex(({ id }) => id === cardId);
+      player.hand.splice(cardIndex, 1);
```
itsalaidbacklife, line 50: Let's cut down on loops by making the findIndex + splice calls the way that we assign the variables for playedCard and targetCard.

Like first findIndex, then

```
const [ playedCard ] = player.hand.splice(playedCardIndex, 1);
```

And same for target card. Then removing captures the variable at the same time

api/helpers/gamestate/moves/jack/execute.js
```diff
@@ -0,0 +1,71 @@
...
+
+      const player = playedBy ? result.p1 : result.p0;
+      const opponent = playedBy ? result.p0 : result.p1;
+
+      const playedCard = player.hand.find(({ id }) => id === cardId);
+      const targetCard = opponent.points.find(({ id }) => id === targetId);
+
+      //add jack(playedCard) to targetcard attachment
+      targetCard.attachments.push(playedCard);
+      //add targetCard to player points
```
itsalaidbacklife, line 46:
```suggestion
      // add targetCard to player's points
```

api/helpers/gamestate/moves/jack/execute.js
```diff
@@ -0,0 +1,71 @@
...
+      player.points.push(targetCard);
+      // remove jack(playedCard) from player hand
```
itsalaidbacklife, line 48:
```suggestion
      // remove jack(playedCard) from player's hand
```

api/helpers/gamestate/moves/jack/execute.js
```diff
@@ -0,0 +1,71 @@
...
+      const cardIndex = player.hand.findIndex(({ id }) => id === cardId);
+      player.hand.splice(cardIndex, 1);
+      //remove target card from oppponent points
```
itsalaidbacklife, line 51:
```suggestion
      // remove target card from oppponent's points
```

api/helpers/gamestate/moves/jack/validate.js
```diff
@@ -0,0 +1,75 @@
...
+
+  fn: ({ currentState, requestedMove, playedBy }, exits) => {
+    try {
+      const player = playedBy ? currentState.p1 : currentState.p0;
+      const opponent = playedBy ? currentState.p0 : currentState.p1;
+
+      const cardPlayed = player.hand.find(({ id }) => id === requestedMove.cardId);
+      const targetCard = opponent.points.find(({ id }) => id === requestedMove.targetId);
+
+      //gameState phase should be MAIN
```
itsalaidbacklife, line 40: Let's add whitespace before the comments

api/helpers/gamestate/moves/jack/validate.js
```diff
@@ -0,0 +1,75 @@
...
+      if (currentState.phase !== GamePhase.MAIN) {
+        throw new Error(`Can only play a jack in main phase, not ${currentState.phase}`);
+      }
+      //is it player turn
+      if (currentState.turn % 2 !== playedBy) {
+        throw new Error('game.snackbar.global.notYourTurn');
+      }
+      //is cardPlayed in player hand
```
itsalaidbacklife, line 48: And a newline between each conditional block
```suggestion

      //is cardPlayed in player hand
```

api/helpers/gamestate/moves/jack/validate.js
```diff
@@ -0,0 +1,75 @@
...
+      if (!cardPlayed) {
+        throw new Error('game.snackbar.global.playFromHand');
+      }
+      //is cardPlayed a jack
+      if (!cardPlayed.rank === '11') {
```
itsalaidbacklife, line 53: The rank should be a number. It looks like this works because the ! Being in front goofs the conditional. That's because `!cardPlayed.rank` casts the rank to boolean and evaluates to `true` which then will not be `=== '11'`, so we never enter this conditional. 

We want 
```suggestion
      if (cardPlayed.rank !== 11) {
```

itsalaidbacklife: This showcases a shortcoming in our tests because we don't have anything that verifies that the backend errors here

api/helpers/gamestate/moves/jack/validate.js
```diff
@@ -0,0 +1,75 @@
...
+        throw new Error('game.snackbar.jack.stealOnlyPointCards');
+      }
+      //is targetcard in the opponent points
+      if (!targetCard) {
+        throw new Error('game.snackbar.jack.stealOnlyPointCards');
+      }
+      //is any card frozen
+      if (cardPlayed.isFrozen || targetCard.isFrozen) {
```
itsalaidbacklife, line 61: Only the cardPlayed being frozen is an issue. Being frozen means you can't play it this turn, but it's not an issue for your target to be frozen, though that's not actually possible because cards are only frozen in the hand 
```suggestion
      if (cardPlayed.isFrozen) {
```

api/helpers/gamestate/moves/jack/validate.js
```diff
@@ -0,0 +1,75 @@
...
+        throw new Error('game.snackbar.global.cardFrozen');
+      }
+      //does opponent have a queen in his faceCards to block the move
```
itsalaidbacklife, line 64:
```suggestion
      // Can't hack it opponent has queen
```

api/policies/hasValidMoveBody.js
```diff
@@ -29,24 +29,46 @@ module.exports = function (req, res, next) {
...
+        if (!targetId) {
+          return res.badRequest({ message: 'Must have a card targeted' });
+        }
+
+        if (!DeckIds.includes(cardId)) {
+          return res.badRequest({ message: `${cardId} is not a valid cardId` });
+        }
+
+        if (!DeckIds.includes(targetId)) {
+          return res.badRequest({ message: `${targetId} is not a valid cardId` });
```
itsalaidbacklife, line 67:
```suggestion
          return res.badRequest({ message: `${targetId} is not a valid targetId` });
```

tests/e2e/specs/in-game/basicMoves.spec.js
itsalaidbacklife, line 1: Can you also unskip the 'plays queens' spec? It turns out that one also required jacks so we couldn't unskip it in the face cards pr

itsalaidbacklife (commented):
Nice work! Only small stuff to fix

itsalaidbacklife:
> thanks, also I am seing some error handling doesn't use the translation, I am going to implement that. do you think that sentence is correct = 'You can only play that card in main phase ' or would it be more accesible to say = 'you cannot play that card as a one-off or countering hasn't been resolved'

How about `'You can only do that in the main phase'` so that it's generic to any move? Then we can reuse it in every relevant validate helper (including draw/pass that don't play cards)

api/helpers/gamestate/moves/jack/execute.js
```diff
@@ -0,0 +1,72 @@
...
+      const opponent = playedBy ? result.p0 : result.p1;
+
+      const cardIndex = player.hand.findIndex(({ id }) => id === cardId);
+      const targetIndex = opponent.points.findIndex(({ id }) => id === targetId);
+
+      //remove card from player's hand
+      const [playedCard] = player.hand.splice(cardIndex, 1);
+
+      //remove target card from oppponent's points
+      const [targetCard] = opponent.points.splice(targetIndex, 1);
```
itsalaidbacklife, line 48: Prefer whitespace around brackets and curly braces. This hasn't been enforced automatically, so I've written up #1049 for the future to bake this into eslint. For now, let's just add space by hand
```suggestion
      const [ playedCard ] = player.hand.splice(cardIndex, 1);

      //remove target card from oppponent's points
      const [ targetCard ] = opponent.points.splice(targetIndex, 1);
```

api/helpers/gamestate/moves/jack/validate.js
```diff
@@ -0,0 +1,80 @@
...
+
+      const cardPlayed = player.hand.find(({ id }) => id === requestedMove.cardId);
+      const targetCard = opponent.points.find(({ id }) => id === requestedMove.targetId);
+
+      //gameState phase should be MAIN
+      if (currentState.phase !== GamePhase.MAIN) {
+        throw new Error('game.snackbar.global.notInMainPhase');
+      }
+
+      //is it player turn
```
itsalaidbacklife, line 45:
```suggestion
      // Must be player's turn
```

api/helpers/gamestate/moves/jack/validate.js
```diff
@@ -0,0 +1,80 @@
...
+      if (currentState.turn % 2 !== playedBy) {
+        throw new Error('game.snackbar.global.notYourTurn');
+      }
+      //is cardPlayed in player hand
```
itsalaidbacklife, line 49:
```suggestion
      // playedCard must be in player's hand
```

api/helpers/gamestate/moves/jack/validate.js
```diff
@@ -0,0 +1,80 @@
...
+      if (!cardPlayed) {
+        throw new Error('game.snackbar.global.playFromHand');
+      }
+
+      //is cardPlayed a jack
```
itsalaidbacklife, line 54: Let's write the comments in the form "X must Y" instead of as questions
```suggestion
      // playedCard must be a jack
```

api/helpers/gamestate/moves/jack/validate.js
```diff
@@ -0,0 +1,80 @@
...
+      if (!targetCard) {
+        throw new Error('game.snackbar.jack.stealOnlyPointCards');
+      }
+
+      //is cardPlayed frozen
+      if (cardPlayed.isFrozen) {
+        throw new Error('game.snackbar.global.cardFrozen');
+      }
+
+      //Can't hack if opponent has queen
```
itsalaidbacklife, line 69:
```suggestion
      // Can't jack if opponent has queen
```

tests/e2e/specs/in-game/basicMoves.spec.js
```diff
@@ -229,7 +229,7 @@ describe('Game Basic Moves - P0 Perspective', () => {
   });
 
   it('Plays Queens', () => {
-    cy.skipOnGameStateApi();
+    //cy.skipOnGameStateApi();
```
itsalaidbacklife, line 232: Remove instead of commenting out
```suggestion
```

src/translations/en.json
```diff
@@ -248,7 +248,8 @@
         "cardFrozen": "That card is frozen! You must wait a turn to play it.",
         "playFromHand": "You can only play a card that is in your hand",
         "blockedByQueen": "Your opponent's queen prevents you from targeting their other cards",
-        "blockedByMultipleQueens": "You cannot play a Targeted One-Off (Two, Nine) when your opponent has more than one Queen"
+        "blockedByMultipleQueens": "You cannot play a Targeted One-Off (Two, Nine) when your opponent has more than one Queen",
+        "notInMainPhase" : "You can only play that card in a main phase"
```
itsalaidbacklife, line 252: Great! Can you please update the existing `validate` functions to throw this same error?

Let's also reword it to not mention playing a card, because this also applies to drawing and passing
```suggestion
        "notInMainPhase" : "You can only do that in the main phase"
```

Please make the same revision in the other language files as well. This will also require updating 
```
it('Displays "Opponent May Counter" when player had neither glasses nor a queen')
```
in `0_countering.spec.js` because it makes an assertion about the error message that comes back, so we'll need to write this message instead. in the assertion of the error

api/helpers/gamestate/moves/jack/validate.js
```diff
@@ -0,0 +1,80 @@
...
+    },
+  },
+  sync: true,
+
+  fn: ({ currentState, requestedMove, playedBy }, exits) => {
+    try {
+      const player = playedBy ? currentState.p1 : currentState.p0;
+      const opponent = playedBy ? currentState.p0 : currentState.p1;
+
+      const cardPlayed = player.hand.find(({ id }) => id === requestedMove.cardId);
```
itsalaidbacklife, line 37: Let's rename this variable to `playedCard` to be more consistent with the `GameStateRow`, `GameState` and the `execute()` helpers. I had this backwards in an early `validate()` example and it propogated to all the validate helpers. I'm changing all the others back to `playedCard` in #1051. Let's follow suit here
```suggestion
      const playedCard = player.hand.find(({ id }) => id === requestedMove.cardId);
```

tests/e2e/specs/in-game/basicMoves.spec.js
itsalaidbacklife, line 1: We've just added a new test case that can be unskipped in this PR as well: 
`1_aces.spec.js` - `it('It scraps attached jacks when an ace resolves')`

api/helpers/gamestate/moves/jack/execute.js
```diff
@@ -0,0 +1,72 @@
...
+        playedBy,
+        playedCard,
+        targetCard,
+        phase: GamePhase.MAIN,
+      };
+
+      return exits.success(result);
+    } catch (err) {
+      return exits.error({ message: 'Cannot play Jack ' + err });
+    }
```
itsalaidbacklife, line 70: I don't believe that we need this catch block because uncaught errors should bubble up to the controller already

api/helpers/gamestate/moves/jack/validate.js
```diff
@@ -0,0 +1,80 @@
...
+
+      // GameState phase should be MAIN
+      if (currentState.phase !== GamePhase.MAIN) {
+        throw new Error('game.snackbar.global.notInMainPhase');
+      }
+
+      // Must be player's turn
+      if (currentState.turn % 2 !== playedBy) {
+        throw new Error('game.snackbar.global.notYourTurn');
+      }
```
itsalaidbacklife, line 48:
```suggestion
      }

```

api/helpers/gamestate/moves/jack/validate.js
```diff
@@ -0,0 +1,80 @@
...
+      // Must playedCard be in player's hand
```
itsalaidbacklife, line 49:
```suggestion
      // playedCard must be in player's hand
```

tests/e2e/specs/in-game/one-offs/0_countering.spec.js
```diff
@@ -514,6 +514,7 @@ describe('Opponent May Counter vs Opponent Must Resolve', () => {
             const errors = [
               "Can't play while waiting for opponent to counter",
               'Can only draw in main phase',
```
itsalaidbacklife, line 516: Can we remove the second error message here? It should only be the first one or the last one you added below since that's what it now says in the GameState API

itsalaidbacklife (approved):
Excellent work. I have a couple nitpicks and then this is ready

itsalaidbacklife (approved):
Excellent work. Thank you for your contribution!

PR #1047, opened 2024-09-18
Feat/resolve gamestate

### LINKED ISSUE #953: [Feature]: Implement the outlined control flow for the `resolve()` endpoint in the GameState API
### Feature Summary

We need to implement Resolving one-offs in the GameState API in order to support the full game logic in this API. See the [spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.wpeuh0qc6ps2) for a general overview of the API, its underlying motivations, and the development approach.

### Detailed Description

This issue is to create the new version of the endpoint and handle the logic for the case where the one-off is countered (even number of twos), and in the case where the one-off resolves, it should `switch()` on the resolving `oneOff`'s rank, and throw an error in each case saying that "Resolving 5's are not yet built" etc for each of the moves. Then we will have separate issues to build out resolving of each individual one-off effect.

The endpoint should be:

POST /api/game/:gameId/move/resolve
and it should expect an empty request body:
```
{}; 
```
Unlike the legacy `/resolve` endpoint, which has all its logic in one large function, we should split this into one helper for each one-off. The helpers should be placed in `api/helpers/moves/resolve` and each be named after their oneoff e.g. `api/helpers/moves/resolve/resolve-ace.js`. Then the large switch based on the rank of the resolving `oneOff` can delegate to the appropriate helper. Each of these individual helpers will be built in its own issue + PR.

In addition to building the endpoint, the frontend consumption of it in the src/stores/game.js file's `requestResolve()` function should be updated to make a request to this endpoint (instead of throwing an error) when the `process.env.VITE_USE_GAME_STATE_API` var equals `"true"`.

The same should be applied to the `resolveOpponent()` test command in `tests/e2e/support/commands.js` when the `CYPRESS_USE_GAMESTATE_API` env var is set to `"true"`. If so, (instead of erroring), it should send a request to the new api endpoint.

### LINKED ISSUE #966: [Feature]: Implement `resolveAce()` helper and call it in general `resolve()` endpoint
### Feature Summary

We need to build the effect of an ace resolving as a helper and call it in `resolve()` in the GameState API to complete the revised game logic. See the [spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.wpeuh0qc6ps2) for a general overview of the API, its underlying motivations, and the development approach.

### Detailed Description

We should create a `resolveAce()` helper in `api/helpers/moves/resolve-ace/execute.js` that inputs a GameState and then returns a new GameState that applies the ace's one-off effect of scrapping all the points.

The new helper should then be called in the `resolve()` endpoint in the case where an ace is resolving without being countered.

itsalaidbacklife:
Implements resolve and resolving an ace
## Issue number

- Resolves #953,
- Resolves #966 

## Please check the following

- For New Features:

### DIFF
--- a/api/helpers/gamestate/moves/draw/validate.js
+++ b/api/helpers/gamestate/moves/draw/validate.js
@@ -33,7 +33,7 @@ module.exports = {
 
       // Must be MAIN phase of the turn
       if (currentState.phase !== GamePhase.MAIN) {
-        throw new Error(`Can only play draw in main phase, not ${currentState.phase}`);
+        throw new Error(`Can only draw in main phase, not ${currentState.phase}`);
       }
 
       // Must be your turn
--- a/api/helpers/gamestate/moves/draw/validate.js
+++ b/api/helpers/gamestate/moves/draw/validate.js
@@ -12,9 +12,9 @@ module.exports = {
       required: true,
     },
     /**
-     * @param {Object} requestedMove - Object describing the request to play points
+     * @param { MoveType.DRAW } requestedMove.moveType - Specifies that this a draw move
+     * @param {Object} requestedMove - Object describing the request to draw
      * @param {1 | 0} requestedMove.playedBy - Which player is playing
-     * @param { MoveType.DRAW } requestedMove.moveType - Specifies that this a Points move
      */
     requestedMove: {
       type: 'ref',
@@ -33,7 +33,7 @@ module.exports = {
 
       // Must be MAIN phase of the turn
       if (currentState.phase !== GamePhase.MAIN) {
-        throw new Error(`Can only play points in main phase, not ${currentState.phase}`);
+        throw new Error(`Can only play draw in main phase, not ${currentState.phase}`);
       }
 
       // Must be your turn
--- a/api/helpers/gamestate/moves/draw/validate.js
+++ b/api/helpers/gamestate/moves/draw/validate.js
@@ -13,8 +13,8 @@ module.exports = {
     },
     /**
      * @param { MoveType.DRAW } requestedMove.moveType - Specifies that this a draw move
-     * @param {Object} requestedMove - Object describing the request to draw
-     * @param {1 | 0} requestedMove.playedBy - Which player is playing
+     * @param { Object } requestedMove - Object describing the request to draw
+     * @param { 1 | 0 } requestedMove.playedBy - Which player is drawing
      */
     requestedMove: {
       type: 'ref',
--- a/api/helpers/gamestate/moves/resolve/ace.js
+++ b/api/helpers/gamestate/moves/resolve/ace.js
@@ -1,0 +1,28 @@
+module.exports = {
+  friendlyName: 'Resolve Ace One-Off',
+
+  description: 'Returns new GameState resulting from resolving an ace, which scraps all points on the field, and all jacks attached to those point cards.',
+
+  inputs: {
+    currentState: {
+      type: 'ref',
+      description: 'The latest GameState before the ace scraps all point cards and their attachments',
+      required: true,
+    },
+  },
+  sync: true, // synchronous helper
+  fn: ({ currentState }, exits) => {
+    let result = _.cloneDeep(currentState);
+
+    // Scrap all points and all their attachments
+    for (let card of [ ...currentState.p0.points, ...currentState.p1.points]) {
+      result.scrap.push(card, ...card.attachments);
+      card.attachments = [];
+    }
+
+    result.p0.points = [];
+    result.p1.points = [];
+
+    return exits.success(result);
+  },
+};
--- a/api/helpers/gamestate/moves/resolve/execute.js
+++ b/api/helpers/gamestate/moves/resolve/execute.js
@@ -42,6 +42,7 @@ module.exports = {
       twos: [],
       turn: result.turn + 1,
       playedBy,
+      resolved: oneOff,
     };
 
     // If one-off was fizzles, make no other changes
@@ -52,6 +53,9 @@ module.exports = {
     }
 
     switch (oneOff.rank) {
+      case 1:
+        result = sails.helpers.gamestate.moves.resolve.ace(result);
+        break;
       default:
         return exits.error(new Error(`${oneOff.rank} is not a valid one-off rank`));
     }
--- a/api/helpers/gamestate/moves/resolve/execute.js
+++ b/api/helpers/gamestate/moves/resolve/execute.js
@@ -38,11 +38,12 @@ module.exports = {
     result = {
       ...result,
       ...requestedMove,
+      phase: GamePhase.MAIN,
+      playedBy,
+      resolved: oneOff,
       oneOff: null,
       twos: [],
       turn: result.turn + 1,
-      playedBy,
-      resolved: oneOff,
     };
 
     // If one-off was fizzles, make no other changes
--- a/api/helpers/gamestate/moves/resolve/execute.js
+++ b/api/helpers/gamestate/moves/resolve/execute.js
@@ -42,6 +42,8 @@ module.exports = {
       playedBy,
       resolved: oneOff,
       oneOff: null,
+      oneOffTarget: null,
+      oneOffTargetType: null,
       twos: [],
       turn: result.turn + 1,
     };
--- a/api/helpers/gamestate/moves/resolve/execute.js
+++ b/api/helpers/gamestate/moves/resolve/execute.js
@@ -30,7 +30,7 @@ module.exports = {
   fn: ({ currentState, requestedMove, playedBy }, exits) => {
     let result = _.cloneDeep(currentState);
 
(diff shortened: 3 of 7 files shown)


api/helpers/gamestate/moves/resolve/ace.js
itsalaidbacklife, line 1: I debated some where to put these sub-helpers and how to name them. I'm leaning towards putting them all in here and naming them after their rank in string form e.g. resolve/ace.js, resolve/two.js etc.

The slightly funky thing is that in addition to these, there will be separate moveType's for `resolveThree`, `resolveFour`, `resolveFive`, and all the seven moves e.g. `sevenPoints`

api/helpers/gamestate/moves/resolve/ace.js
```diff
@@ -0,0 +1,28 @@
...
+  friendlyName: 'Resolve Ace One-Off',
+
+  description: 'Returns new GameState resulting from resolving an ace, which scraps all points on the field, and all jacks attached to those point cards.',
+
+  inputs: {
+    currentState: {
+      type: 'ref',
+      description: 'The latest GameState before the ace scraps all point cards and their attachments',
+      required: true,
+    },
```
itsalaidbacklife, line 11: Notably, unlike the actual `execute()` helpers, the one-off resolutions only require the `currentState` without caring about who is resolving etc.

api/helpers/gamestate/moves/resolve/ace.js
```diff
@@ -0,0 +1,28 @@
...
+    let result = _.cloneDeep(currentState);
+
+    // Scrap all points and all their attachments
+    for (let card of [ ...currentState.p0.points, ...currentState.p1.points]) {
+      result.scrap.push(card, ...card.attachments);
+      card.attachments = [];
+    }
+
+    result.p0.points = [];
+    result.p1.points = [];
```
itsalaidbacklife, line 24: I'm optimistic that similar to the result of splitting all the moves into their own helpers, that having separate one-off resolutions in their own helpers will make the actual logic of each resolving helper fairly lean. I'm pleased with how this one turned out.

It's notable that all the generic things about changing data e.g. updating the turn, scrapping the one-off itself and all the twos, setting the phase back to MAIN, all happens higher up the control flow in `resolve/execute.js` so each one off resolution is really only responsible for doing it's specific effect, without the general one-off cleanup

api/helpers/gamestate/moves/resolve/execute.js
```diff
@@ -0,0 +1,66 @@
...
+    result = {
+      ...result,
+      ...requestedMove,
+      phase: GamePhase.MAIN,
+      playedBy,
+      resolved: oneOff,
+      oneOff: null,
+      twos: [],
+      turn: result.turn + 1,
+    };
```
itsalaidbacklife, line 47: This is all the generic handling shared by all the one-offs that fully resolve without needing additional player input. It will be more complicated for threes, fours, fives, and sevens, which then require the player to do additional stuff. It's unclear to me atm whether to have those cases consume `currentState` instead of `result` to effectively ignore these changes and recreate their own, or whether to have them roll back the unnecessary changes by explicitly setting these values e.g. `turn` differently. I lean towards just having them manipulate `currentState` instead of `result``.

api/helpers/gamestate/moves/resolve/execute.js
```diff
@@ -0,0 +1,66 @@
...
+    playedBy: {
+      type: 'number',
+      description: 'Player number of player requesting move.',
+    },
+  },
+  sync: true, // synchronous helper
+  fn: ({ currentState, requestedMove, playedBy }, exits) => {
+    let result = _.cloneDeep(currentState);
+
+    const { oneOff } = result;
```
itsalaidbacklife, line 33: I believe that 2's and 9's will also need to grab the `targetCard` and `targetType` from the `result` before those values are set to null below, but that isn't used yet so eslint was erroring. We'll re-add them once 2 or 9 is put into place

api/helpers/gamestate/moves/resolve/execute.js
```diff
@@ -0,0 +1,66 @@
...
+    // Move oneOff + twos into scrap and increment turn
+    result.scrap.push(result.oneOff);
+    result.scrap.push(...result.twos);
+    result = {
+      ...result,
+      ...requestedMove,
+      phase: GamePhase.MAIN,
+      playedBy,
+      resolved: oneOff,
+      oneOff: null,
```
itsalaidbacklife, line 44: I think we also need to set `oneOffTarget` and `oneOffTargetType` to `null` here.

api/helpers/gamestate/moves/resolve/execute.js
```diff
@@ -0,0 +1,66 @@
...
+      twos: [],
+      turn: result.turn + 1,
+    };
+
+    // If one-off was fizzles, make no other changes
+    const fizzles = result.twos.length % 2 === 1;
+    if (fizzles) {
+      result.moveType = MoveType.FIZZLE;
+      return exits.success(result);
+    }
```
itsalaidbacklife, line 54: Taking the early return approach in case the one-off fizzles/doesn't happen. I hope the conditional is clear i.e. that an odd number of counters means the one-off doesn't do it's effect

api/helpers/gamestate/moves/resolve/execute.js
```diff
@@ -0,0 +1,66 @@
...
+
+    switch (oneOff.rank) {
+      case 1:
+        result = sails.helpers.gamestate.moves.resolve.ace(result);
+        break;
+      default:
+        return exits.error(new Error(`${oneOff.rank} is not a valid one-off rank`));
+    }
```
itsalaidbacklife, line 62: We could consider using the same technique as the `move` endpoint and dynamically referring to the helper name here rather than switching, but switch allows us to set the input differently for 2's and 9's, which may be helpful. 

Penny for your thoughts @seriouslysean and @Haviles04

api/helpers/gamestate/moves/resolve/validate.js
```diff
@@ -0,0 +1,58 @@
...
+
+      // Must be your chance to counter
+      // It's your chance if it is your turn and there are an odd number of twos
+      // or if it's your opponent's turn and there are an even number of twos
+      // (since the opponent of the person who played the one-off counters first)
+      const isYourTurn = currentState.turn % 2 === playedBy;
+      const numTwosIsEven = currentState.twos.length % 2 === 0;
+      if (isYourTurn === numTwosIsEven) {
+        throw new Error('Waiting for opponent to counter');
+      }
```
itsalaidbacklife, line 51: This took me a while to suss out, so feedback is welcome on the clarity. Whether you are the one who can counter or not right now comes down to whether it's your turn and how many twos there are.

If it is your turn, then you played the one-off and then your opponent should counter first, which means you can only play if there are an odd number of twos. So it's an error if it's your turn and there are an even number of twos i.e. `isYourTurn === numTwosIsEven === true`

If it is **not** your turn, then your opponent played the one-off and _you_ counter first, so it is an error if it is not your turn and there are an odd number of twos i.e. `isYourTurn === numTwosIsEven === false`.

So ultimately that means we should error if both values are true or if both values are false, which we can represent succinctly as error if both values equal each other. It took me a while to work that out, so again, @seriouslysean and @Haviles04 I'd love to hear your thoughts on how it reads.

@Haviles04 this exact logic is also needed to verify if you can play a counter for #952 so we can discuss whether it makes sense to abstract it into a helper

api/helpers/gamestate/moves/draw/validate.js
```diff
@@ -33,7 +33,7 @@ module.exports = {
 
       // Must be MAIN phase of the turn
       if (currentState.phase !== GamePhase.MAIN) {
-        throw new Error(`Can only play points in main phase, not ${currentState.phase}`);
+        throw new Error(`Can only play draw in main phase, not ${currentState.phase}`);
```
Haviles04, line 36:
```suggestion
        throw new Error(`Can only draw in main phase, not ${currentState.phase}`);
```

api/helpers/gamestate/moves/resolve/execute.js
Haviles04: I think this is fine, it's clearer and we don't really need to be dynamic here

tests/e2e/specs/in-game/one-offs/1_aces.spec.js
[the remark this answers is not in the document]
itsalaidbacklife: I think the move is to add + skip a test case in `1_aces.spec` and then unskip it in #1045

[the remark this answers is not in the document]
itsalaidbacklife: Added test case to bottom of this file

PR #1048, opened 2024-09-19
Feat/counter gamestate

### LINKED ISSUE #952: [Feature]: Implement Countering in the GameState API
### Feature Summary

We need to implement Countering in the GameState API in order to support the full game logic in this API. See the [spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.wpeuh0qc6ps2) for a general overview of the API, its underlying motivations, and the development approach.

### Detailed Description

The endpoint should be:

POST /api/game/:gameId/move/counter
```
{ 
  playedCardId: '2H', // countering with the Two of Hearts
}; 
```
In addition to building the endpoint, the frontend consumption of it in the src/stores/game.js file's `requestCounter()` function should be updated to make a request to this endpoint (instead of throwing an error) when the `process.env.VITE_USE_GAME_STATE_API` var equals `"true"`.

The same should be applied to the `counterOpponent()` test command in `tests/e2e/support/commands.js` when the `CYPRESS_USE_GAMESTATE_API` env var is set to `"true"`. If so, (instead of erroring), it should send a request to the new api endpoint.

itsalaidbacklife:
## Issue number

- Resolves #952 

## Please check the following

- For New Features:

### DIFF
--- a/api/helpers/gamestate/get-log.js
+++ b/api/helpers/gamestate/get-log.js
@@ -81,7 +81,7 @@ module.exports = {
 
         case MoveType.FIZZLE:
           return `The ${getFullCardName(
-            game.oneOff,
+            resolved,
           )} is countered, and all cards played this turn are scrapped.`;
 
         case MoveType.RESOLVE:
--- a/api/helpers/gamestate/moves/resolve/execute.js
+++ b/api/helpers/gamestate/moves/resolve/execute.js
@@ -42,6 +42,7 @@ module.exports = {
       twos: [],
       turn: result.turn + 1,
       playedBy,
+      resolved: oneOff,
     };
 
     // If one-off was fizzles, make no other changes
@@ -52,6 +53,9 @@ module.exports = {
     }
 
     switch (oneOff.rank) {
+      case 1:
+        result = sails.helpers.gamestate.moves.resolve.ace(result);
+        break;
       default:
         return exits.error(new Error(`${oneOff.rank} is not a valid one-off rank`));
     }
--- a/api/helpers/gamestate/moves/resolve/execute.js
+++ b/api/helpers/gamestate/moves/resolve/execute.js
@@ -33,6 +33,7 @@ module.exports = {
     const { oneOff } = result;
 
     // Move oneOff + twos into scrap and increment turn
+    const fizzles = result.twos.length % 2 === 1;
     result.scrap.push(result.oneOff);
     result.scrap.push(...result.twos);
     result = {
@@ -48,8 +49,7 @@ module.exports = {
       turn: result.turn + 1,
     };
 
-    // If one-off was fizzles, make no other changes
-    const fizzles = result.twos.length % 2 === 1;
+    // If one-off fizzles, make no other changes
     if (fizzles) {
       result.moveType = MoveType.FIZZLE;
       return exits.success(result);
--- a/api/helpers/gamestate/moves/resolve/execute.js
+++ b/api/helpers/gamestate/moves/resolve/execute.js
@@ -38,11 +38,12 @@ module.exports = {
     result = {
       ...result,
       ...requestedMove,
+      phase: GamePhase.MAIN,
+      playedBy,
+      resolved: oneOff,
       oneOff: null,
       twos: [],
       turn: result.turn + 1,
-      playedBy,
-      resolved: oneOff,
     };
 
     // If one-off was fizzles, make no other changes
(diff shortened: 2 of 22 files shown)


api/helpers/gamestate/get-log.js
```diff
@@ -81,7 +81,7 @@ module.exports = {
 
         case MoveType.FIZZLE:
           return `The ${getFullCardName(
-            game.oneOff,
+            resolved,
```
itsalaidbacklife, line 84: By the time a one-off fizzles, it is in the scrap so we need to use the `resolved` variable which represents which one-off just fizzled/resolved instead of `game.oneOff`

api/helpers/gamestate/moves/resolve/execute.js
```diff
@@ -33,6 +33,7 @@ module.exports = {
     const { oneOff } = result;
 
     // Move oneOff + twos into scrap and increment turn
+    const fizzles = result.twos.length % 2 === 1;
```
itsalaidbacklife, line 36: We were previously checking whether the one-off fizzles _after_ setting the twos to an empty array, so it never fizzled. Need to check beforehand like so

api/helpers/gamestate/moves/resolve/validate.js
```diff
@@ -40,13 +40,9 @@ module.exports = {
...
-      // Must be your chance to counter
-      // It's your chance if it is your turn and there are an odd number of twos
-      // or if it's your opponent's turn and there are an even number of twos
-      // (since the opponent of the person who played the one-off counters first)
-      const isYourTurn = currentState.turn % 2 === playedBy;
-      const numTwosIsEven = currentState.twos.length % 2 === 0;
-      if (isYourTurn === numTwosIsEven) {
+      // Must be your chance to resolve
+      const yourTurnToResolve = sails.helpers.gamestate.yourTurnToCounter(currentState, playedBy);
+      if (!yourTurnToResolve) {
```
itsalaidbacklife, line 45: Pulled this into a helper since it's used in resolve/validate and counter/validate

api/policies/hasValidMoveBody.js
```diff
@@ -17,8 +17,10 @@ module.exports = function (req, res, next) {
     case MoveType.RESOLVE:
       // These moves require no extra data
     return next();
-    
-    case MoveType.FACECARD: {
+
+    case MoveType.POINTS:
+    case MoveType.FACECARD:
+    case MoveType.COUNTER: {
```
itsalaidbacklife, line 23: These are all moves that require specifying one card without a target

src/plugins/sockets/inGameEvents.js
```diff
@@ -65,6 +65,7 @@ export async function handleInGameEvents(evData) {
       gameStore.processFives(evData.discardedCards, evData.game);
       break;
     case SocketEvent.RESOLVE:
+    case SocketEvent.FIZZLE:
```
itsalaidbacklife, line 68: Need a socket event for fizzle in order for the client to respond to it. Or else we could have updated publish-game to use `SocketEvent.RESOLVE`, but I think this is better. The awkward bit is that it checks whether the one-off happened inside the switch case, instead of treating it as a separate case. We can clean this up upon switching to the gamestate api

tests/e2e/specs/in-game/one-offs/0_countering.spec.js
itsalaidbacklife, line 1: Got to unskip everything in this file except `it('Can counter a three', () => {` because threes aren't supported yet. Need to update the corresponding issue to mention this test

tests/e2e/support/commands.js
```diff
@@ -1166,8 +1166,11 @@ Cypress.Commands.add('loadGameFixture', (pNum, fixture) => {
     return cy
       .window()
       .its('cuttle.gameStore.id')
-      .then((gameId) => {
-        return cy.makeSocketRequest(`game/${gameId}`, 'loadFixtureGameState', fixture);
+      .then(async (gameId) => {
+        await cy.makeSocketRequest(`game/${gameId}`, 'loadFixtureGameState', fixture);
+        const playerHandLength = pNum === 0 ? fixture.p0Hand.length : fixture.p1Hand.length;
+        cy.get('[data-player-hand-card]').should('have.length', playerHandLength);
```
itsalaidbacklife, line 1172: The gamestate api case here was missing a DOM assertion to confirm that the request finished processing before ending the command. This way we can immediately call things like `cy.playOneOffOpponent()` because the cards will be in the players' hands before the next line of the test runs.

PR #1052, opened 2024-09-21
Feat/resolve seven gamestate

### LINKED ISSUE #957: [Feature]: Implement sevenPoints() in the GameState API
### Feature Summary

We need to implement playing Points via a seven in the GameState API in order to support the full game logic in this API. See the [spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.wpeuh0qc6ps2) for a general overview of the API, its underlying motivations, and the development approach.

### Detailed Description

The endpoint should be:

POST /api/game/:gameId/move/seven/points
```
{ 
  playedCardId: '6H', // playing the Six of Hearts for points from the top of the deck
}; 
```
This should verify that the move is legal (that it's your turn, a seven is resolving, the specified card is in the top 2 of the deck), then play the card for points and clean up, as expected.

In addition to building the endpoint, the frontend consumption of it in the src/stores/game.js file's `requestScuttle()` function should be updated to make a request to this endpoint (instead of throwing an error) when the `process.env.VITE_USE_GAME_STATE_API` var equals `"true"`.

The same should be applied to the `playPointsFromSevenOpponent()` test command in `tests/e2e/support/commands.js` when the `CYPRESS_USE_GAMESTATE_API` env var is set to `"true"`. If so, (instead of erroring), it should send a request to the new api endpoint.

itsalaidbacklife:
Implements the general setup for multi-phase one-offs and the `sevenPoints` move for playing points from the top of the deck.
## Issue number

- Resolves #957 

## Please check the following

- For New Features:


api/helpers/gamestate/publish-game-state.js
```diff
@@ -44,7 +44,14 @@ module.exports = {
...
       //Last Event, and Extra Socket Variables
       const happened = gameState.moveType !== MoveType.FIZZLE;
-      const { playedBy } = gameState;
+      let { playedBy } = gameState;
+
+      // When resolving, the player who resolves is the opponent
+      // of the one who played the one-off (if the one-off doesn't fizzle)
+      if ( gameState.moveType === MoveType.RESOLVE ) {
+        playedBy = (playedBy + 1) % 2;
+      }
```
itsalaidbacklife, line 53: This feels hacky and I'm not sure what's best. Currently the legacy `/resolve` endpoint emits a socket event with a `playedBy` value that's the _opposite_ of whoever called `resolve`. The idea is that if a one-off does resolve without being countered, then the person who calls resolve is the opponent of the person who played the one-off.

This let's the clientside somewhat cleanly interpret who played the oneOff as `playedBy`, but it's somewhat inconsistent with the way playedBy is used in the rest of the GameState API. Alternatively we could try to rework how it's represented clientside in `inGameEvents`. I don't love either option at the moment

api/helpers/gamestate/moves/resolve/execute.js
```diff
@@ -59,6 +59,19 @@ module.exports = {
...
         break;
+      case 3:
+      case 4:
+      case 5:
+      case 7:
+        // These one-offs require an additional user input after resolving
+        // Each one has its own phase designated by the rank
+        result = {
+          ...result,
+          phase: oneOff.rank,
```
itsalaidbacklife, line 70: This is a bit of a funky move because all the other references to a `GameState`'s `phase` are assigned using the `GamePhase` enum. The trick here is that the values for `GamePhase.RESOLVING_THREE`, `GamePhase.RESOLVING_FOUR`, `GamePhase.RESOLVING_FIVE`, and `GamePhase.RESOLVING_SEVEN` are all equal to their respective one-off rank: 3, 4, 5, and 7. I considered at first making a reverse mapping so this would instead be something like:

```
      phase: rankToGamePhaseMap[oneOff.rank],
```

With the idea being that it would convert the ranks to their respective `GamePhaseValues`, but this would just be an identity mapping each rank to itself. So it seems simpler to use the rank _as_ the `GamePhase`. 

Another alternative would be to separate it into different cases for each of the ranks 3-5 like

```
    case 3:
      result = {
        ...result,
        phase: GamePhase.RESOLVING_THREE,
        oneOff,
        turn: result.turn - 1,
      }
```

But then everything except the phase is assigned exactly the same way for all four cases. This is significantly DRYer so it seems like the best I've come up with, even though it is the only place in the app to reference `phase` without using the `GamePhase` enum

api/helpers/gamestate/moves/resolve/execute.js
```diff
@@ -59,6 +59,19 @@ module.exports = {
...
+          oneOff, // oneOff stays on the stack until next move
+          turn: result.turn - 1, // turn doesn't increment until the next move
```
itsalaidbacklife, line 72: It also feels a little weird to increment the turn above and decremenet it here. Again, this comes down to being DRY. I don't see a way to neatly separate turn incrementing in one-offs ace, 2, 6, and 9 resolving, but not 3-5 + 7 (the multi phase one-offs), while also handling each one individually, without checking the rank twice and repeating the common changes e.g. the turn going up for the single-phase one-offs and the turn not going up for the multi-phase one-offs.

So my approach here is to assign the common values for all the single-stage one-offs up above and then in the switch statement call an individual rank-based helper to actually 'do' the one-off part of the single stage one-offs e.g. `ace()` to scrap all points and `six()` to scrap all jacks, while the multi-phase one-offs are all handled in one cascading switch case here since all of them simply keep the turn + oneOff the same and set the phase to their respective rank. Then the multi phase one-offs can all get their own moveType specific `execute()` and `validate()` helpers to handle their second stage e.g. `sevenPoints()`

api/helpers/gamestate/publish-game-state.js
itsalaidbacklife: Notably this is what the client currently expects and overall we are making minimal changes to the client code for the MVP rollout of the GameState API in the name of downscoping. So when we rework the client data structures this will change regardless, but for now it might be easiest to keep this consistent with legacy?

api/helpers/gamestate/moves/resolve/execute.js
Haviles04: I agree that this is DRYer and better

api/helpers/gamestate/moves/resolve/execute.js
Haviles04: Yea, it's deff a little funky to decrement it, we could possible just increment it after the switch

```
if([1,2,6,9].includes(oneOff.rank){
result.turn++
}
```

api/helpers/gamestate/publish-game-state.js
Haviles04: I think this makes sense for now, but maybe it's better to do it inside the `resolve` helper?

Haviles04 (commented):
LGTM

PR #1053, opened 2024-09-21
 Pass in gamestate

### LINKED ISSUE #1043: [Feature]: Implement Passing in the GameState API
### Feature Summary

We need to implement Passing (ending your turn without playing a card) in the GameState API in order to support the full game logic in this API. See the spec for a general overview of the API, its underlying motivations, and the development approach. Passing is only allowed when the deck is exhausted, and it means incrementing the turn and setting the phase to `GamePhase.MAIN`, without moving any cards.

### Detailed Description

```
POST /api/game/:gameId/move
```
Which move gets made is determined by the request body, which should specify a `moveType` field with a value from the `MoveType` enum. Passing will be triggered by sending a request body like: 
```
{
  moveType: MoveType.PASS,
};
```

The endpoint already exists. When a `moveType: MoveType.PASS` is sent in the req body, the endpoint will attempt to utilize the `pass` helpers: `sails.helpers.gamestate.moves.pass.validate()` and `sails.helpers.gamestate.moves.pass.execute()`. Each of these should follow the paradigm outlined in the [docs](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.vgf103gono5r):

1. update the `api/policies/hasValidMoveBody` policy to allow and validate requests with a moveType of `MoveType.PASS`
2. Implement the `api/helpers/gamestate/moves/pass/validate` helper to verify if the move is legal
3. Implement the `api/helpers/gamestate/moves/pass/execute` helper to create a new GameState that passes the turn

We must also update the policy `api/policies/hasValidMoveBody.js` to accommodate the `MoveType.PASS` moveType. This policy is already applied to the endpoint for making a move and it uses a `switch()` statement on `req.body.moveType` to validate that a given req body has all the data necessary for the given `moveType`. We should add a `MoveType.SCUTTLE` case to this policy.

A `moveType`  of `MoveType.PASS` requires no additional data, because no cards are played. We should `return res.next()` in this switch case without needing any additional conditions. The cleanest way to do this will be to add `case MoveType.PASS` after the existing case for `MoveType.DRAW` so they cascade and share the return statement like

```
    case MoveType.DRAW:
    case MoveType.PASS
      // Draw and pass require no extra data
      return next();
```

### Implementing `validate()`
Create a new `api/helpers/gamestate/moves/pass/validate.js` helper file that matches the general setup of the existing `api/helpers/gamestate/moves/points/validate.js` helper. It should throw errors if the requested move is illegal. It should have three inputs: 
`validate()` should throw if:
* It is not the requesting player's turn (`throw new Error("game.snackbar.global.notYourTurn")`)

### Implementing `execute()`
Create a new `api/helpers/gamestate/moves/pass/execute.js` helper file that matches the general approach of the existing execute helper `api/helpers/gamestate/moves/draw/execute.js`. It should create and return a new `GameState` that is the result of making the `pass` move for the requested player i.e. **it should increment the turn and set the phase to `GamePhase.MAIN`.**

### Unskipping Tests

**Essential tests**
* `winConditions.spec.js` - `it('Passes three times for a Stalemate')`
* `winConditions.spec.js` - `it('Registers stalemate when opponent passes first/last')`
* `5_fives.spec.js` - `it('Cannot to play 5 with an empty deck')`

haaarsh619:
## Issue number #1043

- Resolves #1043

## Please check the following

- For New Features:

### DIFF
--- a/api/helpers/gamestate/check-game-state-for-win.js
+++ b/api/helpers/gamestate/check-game-state-for-win.js
@@ -16,9 +16,14 @@ module.exports = {
       description: 'GameState object',
       required: true,
     },
+    countPasses: {
+      type: 'ref',
+      description: 'Passes Count',
+      required: true,
+    },
   },
 
-  fn: async function ({ game, gameState}, exits) {
+  fn: async function ({ game, gameState, countPasses}, exits) {
     
     const checkWin = (pNum) => {
       const player = pNum ? gameState.p1 : gameState.p0;
@@ -49,7 +54,7 @@ module.exports = {
     const p0Wins = checkWin(0);
     const p1Wins = checkWin(1);
 
-    if (p0Wins || p1Wins) {
+    if (p0Wins || p1Wins || countPasses >= 3) {
       res.gameOver = true;
       const gameUpdates = {};
       gameUpdates.status = GameStatus.FINISHED;
--- a/api/helpers/gamestate/check-game-state-for-win.js
+++ b/api/helpers/gamestate/check-game-state-for-win.js
@@ -23,7 +23,7 @@ module.exports = {
     },
   },
 
-  fn: async function ({ game, gameState, countPasses}, exits) {
+  fn: async function ({ game, gameState, countPasses }, exits) {
     
     const checkWin = (pNum) => {
       const player = pNum ? gameState.p1 : gameState.p0;
--- a/api/helpers/gamestate/check-game-state-for-win.js
+++ b/api/helpers/gamestate/check-game-state-for-win.js
@@ -16,14 +16,14 @@ module.exports = {
       description: 'GameState object',
       required: true,
     },
-    countPasses: {
-      type: 'ref',
+    numPasses: {
+      type: 'number',
       description: 'Passes Count',
       required: true,
     },
   },
 
-  fn: async function ({ game, gameState, countPasses }, exits) {
+  fn: async function ({ game, gameState, numPasses }, exits) {
     
     const checkWin = (pNum) => {
       const player = pNum ? gameState.p1 : gameState.p0;
@@ -54,7 +54,7 @@ module.exports = {
     const p0Wins = checkWin(0);
     const p1Wins = checkWin(1);
 
-    if (p0Wins || p1Wins || countPasses >= 3) {
+    if (p0Wins || p1Wins || numPasses >= 3) {
       res.gameOver = true;
       const gameUpdates = {};
       gameUpdates.status = GameStatus.FINISHED;
--- a/api/helpers/gamestate/moves/draw/execute.js
+++ b/api/helpers/gamestate/moves/draw/execute.js
@@ -30,7 +30,7 @@ module.exports = {
 
     const player = playedBy ? result.p1 : result.p0;
 
-    player.hand.push(currentState.deck.shift());
+    player.hand.push(result.deck.shift());
     result.turn++;
 
     result = {
--- a/api/helpers/gamestate/moves/pass/validate.js
+++ b/api/helpers/gamestate/moves/pass/validate.js
@@ -1,0 +1,49 @@
+const GamePhase = require('../../../../../utils/GamePhase.json');
+
+module.exports = {
+  friendlyName: 'Validate request to Pass',
+
+  description: 'Verifies whether a request to pass a move is legal, throwing explanatory error if not.',
+
+  inputs: {
+    currentState: {
+      type: 'ref',
+      descriptions: 'Object containing the current game state',
+      required: true,
+    },
+    /**
+     * @param { MoveType.PASS } requestedMove.moveType - Specifies that this a passing move
+     * @param { Object } requestedMove - Object describing the request to pass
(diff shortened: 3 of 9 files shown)


api/helpers/gamestate/check-game-state-for-win.js
```diff
@@ -16,9 +16,14 @@ module.exports = {
       description: 'GameState object',
       required: true,
     },
+    countPasses: {
```
itsalaidbacklife, line 19: Let's rename it to numPasses, since that's closer to some of our other counting variable names

api/helpers/gamestate/moves/draw/execute.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
 
     const player = playedBy ? result.p1 : result.p0;
 
-    player.hand.push(currentState.deck.shift());
+    player.hand.push(result.deck.shift());
```
itsalaidbacklife, line 33: Whoopsie daisy. Good catch, thank you. This highlights a shortcoming in the tests since the test for drawing a card doesn't check the number of cards in the deck. That might be worth addressing later on. For now, nice fix

api/helpers/gamestate/moves/pass/validate.js
```diff
@@ -0,0 +1,48 @@
...
+
+  inputs: {
+    currentState: {
+      type: 'ref',
+      descriptions: 'Object containing the current game state',
+      required: true,
+    },
+    /**
+     * @param { MoveType.PASS } requestedMove.moveType - Specifies that this a passing move
+     * @param { Object } requestedMove - Object describing the request to pass
```
itsalaidbacklife, line 16: The object param should go above the individual attributes since they describe its shape
```suggestion

     * @param { Object } requestedMove - Object describing the request to pass
     * @param { MoveType.PASS } requestedMove.moveType - Specifies that this a passing move
```

api/helpers/gamestate/moves/pass/validate.js
```diff
@@ -0,0 +1,48 @@
...
+  fn: ({ currentState, playedBy }, exits) => {
+    try {
+      // Must be your turn
+      if (currentState.turn % 2 !== playedBy) {
+        throw new Error('game.snackbar.global.notYourTurn');
+      }
+
+      // Must be MAIN phase of the turn
+      if (currentState.phase !== GamePhase.MAIN) {
+        throw new Error('Can only draw in main phase');
```
itsalaidbacklife, line 40:
```suggestion
        throw new Error('Can only pass in main phase');
```

api/helpers/gamestate/publish-game-state.js
```diff
@@ -40,6 +38,8 @@ module.exports = {
         return numPasses;
       };
 
+      const victory = await sails.helpers.gamestate.checkGameStateForWin(game, gameState, countPasses());
```
Haviles04, line 41: I'm not a huge fan of passing the function so that it will call twice.

I think the move here is to immediatley invoke countPasses, and then pass it to victory, and the socket call below as just `countPasses`

```suggestion
const countPasses = (function () {
  let numPasses = 0;
  for (const gameState of game.gameStates.slice(-3)) {
    if (gameState.moveType !== MoveType.PASS) {
      return numPasses;
    }
    numPasses++;
  }
  return numPasses;
})();
```

api/helpers/gamestate/check-game-state-for-win.js
```diff
@@ -16,9 +16,14 @@ module.exports = {
...
+      type: 'ref',
```
Haviles04, line 20: If we immediately invoke, this could be type of number

Haviles04 (changes requested):
Nice work, this is looking good. I have one idea about using an IIFE on countPasses

itsalaidbacklife (approved):
Well done. Only a few nitpicks

api/helpers/gamestate/publish-game-state.js
itsalaidbacklife: I'm up for either this or to declare the function in one line and then call it and capture the result in a separate variable afterwards before passing that to the checkWinGame function and using it below in the game object. 

Either way I'd like the variable to be named numPasses

itsalaidbacklife: @haaarsh619 I'll leave it up to you whether to immediately invoke vs use two variables, but let's avoid calling the function twice

api/helpers/gamestate/check-game-state-for-win.js
haaarsh619: @itsalaidbacklife  @Haviles04  how about this way.
``` 
      const countPasses = () => {
        let numPasses = 0;
        for (const gameState of game.gameStates.slice(-3)) {
          if (gameState.moveType !== MoveType.PASS) {
            return numPasses;
          }
          numPasses++;
        }
        return numPasses;
      };

      **const numPasses = countPasses();**

      const victory = await sails.helpers.gamestate.checkGameStateForWin(game, gameState, numPasses);
```

PR #1054, opened 2024-09-23
Feat/resolve two gamestate

### LINKED ISSUE #967: [Feature]: Implement `resolveTwo()` helper and call it in general `resolve()` endpoint
### Feature Summary

We need to build the effect of a two resolving as a helper and call it in `resolve()` in the GameState API to complete the revised game logic. See the [spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.wpeuh0qc6ps2) for a general overview of the API, its underlying motivations, and the development approach.

Resolving a two means scraping target faceCard or glasses 8 from the opponents field

### Detailed Description

```
POST /api/game/:gameId/move
```
Which move gets made is determined by the request body, which should specify a `moveType` field with a value from the `MoveType` enum. Resolving a Two  will be triggered by sending a request body like: 
```
{
  moveType: MoveType.RESOLVE
};
```

The endpoint will already exist. When a `moveType: MoveType.RESOLVE` is sent in the req body, the endpoint will switch on the `oneOff` attempt to utilize the `resolveTwo` helpers: `sails.helpers.gamestate.moves.resolveTwo.validate()` and `sails.helpers.gamestate.moves.resolveTwo.execute()`. Each of these should follow the paradigm outlined in the [docs](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.vgf103gono5r):

1. update the `api/policies/hasValidMoveBody` policy to allow and validate requests with a moveType of `MoveType.RESOLVE`
2. Implement the `api/helpers/gamestate/moves/resolveTwo/validate` helper to verify if the move is legal
3. Implement the `api/helpers/gamestate/moves/resolveTwo/execute` helper to create a new GameState that puts all cards in the opponents `faceCards` into the `scrap`

### Implementing `validate()`
Create a new `api/helpers/gamestate/moves/resolveTwo/validate.js` helper file that matches the general setup of the existing `api/helpers/gamestate/moves/points/validate.js` helper. It should throw errors if the requested move is illegal. It should have three inputs: 
`validate()` should throw if:
* The `playedCard` is not in the player's hand (`"game.snackbar.global.playFromHand"`)
* The `playedCard` is frozen (`game.snackbar.global.cardFrozen`)

### Implementing `execute()`
Create a new `api/helpers/gamestate/moves/resolveTwo/execute.js` helper file that matches the general approach of the existing execute helper `api/helpers/gamestate/moves/resolveTwo/execute.js`. It should create and return a new `GameState` that is the result of making the `resolveTwo` move for the requested player i.e. **it should remove `playedCard` from the player's hand, put `playedCard` in the `scrap` and target card in `faceCards` into the `scrap`. It should also increment the turn and set the phase to `GamePhase.MAIN`. The `playedBy`, `playedCard` attributes should be included in the returned `GameState`.

### Unskipping Tests

**Essential tests**
For this PR, at minimum, we should unskip the following tests, which should be able to pass once it is working correctly. All the following tests are in the `tests/e2e/specs/in-game/one-offs` folder:
* `2_twos-destroy-target-royal.spec.js` - `it('Plays Two to Destroy Face Card')`

`npm run e2e:client:gamestate`./

haaarsh619:
## Issue number #967 

- Resolves #967 

## Please check the following

- For New Features:


Haviles04 (commented):
LGTM, nice work

api/helpers/gamestate/moves/resolve/two.js
```diff
@@ -0,0 +1,26 @@
...
+      required: true,
+    },
+  },
+  sync: true, // synchronous helper
+  fn: ({ currentState }, exits) => {
+    let result = _.cloneDeep(currentState);
+
+    // remove target card from opponent hand 
+    const targetPlayedIndex = result.p1.faceCards.findIndex(({ id }) => id === result.targetCard.id);
+    result.p1.faceCards.splice(targetPlayedIndex, 1);
```
itsalaidbacklife, line 19: It is also possible to target a jack with a 2, and that case is a bit more complicated. We should use `res.oneOffTargetType`  (which will either be 'faceCard' or 'jack') to identify whether we are targeting a jack. If it's a faceCard, do exactly what you have here.

If it's a jack, we need to loop through the opponent's points, and for each point card, check its attachments to look for the targetCard. Then we remove the jack from its point card's attachments, put the jack in the scrap, and move the point card from the opponent's points to the player's points

tests/e2e/specs/in-game/one-offs/2_twos-destroy-target-royal.spec.js
itsalaidbacklife, line 1: We should also unskip the other two tests in this file. They will only work once targeting jacks is built

api/helpers/gamestate/moves/resolve/two.js
haaarsh619: in order to test this i will have to wait for this #1045  to mege so that jack can be playeable as of now it gives movetype off jack is invalid.

api/helpers/gamestate/moves/resolve/execute.js
```diff
@@ -75,6 +66,19 @@ module.exports = {
       default:
         return exits.error(new Error(`${oneOff.rank} is not a valid one-off rank`));
     }
+    
+    result = {
```
itsalaidbacklife, line 70: Some of the data that's being set between lines 70-81 is the same as what needs setting  in the coalesced cases for 3, 4, 5, and 7. Let's split up the assignment to result into three chunks. The general idea is that some data needs setting for every one-off, and some data is specific to the one-offs that require extra user input (e.g. picking a card from the scrap with a three), and then some data gets set only for the single-stage one-offs e.g. ace and six which compeltely finish resolving in this helper e.g. incrementing the turn happens for aces/sixes etc but not for threes/fours/fives/sevens because those need to wait for additional user input in a subsequent request. We should split into 3 times we assign to `result`:

1. One up above back where it was that has just the shared data
2. Another in the cases for 3, 4, 5, and 7 that sets their specific data _and then returns `exits.success(result)`_
3. Another block below, where you have it here, which just sets the data that is shared by the one-offs like ace which finish resolving in a single step.

The first section (where we previously had it) should  remove the line above where we scrap the twos `result.scrap.push(result.oneOff);` (that will be moved below to the third section, and then it should set

```
result = {
      ...result,
      ...requestedMove,
      playedBy,
      resolved: oneOff,
      twos: [],
    };
}
```
because this is the data that's common to all one-offs.

The section, in the cases for 3, 4, 5, and 7 should just set `result.phase = oneOff.rank` and then `return exits.success(result)` because it only sets the phase and waits for the next move to scrap the oneOff and actually do its effect.

The third section is just for the single stage one-offs that fully resolve and it should additionally set:
```
result = {
  ...result,
  oneOff: null,
  oneOffTarget: null,
  oneOffTargetType: null,
  turn: result.turn + 1,
};
```

since this is the data for when the turn ends and a one-off finishes resolving

api/helpers/gamestate/moves/resolve/execute.js
```diff
@@ -71,10 +71,21 @@ module.exports = {
           oneOff, // oneOff stays on the stack until next move
           turn: result.turn - 1, // turn doesn't increment until the next move
```
itsalaidbacklife, line 72: This isn't needed now that we removed the moved the code that increments the turn from line 49 to line 87. We should remove this so that the turn stays the same for this block

api/helpers/gamestate/moves/resolve/two.js
```diff
@@ -0,0 +1,54 @@
...
+      
+      if (cardIndex !== -1) {
+        
+        //remove the card from opponents points
+        const [pointCard] = result.p1.points.splice(cardIndex, 1);
+        
+        //removing the jack from point cards attachment
+        pointCard.attachments = pointCard.attachments.filter((attachment) => {
+          return attachment.id !== result.targetCard.id;
+        });
```
itsalaidbacklife, line 37: The removed jack also needs to be added to the scrap. In order to reduce loops, it might make sense to replace the above .findIndex() calls with regular for loops that set variables for both the pointIndex and jackIndex so that down here, if the point card was found, we can use the jackIndex we already calculated to splice the jack from the point's attachments and then we can add the jack to the scrap

itsalaidbacklife (changes requested):
A couple small fixes

api/helpers/gamestate/moves/resolve/two.js
haaarsh619: in two.js what i did was that after condtion finish for faceCard or jack. it will just take result.targetCard in push it in scrap.  
on line 58. ```result.scrap.push(result.targetCard);```

itsalaidbacklife (changes requested):
Last fix: execute needs to read player and opponent dynamically

api/helpers/gamestate/moves/resolve/two.js
```diff
@@ -0,0 +1,75 @@
...
+        // Add the point card (now without the jack) to the player's points (p0)
+        player.points.push(pointCard);
+      }
+    } else {
+      // remove target card from opponent hand 
+      const targetPlayedIndex = opponent.faceCards.findIndex(({ id }) => id === result.targetCard.id);
+      opponent.faceCards.splice(targetPlayedIndex, 1);
+    }
+      // add target card in scrap
+      result.scrap.push(result.targetCard);
```
itsalaidbacklife, line 71: I hadn't actually realized `result.targetCard` was still set here, and while I believe it's correct, I think it indicates that we need to update the other moves to clear out `targetCard` and other unused data explicitly. That can be a separate issue; I'll write it up

itsalaidbacklife (approved):
Nice work. Just a couple comments to update (I'll commit them) and an issue to write to fix the other moves to clear stale data

api/helpers/gamestate/moves/resolve/two.js
```diff
@@ -0,0 +1,75 @@
...
```
itsalaidbacklife, line 71: result.targetCard is the targetCard from the previous move, which is incorrect in the case where the oneOff was countered an even number of times and still resolved. We need to grab the target from result.oneOfftarget which stays the same during countering
```suggestion
          if (card.attachments[j].id === result.oneOffTarget.id) {
            // Jack found, store the indices and the card
            pointIndex = i;
            jackIndex = j;
            pointCard = card;
            break; // Exit inner loop once jack is found
          }
        }
      
        // Exit outer loop once both point and jack are found
        if (pointCard) {
          break;
        }
      }
      
      // If a point card with the jack was found
      if (pointCard) {
        // Remove the point card from the opponent's points (p1)
      opponent.points.splice(pointIndex, 1);
      
        // Remove the jack from the point card's attachments
        const [targetCard] = pointCard.attachments.splice(jackIndex, 1); // Remove the attachment by index
      
        // Add the point card (now without the jack) to the player's points (p0)
        player.points.push(pointCard);
      }
    } else {
      // remove target card from opponent hand 
      const targetPlayedIndex = opponent.faceCards.findIndex(({ id }) => id === result.oneOffTarget.id);
      opponent.faceCards.splice(targetPlayedIndex, 1);
    }
      // add target card in scrap
      result.scrap.push(targetCard);
```

(This isn't quite right because targetCard needs declaring up above in order to be usable in this scope but that's the idea)

api/helpers/gamestate/moves/resolve/two.js
```diff
@@ -0,0 +1,75 @@
...
+    
+    return exits.success(result);
```
itsalaidbacklife, line 73: Then we should also set the new targetCard on result

itsalaidbacklife (commented):
One more fix

api/helpers/gamestate/moves/resolve/two.js
itsalaidbacklife: I've written up #1061 which is fixed by #1062 so that the `targetCard` and `playedCard` correctly get overwritten. Meanwhile this needs to refer to `oneOffTarget` instead of `targetCard`, as shown in the other comments

tests/e2e/specs/in-game/one-offs/2_twos-destroy-target-royal.spec.js
```diff
@@ -41,8 +40,53 @@ describe('Play TWOS', () => {
       });
     });
 
+    it('Resolves a two to destroy a facecard after being double countered', () => {
```
itsalaidbacklife, line 43: @haaarsh619 I added this testcase to demonstrate the failure based on setting targetCard from the result instead of the oneOffTarget.

When players counter back and forth, the `targetCard` is set to the two they are countering, not the original target. But the `oneOffTarget` stays the same throughout the countering process: the original target, in this case the king.

This test was failing because when players counter and counter back, the targetCard was being treated as the previous counter, rather than the king. The fix was just setting `result.targetCard = result.oneOffTarget` in the beginning of `two()`.

It should be correct now; I'm prepping to merge

api/helpers/gamestate/moves/resolve/two.js
```diff
@@ -0,0 +1,76 @@
...
+    playedBy: {
+      type: 'number',
+      description: 'Player number of player requesting move.',
+    }
+  },
+
+  sync: true, // synchronous helper
+  fn: ({ currentState, playedBy }, exits) => {
+    let result = _.cloneDeep(currentState);
+    result.targetCard = result.oneOffTarget;
```
itsalaidbacklife, line 21: I added this to ensure that the original oneOffTarget (e.g. the opponent king) is correctly set as the `targetCard`, as opposed to the target from the previous move, which could have been a counter. This lets the new test I added pass

tests/e2e/specs/in-game/one-offs/2_twos-destroy-target-royal.spec.js
```diff
@@ -101,6 +145,53 @@ describe('Play TWOS', () => {
       // Should no longer see jack of clubs on screen
       cy.get('[data-player-face-card=11-0]').should('not.exist');
     }); // End playing TWO to destroy jack
+
+    it('Opponent Destroy Jacks', () => {
```
itsalaidbacklife, line 149:
```suggestion
    it('Opponent (p1) plays TWO to destroy a jack', () => {
```

PR #1060, opened 2024-09-30
Feature #1022 reveal opponent hand on exhausted deck

### LINKED ISSUE #1022: [Feature]: Reveal the opponent's hand when the deck is exhausted
### Feature Summary

We should reveal the opponent's hand to both players once the deck is exhausted, so they don't need to spend time sifting through the scrap pile. 

### Detailed Description

To do this, we will need to update `src/routes/game/GameView.vue` so that the computed property `showOpponentHand()` is updated to return true if there is no top card of the deck i.e. `return this.gameStore.hasGlassesEight || this.isSpectating || !this.gameStore.topCard;`

From there, we will need to update the tests so that they can correctly determine whether to expect the opponent hand to be revealed. In order to do so, we will update the testing utility `assertGameState()` to accommodate the possibility of the `topCard` part of the fixture being `null`, and if so, interpret that to mean the deck is exhausted and that we should expect the hands to be revealed.

To do this, we should update the test utility `assertDOMMatchesFixture()` in `tests/e2e/support/helpers.js` so that the both conditional checks for whether to expect the opponent hand to be revealed vs hidden (There are two of them in this function):
```
if (playerHasGlasses || spectating) {
```

are updated to check if `fixture.topCard` is `null`:
```
if (playerHasGlasses || spectating || fixture.topCard === null) {
```

### LINKED ISSUE #1057: [Feature]: Translate the AnnouncementDialog
### Feature Summary

We should pull the text displayed in the `AnnouncementDialog` into the translation system using `i18n` so they can be translated to all supported languages.

### Detailed Description

PR #1056 introduced the `AnnouncementDialog`, which appears on the Home page and is used to display announcements for events and changes to the site. In order to allow non-english-speaking users to read and understand these announcements, we should pull them into the translation system using `vue-18n`. Similarly to #640, we should replace the strings of user-facing text out of its current location in `src/routes/home/components/announcementDialog/data/announcementData.js` with names for the strings to be used as keys in the translation files. We should then put the existing text into `src/translations/en.json` and put Spanish and French translations of the text into `es.json` and `fr.json` respectively.

We should put the translated announcment text inside the `"home"` block under a new `"announcements"` block that goes after `"home.snackbar"`. This way, for each language file e.g. `en.json`, the announcment data should be under `"home.announcements"` like so:

```
"home": {
    "snackbar": {
      "cannotJoin": "Cannot join that game because it's already full",
      "spectateTwoPlayers": "You can only spectate an ongoing game with two players",
      "cantFindGame": "Can't find game",
      "invalidLobbyNumber": "Invalid Lobby Number",
      "cannotSpectate": "Cannot spectate game because you are playing"
    },
  // Just an example. Use whatever announcement text is currently hard coded in announcementData.js
  "announcements": {
    "title": "Keeping Recycler Fives",
    "section1Heading": "Recycler Fives are here to stay",
    "section1Paragraph": "Example text for paragraph 1",
  },
  // other home page text
},
```

Haviles04:
## Issue number

- Resolves #1022
- Resolves #1057

## Please check the following

- For New Features:

- Refactored points, faceCard, and Draw inside the game store
- Added check to look for exhausted deck and show custom snackbar
- Added announcement data and translations

image
image


src/routes/game/GameView.vue
```diff
@@ -768,8 +770,14 @@ export default {
...
       this.snackBarMessage = this.t(messageKey);
       this.showSnackbar = true;
+      this.snackBarColor = 'error';
       this.clearSelection();
     },
+    showCustomSnackbarMessage(messageKey) {
+      this.snackBarMessage = this.t(messageKey);
+      this.showSnackbar = true;
+      this.snackBarColor = 'surface-1';
+    },
```
itsalaidbacklife, line 780: Will this trigger if the opponent makes the move? It looks like it's only called by the requests to make moves. I think we need to trigger it from a socket event instead in order to display it for both players as soon as it's resolved. 

Alternatively, we add a watcher on the top card which calls this func if the oldVal was not null but the new value is null. Then we wouldn't need to call it elsewhere

src/routes/game/GameView.vue
```diff
@@ -850,59 +863,60 @@ export default {
...
-          .then(this.clearSelection)
-          .catch(this.handleError);
-      } else {
-        this.gameStore
-          .requestPlayFaceCard(this.selectedCard.id)
-          .then(this.clearSelection)
-          .catch(this.handleError);
+          });
+        }
+      }catch(messageKey){
```
itsalaidbacklife, line 916:
```suggestion
      } catch(messageKey){
```

src/routes/home/components/announcementDialog/data/announcementData.js
```diff
@@ -1,21 +1,15 @@
...
 export const announcementData = {
-  id: 'recyclerFivesPermanent',
-  activatorText: 'Recycler Fives are here to stay!',
-  title: 'Recycler Fives',
+  id: 'opponentHandRevealed',
+  activatorText: 'announcement.activatorText',
+  title: 'announcement.title',
   imgSrc: null,
-  displayCards: [Card.FIVE_OF_CLUBS, Card.FIVE_OF_DIAMONDS, Card.FIVE_OF_HEARTS, Card.FIVE_OF_SPADES],
+  displayCards: [Card.FIVE_OF_CLUBS, Card.ACE_OF_DIAMONDS, Card.KING_OF_HEARTS, Card.EIGHT_OF_SPADES],
```
itsalaidbacklife, line 8: Let's replace the `displayCards` with an imgSrc to showcase the actual game board

tests/e2e/support/helpers.js
```diff
@@ -197,6 +197,16 @@ export function assertSnackbarError(message, snackName = 'game') {
...
 
+export function assertCustomSnackBar(message, snackName = 'game') {
+  cy.get(`[data-cy=${snackName}-snackbar] .v-snackbar__wrapper`)
+    .should('be.visible')
+    .should('have.class', 'bg-surface-1')
+    .find(`.v-snackbar__content`)
+    .should('contain', message)
+    .get('[data-cy=close-snackbar]')
+    .click();
+}
```
itsalaidbacklife, line 208: Instead of defining a new function for this, we should update the existing `assertSnackBarError()` to generalize by making an optional argument for the color and renaming it to simply `assertSnackBar()`

itsalaidbacklife (approved):
Excellent

PR #1069, opened 2024-10-03
resolve six in gamestate api

### LINKED ISSUE #968: [Feature]: Implement resolving a six in the GameState API
### Feature Summary

We need to build the effect of a six resolving in the GameState API. To do so, we should define a `six()` helper and calling it in general `resolve()` helper. See the [spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.wpeuh0qc6ps2) for a general overview of the API, its underlying motivations, and the development approach.

Resolving a six means scrapping all `faceCards` from the opponents field, scrapping all jacks attached to point cards, and for each point card, moving it to the other player's points if the number of jacks on it was odd before they were scrapped. That is to say if a given point card has 1 jack or 3 jacks attached to it, when all the jacks are scrapped, that point card should be removed from whichever player had control of it and put in the `points` of their opponent.

### Detailed Description

```
POST /api/game/:gameId/move
```
Which move gets made is determined by the request body, which should specify a `moveType` field with a value from the `MoveType` enum. Resolving a Six  will be triggered by sending a request body like: 
```
{
  moveType: MoveType.RESOLVE
};
```

The endpoint already exists. When a `moveType: MoveType.RESOLVE` is sent in the req body, the endpoint will call the `sails.helpers.gamestate.moves.resolve.execute()` helper to execute the one-off. This existing helper switches on the `oneOff` rank in order to execute logic for the specific one-off being resolved. We should define a new helper in `api/helpers/gamestate/moves/resolve/six.js` which returns a new `GameState` that is the result of resolving a six.

There are three pieces to implement:
1. Define `api/helpers/gamestate/moves/resolve/six.js` to scrap all faceCards and jacks, switching control of point cards as needed
2. Update the existing `api/helpers/gamestate/moves/resolve/execute.js` to add a switch case for a rank of `6` and calling the above `six()` helper to create the new `GameState`
3. Unskip the relevant tests by removing the `cy.skipOnOnGameStateApi()` commands from tests which can now pass in the GameState API

### Implementing `six()`
Create a new `api/helpers/gamestate/moves/resolve/six.js` helper file that matches the general approach of the existing one-off helper `api/helpers/gamestate/moves/resolve/ace.js`. It should create and return a new `GameState` that is the result of making the `six` move for the requested player i.e. **it should remove all cards from both players `faceCards` arrays, as well as all cards in the `attachments` of all point cards on the field, and for any point card that had an odd number of jacks (1 or 3 jacks), it should remove that point card from the previous player's `points` array and place it in their opponent's `points` array.**

`six()`  should have one input:

### Unskipping Tests

**Essential tests**
For this PR, at minimum, we should unskip the following tests, which should be able to pass once it is working correctly. All the following tests are in the `tests/e2e/specs/in-game/one-offs` folder:
* `6_sixes.spec.js` - `it('Plays a six to destroy all face cards')`

`npm run e2e:client:gamestate`./

haaarsh619:
## Issue number #968 

- Resolves #968 

## Please check the following

- For New Features:


api/helpers/gamestate/moves/resolve/six.js
```diff
@@ -0,0 +1,50 @@
...
+      description: 'The latest GameState before the six scraps all royals and glasses eights',
+      required: true,
+    },
+  },
+  sync: true, // synchronous helper
+  fn: ({ currentState }, exits) => {
+    let result = _.cloneDeep(currentState);
+
+    const player = result.playedBy ? result.p1 : result.p0;
+    const opponent = result.playedBy ? result.p0 : result.p1;
```
itsalaidbacklife, line 18: It doesn't really matter who played the six since the effect is symmetrical. And actually the `playedBy` is going to be the person who resolved, which is the opponent of the person who played the six. I think we could just grab p0 and p1 and use those instead of player and oponent like
```suggestion
   const { p0, p1 } = result;
```

api/helpers/gamestate/moves/resolve/six.js
```diff
@@ -0,0 +1,50 @@
...
+    
+    let cardToScrap = [];
```
itsalaidbacklife, line 20: Thoughts on the use of `cardsToScrap` as opposed to pushing directly to `result.scap`?

haaarsh619: not any reason for using it just show it in old resolve six code so i used that. should i make it result.scrap ?

itsalaidbacklife: Yes I think that's simpler. The main reason it was previously used in the legacy implementation is that each card needed an individual database write and there is a single sails method for doing that in bulk. So we aggregated all of them before writing to the db. 

But in the GameState API we just need to put the right cards into the right lists in memory so there's no need to create the separate array first

api/helpers/game-states/moves/resolve/six.js
```diff
@@ -0,0 +1,44 @@
...
+  inputs: {
+    currentState: {
+      type: 'ref',
+      description: 'The latest GameState before the six scraps all royals and glasses eights',
+      required: true,
+    },
+  },
+  sync: true, // synchronous helper
+  fn: ({ currentState }, exits) => {
+    let result = _.cloneDeep(currentState);
```
itsalaidbacklife, line 15: Since we're not reassigning to result we can use const
```suggestion
    const result = _.cloneDeep(currentState);
```

api/helpers/game-states/moves/resolve/six.js
```diff
@@ -0,0 +1,44 @@
...
+    collectJacksAndFilterPoints(p0, p1);
+    collectJacksAndFilterPoints(p1, p0);
+      
+    // Remove all played face cards and add in scrap
+    [p0, p1].forEach((card) => {
+      result.scrap.push(...card.faceCards);
+      card.faceCards = [];
+    });
+
+    return exits.success(result);
```
itsalaidbacklife, line 42: This is pretty slick. I don't normally think to leverage side effects in array methods like `.filter()` but it looks like an elegant way to simultaneously remove points and scrap their jacks. 

I'm wondering whether it makes sense to do the scrapping of the faceCards in the same function above like
```suggestion
    const scrapRoyalsAndSwapStolenPoints = (player, opponent) => {
      // Scrap the face cards
      result.scrap.push(...player.faceCards);
      player.faceCards = [];

      // Scrap jacks and return stolen point cards to original owner
      player.points = player.points.filter((point) => {
        const jackCount = point.attachments.length;
        result.scrap.push(...point.attachments);  // Scrap all jacks
        point.attachments = []; // removing attachments from all cards

        if (jackCount % 2 === 1) {
          opponent.points.push(point);  // Move point card to opponent if odd number of jacks
          return false;
        }

        return true;
      });
    };

    scrapRoyalsAndSwapStolenPoints(p0, p1);
    scrapRoyalsAndSwapStolenPoints(p1, p0);

    return exits.success(result);
```

api/helpers/game-states/moves/resolve/six.js
```diff
@@ -0,0 +1,42 @@
...
+    const scrapRoyalsAndSwapStolenPoints = (player, opponent) => {
+      // Scrap the face cards
+      result.scrap.push(...player.faceCards);
+      player.faceCards = [];
+
+      // Scrap jacks and return stolen point cards to original owner
+      player.points = player.points.filter((point) => {
+        const jackCount = point.attachments.length;
+        result.scrap.push(...point.attachments);  // Scrap all jacks
+        point.attachments = []; // removing attachments from all cards
```
itsalaidbacklife, line 28:
```suggestion
        point.attachments = []; // Remove all attachments
```

itsalaidbacklife (approved):
Great work

PR #1072, opened 2024-10-03
feat: implements resolve four in GameState API

### LINKED ISSUE #955: [Feature]: Implement Resolving a Four in the GameState API
### Feature Summary

We need to implement resolving a four in the GameState API in order to support the full game logic in this API. See the [spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.wpeuh0qc6ps2) for a general overview of the API, its underlying motivations, and the development approach.

Resolving a four means moving the `cardId1` and `cardId2` from the requesting player's hand to the `scrap`.

### Detailed Description

```
POST /api/game/:gameId/move
```
Which move gets made is determined by the request body, which should specify a `moveType` field with a value from the `MoveType` enum. Resolving a four will be triggered by sending a request body like: 
```
{
  moveType: MoveType.RESOLVE_FOUR,
  cardId1: '3D',
  cardId2: '5C',
};
```

The endpoint already exists. When a `moveType: MoveType.RESOLVE_FOUR` is sent in the req body, the endpoint will attempt to utilize the `resolveFour` helpers: `sails.helpers.gamestate.moves.resolveFour.validate()` and `sails.helpers.gamestate.moves.resolveFour.execute()`. Each of these should follow the paradigm outlined in the [docs](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.vgf103gono5r).

Unlike resolving most one-offs, fours require additional input from the opponent of the player who first played the one-off, because they make your opponent discard two cards of their choice. The sequence of user actions is:
1. Player plays the four as a one-off with `MoveType.ONE_OFF`
2. Opponent resolves the four with `MoveType.RESOLVE` (this sets the phase and scraps the twos)
3. Opponent discards two cards from their hand with the `MoveType.RESOLVE_FOUR` moveType. (this is what is left to build)

1. Update the `api/policies/hasValidMoveBody` policy to allow and validate requests with a moveType of `MoveType.RESOLVE_FOUR`
2. Implement the `api/helpers/gamestate/moves/resolve-four/validate` helper to verify if the move is legal
3. Implement the `api/helpers/gamestate/moves/resolve-four/execute` helper to create a new GameState that picks the selected cards from the opponent's hand

We need to update the policy `api/policies/hasValidMoveBody.js` to accommodate the `MoveType.RESOLVE_FOUR` moveType. This policy is already applied to the endpoint for making a move and it uses a `switch()` statement on `req.body.moveType` to validate that a given req body has all the data necessary for the given `moveType`. We should add a `MoveType.RESOLVE_FOUR` case to this policy.

A `moveType`  of `MoveType.RESOLVE_FOUR` requires a `cardId1` which is a valid card id. If `cardId` present in req.body and can be found in `utils/DeckIds.json`, then we can `return res.next()` in this switch case. Otherwise we should return `res.badRequest()` with an error that explains whether `req.body.cardId1` attribute is missing or invalid.

### Implementing `validate()`
Create a new `api/helpers/gamestate/moves/resolve-four/validate.js` helper file that matches the general setup of the existing `api/helpers/gamestate/moves/points/validate.js` helper. It should throw errors if the requested move is illegal. It should have three inputs: 
2. `requestedMove` - req body describing the move being made.

The general idea is that the requesting player (opponent of the one who played the four) should discard two cards from their hand, if possible, and one card from their hand if they only have one card.

`validate()` should throw if:
* It **IS** the requesting player's turn (`throw new Error("It's not your turn to discard");`) - because you discard during your opponent's turn
* The GameState's `phase !== GamePhase.RESOLVE_FOUR (`throw new Error("Can't discard unless your opponent played a four");`)
* `cardId1` or `cardId2` are present but not found in the requesting player's hand (`throw new Error('You must select two cards to discard')`)
* `cardId2` is missing and the requesting player has >= 2 cards in hand (`throw new Error('You must select two cards to discard')`)

### Implementing `execute()`
Create a new `api/helpers/gamestate/moves/resolve-four/execute.js` helper file that matches the general approach of the existing execute helper `api/helpers/gamestate/moves/points/execute.js`. It should create and return a new `GameState` that is the result of making the `resolveFour` move for the requested player i.e. **it should remove `cardId1` from the requesting player's hand and place it into the `scrap`, doing the same for `cardId2` if it is set. It should also increment the turn and set the phase to `GamePhase.MAIN`. The `playedBy` attribute should be the requesting player's player number (0 for p0 and 1 for p1). Whatever cards are discarded should also be set in the GameState's `discardedCards` array.

2. `requestedMove` - req body describing the move being made

### Unskipping Tests

**Essential tests**
For this PR, at minimum, we should unskip the following tests, which should be able to pass once it is working correctly.
* `tests/e2e/specs/in-game/one-offs/4_fours.spec.js` -  Unskip all tests in this file
* `tests/e2e/specs/in-game/game-over/winConditions.spec.js` - unskip `it('Concede game while resolving a four')`

Plubedam:
## 955

- Resolves #955

## Please check the following

- For New Features:


api/helpers/gamestate/moves/resolve-four/execute.js
```diff
@@ -0,0 +1,60 @@
...
+    const player = playedBy ? result.p1 : result.p0;
+    const discardedCards = [];
+
+    player.hand.forEach((card, index, object) => {
+      if (card.id === cardId1 || card.id === cardId2) {
+        discardedCards.push(card);
+        result.scrap.push(card);
+        object.splice(index, 1);
+      }
+    });
```
Haviles04, line 43: This is a nice function, and I like that it's all in one, however I think it might be better to stick to the same way we did it in the other helpers (checkout points) where we find the index and splice instead of looping through them all.

api/helpers/gamestate/moves/resolve-four/validate.js
```diff
@@ -0,0 +1,64 @@
...
+
+      if (currentState.phase !== GamePhase.RESOLVING_FOUR) {
+        throw new Error('game.snackbar.oneOffs.four.notResolvingFourPhase');
+      }
+
+      const playerKey = playedBy ? 'p1' : 'p0';
+      const { cardId1, cardId2 } = requestedMove;
+      let selectedCard1, selectedCard2;
+
+      currentState[playerKey].hand.forEach(card =>{
```
Haviles04, line 46: We can use .find here instead of . forEach

api/policies/hasValidMoveBody.js
```diff
@@ -33,6 +33,22 @@ module.exports = function (req, res, next) {
         }
       }
       return next();
+    
+    case MoveType.RESOLVE_FOUR:
+      {
+        if (!cardId1) {
+          return res.badRequest({ message: 'Must specify a card' });
```
Haviles04, line 40: Can we translate these, and the following errors?

Haviles04 (dismissed):
Nice work, I left a couple comments about consistency and translations. :D

api/helpers/gamestate/moves/resolve-four/execute.js
```diff
@@ -0,0 +1,60 @@
...
+
+    result.turn++;
+
+    result = {
+      ...result,
+      ...requestedMove,
+      phase: GamePhase.MAIN,
+      playedBy,
+      playedCard: null,
```
itsalaidbacklife, line 52: We should also set `targetCard: null` to be explicit that there's no target.
```suggestion
      playedCard: null,
      targetCard: null,
```

api/helpers/gamestate/moves/resolve-four/validate.js
itsalaidbacklife, line 46: the playerKey variable is actually less elegant than just assigning to player directly imo. That's also more consistent with the other moves. I'd like to update the `points/validate` move to reflect this as well for consistency.

```suggestion
      const player = playedBy ? result.p1 : result.p0;
      const { cardId1, cardId2 } = requestedMove;
      let selectedCard1, selectedCard2;

      player.hand.forEach(card =>{
```

That said as @Haviles04 said it may be sensible to use two `.find()` calls instead of the `forEach` just for clarity. It's an extra loop, but a fairly clear line.

I'm not opposed to the single-loop approach here with the block-soped `selectedCard1` and `selectedCard2` variables, either. If we keep the single-loop like you have here, let's reformat the conditionals so the blocks live on their own lines. Single line conditionals like that are too cluttered imo.

api/helpers/game-states/moves/resolve-four/execute.js
```diff
@@ -0,0 +1,65 @@
...
+    const { cardId1, cardId2 } = requestedMove;
+    let result = _.cloneDeep(currentState);
+
+    const player = playedBy ? result.p1 : result.p0;
+    const discardedCards = [];
+
+    const cardIndex1 = player.hand.findIndex(({ id }) => id === cardId1);
+
+    discardedCards.push(player.hand[cardIndex1]);
+    result.scrap.push(...player.hand.splice(cardIndex1, 1));
```
itsalaidbacklife, line 40: The test failures look to be due to the oneOff (the four itself) not being added to the scrap.
```suggestion
    result.scrap.push(result.oneOff, ...player.hand.splice(cardIndex1, 1));
```

Incidentally, you can see the test runs in cypress cloud at the url linked in the bottom of the github actions output:
image


In this case this test run is https://cloud.cypress.io/projects/i8bxr8/runs/2077

From there, if you click a test case you can then click 'test replay' to see exactly what happened in CI

image


This is super helpful for looking more precisely at why things are failing w/o sifting through terminal logs from the CI output

Plubedam:
After replaying the tests, it seems like they were failing because the 4 card was added to the scrap pile too early. With the recent changes the 4 one-off it's discarded only after the opponent correctly discards and not in failed attempts.

However, now a test fails when in non-GameState API mode, and I'm not sure why only this one is affected.

tests/e2e/specs/in-game/one-offs/4_fours.spec.js
```diff
@@ -266,7 +258,7 @@ describe('FOURS', () => {
         p1Hand: [Card.TEN_OF_HEARTS, Card.SIX_OF_DIAMONDS],
         p1Points: [],
         p1FaceCards: [],
-        scrap: [Card.FOUR_OF_CLUBS, Card.ACE_OF_DIAMONDS, Card.FOUR_OF_SPADES, Card.FOUR_OF_DIAMONDS],
+        scrap: [Card.FOUR_OF_CLUBS, Card.ACE_OF_DIAMONDS, Card.FOUR_OF_SPADES],
```
itsalaidbacklife, line 261: Okay so this is the crux of the issue with the failing test, and as you've said, it's a byproduct of a legitimate difference between the intended behavior of the GameState API vs legacy: specifically whether the 4 being played for a one-off hits the scrap when the targeted player declines to counter, or when the targeted player discards their cards.

The legacy approach put the oneOff in the scrap as soon as the targeted player declined to counter, and this looks to be the only assertion in the test suite that makes this behavior explicit. I think it actually doesn't matter much which way the system works, but it is more consistent that the oneOff go to the scrap when its effect is resolved since this is a requirement for threes (so you can't play a three and pick it back up in one move).

All that said, I think the most reasonable course of action is to simply remove this assertion from the test by taking the `scrap` out of the asserted fixture. Then it's ignored in this step and we can still assert that the 4 hits the scrap in the next assertGameState() call, which we do after the player discards.
```suggestion
```

itsalaidbacklife (approved):

PR #1059, opened 2024-09-30
feat: adds Resolving Three move support in GameState API

Plubedam:
## 954

Relevant [issue number](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
- Resolves #954

## Please check the following

- [ ] Do the tests still pass? (see [Run the Tests](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#run-the-tests))
- [ ] Is the code formatted properly? (see [Linting (Formatting)](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#linting-formatting))
- For New Features:
  - [ ] Have tests been added to cover any new features or fixes?
  - [ ] Has the documentation been updated accordingly?

## Please describe additional details for testing this change
Can't check locally if tests run properly.

PR #1075, opened 2024-10-04
Fix/resolve increment turn on fizzle

itsalaidbacklife:
Fixes bug in GameState API where the turn doesn't increment when the one-off fizzles
Renames and reworks test case `it('can counter a three')` to `it('Increments turn when one-off fizzles')`

These failures were brought to light by #1059 

## Issue number

N/A

## Please check the following

- For New Features:

### DIFF
--- a/api/helpers/game-states/moves/resolve/execute.js
+++ b/api/helpers/game-states/moves/resolve/execute.js
@@ -50,6 +50,7 @@ module.exports = {
     if (fizzles) {
       result.moveType = MoveType.FIZZLE;
       result.scrap.push(result.oneOff);
+      result.turn++;
       return exits.success(result);
     }
 
--- a/tests/e2e/specs/in-game/one-offs/0_countering.spec.js
+++ b/tests/e2e/specs/in-game/one-offs/0_countering.spec.js
@@ -332,22 +332,21 @@ describe('Countering One-Offs P0 Perspective', () => {
     cy.setupGameAsP0();
   });
 
-  it('Can counter a three', () => {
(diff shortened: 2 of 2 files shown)


api/helpers/game-states/moves/resolve/execute.js
```diff
@@ -50,6 +50,7 @@ module.exports = {
     if (fizzles) {
       result.moveType = MoveType.FIZZLE;
       result.scrap.push(result.oneOff);
+      result.turn++;
```
itsalaidbacklife, line 53: This was a critical bug and I'm surprised to just be catching it now. The turn wasn't incremented when the one-off fizzled.

tests/e2e/specs/in-game/one-offs/0_countering.spec.js
```diff
@@ -332,53 +332,44 @@ describe('Countering One-Offs P0 Perspective', () => {
     cy.setupGameAsP0();
   });
 
-  it('Can counter a three', () => {
-    cy.skipOnGameStateApi();
+  it('Increments turn when one-off fizzles', () => {
```
itsalaidbacklife, line 335: This test case was named totally wrong compared to what it was actually testing since it wasn't actually playing a three that got countered; it was a five. I don't actually know what the heck happened there, but it seems to me that the rank itself won't matter much and that the key thing this test caught when we added it to the gamestate test scope was that the turn needs to increment when a one-off is countered. 

I've renamed and reworked the test case accordingly

PR #1077, opened 2024-10-04
Update (almost) all of the dependencies

seriouslysean:
## Issue number

- Resolves #

## Please check the following

- For New Features:


utils/config-utils.js
```diff
@@ -16,9 +16,12 @@ export async function initCuttleGlobals(app) {
...
     try {
-      const devtools = await import('@vue/devtools');
-      devtools.connect(null, 8098);
-      console.log('Vue devtools connected');
+      document.addEventListener('DOMContentLoaded', () => {
+        const script = document.createElement('script');
+        script.src = 'http://localhost:8098';
+        script.async = true;
+        document.head.appendChild(script);
+      });
```
seriouslysean, line 24: This works with the new version of devtools better than the previous method.

package.json
```diff
@@ -3,63 +3,62 @@
...
+    "rimraf": "6.0.1",
+    "sass": "1.79.4",
+    "storybook": "7.6.0",
+    "vite": "5.4.8",
+    "vite-plugin-vuetify": "^2.0.4",
+    "vitest": "2.1.2"
   },
   "engines": {
-    "node": "18.18"
+    "node": "20.18"
```
seriouslysean, line 61: Bumped node to coincide with one of the latest LTS releases.

seriouslysean (commented):
I updated so, so many things. I think a general regression and test suite pass will be more than enough, but would also appreciate your manual testing, @itsalaidbacklife.

seriouslysean:
@itsalaidbacklife I can't get the sails tests to pass and could use some help figuring them out. Everything else is working though. 🎉

seriouslysean:
@itsalaidbacklife I'm going to take a separate look at the Vitest issue, but we need to be able to upgrade it. As it stands, the paralellization is wreaking havoc on it.

tools/msg-bump-version.js
itsalaidbacklife, line 1: RIP lil guy

utils/config-utils.js
itsalaidbacklife: Is it catchable? Historically we haven't been actually able to hit that `console.warn()` when the devtools fail to connect. It would be nice if we could swallow whatever vue would output and just send this warning ourselves. Either way it wouldn't be a regression b/c the catch never worked

api/controllers/HealthController.js
```diff
@@ -15,4 +15,16 @@ module.exports = {
       version: packageVersion,
     });
   },
+  getDevtoolsHealth: async function (req, res) {
```
seriouslysean, line 18: @itsalaidbacklife new devtools health endpoint, this gives the frontend awareness and lets us properly log the status. 😎

utils/config-utils.js
```diff
@@ -15,12 +36,6 @@ export async function initCuttleGlobals(app) {
...
   // Connect the devtools -- non-prod only
   if (import.meta.env.DEV) {
-    try {
-      const devtools = await import('@vue/devtools');
-      devtools.connect(null, 8098);
-      console.log('Vue devtools connected');
-    } catch (err) {
-      console.warn('Failed to connect vue devtools - try running npm run start:devtools');
-    }
+    document.addEventListener('DOMContentLoaded', initDevtools);
```
seriouslysean, line 39: Hoisted the logic up and added that sails controller check.

vite.config.mjs
```diff
@@ -2,7 +2,7 @@ import os from 'os';
 import { dirname, resolve } from 'path';
 import { fileURLToPath } from 'url';
 
-// vite.config.js
+// vite.config.mjs
```
seriouslysean, line 5: Renamed to `.mjs` since we're using this file as a valid module anyway to get rid of the CJS warning from the console.

utils/config-utils.js
```diff
@@ -1,5 +1,26 @@
 import { version } from '_/package.json';
 
+async function initDevtools() {
+  try {
+    // Check devtools availability
+    const response = await fetch('/api/devtools-health');
```
itsalaidbacklife, line 6: Why go to the backend here instead of the devtools directly? Like why not just

```
const response = await fetch('localhost:8098');
if (response.status !== 200) {
  throw new Error('Devtools not available');
}
```

seriouslysean: That doesn't work due to a CORS issue, so we need to proxy it. I also don't want the full script, only a true/false boolean as a health directive.

utils/config-utils.js
```diff
@@ -1,5 +1,26 @@
...
+    // Add devtools script to the DOM
+    const script = document.createElement('script');
+    script.src = 'http://localhost:8098';
+    script.async = true;
+    script.onerror = () => {
+      console.error('Failed to load the devtools script');
+    };
+    document.head.appendChild(script);
+  } catch (err) {
+    console.error('Error checking devtools availability:', err.message);
```
itsalaidbacklife, line 20: I think this should probably be a warning instead of an error
```suggestion
    console.warn('Error checking devtools availability:', err.message);
```

itsalaidbacklife (approved):

PR #1081, opened 2024-10-06
Fix/empty deck snackbar conditional

itsalaidbacklife:
Fixes bug where snackbar to alert players of an empty deck would appear at the beginning of a game during a rematch:
image

## Issue number

N/A

## Please check the following

- For New Features:

### DIFF
--- a/src/routes/game/GameView.vue
+++ b/src/routes/game/GameView.vue
@@ -750,7 +750,7 @@ export default {
       });
     },
     topCard(newTopCard, oldTopCard) {
-      if (oldTopCard && !newTopCard) {
+      if (this.gameStore.id && oldTopCard && !newTopCard) {
         this.showCustomSnackbarMessage('game.snackbar.draw.exhaustedDeck');
       }
     }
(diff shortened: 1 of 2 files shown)


src/routes/game/GameView.vue
```diff
@@ -750,7 +750,7 @@ export default {
       });
     },
     topCard(newTopCard, oldTopCard) {
-      if (oldTopCard && !newTopCard) {
+      if (this.gameStore.id && oldTopCard && !newTopCard) {
```
itsalaidbacklife, line 753: This is the fix. The snackbar would trigger during the rematch flow because there was a moment where the gameStore would be reset so the topCard was null and it was previously not null (during the last game), but at that moement the game's id is also set to null. So this check ensures that we only show the snackbar duing an actual game

tests/e2e/specs/in-game/game-over/rematch.spec.js
```diff
@@ -269,6 +269,10 @@ describe('Creating And Updating Casual Games With Rematch', () => {
     // Game 2: Player concedes
     startRematchPlayerFirst();
     cy.get('[data-player-hand-card]').should('have.length', 6);
+
+    // Snackbar about empty deck should not appear
+    cy.get('[data-cy=game-snackbar] .v-snackbar__wrapper').should('not.exist');
```
itsalaidbacklife, line 274: Added an assertion which fails without this fix.

Haviles04 (approved):
Nice fix!

PR #1083, opened 2024-10-07
Enforced newline-per-chained-call

### LINKED ISSUE #1019: [DevEx]: Configure eslint rule: newline-per-chained-call
### Improvement Summary

In order to reduce stylistic discrepancies between contributors regarding line length, we should apply the newline-per-chained-call eslint rule

### Detailed Description

Here is the rule definition: https://eslint.style/rules/js/newline-per-chained-call. We should use the default value of 2 for `ignoreChainWithDepth` to allow short chains of 2 commands to use a single line. We should run `npm run lint:fix` and commit the changes after configuring the rule to normali3 line length across the project.

This may require installing an additional dev dependency for style rules.

Mang0codes:
## Issue number

- Resolves #1019

## Please check the following

- For New Features:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cuttle",
-  "version": "10.28.0",
+  "version": "10.28.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "cuttle",
-      "version": "10.28.0",
+      "version": "10.28.1",
       "dependencies": {
         "chart.js": "4.4.4",
         "connect-redis": "6.1.3",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "10.28.0",
+  "version": "10.28.1",
   "dependencies": {
     "chart.js": "4.4.4",
     "connect-redis": "6.1.3",


Mang0codes:
@itsalaidbacklife I closed that pull request and also created a PR for the LF issue you raised. Is this okay now?

api/helpers/ find-spectatable-games.js
```diff
@@ -8,7 +8,8 @@ module.exports = {
   description: 'Finds all the games that are available to spectate',
 
   fn: async (_, exits) => {
-    const recentUpdateThreshhold = dayjs.utc().subtract(5, 'minute').toDate();
+    const recentUpdateThreshhold = dayjs.utc().subtract(5, 'minute')
+.toDate();
```
seriouslysean, line 12: I would expect these to indent properly, and not just be flush far left
```
    const recentUpdateThreshhold = dayjs.utc().subtract(5, 'minute')
        .toDate();
```

.eslintrc.js
```diff
@@ -55,6 +55,7 @@ module.exports = {
         }
       }
     ],
+    'newline-per-chained-call': ['error', { ignoreChainWithDepth: 2 }],
```
seriouslysean, line 58: It seems like you may also need something like this to enforce proper indenting, but you'd need to test it.
```
'indent': ['error', 2],
```

Mang0codes: Yes, the indentation seemed odd to me but adding this `'indent': ['error', 2],` will affect the entire codebase which means some of the existing codes will be changed too. If you're okay with this I will commit the changes.
Screenshot 2024-10-09 192115

seriouslysean: That feels ok with me, but test it and we’ll let @itsalaidbacklife have final say.

Mang0codes: > That feels ok with me, but test it and we’ll let @itsalaidbacklife have final say. Seems like a bit of research is needed first.

Is this okay? @seriouslysean @itsalaidbacklife

itsalaidbacklife: Yes I think that's what we want. We'll have to see how it goes and it's plausible that we may need to tweak or backpedal at that point but let's try it and see

Mang0codes: I've already committed the changes... You may review them

api/controllers/game/move.js
```diff
@@ -13,14 +13,14 @@ module.exports = async function (req, res) {
...
-      default:
-        throw new Error('You are not a player in this game!');
+    case game.p0.id:
+      playedBy = 0;
+      break;
+    case game.p1.id:
+      playedBy = 1;
+      break;
+    default:
+      throw new Error('You are not a player in this game!');
```
itsalaidbacklife, line 23: I'd like to preserve indenting one level for switch cases, so I believe we need to set `SwitchCase: 1` on the indent rule in .eslintrc

.eslintrc.js
itsalaidbacklife: Let's additionally configure the SwitchCase option to preserve the single level of indentation applied to switch cases. It'll mean using the object definition of the rule and setting 

`SwitchCase: 1`

itsalaidbacklife (commented):
This looks good, but let's preserve the existing switch case indentation using the SwitchCase option of the indent rule

itsalaidbacklife:
@Mang0codes did you mean to close this? I see that there was another force push and the PR closed

Mang0codes:
> @Mang0codes did you mean to close this? I see that there was another force push and the PR closed

I didn't mean to, I guess this occured due to the git reset command. I'll go ahead and create another pull request including the updates you mentioned.

PR #1086, opened 2024-10-08
Revert "Added .gitattributes to enforce LF line endings"

itsalaidbacklife:
Reverts cuttle-cards/cuttle#1084

Reverting because this somehow seems to have corrupted all png's and caused other git problems. This needs further investigation

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,0 @@
-* text eol=lf


itsalaidbacklife:
Replacing with #1087

PR #1090, opened 2024-10-08
Upgrade vite, isolate all sails tests to remove parallelization

seriouslysean:
## Issue number

- Resolves #

## Please check the following

- For New Features:

1. Run `npm run test` locally and ensure all tests pass


.github/workflows/pr-actions.yml
```diff
@@ -46,12 +46,12 @@ jobs:
       script: 'npm run lint'
 
   # Run Jest Unit Tests
-  test-unit:
+  test:
```
seriouslysean, line 49: Renamed this to test since it comprises multiple types now.

package.json
```diff
@@ -97,6 +97,8 @@
     "storybook": "storybook dev -p 6006",
     "storybook:build": "storybook build",
     "test:unit": "NODE_ENV=development vitest run",
+    "test:sails": "NODE_ENV=development vitest run --config tests/unit/vitest-sails.config.mjs",
+    "test": "npm run test:unit && npm run test:sails",
```
seriouslysean, line 101: `test:unit` is only for pure unit tests now, `test:sails` is for the sails specific tests and `test` will now run both.

tests/unit/specs/client-config.spec.js
```diff
@@ -31,11 +31,6 @@ describe('initCuttleGlobals', () => {
     vi.restoreAllMocks();
   });
 
-  it('should execute sails helpers', async () => {
-    const serverIsUp = await sails.helpers.getApiHealth();
-    expect(serverIsUp).toBe(true);
-  });
```
seriouslysean, line 37: Duplicate test doesn't belong here.

tests/unit/specs/sails/game-state.spec.js
```diff
@@ -1,7 +1,7 @@
 import { describe, it, expect, beforeEach } from 'vitest';
-import { points } from '../fixtures/gameStates/points';
-import { resolveThree } from '../fixtures/gameStates/resolveThree';
-import { resolveNine } from '../fixtures/gameStates/resolveNine';
+import { points } from '../../fixtures/gameStates/points';
+import { resolveThree } from '../../fixtures/gameStates/resolveThree';
+import { resolveNine } from '../../fixtures/gameStates/resolveNine';
```
seriouslysean, line 4: Moved the sails test in to a sails directory in the unit spec folder to better isolate and match them.

tests/unit/vitest-sails.config.mjs
```diff
@@ -0,0 +1,28 @@
...
+        '@': resolve(__dirname, './src'),
+      },
+    },
+  },
+  test: {
+    // vitest includes tests from the project root
+    include: ['tests/unit/specs/sails/**/*'],
+    globals: true,
+    environment: 'node',
+    setupFiles: ['tests/unit/setup-sails.vitest.js'],
```
seriouslysean, line 22: New sails specific file, that ONLY handles sails setup and runs the sails specific files. I think there's a way to parallelize these properly, but the point of this PR is to upgrade Vitest. I think improving this should be a follow up.

vite.config.mjs
```diff
@@ -57,13 +57,6 @@ export default defineConfig(({ mode }) => {
         usePolling: os.platform() === 'win32',
       },
     },
-    test: {
-      include: ['**/tests/unit/specs/**/*.spec.{j,t}s?(x)'],
-      threads: false,
-      globals: true,
-      environment: 'node',
-      setupFiles: ['tests/unit/setup.vitest.js'],
-    },
```
seriouslysean, line 66: Moved to discrete Vitest configs.

vitest.config.mjs
```diff
@@ -0,0 +1,23 @@
...
+
+// https://vitest.dev/config/
+export default defineConfig({
+  resolve: {
+    alias: {
+      _: resolve(__dirname),
+      '@': resolve(__dirname, './src'),
+    },
+  },
+  test: {
```
seriouslysean, line 16: Standard, pure unit config.

vitest.config.mjs
```diff
@@ -0,0 +1,23 @@
...
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+// https://vitest.dev/config/
+export default defineConfig({
+  resolve: {
+    alias: {
+      _: resolve(__dirname),
+      '@': resolve(__dirname, './src'),
+    },
```
seriouslysean, line 14: There's probably an opportunity to DRY this up, but it's fine to duplicate aliases for now.

itsalaidbacklife: Agreed; fine for now

.github/workflows/pr-actions.yml
itsalaidbacklife: My only concern is clarity about whether this runs the e2e tests. Not a hard sticking point, but I wonder what's an appropriate terminology. At work we use 'unit test' to describe tests of individual functions which require the backend to be booted, so calling the sails tests unit tests feels fairly natural to me. Would you say they're more backend integration tests?

seriouslysean: I was wondering that, too. I think it’s fair to say they’re both unit tests, we could just change the sails name to be something like test:unit-sails

itsalaidbacklife: Per our discord convo, we'll go with test:unit for testing both and test:unit:server and test:unit:client for the separate categories of unit tests

itsalaidbacklife (approved):

PR #1098, opened 2024-10-11
feature_seven_scuttle

danielmantic:
## Issue number

- Resolves #

## Please check the following

- For New Features:


api/helpers/game-states/moves/seven-scuttle/validate.js
```diff
@@ -0,0 +1,71 @@
...
+  description: 'Verifies whether a request to scuttle with a seven is legal, throwing an explanatory error if not.',
+
+  inputs: {
+    currentState: {
+      type: 'ref',
+      description: 'Object containing the current game state',
+      required: true,
+    },
+    /**
+     * @param { Object } requestedMove - Object describing the request to scuttle
```
itsalaidbacklife, line 15:
```suggestion
     * @param { Object } requestedMove - Object describing the request to scuttle from the top of the deck
```

api/helpers/game-states/moves/seven-scuttle/validate.js
```diff
@@ -0,0 +1,71 @@
...
+     * @param { String } requestedMove.cardId - Card Played for scuttle
+     * @param { String } requestedMove.targetId - Card Targeted for scuttle
+     * @param { MoveType.SEVEN_SCUTTLE } requestedMove.moveType - Specifies that this a scuttle
```
itsalaidbacklife, line 18:
```suggestion
     * @param { MoveType.SEVEN_SCUTTLE } requestedMove.moveType - Specifies that this a sevenScuttle
```

api/helpers/game-states/moves/seven-scuttle/validate.js
```diff
@@ -0,0 +1,71 @@
...
+        throw new Error('game.snackbar.global.notInMainPhase');
+      }
+
+      if (!playedCard) {
+        throw new Error('game.snackbar.global.playFromHand');
+      }
+
+      if (playedCard.isFrozen) {
+        throw new Error('game.snackbar.global.cardFrozen');
+      }
```
itsalaidbacklife, line 56: Only cards in a player's hand can be frozen so we don't need this check.

What it means to be frozen is that  your opponent just played a nine to return a card to your hand and now you can't play it next turn. That's why it's impossible for either of the top cards in the deck to be frozen
```suggestion
```

api/helpers/game-states/moves/seven-scuttle/validate.js
```diff
@@ -0,0 +1,71 @@
...
+  },
+
+  sync: true,
+
+  fn: ({ requestedMove, currentState, playedBy }, exits) => {
+    try {
+      const player = playedBy ? currentState.p1 : currentState.p0;
+      const opponent = playedBy ? currentState.p0 : currentState.p1;
+
+      const playedCard = player.hand.find(({ id }) => id === requestedMove.cardId);
```
itsalaidbacklife, line 39: Because we are resolving a seven, the played card should be taken from the top two cards of the deck, not the player's hand. Playing a seven lets you pick one of the top two cards from the deck to play any way you choose. So all the different sevenSomething move typed e.g. sevenScuttle are finishing your seven effect by playing a card from the top of the deck.

That means the card used to scuttle (the playedCard) must come from the top of the deck rather than the player's hand. It also means that this effect is only allowed in the situation where we are in the middle of resolving a seven (meaning the player just played a seven that their opponent declined to counter and the player hasn't yet chosen which way they will play a card from the top of the deck).

All this means we need an additional check to confirm that the gamephase is `GamePhase.RESOLVING_SEVEN` and that when we capture the playedCard that we should do so from the top two cards of the deck instead of the player hand. Check out the existing `api/helpers/gameStates/moves/seven-points/validate.js` helper for examples of both of these

api/helpers/game-states/moves/seven-scuttle/validate.js
```diff
@@ -0,0 +1,71 @@
...
+      const targetCard = opponent.points.find(({ id }) => id === requestedMove.targetId);
+
+      if (currentState.turn % 2 !== playedBy) {
+        throw new Error('game.snackbar.global.notYourTurn');
+      }
+
+      if (currentState.phase !== GamePhase.MAIN) {
+        throw new Error('game.snackbar.global.notInMainPhase');
+      }
```
itsalaidbacklife, line 48: The phase needs to be GamePhase.RESOLVING_SEVEN instead of main to make this move

api/helpers/game-states/moves/seven-scuttle/execute.js
```diff
@@ -0,0 +1,58 @@
...
+      description: 'Player number of player requesting move.',
+    },
+  },
+  sync: true,
+  fn: ({ currentState, requestedMove, playedBy }, exits) => {
+    let result = _.cloneDeep(currentState);
+
+    const { cardId, targetId } = requestedMove;
+
+    const player = playedBy ? result.p1 : result.p0;
```
itsalaidbacklife, line 35: We don't actually need to the player as a variable because the playedCard comes from the deck not their hand

api/helpers/game-states/moves/seven-scuttle/execute.js
```diff
@@ -0,0 +1,58 @@
...
+    const opponent = playedBy ? result.p0 : result.p1;
+
+    const cardPlayedIndex = player.hand.findIndex(({ id }) => id === cardId);
+    const [ playedCard  ] = player.hand.splice(cardPlayedIndex, 1);
```
itsalaidbacklife, line 39: The playedCard should come from the deck instead of the player hand

itsalaidbacklife (changes requested):
Good start. Currently this is almost an exact clone of the scuttle move, but it needs to become more of a hybrid of scuttling and general seven handling as seen in the existing sevenPoints move.

The key differences are that the seven moves e.g. sevenPoints and secenScuttle can only be played during the `RESOLVING_SEVEN` game phase and that these moves all play from the top two cards of the deck instead of the player hand

api/helpers/game-states/moves/seven-scuttle/execute.js
```diff
@@ -0,0 +1,57 @@
...
+      ...result,
+      ...requestedMove,
+      phase: GamePhase.MAIN,
+      playedBy,
+      playedCard,
+      targetCard
+    };
+    return exits.success(result);
+  },
+};
```
itsalaidbacklife, line 57:
```suggestion
};

```

api/helpers/game-states/moves/seven-scuttle/validate.js
```diff
@@ -0,0 +1,73 @@
...
+      if (currentState.turn % 2 !== playedBy) {
+        throw new Error('game.snackbar.global.notYourTurn');
+      }
+
+      if (currentState.phase !== GamePhase.RESOLVING_SEVEN) {
+        throw new Error('game.snackbar.global.notInMainPhase');
+      }
+
+      if (!playedCard) {
+        throw new Error('game.snackbar.global.playFromHand');
```
itsalaidbacklife, line 57: This error message should be the one from sevenPoints/validate when the card isn't found, because they should pick from the top of the deck instead of their hand

api/helpers/game-states/moves/seven-scuttle/validate.js
```diff
@@ -0,0 +1,73 @@
...
+      }
+
+      if (playedCard.rank !== 7) {
+        throw new Error('game.snackbar.sevenScuttle.mustPlaySeven');
+      }
```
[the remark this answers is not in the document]
itsalaidbacklife: The error message itself should be the one from scuttle/validate for the same issue

api/policies/hasValidMoveBody.js
```diff
@@ -89,6 +89,20 @@ module.exports = function (req, res, next) {
       }
       return next();
 
+    case MoveType.SEVEN_SCUTTLE:
```
itsalaidbacklife, line 92: Because this is the exact same logic as the above SCUTTLE case, we should combine them so instead of adding logic here we just do 
```
  case MoveType.SCUTTLE:
  case MoveType.SEVEN_SCUTTLE:
```
Before the existing logic.

itsalaidbacklife: While we are at it, let's also consolidate the `MoveType.JACK` case with these as well since they all have the same requirements

api/helpers/game-states/moves/seven-scuttle/execute.js
```diff
@@ -0,0 +1,57 @@
...
+
+    result.turn++;
+
+    result = {
+      ...result,
+      ...requestedMove,
+      phase: GamePhase.MAIN,
+      playedBy,
+      playedCard,
+      targetCard
```
itsalaidbacklife, line 53: We also need to do some cleanup from the fact that this is ending a seven. Specifically, we need to clear out the systems `oneOff`, putting it in the scrap, and we need to set the state's `resolved` attribute to be that same card as well.

This is all logic that's shared by all seven moves so you can see it in sevenPoints/execute.js

itsalaidbacklife (changes requested):
Good progress. Here's another round of comments with fixes

api/policies/hasValidMoveBody.js
```diff
@@ -51,7 +51,9 @@ module.exports = function (req, res, next) {
       }
       return next();
 
-    case MoveType.SCUTTLE:
+      case MoveType.JACK:
+      case MoveType.SCUTTLE:
+      case MoveType.SEVEN_SCUTTLE:
```
itsalaidbacklife, line 56: This is correct, but:
* it looks like the indentation should be decreased one level
* we should delete the below `case MoveType.JACK` from lines 91-108 now that it's being handled here

api/helpers/game-states/moves/seven-scuttle/validate.js
```diff
@@ -0,0 +1,68 @@
...
+  sync: true,
+
+  fn: ({ requestedMove, currentState, playedBy }, exits) => {
+    try {
+      const opponent = playedBy ? currentState.p0 : currentState.p1;
+
+      const topTwoCards = currentState.deck.slice(0, 2);
+      const targetCard = opponent.points.find(({ id }) => id === requestedMove.targetId);
+      const playedCard = topTwoCards.find(({ id }) => id === requestedMove.cardId);
+      const cardPlayedIndex = topTwoCards.findIndex(({ id }) => id === requestedMove.cardId);
```
itsalaidbacklife, line 41: Nitpick: we don't need the index and can just use playedCard, since we aren't doing any array manipulation in `validate()`. The `execute()` helpers use `.findIndex()` so that they can call `.splice()` to actually move cards around.

So keep line 40 as it is and remove 41
```suggestion
      const playedCard = topTwoCards.find(({ id }) => id === requestedMove.cardId);
```

api/helpers/game-states/moves/seven-scuttle/validate.js
```diff
@@ -0,0 +1,68 @@
...
+        throw new Error('game.snackbar.global.notYourTurn');
+      }
+
+      if (currentState.phase !== GamePhase.RESOLVING_SEVEN) {
+        throw new Error('game.snackbar.seven.pickAndPlay');
+      }
+
+      if (cardPlayedIndex === -1) {
+        throw new Error('game.snackbar.seven.pickAndPlay');
+      }
```
itsalaidbacklife, line 53: Then this can use playedCard instead 
```suggestion
      if (!playedCard) {
        throw new Error('game.snackbar.seven.pickAndPlay');
      }
```

api/helpers/game-states/moves/seven-scuttle/execute.js
```diff
@@ -0,0 +1,55 @@
...
+      type: 'number',
+      description: 'Player number of player requesting move.',
+    },
+  },
+  sync: true,
+  fn: ({ currentState, requestedMove, playedBy }, exits) => {
+    const { cardId } = requestedMove;
+    let result = _.cloneDeep(currentState);
+
+    const player = playedBy ? result.p1 : result.p0;
```
itsalaidbacklife, line 34: We don't need player, but we do need opponent. This is because the cards in the player hand and points don't change for this move, but the cards from the opponent do since we move the targetCard from their points to the scrap
```suggestion
    const opponent = playedBy ? result.p0 : result.p1;
```

api/helpers/game-states/moves/seven-scuttle/execute.js
```diff
@@ -0,0 +1,55 @@
...
+    const cardIndex = result.deck.findIndex(({ id }) => id === cardId);
+    const [ playedCard ] = result.deck.splice(cardIndex, 1);
+    const { oneOff } = result;
+    player.points.push(playedCard);
```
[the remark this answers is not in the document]
itsalaidbacklife: Concretely, we need to find the targetCard in the opponent's points, and add it to the scrap, along with the playedCard

api/helpers/game-states/moves/seven-scuttle/execute.js
```diff
@@ -0,0 +1,55 @@
+const GamePhase = require('../../../../../utils/GamePhase.json');
+
+module.exports = {
+  friendlyName: 'Seven_Scuttle',
```
itsalaidbacklife, line 4:
```suggestion
  friendlyName: 'Seven Scuttle',
```

itsalaidbacklife (changes requested):
Getting there. I think there is some confusion about what the actual move is supposed to be doing. 

I've done my best to explain the difference in the comments. Please feel free to ask any questions you might have here in GitHub or on our [discord](https://discord.gg/P8wCwxNddJ).

I encourage you to try running the tests that are being unskipped here locally by booting the application with

```
npm run start:dev:gamestate
```

And then opening cypress with 

```
npm run e2e:gui:gamestate

```

And running one of the test files in this PR to see if the unskipped test passes. This might help you build up your intuition of what this move is trying to do by seeing it visually.

danielmantic:
So after the editing the validate.js and hasValidMoveBody.js is done and the problem is in execute.js?

api/helpers/game-states/moves/seven-points/validate.js
```diff
@@ -45,8 +45,8 @@ module.exports = {
         throw new Error('game.snackbar.seven.pickAndPlay');
       }
 
-      if (playedCard.rank > 10) {
-        throw new Error('game.snackbar.points.numberOnlyForPoints');
+      if (playedCard.rank >= 11) {
+        throw new Error('game.snackbar.sevenScuttle.mustPlaySeven');
```
itsalaidbacklife, line 49: The error message should stay the same as it was. This is a pre-existing error message that's defined in all the translation files eg src/translations/en.json. 


It is an error message that applies to the points move and sevenPoints move. Both of them should use the previous version of this error string

api/helpers/game-states/moves/seven-scuttle/execute.js
```diff
@@ -0,0 +1,58 @@
...
+  sync: true,
+  fn: ({ currentState, requestedMove, playedBy }, exits) => {
+    const { cardId, targetId } = requestedMove;
+    let result = _.cloneDeep(currentState);
+
+    const opponent = playedBy ? result.p0 : result.p1;
+
+    // Remove playedCard from the top of the deck
+    const cardIndex = result.deck.findIndex(({ id }) => id === cardId);
+    const [playedCard] = result.deck.splice(cardIndex, 1);
```
itsalaidbacklife, line 38: Linter requires whitespace around bracket contents
```suggestion
    const [ playedCard ] = result.deck.splice(cardIndex, 1);
```

api/helpers/game-states/moves/seven-scuttle/execute.js
```diff
@@ -0,0 +1,58 @@
...
+
+    // Remove targetCard from opponent's points
+    const targetIndex = opponent.points.findIndex(({ id }) => id === targetId);
+    const [targetCard] = opponent.points.splice(targetIndex, 1);
```
itsalaidbacklife, line 42:
```suggestion
    const [ targetCard ] = opponent.points.splice(targetIndex, 1);
```

itsalaidbacklife (commented):
This is really close. Two linter errors and changing back an error message and then I believe this will be ready

danielmantic:
is there something more to change?

itsalaidbacklife:
@danielmantic there is still the error message to roll back in the sevenPoints/validate.js file

danielmantic:
Have you any idea where could be the problem of the last test. If I run e2e locally 
cy.get('#waiting-for-opponent-play-from-deck-scrim').should('not.exist'); this line of code is error, but I dont know what it is doing and how to fix it. I edit validate.js for the rank and suit control from issiu documentation, but it dont help.

Haviles04:
> Have you any idea where could be the problem of the last test. If I run e2e locally cy.get('#waiting-for-opponent-play-from-deck-scrim').should('not.exist'); this line of code is error, but I dont know what it is doing and how to fix it. I edit validate.js for the rank and suit control from issiu documentation, but it dont help.

Are you getting a certain error message? A screenshot would be helpful if you have one

danielmantic:
Obrázok1
this is screenshot of error message

api/helpers/game-states/moves/seven-scuttle/execute.js
```diff
@@ -0,0 +1,58 @@
...
+
+    // Seven one-off cleanup
+    const { oneOff } = result;
+    result.scrap.push(oneOff);
+    result.oneOff = null;
+    result.resolved = oneOff;
+
+    result.turn++;
+    result.phase = GamePhase.MAIN;
+
```
Haviles04, line 55: lets make the result object like 
```
    result = {
      ...result,
      ...requestedMove,
      phase: GamePhase.MAIN,
      oneOff: null,
      playedBy,
      playedCard,
      targetCard: null,
      resolved: oneOff,
    };
    ```

I think adding the requestedMove might fix your problem

Comment (danielmantic) [reply api/helpers/game-states/moves/seven-scuttle/execute.js:55]: Thank you, locally all test succeeded

File: api/helpers/game-states/moves/seven-scuttle/execute.js [hunk 23]
```diff
@@ -0,0 +1,69 @@
...
+    result.scrap.push(playedCard, targetCard);
+
+    // Seven one-off cleanup
+    const { oneOff } = result;
+    result.scrap.push(oneOff);
+    result.oneOff = null;
+    result.resolved = oneOff;
+
+    result.turn++;
+    result.phase = GamePhase.MAIN;
```
Haviles04, line 54: Let's remove the ones that we have in the result object below. No reason to set them twice.

We can either keep the turn increment here or we can put it in the new object

api/helpers/game-states/moves/seven-scuttle/validate.js
```diff
@@ -0,0 +1,82 @@
...
+      if (!targetCard) {
+        throw new Error('game.snackbar.scuttle.mustTargetPointCard');
+      }
+
+      if (playedCard.rank < targetCard.rank) {
+        throw new Error('game.snackbar.scuttle.rankTooLow');
+      }
+
+      if(playedCard.rank == targetCard.rank && playedCard.suit < targetCard.suit){
+        throw new Error('game.snackbar.scuttle.rankTooLow');
```
Haviles04, line 74: I believe this is the wrong error, it should be suit

api/helpers/game-states/moves/seven-scuttle/validate.js
```diff
@@ -0,0 +1,82 @@
...
+      const playedCard = topTwoCards.find(({ id }) => id === requestedMove.cardId);
+
+      // Check if it's the player's turn
+      if (currentState.turn % 2 !== playedBy) {
+        throw new Error('game.snackbar.global.notYourTurn');
+      }
+
+      // Check if the game phase is RESOLVING_SEVEN
+      if (currentState.phase !== GamePhase.RESOLVING_SEVEN) {
+        throw new Error('game.snackbar.seven.pickAndPlay');
```
Haviles04, line 51: I think this is likely the wrong error

api/helpers/game-states/moves/seven-scuttle/execute.js
```diff
@@ -0,0 +1,69 @@
...
+    // Remove targetCard from opponent's points
+    const targetIndex = opponent.points.findIndex(({ id }) => id === targetId);
+    const [ targetCard ] = opponent.points.splice(targetIndex, 1);
+
+    // Move both cards to scrap
+    result.scrap.push(playedCard, targetCard);
+
+    // Seven one-off cleanup
+    const { oneOff } = result;
+    result.scrap.push(oneOff);
```
Haviles04, line 49: Let's combine these two pushes into one line 45,49

Haviles04 (changes requested):
This is getting close, I left a few comments about error messages, and setting values twice

api/helpers/game-states/moves/seven-scuttle/validate.js
itsalaidbacklife: @Haviles04 it's a little silly but atm it's combined into a single error which handles the wrong phase in both directions:
```"pickAndPlay": "You can only play cards from the top of the deck while resolving a seven."```

idea being that it's an if and only if situation. It might make sense to pull out a separate wrongPhase move here but I don't feel strongly about it. Thoughts?

File: api/helpers/game-states/moves/seven-scuttle/validate.js (hunk 24 above)
Comment (itsalaidbacklife) [reply api/helpers/game-states/moves/seven-scuttle/validate.js:74]: The name is correct but the logic isn't. This error actually accounts for any scuttle where the target is too big. It should be the same as the logic for the `api/helpers/game-states/moves/scuttle/validate.js` helper. The requirement is actually that the scuttle is illegal if:

1. `targetCard.rank > playedCard.rank`  OR
2. `targetCard.rank === playedCard.rank && targetCard.suit > playedCard.rank`

This actually indicates a gap in our testing because we're not currently validating that seven scuttles are illegal under the right circumstances. I've created #1106 to add this test case.

File: api/helpers/game-states/moves/seven-scuttle/validate.js [hunk 27]
```diff
@@ -0,0 +1,82 @@
...
+      // Check if the targetCard is in the opponent's points
+      if (!targetCard) {
+        throw new Error('game.snackbar.scuttle.mustTargetPointCard');
+      }
+
+      if (playedCard.rank < targetCard.rank) {
+        throw new Error('game.snackbar.scuttle.rankTooLow');
+      }
+
+      if(playedCard.rank == targetCard.rank && playedCard.suit < targetCard.suit){
```
itsalaidbacklife, line 73: This is close but only handles one of the two cases. I think it reads most clearly like we have it in the `/scuttle/validate` helper:
```suggestion
      const lowerRank = playedCard.rank < targetCard.rank;
      const sameRankLowerSuit = playedCard.rank === targetCard.rank && playedCard.suit < targetCard.suit;

      if (lowerRank || sameRankLowerSuit) {

```

itsalaidbacklife (commented):
Good work. Last couple tweaks and this will be ready

api/helpers/game-states/moves/seven-scuttle/validate.js
itsalaidbacklife: Switched to `.wrongPhase`

api/helpers/game-states/moves/seven-scuttle/validate.js
itsalaidbacklife: This is now correct

itsalaidbacklife (approved):
Great work. I know this took some iterating. Thanks for sticking with it, and thanks for your contribution!

PR #1104, opened 2024-10-14
Implement resolve-five functionality

### LINKED ISSUE #956: [Feature]: Implement resolving a Five in GameState API
### Feature Summary

We need to implement resolving a five in the GameState API in order to support the full game logic in this API. See the [spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.wpeuh0qc6ps2) for a general overview of the API, its underlying motivations, and the development approach.

Resolving a five means moving the `cardId` to the scrap (if present) from the requesting player's hand, and drawing three cards from the deck to the requesting player's hand, stopping at the hand limit of 8. If the requesting player has 0 cards in hand when the five resolves, they draw three cards without discarding.

### Detailed Description

```
POST /api/game/:gameId/move
```
Which move gets made is determined by the request body, which should specify a `moveType` field with a value from the `MoveType` enum. Resolving a four will be triggered by sending a request body like: 
```
{
  moveType: MoveType.RESOLVE_FIVE,
  cardId: '3D',
};
```

The endpoint already exists. When a `moveType: MoveType.RESOLVE_FIVE` is sent in the req body, the endpoint will attempt to utilize the `resolveFive` helpers: `sails.helpers.gamestate.moves.resolveFive.validate()` and `sails.helpers.gamestate.moves.resolveFive.execute()`. Each of these should follow the paradigm outlined in the [docs](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.vgf103gono5r).

Unlike resolving most one-offs, fives require additional input from the player who first played the one-off, because they choose a card to discard before drawing 3 cards. The sequence of user actions is:
1. Player plays the five as a one-off with `MoveType.ONE_OFF`
2. Opponent resolves the five with `MoveType.RESOLVE` (this sets the phase and scraps the twos)
3. Original player discards a card from their hand and then draws 3 cards via the `MoveType.RESOLVE_FIVE` moveType. (this is what is left to build)

1. Update the `api/policies/hasValidMoveBody` policy to allow and validate requests with a moveType of `MoveType.RESOLVE_FIVE`
2. Implement the `api/helpers/gamestate/moves/resolve-five/validate` helper to verify if the move is legal
3. Implement the `api/helpers/gamestate/moves/resolve-five/execute` helper to create a new GameState that discards one card and draws 3 cards.

We need to update the policy `api/policies/hasValidMoveBody.js` to accommodate the `MoveType.RESOLVE_FIVE` moveType. This policy is already applied to the endpoint for making a move and it uses a `switch()` statement on `req.body.moveType` to validate that a given req body has all the data necessary for the given `moveType`. We should add a `MoveType.RESOLVE_FIVE` case to this policy.

A `moveType`  of `MoveType.RESOLVE_FIVE` does not require any additional input (because cardId is optional based on the number of cards in the player's hand). We should add a case for `MoveType.RESOLVE_FIVE` to the cascading cases along with POINTS and PASS for moves that do not require additional data.

### Implementing `validate()`
Create a new `api/helpers/gamestate/moves/resolve-five/validate.js` helper file that matches the general setup of the existing `api/helpers/gamestate/moves/points/validate.js` helper. It should throw errors if the requested move is illegal. It should have three inputs: 
2. `requestedMove` - req body describing the move being made.

The general idea is that the requesting player (the one who originally played the five) should discard 1 card from their hand if they have at least 1 card in hand, so we should ensure that it is their turn and that they are discarding in necessary.

`validate()` should throw if:
* It is not the requesting player's turn (`throw new Error('game.snackbar.global.notYourTurn');`)
* The GameState's `phase !== GamePhase.RESOLVE_FIVE (`throw new Error('game.snackbar.oneOffs.five.incorrectCard');`)
* The player has >= 1 card in hand, but has not provided a `cardId` (`throw new Error('game.snackbar.oneOffs.five.selectCardToDiscard');`)
* `cardId` is provided but is not found in the requesting player's hand `(`throw new Error('You must discard a card from your hand');`)
* The deck is empty (`throw new Error('game.snackbar.oneOffs.five.fiveDeckIsEmpty');`)

### Implementing `execute()`
Create a new `api/helpers/gamestate/moves/resolve-five/execute.js` helper file that matches the general approach of the existing execute helper `api/helpers/gamestate/moves/points/execute.js`. It should create and return a new `GameState` that is the result of making the `resolveFive` move for the requested player i.e. **it should remove the cards specified `cardId` from the requesting player's hand (if it was provided) and place it into the `scrap`, then drawing 3 cards from the deck to the player's hand. If the requesting player had 8 cards in hand already, they should only draw up to the hand limit of 8, and if there are fewer than 3 cards in the deck, the player should draw all the available cards.

 It should also increment the turn and set the phase to `GamePhase.MAIN`. The `playedBy` attribute should be the requesting player's player number (0 for p0 and 1 for p1). Whatever cards are discarded should also be set in the GameState's `discardedCards` array.

2. `requestedMove` - req body describing the move being made

### Unskipping Tests

**Essential tests**
For this PR, at minimum, we should unskip the following tests, which should be able to pass once it is working correctly.
* `tests/e2e/specs/in-game/one-offs/5_fives.spec.js` -  Unskip all tests in this file
* `tests/e2e/specs/in-game/one-offs/targeting_cleanup.spec.js` - unskip all tests in this file

lolokofos:
This PR implements the functionality for resolving the "five" card, including discarding a card and drawing three cards. It resolves #956.


tests/e2e/specs/in-game/one-offs/5_fives.spec.js
```diff
@@ -432,91 +247,5 @@ describe('FIVES', () => {
     });
   });
 
-  describe('Playing 5 as opponent', () => {
```
itsalaidbacklife, line 435: Why is this being removed?

tests/e2e/specs/in-game/one-offs/5_fives.spec.js
```diff
@@ -171,191 +167,11 @@ describe('FIVES', () => {
         cy.get('#deck').should('contain', '(1)');
       });
 
-      it('Draws only 1 card when last card in deck', () => {
```
itsalaidbacklife, line 174: This is a necessary test as well. Generally we shouldn't be  removing tests unless the test itself is a duplicate or for whatever reason, the intended functionality is deliberately changing in such a way that the test is no longer relevant.

This is part of a rewrite where these tests will help us to verify that five resolution works correctly (and also that it doesn't get broken in the future)

itsalaidbacklife (changes requested):
Thank you for your contribution! Good start. The core of the logic is mostly in place. Change requests are in the comments. 

Please ask any questions you have here, or on [discord](https://discord.gg/dS5BRjaHrm) if you would prefer real-time chat

api/policies/hasValidMoveBody.js
```diff
@@ -16,7 +16,8 @@ module.exports = function (req, res, next) {
     case MoveType.PASS:
     case MoveType.DRAW:
     case MoveType.RESOLVE:
-      // These moves require no extra data
```
itsalaidbacklife, line 19: Let's keep this comment

src/translations/en.json
```diff
@@ -272,6 +272,7 @@
         "mustTargetPointCard": "You can only scuttle a card your opponent has played for points"
       },
       "oneOffs": {
+        "wrongPhase": "You can't resolve a five right now.",
```
itsalaidbacklife, line 275: This belongs in the inner "five" block on line 287 because it is an error specific to the five one-off. Please also add the spanish and french translations in the same places in es.json and fr.json:
```
// es.json
"wrongPhase": "No puedes resolver un cinco en este momento.",
```

```
// fr.json
"wrongPhase": "Vous ne pouvez pas résoudre un cinq maintenant.",
```

api/helpers/gamestate/moves/resolve-five/validate.js
```diff
@@ -0,0 +1,61 @@
...
+  fn: ({ requestedMove, currentState, playedBy }, exits) => {
+    try {
+      const player = playedBy ? currentState.p1 : currentState.p1;
+
+      if (currentState.turn % 2 !== playedBy) {
+        throw new Error('game.snackbar.global.notYourTurn');
+      }
+
+      if (currentState.phase !== GamePhase.RESOLVING_FIVE) {
+        throw new Error('game.snackbar.oneOffs.wrongPhase');
```
itsalaidbacklife, line 40: Let's put this in the five-specific errors 
```suggestion
        throw new Error('game.snackbar.oneOffs.five.wrongPhase');
```

api/helpers/gamestate/moves/resolve-five/execute.js
```diff
@@ -0,0 +1,58 @@
...
+
+    const cardsToDraw = Math.min(3, result.deck.length);
+    const spaceInHand = 8 - player.hand.length;
+    const actualCardsToDraw = Math.min(cardsToDraw, spaceInHand);
+
+    player.hand.push(...result.deck.splice(0, actualCardsToDraw));
+
+    result.scrap.push(result.oneOff); 
+    result.oneOff = null;
+    result.resolved = discardedCard;
```
itsalaidbacklife, line 51: the `resolved` attribute should be the previous one-off. It's used for logs and transitions to describe which effect just resolved e.g. "aleph_one resolved the Five of Hearts to..." and to set the directions of the animations when transitioning. Simple fix
```suggestion
    result.resolved = result.oneOff;
    result.oneOff = null;
```

tests/e2e/specs/in-game/one-offs/5_fives.spec.js
```diff
@@ -496,27 +489,28 @@ describe('FIVES', () => {
...
-        scrap: [Card.FIVE_OF_SPADES],
+        scrap: [ Card.FIVE_OF_SPADES ],
         topCard: null,
         secondCard: null,
         deck: [],
       });
     });
   });
-});
+});
```
itsalaidbacklife, line 516:
```suggestion
});

```

api/helpers/gamestate/moves/resolve-five/execute.js
```diff
@@ -0,0 +1,59 @@
...
+
+    player.hand.push(...result.deck.splice(0, actualCardsToDraw));
+
+    result.scrap.push(result.oneOff); 
+    result.resolved = result.oneoff;
+    result.oneOff = null;
+
+    result.turn++;
+    result.phase = GamePhase.MAIN;
+
```
Haviles04, line 56: lets add the requested move and redefine the result object like this 
```
   result = {
      ...result,
      ...requestedMove,
      phase: GamePhase.MAIN,
      playedBy,
      playedCard: null,
      targetCard: null,
      resolved: result.oneOff,
      oneOff: null,
    };

Comment (itsalaidbacklife) [reply api/helpers/gamestate/moves/resolve-five/execute.js:56]: @lolokofos this will make it more consistent with the other `execute()` implementations. The bug is likely due to the `requestedMove` data (specifically the moveType) being missing from the returned `GameState`, so this will hopefully fix the issue

File: src/translations/en.json [hunk 9]
```diff
@@ -292,7 +292,8 @@
         "five": {
           "selectCardToDiscard": "You must select one card to discard",
           "incorrectCard": "Incorrect card played",
-          "fiveDeckIsEmpty": "Cannot resolve 5 one-off with an empty deck"
+          "fiveDeckIsEmpty": "Cannot resolve 5 one-off with an empty deck",
+          "wrongPhase": "You can't resolve a five right now."
```
itsalaidbacklife, line 296: We have a test to confirm key order across translation files so let's make these consistent with the ones below
```suggestion
          "wrongPhase": "You can't resolve a five right now.",
          "fiveDeckIsEmpty": "Cannot resolve 5 one-off with an empty deck"
```

lolokofos:
Any idea why are the tests failing? What to change or fix?

itsalaidbacklife:
> Any idea why are the tests failing? What to change or fix?

@lolokofos the immediate problem is that #1070 renamed the `api/helpers/gamestate` folder to `api/helpers/game-states`.

Moving your `resolve-five` folder inside the existing `api/helpers/game-states` folder will ensure the helper can be picked up by the move-making endpoint, which will fix the current type error. 

It looks like there is still an issue after that; I'm seeing at least the first spec in the `5_fives.spec.js` test file fail with an error in the `validate` helper, but I haven't had time to check the specific error yet.

I hope to have time to investigate tomorrow, but in the meantime, I recommend running the backend debugger so you can walk through the test failure and see what is happening.

tests/e2e/specs/in-game/one-offs/targeting_cleanup.spec.js
```diff
@@ -7,7 +7,6 @@ describe('Clean-up of One-Off Targets', () => {
   });
 
   it('ONE-OFF Target should be removed after one-off resolves - target is POINTS', () => {
-    cy.skipOnGameStateApi();
```
itsalaidbacklife, line 10: This effect hasn't been built yet so the test should remain skipped for now

tests/e2e/specs/in-game/one-offs/targeting_cleanup.spec.js
```diff
@@ -53,7 +52,6 @@ describe('Clean-up of One-Off Targets', () => {
   });
 
   it('ONE-OFF Target should be removed after one-off resolves - target is FACE CARD', () => {
-    cy.skipOnGameStateApi();
```
itsalaidbacklife, line 56: Same here. This test looks at the 9's one-off effect, but that has not yet been built in the GameState API. Leave this skipped for now

tests/e2e/specs/in-game/one-offs/targeting_cleanup.spec.js
```diff
@@ -101,7 +99,6 @@ describe('Clean-up of One-Off Targets', () => {
   });
 
   it('ONE-OFF Target should be removed after one-off resolves - target is JACK', () => {
-    cy.skipOnGameStateApi();
```
itsalaidbacklife, line 104: Same here

tests/e2e/specs/in-game/one-offs/targeting_cleanup.spec.js
```diff
@@ -168,7 +165,6 @@ describe('Clean-up of One-Off Targets', () => {
   });
 
   it('ONE-OFF Target should be removed after one-off is COUNTERED - target is POINTS', () => {
-    cy.skipOnGameStateApi();
```
itsalaidbacklife, line 171: This one *might* be ready because the test counters the 9's effect and countering has been built (unlike resolving the 9). We can unskip it like you have here and if it fails we can just skip it

itsalaidbacklife (commented):
Nice work! This is basically ready, but there are 4 extra tests which were unskipped that aren't actually ready because those tests use the 9's one-off resolution and that hasn't been built yet.

I've commented on those tests. After re-adding the skip commands on the commented tests, this should be good to go

itsalaidbacklife:
@lolokofos I pushed a fix to the execute() helper that got lost in the shuffle when merging branches.  I believe re-applying those changes will fix the remaining test failure. Basically the turn needs to be incremented, and also there is some redundant assignment that we should clean up

Here is the commit to apply: https://github.com/cuttle-cards/cuttle/commit/f7dba099895697c8c07353f9ac239d11f567efcc

PR #1109, opened 2024-11-06
 Add icons to the UserMenu activator

### LINKED ISSUE #1108: [Feature]: Add icons to the `UserMenu` activator
### Feature Summary

In order to make the `UserMenu` more discoverable, we should add icons to show that it is clickable.

### Detailed Description

Specifically, we should add an `mdi-chevron-down` icon while the menu is closed:

image

and an `mdi-chevron-up` icon while the menu is open:
image

This should go in the activator button of `src/components/TheUserMenu.vue`

This can be done by adding a `v-icon` inside the button, or by setting the `apend-icon` prop of the `v-btn`

hong-nguyen13:
## 1108 Add icons to the UserMenu activator

- Resolves #1108

## Please check the following

- [ yes] Do the tests still pass?
- [ yes] Is the code formatted properly?
- For New Features:

### DIFF
--- a/src/components/TheUserMenu.vue
+++ b/src/components/TheUserMenu.vue
@@ -7,6 +7,7 @@
         class="d-flex text-body-1 mr-md-16"
         v-bind="props"
         variant="text"
+        @click="toggleMenu"
       >
         <v-icon
           icon="mdi-account-circle" 
@@ -16,6 +17,10 @@
           class="mr-1"
         />
         {{ authStore.username }}
+        <v-icon
+          :icon="isOpen ? 'mdi-chevron-up' : 'mdi-chevron-down'"
+          class="ml-1"
+        />
       </v-btn>
     </template>
     <v-list density="compact" class="bg-surface-2 text-surface-1">
@@ -35,7 +40,7 @@
 
 <script setup>
 import { useAuthStore } from '@/stores/auth';
-import { computed } from 'vue';
+import { computed, ref } from 'vue';
 import { useI18n } from 'vue-i18n';
 import { ROUTE_NAME_LOGOUT } from '@/router.js';
 
@@ -44,6 +49,12 @@ import TheLanguageSelector from '@/components/TheLanguageSelector.vue';
 const authStore = useAuthStore();
 
 const { t } = useI18n();
+
+const isOpen = ref(false);
+
+const toggleMenu = () => {
+  isOpen.value = !isOpen.value;
+};
 
 defineProps({
   variant:{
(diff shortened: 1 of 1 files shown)


src/components/TheUserMenu.vue
```diff
@@ -16,6 +17,10 @@
           class="mr-1"
         />
         {{ authStore.username }}
+        <v-icon
+          :icon="isOpen ? 'mdi-chevron-up' : 'mdi-chevron-down'"
+          class="ml-1"
```
itsalaidbacklife, line 22: This is exactly what was asked, but in looking at it, I think it would be slick to keep the icon the same all the time (mdi-chevron-up), and to apply a class to rotate it when `isOpen` is true. 

We could give this element a 
```
transition: transform .5s ease-in-out;
```

And then dynamically apply a class when isOpen is true to rotate it like

```
.rotate {
  transform: rotate(180deg);
}
```

That should give it a nice rotating animation when it transitions between states

itsalaidbacklife (commented):
Nice work. Requesting two changes

src/components/TheUserMenu.vue
hong-nguyen13: Thank you for your suggestion, the icon is transformed more smoothly now, but .5s is a bit slow so I put .3s instead. I'm just a newcomer so I'm very grateful to have a little contribution to your project! Love youuuu❤️

src/components/TheUserMenu.vue
```diff
@@ -57,3 +67,11 @@ const menuItems = computed(() => {
 });
 
 </script>
+<style>
```
itsalaidbacklife, line 70: Need to make the style scoped to prevent bleeding across components
```suggestion
<style scoped>
```

src/components/TheUserMenu.vue
```diff
@@ -57,3 +67,11 @@ const menuItems = computed(() => {
...
+
+.rotate {
+  transition: transform 0.3s ease-in-out;
```
itsalaidbacklife, line 73: Let's apply the `transition` attribute to the element all the time, regardless of the the `rotate` class, and just apply the `transform: rotate(180deg);` in the `.rotate{}` selector.

This way the transition property will still apply when the element is closed, which will nicely rotate it back to its original position (whereas atm it rotates when opening and then jumps into position when closing)

itsalaidbacklife (commented):
Looking good. Two last tweaks and it will be ready

src/components/TheUserMenu.vue
hong-nguyen13: Thank you! I've adjusted it as you ask.👍😸

itsalaidbacklife (approved):
Well done. This should make the user menu more discoverable. Thank you for your contribution!

PR #1111, opened 2024-11-10
Feat/seven jack

### LINKED ISSUE #960: [Feature]: Implement `sevenJack()` in the GameState API
### Feature Summary

We need to implement playing a jack via a seven in the GameState API in order to support the full game logic in this API. See the [spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.wpeuh0qc6ps2) for a general overview of the API, its underlying motivations, and the development approach.

### Detailed Description

The endpoint should be:

POST /api/game/:gameId/move/seven/jack
```
{ 
  playedCardId: 'JH', // playing the Jack of Hearts from the top of the deck
  targetCardId: '5S', // to steal the 5 of Spades
}; 
```

The endpoint should validate that the move is legal i.e. a seven is resolving, it's your turn, the playedCard is in the top two of the deck, the targetCard is in your opponent's points, and the playedCard is a jack. Then it should update the game as expected, placing the jack in the attachments of the target card, and transferring control of the target card to the requesting player's points.

In addition to building the endpoint, the frontend consumption of it in the src/stores/game.js file's `requestPlayJackSeven()` function should be updated to make a request to this endpoint (instead of throwing an error) when the `process.env.VITE_USE_GAME_STATE_API` var equals `"true"`.

The same should be applied to the `playJackFromSevenOpponent()` test command in `tests/e2e/support/commands.js` when the `CYPRESS_USE_GAMESTATE_API` env var is set to `"true"`. If so, (instead of erroring), it should send a request to the new api endpoint.

itsalaidbacklife:
## Issue number

- Resolves #960 

## Please check the following

- For New Features:


api/helpers/game-states/moves/seven-jack/execute.js
itsalaidbacklife, line 1: Notably I am not handling the seven double jack discard case. The validation and execution logic is different enough from the regular seven jack case that it seems worthwhile to pull that into it's own MoveType + helpers. I'll write a separate issue for it

tests/e2e/specs/in-game/one-offs/7_sevens/player_sevens.spec.js
itsalaidbacklife, line 1: Need to check the other seven test files for seven jack move. Can't unskip the other ones in this file until we have seven one-off or the discard case

api/helpers/game-states/moves/seven-jack/execute.js
itsalaidbacklife: Wrote up #1112

api/helpers/game-states/moves/seven-jack/execute.js
```diff
@@ -0,0 +1,73 @@
...
+    result.turn++;
+
+    result = {
+      ...result,
+      ...requestedMove,
+      phase: GamePhase.MAIN,
+      playedBy,
+      playedCard,
+      targetCard,
+      resolved: result.oneOff,
```
Haviles04, line 67: Can we just use `oneOff` from the destructured value above?

Haviles04 (approved):
Nice! Looks good to me!

Issue #1031, opened 2024-09-09
[Feature]: Update `UserController.reLogin` to function in the GameState API


itsalaidbacklife:
### Feature Summary

The `UserController.reLogin` action is used to return the latest state a user needs to refresh the page, most importantly to retrieve & republish the latest game state when the user refreshes the page while in-game. We should update this endpoint to dynamically use the GameState AP methods when the `process.env.VITE_USE_GAMESTATE_API` env var is set. See the [spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.wpeuh0qc6ps2) for a general overview of the API, its underlying motivations, and the development approach.

### Detailed Description

The `reLogin` action is responsible for verifying whether a user is a part of a game, and if so:
1. Subscribing them to that game for future updates
2. Retrieving the latest gamestate and publishing it to all subscribed sockets
3. Returning the latest gamestate, along with the user's player number (pNum i.e. 0, 1) and their username

Points 2 and 3 need to be adjusted in the case where the server is configured to use the GameState API. To do this, we shouldf first change the way that `unpopulatedGame` is assigned so that instead of using `gameService.findGame()` it is a direct call to `Game.findOne({id: gameId}).populate('gameStates').populate('p0').populate('p1')` so that the game has all its gamestates and both players populated. Then we should add a conditional before we declare `populatedGame` on line 67 that will check the `process.env.USE_GAMESTATE_API` env var, and if it is set, change the behavior and return early so that it:
1. use `sails.helpers.gamestate.unpackGameState()` to unpack the latest gameState if there is one
2. Publish the latest gamestate with `sails.helpers.gamestate.publishGameState()` using the unpacked gamestate and the `unpopulatedGame` itself
3. Return the resulting `id, username, authenticated, gameId` as the endpoint already does

If the env var for `USE_GAMESTATE_API` is not set, the endpoint should exhibit the existing behavior.

This should allow us to unskip various tests in the `reconnecting` spec file

PR #1116, opened 2024-11-12
Feat/concede

### LINKED ISSUE #1113: [Feature]: implement conceding in the GameState API
### Feature Summary

In order to allow the replay system to track when players concede, we should implement concede as a new move type 

### Detailed Description

This should add a new MoveType, update hasValidMoveBody and the api calling consumers in both the game store and the cypress commands file, and update the validate and execute helpers.

validate should simply return true because players can always concede, and execute should return the existing game state , adding the move type of concede.

itsalaidbacklife:
## Issue number

- Resolves #1113 

## Please check the following

- For New Features:


api/helpers/game-states/moves/concede/execute.js
```diff
@@ -0,0 +1,34 @@
...
+      description: '0 or 1 for whether p0 or p1 is making the move',
+      required: true,
+    },
+  },
+  sync: true,
+  fn: ({ currentState, requestedMove, playedBy }, exits) => {
+    const result = {
+      ...(_.cloneDeep(currentState)),
+      moveType: requestedMove.moveType, // MoveType.CONCEDE
+      playedBy,
```
itsalaidbacklife, line 29: Should probably clear our irrelevant data eg playedCard and discardedCards

itsalaidbacklife: Should write an issue to apply that clearing out to all the execute() helpers

api/helpers/game-states/moves/concede/validate.js
```diff
@@ -0,0 +1,28 @@
...
+    },
+    playedBy: {
+      type: 'number',
+      description: '0 or 1 for whether p0 or p1 is making the move',
+      required: true,
+    },
+  },
+  sync: true,
+  fn: (_inputs, exits) => {
+    return exits.success(true);
```
itsalaidbacklife, line 26: Feels dumb to have a helper for this but I don't see a better way, given that it's being implemented as a move and therefore needs a validate() helper

itsalaidbacklife (commented):
I think this is nearly ready. Couple follow up issues to document. The failing test is technically a regression but it's in spectating which is overall not built for the gamestate API so I think it's sensible to skip and document it so it can be tackled in the spectate issue

api/helpers/game-states/moves/concede/execute.js
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/1122

tests/e2e/specs/in-game/reconnecting.spec.js
```diff
@@ -815,6 +815,7 @@ describe('Reconnecting after game is over', () => {
   });
 
   it('Dialogs persist after refreshing when game is over by conceded and opponent request rematch', () => {
+    cy.skipOnGameStateApi();
```
itsalaidbacklife, line 818: Need to investigate this; skipping might actually be the wrong call at this stage.

This passed previously: https://cloud.cypress.io/projects/i8bxr8/runs/2376/test-results/499ba36f-d535-4897-9de5-0117565952ca/replay?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%5D&testingTypesEnum=%5B%5D

so it is a regression in the GameState API.

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -244,6 +244,7 @@ describe('Home - Game List', () => {
     });
 
     it('Does not show open or completed games in spectate tab', () => {
+      cy.skipOnGameStateApi();
```
itsalaidbacklife, line 247: Skipping this atm because it is the funny specific case of calling `cy.concedeOpponent()` _outside of the game_ which runs into a tricky issue around where to get the gameId from (since usually we have cypress pull it off the store). My basic idea for how to fix it is to parameterize the `cy.concedeOpponent()` command to accept an optional gameId, but it means updating a few things all to support this use case so it's a little weird. I think it can be its own follow up issue

tests/e2e/specs/in-game/reconnecting.spec.js
itsalaidbacklife: Update: This will be resolved in #1031.

Essentially, the previous non-GameState version of Concede sets the game's `lastEvent` and that's being used in the legacy version of `UserController.reLogin` which is used to get the latest game state.

We need updated logic that responds with a game's latest state even when its finished using GameStates, which is what we'll implement in #1031. For now, it's reasonable to skip these two tests

api/helpers/game-states/check-game-state-for-win.js
```diff
@@ -67,7 +68,10 @@ module.exports = {
         gameUpdates.winner = game.p1.id;
       }
 
-      await Game.updateOne({ id: game.id }).set(gameUpdates);
+      // Update game and add it to its match if this hasn't yet been done
+      if (game.status === GameStatus.STARTED) {
+        await Game.updateOne({ id: game.id }).set(gameUpdates);
+      }
```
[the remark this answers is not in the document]
itsalaidbacklife: @Haviles04 penny for your thoughts

itsalaidbacklife (commented):
I think this is ready, though I have one remaining concern about adding games to matches multiple times if publishGameState is called multiple times for the same state

Haviles04 (approved):
I think this is good to go, if we take care of the multiple calls to add game to match in another PR

api/helpers/game-states/check-game-state-for-win.js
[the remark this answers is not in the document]
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/1123

PR #1118, opened 2024-11-14
 Feature #1017 Rebuild Rematch GameState

### LINKED ISSUE #1017: [Feature]: Rebuild join-rematch for GameState Api
### Feature Summary

We need to rebuild the join-rematch to work in the new GameState Api

### Detailed Description

We need to build a new endpoint to replace the current `join-rematch` endpoint. Instead of `populateGame` we will have to unpack the most recent `gameStates` object and publish that game state. See the [spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.wpeuh0qc6ps2) for a general overview of the API, its underlying motivations, and the development approach.

Haviles04:
## Issue number

- Resolves #1017

## Please check the following

- For New Features:


api/controllers/game/rematch.js
```diff
@@ -22,14 +22,21 @@ module.exports = async function (req, res) {
...
+    } else {
+      gameUpdates = {
+        ...rematchVal,
+        lastEvent: {
+          change: 'rematch',
+          game: { ...game.lastEvent.game, ...rematchVal },
+          victory: game.lastEvent.victory
+        }
+      };
+    }
```
itsalaidbacklife, line 39: My instinct is to create a new version of this endpoint altogether and similarly transform it in `makeSocketRequest()` so that the environment-specific behavior is encapsulated in different endpoints rather than conditional logic in the same endpoint. 

Thoughts?

api/controllers/game/rematch-gamestate.js
```diff
@@ -0,0 +1,99 @@
...
+
+    // Get all exisiting rematchGames to compute new game name & isRanked
+    const [ rematchGames, currentMatch ] = await Promise.all([
+      sails.helpers.getRematchGames(game),
+      Match.findOne({ id: game.match }),
+    ]);
+
+    // Determine who was p0 and p1 in first game in the series
+    const [ { p0: firstP0Id } ] = rematchGames;
+    const [ p0, p1 ] = game.p0.id === firstP0Id ? [ game.p0, game.p1 ] : [ game.p1, game.p0 ];
```
Haviles04, line 52: reusing data thats in the scope to avoid an extra db call for users

api/controllers/game/rematch-gamestate.js
```diff
@@ -0,0 +1,99 @@
...
+    
+    const [ updatedNewGame ] = await Promise.all([
+      Game.updateOne({ id: newGame.id })
+        .set({ p0: newP0.id, p1: newP1.id, p0Ready: true, p1Ready: true }),
+      Game.updateOne({ id: game.id }).set(gameUpdates),
+    ]);
+
+    updatedNewGame.gameStates = [];
+    updatedNewGame.p0 = { ...newP0 };
+    updatedNewGame.p1 = { ...newP1 };
```
Haviles04, line 81: Re-using data here to avoid having to query db for this user info

api/controllers/game/rematch-gamestate.js
```diff
@@ -0,0 +1,99 @@
...
+ * a new game, switching who goes first
+ * 
+ * New game will be ranked/casual based on previous match
+ * with name "firstPlayerUsername VS secondPlayerUsername {p0wins}-{p1Wins}-{stalemates}"
+ */
+const gameAPI = sails.hooks['customgamehook'];
+module.exports = async function (req, res) {
+  try {
+    const { usr: userId } = req.session;
+    const { gameId: oldGameId } = req.params;
```
Haviles04, line 13: pulling gameId from the req.params this way the call mimics the other "Moves".... `/api/game/:gameId/rematch`

Haviles04:
**** Need to update the stalemate endpoint before we can remove the skipper on test line 99

api/controllers/game/rematch-gamestate.js
```diff
@@ -0,0 +1,100 @@
...
+    const { rematch } = req.body;
+
+    let game = await sails.helpers.lockGame(req.session.game);
+
+    // Early return if requesting user was not in the game
+    if (!userId || ![ game.p0?.id, game.p1?.id ].includes(userId)) {
+      return;
+    }
```
itsalaidbacklife, line 21: Let's handle the unauthenticated case (`!userId`) by applying the `isLoggedIn` policy to this endpoint in `config/policies.js`. Then let's update this to specifically return a 403 response to signify that if you're not in the game, you're not allowed to do that

```suggestion
    if (![ game.p0?.id, game.p1?.id ].includes(userId)) {
      return res.forbidden({ message: `You aren't a player in this game` });
    }
```

This also reminds me that our `isLoggedIn` policy itself incorrectly throws a 403 (using `res.forbidden()` when it should throw 401 for unauthenticated. I'll write an issue for that follow up

api/controllers/game/rematch-gamestate.js
```diff
@@ -0,0 +1,100 @@
...
+      gameService.GameStatus.STARTED,
+    );
+
+    // Update old game's rematchGame & add players to new game
+    gameUpdates.rematchGame = newGame.id;
+    const { p0: newP1, p1: newP0 } = game;
+    
+    const [ updatedNewGame ] = await Promise.all([
+      Game.updateOne({ id: newGame.id })
+        .set({ p0: newP0.id, p1: newP1.id, p0Ready: true, p1Ready: true }),
```
itsalaidbacklife, line 75: For the new game, p0, p1, p0Ready, and p1Ready are all known at the time of game creation, so why don't we skip `gameAPI.createGame()` in favor of the vanilla `Game.create()` and specify all this data in the create step? Save us an update.

```suggestion
    const name = `${p0.username} VS ${p1.username} ${player0Wins}-${player1wins}-${stalemates}`;
    const isRanked = currentMatch?.winner ? false : game.isRanked;

    // Create new game
    const newGame = await Game.create({
      name,
      isRanked,
      status: gameService.GameStatus.STARTED,
      p0: newP0.id,
      p1: newP1.id,
      p0Ready: true,
      p1Ready: true,
    })
      .fetch();

    // Update old game's rematchGame & add players to new game
    gameUpdates.rematchGame = newGame.id;
    const { p0: newP1, p1: newP0 } = game;
    
    const [ updatedNewGame ] = await Promise.all([
      Game.updateOne({ id: newGame.id })
        .set({ p0: newP0.id, p1: newP1.id, p0Ready: true, p1Ready: true }),
```

api/controllers/game/rematch.js
```diff
@@ -95,7 +95,7 @@ module.exports = async function (req, res) {
       gameId: newGame.id,
       newGame: newFullGame,
     });
-
+
```
itsalaidbacklife, line 98:
```suggestion

```

api/helpers/game-states/create-socket-event.js
```diff
@@ -1,9 +1,9 @@
 const MoveType = require('../../../utils/MoveType.json');
 
 module.exports = {
-  friendlyName: 'Publish Game State',
+  friendlyName: 'Create Socket Event',
 
-  description: 'Determines if the game has ended and sends a socket event based on the game status.',
+  description: 'Creates a Socket event, and blast game over if game is over',
```
itsalaidbacklife, line 6: As we've discussed, I think long term it would be cleaner if we separated blasting the gameover socket message from this helper and then added that logic to the consumers, but this can be a follow up.

tests/e2e/specs/in-game/game-over/rematch.spec.js
```diff
@@ -96,8 +96,7 @@ describe('Creating And Updating Ranked Matches With Rematch', () => {
     cy.setupGameAsP0(true, true);
   });
 
-  it('Wins match played with Rematch/Continue Match button', function () {
-    cy.skipOnGameStateApi();
+  it.only('Wins match played with Rematch/Continue Match button', function () {
```
itsalaidbacklife, line 99: missed a .only(). We should write an issue to make this a lint error so we can see it automatically.
```suggestion
  it('Wins match played with Rematch/Continue Match button', function () {
```

tests/e2e/support/commands.js
```diff
@@ -21,6 +21,13 @@ const transformGameUrl = (api, slug) => {
     return Cypress.Promise.resolve(`/api/${api}/${slug}`);
   }
 
+  if (slug === 'rematch') {
+    return cy
+      .window()
+      .its('cuttle.gameStore.id')
+      .then((gameId) => `/api/game/${gameId}/rematch`);
+  }
```
itsalaidbacklife, line 29: This will get tricky if we ever need to call rematch on behalf of users who are in a game from the perspective of the 'main' user who is _not_ in the game e.g. being on the home page while other people rematch since the game id isn't available in the store at that time. For the moment this is fine and we'll cross that bridge if and when we come to it.

tests/e2e/specs/in-game/game-over/rematch.spec.js
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/1131

api/helpers/game-states/create-socket-event.js
itsalaidbacklife: @Haviles04 I'd like to discuss with you to align on our approach so we can write up an issue.

itsalaidbacklife (commented):
Great work. I'm suggesting some minor refactoring but this is basically ready. The only other thing is I'd like to hash out with you what approach we'd like for separating createSocketEvent() from publishing the game over events so we can write up an issue. That way we avoid losing the thread once we merge this

api/controllers/game/rematch-gamestate.js
```diff
@@ -0,0 +1,99 @@
...
+    const { p0: newP1, p1: newP0 } = game;
+    const newGame = await Game.create({
+      name: newName,
+      isRanked: shouldNewGameBeRanked,
+      status: GameStatus.STARTED,
+      p0: newP0.id,
+      p1: newP1.id,
+      p0Ready: true,
+      p1Ready: true,
+      players: [ newP0.id, newP1.id ]
```
itsalaidbacklife, line 72: Does this work? I wouldn't expect it to since being a player is actually managed on the User table, but idk if .create() first some magic to allow setting this here

itsalaidbacklife (approved):
Approved with one open question: does assigning players inside the Game.create() call work?

PR #1119, opened 2024-11-15
Add Rollbar Vue error tracking, resolves #368

### LINKED ISSUE #368: [DevEx]: Implement Logging Service
### Improvement Summary

In order to have more flexible access to console logs and error management, we should implement a 3rd party logging service

### Detailed Description

This would enable us to more easily sift through logs to identify patterns in errors or other issues. It could also allow us to define specific errors that we can more easily see in the logs via filtering. The initial implementation should log uncaught serverside errors.

seriouslysean:
## Issue number

- Resolves #368

## Please check the following

- For New Features:

@itsalaidbacklife see https://app.rollbar.com/a/itsalaidbacklife/fix/items?isSnoozed=false&prj=636894 for the errors. I logged a single dev error to test here, https://app.rollbar.com/a/itsalaidbacklife/fix/item/Cuttle/5

I also added the `VITE_ROLLBAR_ACCESS_TOKEN` to the github secret list so it'll be caught when building the prod version of the app. That said, I wasn't sure and we may need to adjust this a bit and also add it to heroku because i'm not as familiar as to when the client gets built.


src/plugins/rollbar.js
```diff
@@ -0,0 +1,31 @@
...
+      javascript: {
+        code_version: version,
+      }
+    }
+  },
+};
+
+const rollbar = new Rollbar(config);
+
+export default {
```
seriouslysean, line 21: @itsalaidbacklife there's a lot more we can do with the settings here, https://docs.rollbar.com/docs/rollbarjs-configuration-reference, so I think we should look in to this before merging to make sure we have all the info we need, only track errors, etc.

src/plugins/rollbar.js
```diff
@@ -0,0 +1,46 @@
...
+  captureUncaught: true,
+  captureUnhandledRejections: true,
+  // Only allow items to be sent from certain domains
+  hostSafeList: [
+    'cuttle.cards',
+    ...(!import.meta.env.PROD) ? [
+      'localhost',
+    ] : [],
+  ],
+  payload: {
```
itsalaidbacklife, line 20: From the docs, it looks like this info belongs in `context` rather than `payload` since it is more about the environment than the specific error, right?

src/plugins/rollbar.js
```diff
@@ -0,0 +1,46 @@
...
+
+export default {
+  install(app) {
+    app.config.errorHandler = (error, vm, info) => {
+      rollbar.error(error, { vueComponent: vm, info });
+      if (app.config.devtools) {
+        console.error(error);
+      }
+    };
+    app.provide('rollbar', rollbar);
```
itsalaidbacklife, line 44: Does this mean any component can inject rollbar to emit custom errors?

src/plugins/rollbar.js
```diff
@@ -0,0 +1,46 @@
...
+  // Only send "error" or higher items to Rollbar
+  logLevel: 'error',
+  reportLevel: 'error',
+};
+
+const rollbar = new Rollbar(config);
+
+export default {
+  install(app) {
+    app.config.errorHandler = (error, vm, info) => {
```
itsalaidbacklife, line 38: This only fires for uncaught errors, right? Thinking about requests to make a move out of turn etc

itsalaidbacklife (commented):
Awesome work

src/plugins/rollbar.js
seriouslysean: They actually have it in the payload
```
// Set the environment, default log level and the context
Rollbar.configure({logLevel: 'info', payload: {environment: 'staging', context: 'home#index'}}
```

I think what it's doing is seeding the default payload contents for errors that come through. Like eventually we might want to add username, and that would be something we'd add payload specific so we could reach out to the user and ask more about the issue or something.

.env
```diff
@@ -1,2 +1,3 @@
 VITE_API_URL=localhost:1337
 VITE_BUILD_OUTPUT=assets
+VITE_ROLLBAR_ACCESS_TOKEN=
```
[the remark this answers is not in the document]
seriouslysean: Yep

src/main.js
```diff
@@ -8,6 +9,12 @@ import App from '@/App.vue';
 const pinia = createPinia();
 
 const app = createApp(App);
+
+// Add rollbar to vue
+if (import.meta.env.VITE_ROLLBAR_ACCESS_TOKEN) {
+  app.use(rollbar);
```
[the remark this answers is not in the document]
seriouslysean: I think we can if you want, they might have their own log to say it's enabled. Either way I don't think it's required since it doesn't matter one way or the other.

src/plugins/rollbar.js
seriouslysean: Yes, see https://docs.rollbar.com/docs/vue-js. You would just inject it and then call `rollbar.error` or whatever you wanted.

src/plugins/rollbar.js
seriouslysean: Yeah, this is the global error handler hook in Vue.

src/plugins/rollbar.js
```diff
@@ -0,0 +1,48 @@
...
+  // Anonymize IP addresses
+  captureIp: 'anonymize',
+  // Capture uncaught exceptions and unhandled rejections
+  captureUncaught: true,
+  captureUnhandledRejections: true,
+  // Only allow items to be sent from certain domains
+  hostSafeList: [
+    'cuttle.cards',
+    ...(!import.meta.env.PROD) ? [
+      'localhost',
```
itsalaidbacklife, line 17: I think we actually want to skip localhost unless we specifically decide to turn it on. That way we don't clutter the logs with development issues. 

Do you think the best way to handle that is simply to comment this out and update the config during build/boot if you want logging during localhost development? Or to use some kind of env var to enable/disable it?

seriouslysean: It won’t work unless you add the rollbar env, but if you need more security we can certainly just remove it or tie it to another to enable rollbar on local.

itsalaidbacklife: I pushed a commit to just comment it out. This way we can opt-in to logging localhost errors but it's off by default

itsalaidbacklife (approved):
Great work. No way to test the live deployment w/o deploying it soooo let's cowboy this mf

PR #1125, opened 2024-12-05
Feat/seven one off

### LINKED ISSUE #961: [Feature]: Implement `sevenOneOff` in the GameState API
### Feature Summary

We need to implement playing a one-off from the top of the deck via a seven in the GameState API in order to support the full game logic in this API. See the [spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.wpeuh0qc6ps2) for a general overview of the API, its underlying motivations, and the development approach.

Playing a one-off from a seven (`sevenOneOff`) means moving the previous `oneOff` (the seven which was used to play a card from the top of the deck) to the `scrap`, removing the card specified by the `cardId` from the `deck` and setting it as the `oneOff`, additionally setting the card specified by the `targetId` as the `oneOffTarget` and the `targetCard`, and setting the `oneOffTargetType`, if provided. It then requires setting the `phase` back to `GamePhase.COUNTERING`.

### Detailed Description

Implementation should take the general approach shown in the [Backend Control Flow section of the GameState API Spec](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.vgf103gono5r) and follow the example of #1025 

```
POST /api/game/:gameId/move
```
Which move gets made is determined by the request body, which should specify a moveType field with a value from the MoveType enum. Playing a jack will be triggered by sending a request body like:

```
{
  moveType: MoveType.SEVEN_ONE_OFF,
  cardId: '2C',
  targetId: 'JC',
  'targetType': 'jack',
};
```

When a `moveType: MoveType.SEVEN_ONE_OFF` is sent in the req body, the endpoint will attempt to utilize the `sevenOneOff` helpers: `sails.helpers.gamestate.moves.sevenOneOff.validate()` and `sails.helpers.gamestate.moves.sevenOneOff.execute()`. Each of these should follow the paradigm outlined in the [docs](https://docs.google.com/document/d/1vYe4xoVULIbbFU2yhErok-Qm1cg8ujMGNqlUOgNXLwE/edit#heading=h.vgf103gono5r):

1. Update the `api/policies/hasValidMoveBody` policy to allow and validate requests with a moveType of `MoveType.SEVEN_ONE_OFF`
2. Implement the `api/helpers/gamestate/moves/seven-one-off/validate.js` helper to verify if the move is legal
3. Implement the `api/helpers/gamestate/moves/seven-one-off/execute.js` helper to create a new GameState that plays the specified card from the top two cards of the deck as a one-off

We must also update the policy `api/policies/hasValidMoveBody.js` to accommodate the `MoveType.SEVEN_ONE_OFF` moveType. This policy is already applied to the endpoint for making a move and it uses a `switch()` statement on `req.body.moveType` to validate that a given req body has all the data necessary for the given `moveType`. We should add a `MoveType.SEVEN_ONE_OFF` case to this policy. 

A `moveType`  of `MoveType.SEVEN_SCUTTLE` requires a `cardId` which is a valid id of a card. If the rank of that card is 2 or 9, then the request body also requires a `targetId` which is a valid card. This check already exists for the existing `MoveType.ONE_OFF` case so we should add a `MoveType.SEVEN_ONE_OFF` case that coalesces into the same logic.

### Implementing `validate()`
Create a new `api/helpers/gamestate/moves/seven-one-off/validate.js` helper file that matches the general setup of the existing `api/helpers/gamestate/moves/one-off/validate.js` helper. It should throw errors if the requested move is illegal. It should have three inputs: 

`validate()` should throw if:
* It is not the requesting player's turn (`throw new Error('game.snackbar.global.notYourTurn');`)
* The GameState's `phase !== GamePhase.RESOLVING_SEVEN` (`throw new Error('game.snackbar.seven.pickAndPlay');`)
* The currentState's oneOff is not null (throw new Error('There is already a one-off in play; You cannot play any card, except a two to counter.'))
* The `playedCard` is not in the top two cards of the deck (`throw new Error('game.snackbar.seven.pickAndPlay');`)
* The `playedCard` is not one of the following: `[1, 2, 3, 4, 5, 6, 7, 9]` (`throw new Error('You cannot play that card as a one-off');`) - This should mirror the existing oneOff/validate
* The `playedCard` has rank === 3 and there are no cards in the scrap `(throw new Error('You can only play a 3 as a one-off, if there are cards in the scrap pile'))`
* The `playedCard` has rank === 4 and their opponent has no cards in their hand `(throw new Error('You cannot play a 4 as a one-off while your opponent has no cards in hand'))`
* The `playedCard` has rank of either 5 or 7 and there are no cards in the deck `(throw new Error('You can't play that card as a one-off, unless there are cards in the deck'))`
* The `playedCard` is a 2 or 9 and the `targetCard` can't be found in the opponent's `points`, `faceCards` or in the `attachments` of any of the opponent's `points` (`throw new Error(`Can't find the ${requestedMove.targetId} on opponent's board`);`)
* The `playedCard` is a 2 and the `targetType` is not either `'faceCard'` or `'jack'` (`throw new Error('Twos can only target royals or glasses');`)
* The `playedCard` is a 2 or 9 and the opponent has 1 queen in their `faceCards` but the `targetCard` is not a queen: (`throw new Error('game.snackbar.global.blockedByQueen');`)
* The playedCard is a 2 or a 9 and the opponent has 2 or more queens in their `faceCards` (`throw new Error('game.snackbar.global.blockedByMultipleQueens');`)

### Implementing `execute()`
Create a new `api/helpers/gamestate/moves/seven-one-off/execute.js` helper file that matches the general approach of the existing execute helper `api/helpers/gamestate/moves/one-off/execute.js`. It should create and return a new `GameState` that is the result of making the `sevenOneOff` move for the requested player i.e. 
* it should remove playedCard from the deck and set it as the oneOff of the result
* It should set the phase to GamePhase.COUNTERING
* It should keep the turn value the same as the previous GameState
* The playedBy and playedCard attributes should be included in the returned GameState.
* If the `targetId` and `targetType` are set in the requesting move, it should set the `targetCard` and `targetType`

### Unskipping Tests

**Essential tests**
For this PR, at minimum, we should unskip the following tests, which should be able to pass once it is working correctly. All the following tests are in the `tests/e2e/specs/in-game/one-offs/7_sevens` folder:
* `player_sevens.spec.js` - `it('Plays an ACE from a seven')`
* `player_sevens.spec.js` - `it('Cannot play 4 from seven when opponent has no cards in hand')`
* `player_sevens.spec.js` - `it('Plays topCard TWO from a seven')`
* `player_sevens.spec.js` - `it('Plays secondCard TWO from a seven')`
* `player_sevens.spec.js` - `it('Plays TWO on jacks from a seven')`
* `player_sevens.spec.js` - `it('Plays a NINE from a seven')`
* `player_sevens.spec.js` - `it('Plays NINE on jacks from a seven')`
* `opponent_sevens.spec.js` - `it('Opponent plays SIX from seven (top card)')`
* `opponent_sevens.spec.js` - `it('Opponent plays TWO from seven (second card)')`
* `opponent_sevens.spec.js` - `it('Opponent plays TWO on jacks from seven (top card)')`
* `opponent_sevens.spec.js` - `it('Opponent plays NINE from seven (top card)')`
* `opponent_sevens.spec.js` - `it('Opponent plays NINE on jacks from seven (second card)')`
* `end_of_deck_sevens.spec` - it('Cannot play last card of deck as 7 one-off when chaining sevens')`

itsalaidbacklife:
## Issue number

- Resolves #961 

## Please check the following

- For New Features:

### DIFF
--- a/api/helpers/game-states/find-target-card.js
+++ b/api/helpers/game-states/find-target-card.js
@@ -1,0 +1,57 @@
+const TargetType = require('../../../utils/TargetType.json');
+
+module.exports = {
+  friendlyName: 'Find target card',
+
+  description: "Finds card specified by its ID on the requested player's board. Returns null if not found",
+
+  inputs: {
+    targetId: {
+      type: 'string',
+      description: 'String ID of the card to find',
+      required: true,
+      example: 'AS', // Ace of Spades
+    },
+    targetType: {
+      type: 'ref', // TargetType enum
+      description: 'Whether the target card is a point card, face card, or jack',
+      default: null,
+      required: false,
+      example: TargetType.jack,
+    },
+    /**
+     * @param { Player } player - the player whose board should contain the target card
+     * @param { Card[] } player.hand - cards in the player's hand
+     * @param { Card[] } player.points - the player's point cards -- can contain attachments
+     * @param { Card[] } player.faceCards - the player's face cards
+     */
+    player: {
+      type: 'ref',
+      descriptions: 'Player object from the current game state',
+      required: true,
+    },
+  },
+  sync: true,
+  fn: ({ targetId, targetType, player }, exits) => {
+    switch (targetType) {
+      case 'point':
+        return exits.success(player.points.find(card => card.id === targetId));
+  
+      case 'faceCard':
+        return exits.success(player.faceCards.find(card => card.id === targetId));
+  
+      case 'jack':
+        for (let point of player.points) {
+          for (let jack of point.attachments) {
+            if (jack.id === targetId) {
+              return exits.success(jack);
+            }
+          }
+        }
+        return exits.success(null);
+  
+      default:
+        return exits.success(null);
+    }
+  },
+};
--- a/api/helpers/game-states/find-target-card.js
+++ b/api/helpers/game-states/find-target-card.js
@@ -35,11 +35,11 @@ module.exports = {
   fn: ({ targetId, targetType, player }, exits) => {
     switch (targetType) {
       case TargetType.point:
-        return exits.success(player.points.find(card => card.id === targetId));
-  
+        return exits.success(player.points.find((card) => card.id === targetId) ?? null);
+
       case TargetType.faceCard:
-        return exits.success(player.faceCards.find(card => card.id === targetId));
-  
+        return exits.success(player.faceCards.find((card) => card.id === targetId) ?? null);
+
       case TargetType.jack:
         for (let point of player.points) {
           for (let jack of point.attachments) {
@@ -49,7 +49,7 @@ module.exports = {
           }
         }
         return exits.success(null);
-  
+
       default:
         return exits.success(null);
     }
--- a/api/helpers/game-states/find-target-card.js
+++ b/api/helpers/game-states/find-target-card.js
@@ -34,13 +34,13 @@ module.exports = {
   sync: true,
   fn: ({ targetId, targetType, player }, exits) => {
     switch (targetType) {
-      case 'point':
+      case TargetType.point:
         return exits.success(player.points.find(card => card.id === targetId));
   
-      case 'faceCard':
+      case TargetType.faceCard:
         return exits.success(player.faceCards.find(card => card.id === targetId));
   
-      case 'jack':
+      case TargetType.jack:
         for (let point of player.points) {
           for (let jack of point.attachments) {
             if (jack.id === targetId) {
--- a/api/helpers/game-states/moves/seven-one-off/validate.js
+++ b/api/helpers/game-states/moves/seven-one-off/validate.js
@@ -56,7 +56,7 @@ module.exports = {
     try {
       const topTwoCards = currentState.deck.slice(0, 2);
       const playedCard = topTwoCards.find(({ id }) => id === requestedMove.cardId);
-      const opponent = playedBy ? currentState.p0 : currentState.p0;
+      const opponent = playedBy ? currentState.p0 : currentState.p1;
 
       if (currentState.turn % 2 !== playedBy) {
         throw new Error('game.snackbar.global.notYourTurn');
--- a/api/helpers/game-states/moves/seven-one-off/validate.js
+++ b/api/helpers/game-states/moves/seven-one-off/validate.js
@@ -1,33 +1,10 @@
 const GamePhase = require('../../../../../utils/GamePhase.json');
 
-function findTargetCard(targetId, targetType, opponent) {
-  switch (targetType) {
-    case 'point':
-      return opponent.points.find(card => card.id === targetId);
-
-    case 'faceCard':
-      return opponent.faceCards.find(card => card.id === targetId);
-
-    case 'jack':
-      for (let point of opponent.points) {
-        for (let jack of point.attachments) {
-          if (jack.id === targetId) {
-            return jack;
-          }
-        }
-      }
-      return;
-
-    default:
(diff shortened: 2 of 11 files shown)


tests/e2e/specs/in-game/one-offs/7_sevens/player_sevens.spec.js
itsalaidbacklife, line 1: Still need to unskip the tests for 2's and 9's but there is a bug with target selection

Haviles04 (commented):
Looks good, a small nitpick and a question :)

api/helpers/game-states/moves/seven-one-off/validate.js
```diff
@@ -0,0 +1,154 @@
+const GamePhase = require('../../../../../utils/GamePhase.json');
+
+function findTargetCard(targetId, targetType, opponent) {
+  switch (targetType) {
+    case 'point':
```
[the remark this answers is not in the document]
itsalaidbacklife: Yes and it's actually worse we use it 4 times because the non seven one-off validate/execute helpers also duplicate it atm

api/helpers/game-states/find-target-card.js
```diff
@@ -0,0 +1,57 @@
...
+      type: 'ref',
+      descriptions: 'Player object from the current game state',
+      required: true,
+    },
+  },
+  sync: true,
+  fn: ({ targetId, targetType, player }, exits) => {
+    switch (targetType) {
+      case 'point':
+        return exits.success(player.points.find(card => card.id === targetId));
```
Haviles04, line 38: What happens if player is somehow undefined, will this still come through as a success?

api/helpers/game-states/moves/seven-one-off/validate.js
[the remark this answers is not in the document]
itsalaidbacklife: Done

api/helpers/game-states/find-target-card.js
itsalaidbacklife: Yes atm it will succeed with a null value. My thought was to then catch that in the consuming validate helper (theoretically shouldn't happen in execute() because we'd already try in validate).

The idea is that callers do a null check like they do for the `.find()` calls on points etc:

```
          const targetCard = sails.helpers.gameStates.findTargetCard(
            requestedMove.targetId,
            requestedMove.targetType,
            opponent,
          );
          // Must have target
          if (!targetCard) {
            throw new Error(`Can't find the ${requestedMove.targetId} on opponent's board`);
          }
```

That said we could consider throwing in `findTargetCard()` and then we could even remove the `if (!targetCard)` from the consuming validate/execute helpers.

But my instinct is to keep it like this so that the error gets thrown in validate() itself. That way we can rely on `validate()` to be the place to look to see all the errors for any given move

tests/e2e/specs/in-game/one-offs/7_sevens/player_sevens.spec.js
itsalaidbacklife: done

api/helpers/game-states/find-target-card.js
Haviles04: Okay that makes sense, I do think we should explicitly send back null if not found though because `.find` returns undefined if nothing is found.

PR #1128, opened 2024-12-09
Modified the GameListItem.vue component to render the players' usernames in the home page list of games

### LINKED ISSUE #1103: [Feature]: Replace the player counts in the Home Page with the player usernames
### Feature Summary

Each game in the home page’s list of open games show the number of players in it e.g. 0/2, 1/2, 2/2. We should replace this with text describing who specifically is in the game e.g. “vs aleph_one”.

### Detailed Description

There can be 0, 1, or 2 players in a game at any time, so we should update their display to handle each of these cases as follows

### 0 players 
The text should say: **Empty**

### 1 player
The text should say: **vs <username>**
 e.g. **vs aleph_one**
 
### 2 players
The text should say: **<p0_username> vs <p1_username>**
e.g. **aleph_one vs SUBMARINO**

## Implementation
Building this feature will require backend updates to ensure the data is provided, and frontend updates to appropriately consume them, and test updates to describe the changed behavior.

### Backend
There is one endpoint that provides the list of games displayed by the frontend: `api/controllers/game/get-list.js`

This endpoint uses two helpers to generate the delivered data:
* `api/hooks/customGameHook/index.js`’s `findOpenGames()` function
* `api/helpers/find-spectatable-games.js`

These two should be updated so that they both include the `.players` on each game, but with each user transformed to only include the id and username.

#### customeGameHook.findOpenGames()
This function already includes the players as an array on each game because it calls `.populate(‘players’)` when querying for the games on line 39. This queries for all the user records associated with each game and adds them to the `players` array of the respective game.

The issue with this is that it includes *all* data for each user record, when we should instead include only the `id` and `username`. To address this, we should update the processing logic on line 46 so that the returned `openGames` variable has each of its games transformed so that the each user in the game’s `players` array consists only of their `id` and username.

#### findSpectatableGames()
This helper doesn’t yet include the players associated with a game, so we should update it to call `.populate(‘players’)` on the existing query for games, and then use the same technique as above to convert each game so that each of its `players` have only an `id` and `username`.

### Frontend
The frontend requires a change to the `GameListItem` component, which renders a game in the list, and the `gameList` store, which handles processing data for the list of available games.

#### GameListItem Component
The primary frontend change will be in `src/routes/home/components/GameListItem.vue` which is the component that renders an individual game that is available to play or to spectate. We should update it so that it expects the game to have an array of `players` and that it renders the text described above based on the number of players in the game.

#### gameList Store
In addition to the component change, we need to update the pinia store `src/stores/gameList.js` so that it works with the list of players in the game, rather than just the number of players in the game. Specifically, we need to:

1. Update the `GameSummary` constructor to use the `players` array itself instead of creating a `numPlayers` attribute using the length of the players list.
2. Update the `joinGame()` and `otherLeftGame()` attributes to `.push()`  and `.pop()` players from the relevant game’s `.players()` array rather than incrementing/decrementing the player count.

AlAnoud2003:
## Issue number

- Resolves #1103

## Please check the following

- For New Features:

This pull request updates the GameListItem.vue component file in the frontend to ensure that it expects the game to have an array of players and that on the home page's list of open games, it renders text displaying who specifically is in the game (players' usernames) based on the number of players in the game.

In GameListItem.vue:
* Props: Added a ‘players’ prop to the component, which is a required array
* Computed Property: Added a ‘playersText’ computed property to generate the text based on the number of players in the game.
* Template: Updated the template to use the ‘playersText’ computed property to display the text describing who specifically is in the game.

Mock Tests: GameListItem.spec.js
* Created a file titled: GameListItem.spec.js in tests > unit > specs. It imports Vitest functions and mocks the GameListItem.vue component by defining a mock version with props and a render method. It also includes two test cases: 
    * Test Case 1: Checks if the component displays "Empty" when there are no players.
    * Test Case 2: Checks if the component displays player names when there are players.


src/routes/home/components/GameListItem.vue
```diff
@@ -106,11 +98,16 @@ export default {
...
+    playersText() {
+      const numPlayers = this.players.length;
+      if (numPlayers === 0) {
+        return 'Empty';
+      } else if (numPlayers === 1) {
+        return `vs ${this.players[0].username}`;
+      } else if (numPlayers === 2) {
+        return `${this.players[0].username} vs ${this.players[1].username}`;
+      }
+      return '';
```
itsalaidbacklife, line 110: This looks right, but stylistically we generally use switch statements for multiple conditions on the same variable

tests/unit/specs/GameListItem.spec.js
itsalaidbacklife, line 1: Presently we do all our frontend testing end-to-end with cypress. The key test file to update is
```
tests/e2e/specs/out-of-game/home.spec.js
```

In addition to whatever else may fail after this change i.e. anywhere that's currently making assertions about the player text in `GameListItem`.

package.json
itsalaidbacklife, line 1: We shouldn't need any additional dependencies for this PR. The tests should be done with the existing cypress setup, which is how we manage UI assertions. We use vitest for unit testing, which currently is only backend tests. It's not out of the question to add component UI tests for something like this but:

1. If we did so, we should either use cypress component testing or vitest as these are the testing tools already in place
2. We already have fairly solid coverage of the expected text in the e2e suite for the home.spec file and it makes the most sense to continue to test these features there. That way we don't simply confirm that the component accepts and renders props as expected, but that the entire application produces and consumes data correctly i.e. it catches bugs where the props are consistent but the front and backend are misaligned

package.json
```diff
@@ -48,12 +50,14 @@
...
     "eslint-plugin-vue": "9.28.0",
     "eslint-plugin-vuetify": "2.4.0",
+    "jest": "^29.7.0",
     "prettier": "3.3.3",
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "rimraf": "6.0.1",
     "sass": "1.79.4",
     "storybook": "7.6.0",
+    "typescript": "^5.7.2",
```
itsalaidbacklife, line 60: I very much want to get typescript up and running for Cuttle but that is a separate project and I'd prefer to keep updates minimally scoped to the essential changes for a given issue, so lets leave dependency changes out of this PR

itsalaidbacklife (commented):
It looks like this is a work in progress, which is totally fine. I left some initial feedback just to give you a sense of changes that will be necessary as this comes along. A couple points to consider:
1. The frontend and backend will need to be PR'd simultaneously for this to happen, so it should be in one PR. Our CI setup will immediately deploy merged code to production so any given merge needs to be fully functional across the front and back end. It's fine to manage it separately in your own fork if that is helpful to you, but when it comes into the main repo, the entire change needs to happen at once (otherwise we'll see failures in the tests)
2. We do most of our testing end-to-end with cypress, so we'll want to continue to use those tests and update them to match the new behavior. 


All in all it looks like you're off to a good start! Feel free to reach out here or on [discord](https://discord.gg/RHqzRMzds3) if you have any questions.

package.json
itsalaidbacklife, line 1: @AlAnoud2003 you resolved the comments here but it looks like `package.json` is still including dependency changes which are unnecessary for this PR. Please set `package.json` and `package-lock.json` to match `main`

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -97,6 +97,21 @@ describe('Home - Game List', () => {
         .contains('Game made by other player');
     });
 
+    it('Displays "Empty" when there are no players', () => {
+      const players = [];
+      cy.createGamePlayer({ gameName: 'Game with no players', isRanked: false, players });
```
itsalaidbacklife, line 102: `cy.createGamePlayer()` does not accept a `players` argument. It makes a single api call to create the game, at which point there are never any players in it. So we don't need this third argument for this test.
```suggestion
      cy.createGamePlayer({ gameName: 'Game with no players', isRanked: false });
```

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -97,6 +97,21 @@ describe('Home - Game List', () => {
...
+      cy.get('[data-cy=game-list-item]').contains('Empty');
```
itsalaidbacklife, line 103: Let's additionally check that there is only one game
```suggestion
      cy.get('[data-cy=game-list-item]').should('have.length', 1).contains('Empty');
```

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -97,6 +97,21 @@ describe('Home - Game List', () => {
...
+    });
```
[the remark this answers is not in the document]
itsalaidbacklife:
```suggestion
```

PR #1130, opened 2024-12-12
Player array task

Hakaabi:
## Issue number

- Resolves #

## Please check the following

- For New Features:

### DIFF
--- a/api/hooks/customGameHook/index.js
+++ b/api/hooks/customGameHook/index.js
@@ -30,24 +30,37 @@ module.exports = function gameHook() {
     },
     findOpenGames: function () {
       return new Promise(function (resolve, reject) {
-        const recentUpdateThreshhold = dayjs.utc().subtract(1, 'day')
+        const recentUpdateThreshold = dayjs.utc().subtract(1, 'day')
           .toDate();
+    
         Game.find({
           status: gameService.GameStatus.CREATED,
-          createdAt: { '>=': recentUpdateThreshhold },
+          createdAt: { '>=': recentUpdateThreshold },
         })
-          .populate('players')
+          .populate('players') // Populate the players array
           .exec(function (error, games) {
             if (error) {
               return reject(error);
-            } else if (!games) {
+            } else if (!games || games.length === 0) {
               return reject({ message: "Can't find games" });
             }
-            const openGames = games.filter(({ players }) => players.length < 2);
-            return resolve(openGames);
+    
+            // Filter games with fewer than 2 players and transform the players array
+            const openGames = games
+              .filter(({ players }) => players.length < 2) // Keep games with fewer than 2 players
+              .map((game) => {
+                // Transform players array to include only id and username
+                game.players = game.players.map((player) => ({
+                  id: player.id, // Use 'id' since Sails.js generates it automatically
+                  username: player.username,
+                }));
+                return game; // Return the updated game object
+              });
+    
+            return resolve(openGames); // Resolve with transformed open games
           });
       });
-    },
+    },    
     findGame: function (id) {
       return new Promise(function (resolve, reject) {
         Game.findOne(id)
--- a/src/routes/home/components/GameListItem.vue
+++ b/src/routes/home/components/GameListItem.vue
@@ -6,7 +6,7 @@
           {{ name }}
         </p>
         <p v-if="!isSpectatable" class="text-surface-1">
-          {{ readyText }} {{ t('home.players') }}
+          {{ playersText }}
         </p>
       </v-col>
       <v-col lg="6" class="list-item__button pr-md-0">
@@ -81,6 +81,10 @@ export default {
       type: Number,
       required: true,
     },
+    players: {
+      type: Array,
+      required: true,
+    },
     isRanked: {
       type: Boolean,
       default: false,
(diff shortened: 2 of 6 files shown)


api/hooks/customGameHook/index.js
```diff
@@ -30,24 +30,37 @@ module.exports = function gameHook() {
...
+                  username: player.username,
+                }));
+                return game; // Return the updated game object
+              });
+    
+            return resolve(openGames); // Resolve with transformed open games
           });
       });
-    },
+    },
```
itsalaidbacklife, line 63:
```suggestion
    },
```

api/hooks/customGameHook/index.js
```diff
@@ -30,24 +30,37 @@ module.exports = function gameHook() {
...
-          createdAt: { '>=': recentUpdateThreshhold },
+          createdAt: { '>=': recentUpdateThreshold },
         })
-          .populate('players')
+          .populate('players') // Populate the players array
           .exec(function (error, games) {
             if (error) {
               return reject(error);
-            } else if (!games) {
+            } else if (!games || games.length === 0) {
```
itsalaidbacklife, line 44: We can combine these with optional chaining and utilize the fact that `0` is falsey like so
```suggestion
            } else if (!games?.length) {
```

api/hooks/customGameHook/index.js
```diff
@@ -30,24 +30,37 @@ module.exports = function gameHook() {
...
+              .map((game) => {
+                // Transform players array to include only id and username
+                game.players = game.players.map((player) => ({
+                  id: player.id, // Use 'id' since Sails.js generates it automatically
+                  username: player.username,
+                }));
+                return game; // Return the updated game object
+              });
+    
+            return resolve(openGames); // Resolve with transformed open games
```
itsalaidbacklife, line 60: Looks good. I think the above comments are helpful but the ones on the same line are not adding much clarity beyond what the line already says. Let's remove these.

Generally, our approach to commenting is to either:
1. Explain anything weird
3. Break out a larger thing into sections for quick scanning
```suggestion
                  id: player.id,
                  username: player.username,
                }));
                return game;
              });
    
            return resolve(openGames);
```

api/hooks/customGameHook/index.js
```diff
@@ -30,24 +30,37 @@ module.exports = function gameHook() {
...
-            } else if (!games) {
+            } else if (!games || games.length === 0) {
               return reject({ message: "Can't find games" });
             }
-            const openGames = games.filter(({ players }) => players.length < 2);
-            return resolve(openGames);
+    
+            // Filter games with fewer than 2 players and transform the players array
+            const openGames = games
+              .filter(({ players }) => players.length < 2) // Keep games with fewer than 2 players
```
itsalaidbacklife, line 50: I think this comment is sufficiently explained by the one above the `openGames` assignment
```suggestion
              .filter(({ players }) => players.length < 2)
```

src/routes/home/components/GameListItem.vue
```diff
@@ -81,6 +81,10 @@ export default {
       type: Number,
       required: true,
     },
+    players: {
+      type: Array,
+      required: true,
+    },
```
itsalaidbacklife, line 87: We should also remove `numPlayers` as a prop and replace it with a computed property that returns the length of the `players` array. Or alternatively remove it altogether and just check `this.players.length` wherever we need the numPlayers

src/stores/gameList.js
```diff
@@ -8,6 +8,7 @@ class GameSummary {
     this.id = obj.id ? obj.id : null;
     this.name = obj.name ? obj.name : null;
     this.numPlayers = Object.prototype.hasOwnProperty.call(obj, 'players') ? obj.players.length : 0;
+    this.players = Object.prototype.hasOwnProperty.call(obj, 'players') ? obj.players : [];
```
itsalaidbacklife, line 11: Let's also remove `numPlayers` above since its just computed from `players.length`

itsalaidbacklife: We will additionally need to update the places `numPlayers` are referenced and remove them accordingly

src/stores/gameList.js
```diff
@@ -53,13 +54,15 @@ export const useGameListStore = defineStore('gameList', {
...
         updatedGame.numPlayers = Math.min(2, updatedGame.numPlayers + 1);
+        updatedGame.players.push(data.newPlayer);
         updatedGame.status = data.newStatus;
       }
     },
     otherLeftGame(gameId) {
       const updatedGame = this.openGames.find((game) => game.id === gameId);
       if (updatedGame) {
         updatedGame.numPlayers = Math.max(0, updatedGame.numPlayers - 1);
+        updatedGame.players.pop();
```
itsalaidbacklife, line 65: This is actually a little trickier than you have it here and will require a full stack change. Specifically, we don't know which player left, so we can't just assume it was the one at the end of the array. We need to update
```
api/controllers/game/leave-lobby.js
```

So that its `sails.sockets.blast('leftGame', { id: values[0].id }, req);` call additionally specifies _which user id left_ and then we need to update this `otherLeftGame` method to look up which specific user left the game and remove that one. It could look like this

```
    otherLeftGame(gameId, userId) {
      if (updatedGame) {
        updatedGame.players. = updatedGame.players.filter((player) => player.id !== userId);
      }
  }
```

itsalaidbacklife (commented):
Overall this looks right. I've commented with some change requests. Two things to consider:
1. This PR, along with #1128 and all the other changes necessary to complete #1103 need to be consolidated into a **single PR** from your fork into the original repo. This is because merging a PR into the `main` branch of the original repo immediately builds and deploys the result to production _and the separate pieces will not work if merged individually_. I fully understand wanting to work in small chunks, and also to get feedback along the way, so I am happy to review these individually as I have been doing, but it's important to understand that all of your team's changes need to be merged into a single branch which is PR'd into the original repo, or they can't be used.
2. The `otherLeftGame` update is a little more complicated than what you've done so far because we need to look up which specific user left the game and therefore to update the backend to send that information in the socket event so that the frontend can remove the right user. Details on how to do this are in my comments.

Overall this is solid work that gets us close to what we need

PR #1140, opened 2024-12-30
Feat #1031 Usercontroller.relogin gamestate rework

### LINKED ISSUE #1031: [Feature]: Update `UserController.reLogin` to function in the GameState API

Haviles04:
## Issue number

- Resolves #1031

## Please check the following

- For New Features:

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -63,6 +63,27 @@ module.exports = {
       }
       // Query for game if user is in one
       const gameId = (user.game ?? req.session.game) ?? null;
+
+      if (process.env.VITE_USE_GAMESTATE_API) {
+        const { unpackGamestate, createSocketEvent } = sails.helpers.gameStates;
+        const game = await Game.findOne(gameId).populate('gameStates');
+        const gameObject = await unpackGamestate(game.gameStates.at(-1));
+        const socketEvent = await createSocketEvent(game, gameObject);
+        
+        Game.subscribe(req, [ game.id ]);
+        req.session.loggedIn = true;
+        req.session.usr = user.id;
+        req.session.game = game.id;
+        req.session.pNum = user.pNum ?? undefined;
+        Game.publish([ game.id ],socketEvent);
+
+        return res.ok({
+          game: socketEvent.game,
+          username: user.username,
+          pNum: user.pNum,
+        });
+      }
+      // Remove everything between here and catch AFTER gamestate is deployed
       const unpopulatedGame = gameId ? await gameService.findGame({ gameId }) : null;
       const populatedGame =
         unpopulatedGame?.status === GameStatus.STARTED
@@ -97,6 +118,7 @@ module.exports = {
         username: user.username,
         pNum: user.pNum,
       });
+      
     } catch (err) {
       return res.badRequest(err);
     }
--- a/tests/e2e/specs/in-game/reconnecting.spec.js
+++ b/tests/e2e/specs/in-game/reconnecting.spec.js
@@ -4,7 +4,6 @@ import { Card } from '../../fixtures/cards';
 
 describe('Reconnecting to a game', () => {
   it('Refreshs Game from Game Menu', () => {
-    cy.skipOnGameStateApi();
     cy.setupGameAsP1();
     cy.loadGameFixture(1, {
       p0Hand: [ Card.ACE_OF_CLUBS, Card.SEVEN_OF_DIAMONDS ],
@@ -44,7 +43,6 @@ describe('Reconnecting to a game', () => {
   });
 
   it('Persists session after refreshing the page', () => {
-    cy.skipOnGameStateApi();
     cy.setupGameAsP0();
 
     cy.loadGameFixture(0, {
@@ -74,7 +72,6 @@ describe('Reconnecting to a game', () => {
   });
 
   it('Reconnects after refreshing the page', () => {
-    cy.skipOnGameStateApi();
     cy.setupGameAsP0();
 
     cy.loadGameFixture(0, {
@@ -105,7 +102,6 @@ describe('Reconnecting to a game', () => {
 
   describe('Reconnecting into Cannot Counter Dialog', () => {
     it('oneOff - Reconnect into cannot counter dialog', () => {
-      cy.skipOnGameStateApi();
       cy.setupGameAsP1();
 
       cy.loadGameFixture(1, {
@@ -141,7 +137,6 @@ describe('Reconnecting to a game', () => {
     });
 
     it('targetedOneOff -- reconnect into cannot counter dialog', () => {
-      cy.skipOnGameStateApi();
       cy.setupGameAsP1();
 
       cy.loadGameFixture(1, {
@@ -177,7 +172,6 @@ describe('Reconnecting to a game', () => {
     });
 
     it('counter -- Reconnect into cannot counter dialog', () => {
-      cy.skipOnGameStateApi();
       cy.setupGameAsP0();
       cy.loadGameFixture(0, {
         p0Hand: [ Card.ACE_OF_CLUBS ],
@@ -215,7 +209,6 @@ describe('Reconnecting to a game', () => {
     });
 
     it('sevenOneOff -- Reconnect into cannot counter dialog', () => {
-      cy.skipOnGameStateApi();
       cy.setupGameAsP1();
       cy.loadGameFixture(1, {
         p0Hand: [ Card.SEVEN_OF_CLUBS ],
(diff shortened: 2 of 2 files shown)


api/controllers/UserController.js
```diff
@@ -63,6 +63,27 @@ module.exports = {
...
+        req.session.pNum = user.pNum ?? undefined;
+        Game.publish([ game.id ],socketEvent);
+
+        return res.ok({
+          game: socketEvent.game,
+          username: user.username,
+          pNum: user.pNum,
+        });
+      }
+      // Remove everything between here and catch AFTER gamestate is deployed
```
itsalaidbacklife, line 86:
```suggestion
      // FIXME: #965
      // Remove everything between here and catch AFTER gamestate is deployed
```

api/controllers/UserController.js
```diff
@@ -63,6 +63,27 @@ module.exports = {
...
       const gameId = (user.game ?? req.session.game) ?? null;
+
+      if (process.env.VITE_USE_GAMESTATE_API) {
+        const { unpackGamestate, createSocketEvent } = sails.helpers.gameStates;
+        const game = await Game.findOne(gameId).populate('gameStates');
+        const gameObject = await unpackGamestate(game.gameStates.at(-1));
+        const socketEvent = await createSocketEvent(game, gameObject);
+        
+        Game.subscribe(req, [ game.id ]);
+        req.session.loggedIn = true;
```
itsalaidbacklife, line 74: Are we still using session.loggedIn? I thought we had switched to only reading session.usr instead

api/controllers/UserController.js
```diff
@@ -63,6 +63,27 @@ module.exports = {
...
+        req.session.usr = user.id;
+        req.session.game = game.id;
+        req.session.pNum = user.pNum ?? undefined;
```
itsalaidbacklife, line 77:
```suggestion
        // FIXME: #965 - remove game and pNum
        req.session.game = game.id;
        req.session.pNum = user.pNum ?? undefined;
```

itsalaidbacklife (approved):

api/controllers/UserController.js
```diff
@@ -63,6 +63,30 @@ module.exports = {
       }
       // Query for game if user is in one
       const gameId = (user.game ?? req.session.game) ?? null;
+
+      if (process.env.VITE_USE_GAMESTATE_API) {
+        const { unpackGamestate, createSocketEvent } = sails.helpers.gameStates;
+        const game = await Game.findOne(gameId).populate('gameStates')
+          .populate('p0')
+          .populate('p1');
+        const gameObject = await unpackGamestate(game.gameStates.at(-1));
```
itsalaidbacklife, line 72: We need to populate the gameStates for them to be available for the createSocketEvent call. Additionally we should handle the case where there are no gameStates because you're in the lobby

Haviles04: It's being populated on line 69, or do we need to do something different?

itsalaidbacklife: Ah silly me I just missed it because the other two were on new lines and I reviewed on my phone. This is right. The only thing is we should handle the case where there are no gamestates yet ie toy refresh from the lobby

Haviles04: :) working on that at the moment

tests/e2e/specs/in-game/reconnecting.spec.js
```diff
@@ -1029,31 +1003,30 @@ describe('Reconnecting after game is over', () => {
...
-    cy.skipOnGameStateApi();
     cy.loadGameFixture(0, {
       p0Hand: [ Card.SIX_OF_DIAMONDS, Card.SEVEN_OF_SPADES ],
       p0Points: [ Card.SEVEN_OF_HEARTS, Card.SEVEN_OF_DIAMONDS ],
       p0FaceCards: [],
       p1Hand: [],
       p1Points: [],
       p1FaceCards: [],
-      topCard: Card.SEVEN_OF_SPADES,
+      topCard: Card.SEVEN_OF_CLUBS,
```
Haviles04, line 1021: fixing duplicate card

api/helpers/game-states/create-socket-event.js
```diff
@@ -14,19 +15,29 @@ module.exports = {
     gameState: {
       type: 'ref',
       description: 'Game state object',
-      required: true,
     },
   },
 
   fn: async function ({ game, gameState }, exits) {
     try {
+      if (!game.status === GameStatus.CREATED && !gameState) {
```
Haviles04, line 23: added this check here so we can remove the required attribute and still use this helper if the user is not yet in a game

api/helpers/game-states/create-socket-event.js
```diff
@@ -14,19 +15,29 @@ module.exports = {
...
-      const p0 = { ...game.p0, ...gameState.p0 };
+      const p0 = { ...game.p0, ...gameState?.p0 };
       delete p0.encryptedPassword;
-      const p1 = { ...game.p1, ...gameState.p1 };
+      const p1 = { ...game.p1, ...gameState?.p1 };
       delete p1.encryptedPassword;
       const players = [ p0, p1 ];
 
+      // Early return if not in game yet
+      if (!gameState) {
```
Haviles04, line 34: Early return here, since the game table has almost everything the client needs for the lobby, adding players to add the players to the game store.


wondering if we should make a new variable called something like `gameCreated` and use that in the spot above and here

api/controllers/UserController.js
```diff
@@ -63,6 +63,32 @@ module.exports = {
...
+        const socketEvent = await createSocketEvent(game, gameObject);
+        
+        Game.subscribe(req, [ game.id ]);
+        req.session.usr = user.id;
+        // FIXME: #965 - remove game and pNum
+        req.session.game = game.id;
+        req.session.pNum = user.pNum ?? undefined;
+        Game.publish([ game.id ],socketEvent);
+
+        const pNum = game.p0?.id === user.id ? 0 : 1;
```
Haviles04, line 83: Making sure the pnum is correct based on the game.

tests/e2e/specs/in-game/game-over/winConditions.spec.js
```diff
@@ -401,6 +401,7 @@ describe('Stalemates', () => {
     });
 
     it('Player requests stalemate, then reloads before opponent accepts', () => {
+      cy.skipOnGameStateApi();
```
Haviles04, line 404: Should be removed in #1146

api/helpers/game-states/create-socket-event.js
itsalaidbacklife: This feels strange. Why would we call this method if there is no gamestate? And why restrict it to only games that are started but not finished ones?

Haviles04: This is so we can use this method for games not started yet in the lobby, this check basically makes sure that there is a gamestate available for every other time this will be called besides when we are in a lobby

Haviles04: It deff should be !== though 
```suggestion
      if (game.status !== GameStatus.CREATED && !gameState) {
```

itsalaidbacklife (approved):

PR #1144, opened 2025-01-03
Feat: Add env support to remove hardcoding the prod urls

seriouslysean:
## Issue number

- Resolves #

## Please check the following

- For New Features:


index.html
```diff
@@ -4,28 +4,28 @@
     <meta charset="utf-8" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width,initial-scale=1.0" />
-    <link rel="icon" href="/favicon.ico" />
+    <link rel="icon" href="%VITE_APP_URL%/favicon.ico" />
```
seriouslysean, line 7: @Haviles04 I'm looking at you, shame! You hardcoded all these productions urls, hahaha. @itsalaidbacklife we should get rid of these and switch to this env method. @jerryh-jr will need it in his unhead PR anyway.

.env
```diff
@@ -1,3 +1,4 @@
+VITE_APP_URL=http://localhost:8080
```
seriouslysean, line 1: I probably missed some, someone really double check me. @Haviles04 @itsalaidbacklife

index.html
seriouslysean: Actually, Mal did it here, https://github.com/cuttle-cards/cuttle/commit/7826a8327d784359a8344d18812fd93cb3f54edd, but no one fixed it hahahaha.

.env
itsalaidbacklife: Additional files using localhost:
* `vite.config.mjs`
* `config/env/staging.js`

Places I'm unsure whether we want to change the hard coded references:
* tests - I'd think we want to keep the tests commands e.g. `rankedMatches.spec.js` to use localhost in `cy.request()` to avoid accidentally testing against prod and generating bs data
* `src/plugins/sails.js` - fallback in case of missing env; seems reasonable to keep it

index.html
seriouslysean: C'mon, bro, read the docs https://vite.dev/guide/env-and-mode.html#html-env-replacement 😏

.env
seriouslysean: We should not change `HOST_SERVER_URL` in `vite.config` as it's used for the docker version, and server only. Same for `config/env/staging.js`, they should stay hardcoded since it's based on hard url location. I also think tests are fine staying hardcoded for that same reason.

index.html
itsalaidbacklife: I mean touché but given that we have 4 different tools for rendering dynamic html between Vue, vite, sails and ejs, it's not always clear _which_ docs to read when seeing a new templating syntax 🤷

Haviles04: *puts hands up* it wasn't me!

.env
Haviles04: It's slightly confusing that this is the only place that `VITE_API_URL` and `VITE_APP_URL` differ

seriouslysean: It's very confusing but has to do with where Cuttle needs these values. The server only ever cares about 1337. The client cares about both.

That's why I said I may have missed something, but this all works from my testing.

Issue #929, opened 2024-05-30
[Feature]: Add robots.txt and add a canonical link to the Signup page


itsalaidbacklife:
### Feature Summary

We should add a robots.txt file to tell SEO crawlers how to navigate the site

### Detailed Description

The site only has two pages navigatable without authenticating: 
/signup
/rules

We should also add a `<link>` that specifies `/login` as canonically equivalent to /signup

Shubh-Dev:
Hi, Please assign me this task.

itsalaidbacklife:
> Hi, Please assign me this task.

Hi @Shubh-Dev, thank you. Issues can't be officially assigned outside the organization, but you are totally welcome to pick this one up. Have at it!

Shubh-Dev:
@itsalaidbacklife Thank you so much, I am on it.

jerryh-jr:
@itsalaidbacklife I am going to work on this.

casens5:
@itsalaidbacklife this feature appears to be resolved?

PR #1145, opened 2025-01-04
Feat 929 add robot.txt file

jerryh-jr:
Relevant [issue number ](929)

Relevant [[Feature]: Add robots.txt and add a canonical link to the Signup page](https://github.com/cuttle-cards/cuttle/issues/929#event-15795204400)
- Resolves #

## Please check the following

- For New Features:

Verified link shows in console window for both signup/login page.


jerryh-jr:
Will fix lint issues and update my branch.

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -682,3 +682,11 @@ describe('Announcement Dialogs', () => {
     cy.get('[data-cy=announcement-dialog]').should('not.exist');
   });
 });
+
+// Check for canonical link in head
+describe('Canonical Link', () => {
```
itsalaidbacklife, line 687: Let's move the test into `tests/e2e/specs/out-of-game/authentication.spec.js` in the existing describe block for 'page contents'

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -682,3 +682,11 @@ describe('Announcement Dialogs', () => {
...
+  it('Checks for the canonical link in the head meta data on login/signup page', () => {
+    cy.visit('/login');
+    cy.get('head link[rel="canonical"]').should('have.attr', 'href').and('contain', 'https://www.cuttle.cards/signup');
```
itsalaidbacklife, line 690: To wrangle nuance around localhost vs production, let's just make it check for the end of the url without the domain 
```suggestion
    cy.get('head link[rel="canonical"]').should('have.attr', 'href').and('contain', '/signup');
```

src/main.js
```diff
@@ -30,4 +35,7 @@ app.use(i18n);
 // Add Cuttle window object
 initCuttleGlobals(app);
 
+// Add unHead to vue
+app.mixin(VueHeadMixin);
```
itsalaidbacklife, line 39: We are slowly switching from the options API to composition, so instead of using this mixin, let's  use the composable version. Here are the [docs](https://unhead.unjs.io/setup/vue/installation#demos).

Basically we want to have the signup page call `useHead()` instead of this.head. That require s putting the useHead() call inside the setup() block of LoginView

itsalaidbacklife (commented):
Looks good. Requesting a few tweaks

public/robots.txt
```diff
@@ -0,0 +1,6 @@
+User-agent: *
+Allow: /signup
+Allow: /rules
+Disallow: /stats
+Disallow: /lobby/*
+Disallow: /game/*
```
seriouslysean, line 6: I think you need a trailing return in this file.

src/main.js
seriouslysean: I agree with this 👍🏼

public/robots.txt
jerryh-jr: [Googles](https://developers.google.com/search/docs/crawling-indexing/robots/create-robots-txt) docs does not show a return. I will research this more and verify.

Thanks!

itsalaidbacklife: @jerryh-jr it's not a specific thing for robots.txt but a general linting rule we have for the project that all files end with a trailing newline. That's also what the red circle symbol 🚫 in github is indicating; lack of newline ad space bad of file

seriouslysean: Yeah, my point was he might need one, or if google doesn’t want it to add an exception to the rule. Thanks @itsalaidbacklife 👍🏼

tests/e2e/specs/out-of-game/home.spec.js
jerryh-jr: Hey @itsalaidbacklife I have moved the location of the test and shorted the condition to only the route and not the entire url. [commit](https://github.com/cuttle-cards/cuttle/pull/1145/commits/3d1f075eba5e6e6b19d80f0a9dfb9bc6cbafa3f8)

src/main.js
jerryh-jr: I have switched unhead from using the options api to composable. Moving the useHead() call into the setup() and using a computed getter to get it to work.

[Here](https://github.com/cuttle-cards/cuttle/pull/1145/commits/7edb50266b01fdc6fa4b56043691c24c99d8fb5a)

src/main.js
```diff
@@ -6,10 +6,15 @@ import router from '@/router';
 import i18n from '@/i18n';
 import { initCuttleGlobals } from '_/utils/config-utils';
 import App from '@/App.vue';
+import { createHead } from '@unhead/vue';
```
itsalaidbacklife, line 9: Please move this import up higher, so it goes with the other external imports. It can go right under i18n

src/routes/login/LoginView.vue
```diff
@@ -185,6 +185,7 @@ import BaseSnackbar from '@/components/BaseSnackbar.vue';
 import MarkdownContent from '@/components/MarkdownContent.vue';
 import BaseVideo from '@/components/BaseVideo.vue';
 import TheLanguageSelector from '@/components/TheLanguageSelector.vue';
+import { useHead } from '@unhead/vue';
```
itsalaidbacklife, line 188: Please move the import above the internal ones so it goes with the external deps

itsalaidbacklife (approved):
Looks good! Just nitpicking import order then it's ready

src/main.js
seriouslysean: Seems like a follow up issue to add sort imports as an eslint rule :) https://eslint.org/docs/latest/rules/sort-imports

src/routes/login/LoginView.vue
jerryh-jr: Hey @itsalaidbacklife I moved the import above the internal imports. Here is the [link](https://github.com/cuttle-cards/cuttle/pull/1145/commits/48d0a2b74769d1349504ff790ff284668be7a75b)

src/main.js
jerryh-jr: Hey @itsalaidbacklife I have updated this import as well to be under the i18n import. Here is the [link](https://github.com/cuttle-cards/cuttle/pull/1145/commits/86f6bc115aa620ac281bc9f6e9125a3f78d3efb5).

PR #1156, opened 2025-01-23
feat:spectate tab url parameter

### LINKED ISSUE #1154: [Feature]: HomeView - control play/spectate tab choice url path parameter
### Feature Summary

We should revise how the HomeView tracks whether you’re viewing the list of open games or spectatable games so that it’s managed via a path param. clicking the tab to view the list of spectatable games should update the url to /spectate-list so that when a user is looking at the games to spectate and they refresh the page, they “stay on” the list of spectateable games instead of defaulting back to the open/playable games 

### Detailed Description

Currently the “Play” and “Spectate” tabs of the HomeView only track state via a ref in the HomeView and so if you click “spectate”, then refresh the page, the page switches back to the “Play” tab. We should fix this so that users can refresh the page and stay viewing whichever list of games they were previously looking at by tracking this state via the url.

We could either do this with a router-view by making the area displaying the list of games one of two sub-pages, or by using the HomeView’s created() hook to initialize the existing variable from the url and then using a watcher to update the url when the variable updates

haaarsh619:
## Issue number #1154

- Resolves #1154

## Please check the following

- For New Features:


src/components/TheHeader.vue
```diff
@@ -83,6 +83,11 @@ const linkColor = computed(() => variant.value === 'light' ? 'text-surface-1' :
 const logoColor = computed(() => variant.value === 'light' ? 'brown' : 'white');
 
 const tabColor = (page) => {
+
+  if(route.name === 'SpectateList') {
+    return page === 'Home' ? 'text-newPrimary' : linkColor.value;
+  }
```
itsalaidbacklife, line 89: This feels a bit awkward, though I see why you're doing it. I think we may be stuck with this if we go the separate routes approach.

src/routes/home/HomeView.vue
```diff
@@ -55,7 +55,7 @@
               </v-row>
             </div>
             <v-window v-else v-model="tab" class="pa-4 overflow-y-auto">
-              <v-window-item :value="TABS.PLAY">
+              <v-window-item>
```
itsalaidbacklife, line 58: By itself I wouldn't expect this to work, as I don't see anything that relatest he route `/spectate-list` to telling this component which list to render. To check, we should update the test

```
it('Does not show open or completed games in spectate tab')
```
in `home.spec` so that on line 306, instead of `cy.visit('/')`  and `cy.get('[data-cy-game-list-selector=spectate]').click();`, we call `cy.reload()` then confirm we are on the same list of games (by checking which tab is highlighted).

Likewise on the next test 
```
it('Shows ongoing games as available to spectate when user navigates to home page')
```
we should add a `cy.reload()` after navigating to the list of spectatable games to confirm that we stay on the list of spectatable games and can still join one.

src/routes/home/HomeView.vue
```diff
@@ -154,11 +154,6 @@ import HowItWorksDialog from '@/routes/home/components/HowItWorksDialog.vue';
 import GameStatus from '_/utils/GameStatus.json';
 import AnnouncementDialog from './components/announcementDialog/AnnouncementDialog.vue';
 
-const TABS = {
-  PLAY: 'play',
-  SPECTATE: 'spectate',
-};
```
itsalaidbacklife, line 160: I'd prefer to keep this enum for clarity that these are the only expected tab values. We can revise their values to match the values you're using currently like
```
const tabs = {
  PLAY: '/',
  SPECTATE: '/spectate-list'
};
```

src/routes/home/HomeView.vue
```diff
@@ -211,8 +208,13 @@ export default {
   async created() {
     await this.gameListStore.requestGameList();
     this.loadingData = false;
+    this.tab = this.$route.path;
   },
   methods: {
+    gameListWindow(url){
+      this.$router.replace(url);
```
itsalaidbacklife, line 215: An awkward thing about using `.replace()` instead of `.push()` is that using the back button won't bring you back to the previous page. If we use `$router.push()` is the rest of the UX significantly different?

itsalaidbacklife (commented):
Looking good. A couple nitpicks

src/routes/home/HomeView.vue
```diff
@@ -198,21 +198,29 @@ export default {
...
       }
     }
   },
   async created() {
     await this.gameListStore.requestGameList();
     this.loadingData = false;
+    this.tab = this.$route.path;
   },
   methods: {
+    gameListWindow(url){
```
itsalaidbacklife, line 220: Let's rename this function to `setGameListUrl()`

itsalaidbacklife (commented):
Looks good. Two last chores and it's ready to merge

src/routes/home/HomeView.vue
```diff
@@ -198,21 +198,29 @@ export default {
...
+    tab(newVal){
+      this.setGameListUrl(newVal);
+    },
     $route: {
       immediate: true,
       handler() {
         if (this.$route.query?.error) {
           this.handleSubscribeError(Number(this.$route.query.gameId), this.t(this.$route.query.error));
-          this.$router.replace('/');
+          this.$router.push('/');
```
itsalaidbacklife, line 209: This can stay replace
```suggestion
          this.$router.replace('/');
```

itsalaidbacklife (approved):
Great work

itsalaidbacklife (approved):
Great work

PR #1135, opened 2024-12-25
Feat/spectate gamestate

itsalaidbacklife:
## Issue number

Relevant [issue number](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
- Resolves #1126 

## Please check the following

- [ ] Do the tests still pass? (see [Run the Tests](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#run-the-tests))
- [ ] Is the code formatted properly? (see [Linting (Formatting)](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#linting-formatting))
- For New Features:
  - [ ] Have tests been added to cover any new features or fixes?
  - [ ] Has the documentation been updated accordingly?

PR #1161, opened 2025-01-26
Fix/rollbar crash

### LINKED ISSUE #1129: [Bug]: Tab becomes unresponsive when hitting "Go Home" after spectating game
### Bug Summary

After spectating a game and going home, the tab freezes and becomes unresponsive, failing to load the list of games

### Detailed Description

Spectating a game, then going home from the GameOverDialog causes the tab to hang and the game list to fail to load

### Current Behavior

The game list fails to load and the tab freezes

### Expected Behavior

The game list loads and the tab functions as normal

### Steps to Reproduce

1. Log in
2. Spectate a game
3. Wait for game to be finished
4. Hit "Go Home"
5. Observe that game list doesn't load and page hangs

itsalaidbacklife:
Fixes 2 bugs:
1. `rollbar` was crashing the app when logging some errors due to poor serialization of the `vm`
2. request to leave as spectator had wrong url, causing error which then crashed the app due to the above

Fix for the `rollbar` error was to update the serialization of `vm` so that we grab specific data we want included, rather than the entire component instance.

Fix for the spectator error was to use the correct url.
## Issue number

- Resolves #1129 

## Please check the following

- For New Features:


src/main.js
itsalaidbacklife, line 1: We can re-enable logging now that we've fixed the cause of the crash

src/plugins/rollbar.js
```diff
@@ -36,10 +36,20 @@ const rollbar = new Rollbar(config);
 export default {
   install(app) {
     app.config.errorHandler = (error, vm, info) => {
-      rollbar.error(error, { vueComponent: vm, info });
```
itsalaidbacklife, line 39: `{ vueComponent: vm }` was the root cause of the crash. Despite literally coming straight out of the [docs](https://docs.rollbar.com/docs/vue-js), attempting to serialize the entire component instance would consistently cause rollbar to crash the entire application with an out of memory error.

src/plugins/rollbar.js
```diff
@@ -36,10 +36,20 @@ const rollbar = new Rollbar(config);
...
         console.error(error);
       }
+
+      const vueComponent = {
+        name: vm.$options?.name ?? 'AnonymousComponent',
+        props: vm.$props ?? {},
+        data: vm.$data ?? {},
+        attributes: vm.$attributes ?? {},
+        route: vm.$route ?? {},
+      };
```
itsalaidbacklife, line 50: The fix is simply to hand-pick the data we want off the `vm`. This seems a good starting point for understanding the context of an error, and it falls back where data is missing

src/stores/game.js
```diff
@@ -462,7 +462,7 @@ export const useGameStore = defineStore('game', {
     },
     async requestSpectateLeave() {
       return new Promise((resolve, reject) => {
-        io.socket.get('/api/game/spectateLeave', (res, jwres) => {
+        io.socket.post(`/api/game/${this.id}/spectate/leave`, (_res, jwres) => {
```
itsalaidbacklife, line 465: This was the cause of the error that was being logged and therefore crashing the app. Url for this endpoint was switched in #1135

src/plugins/rollbar.js
itsalaidbacklife: See this issue in rollbar: https://github.com/rollbar/rollbar.js/issues/1126#issuecomment-1829327874

api/controllers/TestController.js
```diff
@@ -95,4 +95,21 @@ module.exports = {
...
+          ...spectatingRecord,
+          spectator: spectatingRecord.spectator.username
+        };
+      });
+
+      return res.json(result);
+    } catch (err) {
+      return res.serverError(err);
+    }
+  }
```
itsalaidbacklife, line 114: Added this to make it easier to test whether spectators are inactivated when they leave (this was failing because `stores/game.js` had the wrong url for the spectate/leave endpoint). In addition to fixing the behavior, we should test that it works. This way we can get all the spectators, assert how many there are, and that they have the correct date

src/main.js
```diff
@@ -18,7 +18,7 @@ const head = createHead();
 // Add rollbar to vue
 if (import.meta.env.VITE_ROLLBAR_ACCESS_TOKEN) {
   // TODO #1129 - re-enable rollbar when it doesn't crash the tab
-  // app.use(rollbar);
+  app.use(rollbar);
```
itsalaidbacklife, line 21: Able to turn this on now that the error logging works

PR #1167, opened 2025-02-18
added ability to click enter on rank toggle

nick3v:
Issue #1166

Relevant [#1166](https://github.com/cuttle-cards/cuttle/issues/1166)
- Resolves #

## Please check the following

- [ ] Do the tests still pass? (see [Run the Tests](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#run-the-tests))
- [x] Is the code formatted properly? (see [Linting (Formatting)](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#linting-formatting))
- For New Features:
  - [ ] Have tests been added to cover any new features or fixes?
  - [ ] Has the documentation been updated accordingly?

## Please describe additional details for testing this change
I saw another pull request for this but already had this solution done a little differently. Same issue regarding local tests not passing and I cannot see a reason why it is. First attempted contribution so this is a learning process for me. Please let me know what I can improve upon. Thanks.

PR #1168, opened 2025-02-18
Enter togggle ranked/1166

### LINKED ISSUE #1166: [Bug]: Can't change from ranked/casual setting on games via keyboard
### Bug Summary

You can focus the swtich with 'tab' but hitting 'enter' doesn't actually trigger it like a click

### Detailed Description

See the switch in the `LobbyView`:
Image

We can successfully bring the switch into focus with 'tab', but hitting 'enter' should then trigger the switch the same as clicking it. This should switch the game from ranked to casual (or back). 

We should update the test case in the lobby.spec `it('Changes games to ranked and casual from the lobby')` so that it additionally tests hitting 'enter' while the switch is in focus.

### Current Behavior

Hitting 'enter' while the switch is in focus does not switch the game between ranked and casual.

### Expected Behavior

Hitting 'enter/return' while the switch is in focus should switch the game between ranked and casual.

### Steps to Reproduce

1. Sign up/log in
2. Create or join a game
3. Hit tab until the switch for ranked/casual is in focus
4. Hit 'enter'/'return'

abebanos:
## Issue number

- Resolves #1166

## Please check the following

- [ x] Is the code formatted properly? (see [Linting (Formatting)](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#linting-formatting))
- For New Features:
  - [x ] Have tests been added to cover any new features or fixes?

### DIFF
--- a/src/routes/lobby/LobbyView.vue
+++ b/src/routes/lobby/LobbyView.vue
@@ -60,6 +60,8 @@
                   color="primary"
                   hide-details
                   @update:model-value="setIsRanked"
+                  @keydown.enter="toggleRanked"
+                  tabindex="0"
                 />
                 <v-icon
                   class="mr-2 mr-md-4"
@@ -148,6 +150,11 @@ async function setIsRanked() {
   await gameStore.requestSetIsRanked({
     isRanked: gameStore.isRanked,
   });
+}
+
+async function toggleRanked() {
+   gameStore.isRanked = !gameStore.isRanked;
+   setIsRanked();
 }
 
 async function leave() {
--- a/src/routes/lobby/LobbyView.vue
+++ b/src/routes/lobby/LobbyView.vue
@@ -59,9 +59,9 @@
                   data-cy="edit-game-ranked-switch"
                   color="primary"
                   hide-details
+                  tabindex="0"
                   @update:model-value="setIsRanked"
                   @keydown.enter="toggleRanked"
-                  tabindex="0"
                 />
                 <v-icon
                   class="mr-2 mr-md-4"
@@ -153,8 +153,8 @@ async function setIsRanked() {
 }
 
 async function toggleRanked() {
-   gameStore.isRanked = !gameStore.isRanked;
-   setIsRanked();
+  gameStore.isRanked = !gameStore.isRanked;
+  setIsRanked();
 }
 
 async function leave() {
--- a/tests/e2e/specs/out-of-game/lobby.spec.js
+++ b/tests/e2e/specs/out-of-game/lobby.spec.js
@@ -94,6 +94,32 @@ describe('Lobby - Page Content (Ranked)', () => {
     checkRanked(true);
     cy.get('[data-cy=ready-button-sword-cross-icon]').should('exist');
   });
+
+  it('Toggles ranked/casual switch with "Enter" key', () => {
+    // Focus the ranked/casual switch
+    cy.get('[data-cy=edit-game-ranked-switch]').focus();
+
+    // Check the initial state (ranked mode)
+    checkRanked(true);
+    cy.get('[data-cy=ready-button-sword-cross-icon]').should('exist');
+
+    // Simulate pressing the 'Enter' key
+    cy.get('[data-cy=edit-game-ranked-switch]').type('{enter}');
+
+    // Verify the switch toggles to casual mode
+    checkRanked(false);
+    cy.get('[data-cy=ready-button-coffee-icon]').should('exist');
+    cy.contains('Game Mode changed to').should('exist');
+    cy.get('[data-cy="close-snackbar"]').click();
+
+    // Simulate pressing the 'Enter' key again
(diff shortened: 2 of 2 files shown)


abebanos:
Manual testing was done and the issue works however, I was unable to get cypress to work properly as just running the e2e test after starting the dev server still leads in test failing before my changes.

Haviles04:
> Manual testing was done and the issue works however, I was unable to get cypress to work properly as just running the e2e test after starting the dev server still leads in test failing before my changes.

Which test is failing and what does it say? Are you on windows or Mac?

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -94,6 +94,32 @@ describe('Lobby - Page Content (Ranked)', () => {
     checkRanked(true);
     cy.get('[data-cy=ready-button-sword-cross-icon]').should('exist');
   });
+
+  it('Toggles ranked/casual switch with "Enter" key', () => {
+    // Focus the ranked/casual switch
+    cy.get('[data-cy=edit-game-ranked-switch]').focus();
```
itsalaidbacklife, line 100: Good idea, though focusing the switch is a little more nuanced, unfortunately, due to the fact that cypress can only focus the input element _inside_ the switch, and that the data-cy attribute gets put on a containing div, not the `input` itself. Try this
```suggestion
    cy.get('[data-cy=edit-game-ranked-switch] input').focus();
```

to select the descendant `<input>` element of the div with the chosen selector.

FOr reference, here is the [cypress cloud run showing the test failure](https://cloud.cypress.io/projects/i8bxr8/runs/2632/overview/38e29299-ad4c-405c-9745-67dbff1fd536/replay?roarHideRunsWithDiffGroupsAndTags=1&runStatus=FAILED&runStatus=RUNNING&runStatus=CANCELLED&runStatus=ERRORED&runStatus=OVERLIMIT&runStatus=PASSED&ts=1739900551207.3&att=1).

You can see the test runs in the "Checks" tab of the PR, and then at the bottom of the logs for a given test run you can find the link to its cypress cloud run, which will let you replay all the tests interactively:
image

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -94,6 +94,32 @@ describe('Lobby - Page Content (Ranked)', () => {
...
+
+    // Check the initial state (ranked mode)
+    checkRanked(true);
+    cy.get('[data-cy=ready-button-sword-cross-icon]').should('exist');
+
+    // Simulate pressing the 'Enter' key
+    cy.get('[data-cy=edit-game-ranked-switch]').type('{enter}');
```
itsalaidbacklife, line 107: Same idea for all the references 
```suggestion
    cy.get('[data-cy=edit-game-ranked-switch] input').type('{enter}');
```

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -94,6 +94,32 @@ describe('Lobby - Page Content (Ranked)', () => {
...
+
+    // Verify the switch toggles to casual mode
+    checkRanked(false);
+    cy.get('[data-cy=ready-button-coffee-icon]').should('exist');
+    cy.contains('Game Mode changed to').should('exist');
+    cy.get('[data-cy="close-snackbar"]').click();
+
+    // Simulate pressing the 'Enter' key again
+    cy.get('[data-cy=edit-game-ranked-switch]').type('{enter}');
```
itsalaidbacklife, line 116:
```suggestion
    cy.get('[data-cy=edit-game-ranked-switch] input').type('{enter}');
```

src/routes/lobby/LobbyView.vue
```diff
@@ -59,7 +59,9 @@
                   data-cy="edit-game-ranked-switch"
                   color="primary"
                   hide-details
+                  tabindex="0"
```
itsalaidbacklife, line 62: Why the tabindex? You can already focus the element with the tab key in the regular order of the document flow, and actually the 'Ready' button above is the more primary action which should be focused first. I think we can skip this
```suggestion
```

src/routes/lobby/LobbyView.vue
```diff
@@ -59,7 +59,9 @@
...
                   @update:model-value="setIsRanked"
+                  @keydown.enter="toggleRanked"
```
itsalaidbacklife, line 64: This works, but I think it would be simpler to fire a click. Another contributor opened #1167 to address the same issue. They have an elegant one-liner for the fix, but they didn't add a test case. I think the ideal combination would be to use your test case (plus the changes to account for focusing the switch correctly) and their solution to fire a click. I've asked them to PR into your fork to combine your PRs so you can both get contribution credit for the fix.

Would that work for you? If so, stay tuned for a PR into your fork updating this part, and if it isn't done soon then we can just update your fix along with the test case to match the one-liner and use the test to confirm if that works
```suggestion
                  @keydown.enter.stop="(e) => e.target.click()"
```

PR #1177, opened 2025-03-28
[Feature]: Translate all error messages in the GameState API

### LINKED ISSUE #1094: [Feature]: Translate all error messages in the GameState API
### Feature Summary

In order to ensure users speaking all supported languages can read the error messages while they play, we should make sure all the server-provided error messages in the GameState API are keys used in the i18n translation system.

### Detailed Description

This is mostly in place, but we should do a pass through all the existing errors thrown inside of `api/helpers/gameStates` to ensure that all errors thrown are translation keys and not raw english text. Correspondingly all of those keys should be found in `src/translations/en.json`, `src/translations/fr.json` and `src/translations/es.json` so they can be properly translated by the client at runtime.

For example, in `api/helpers/gameStates/moves/counter/validate.js`:
```
      if (!currentState.oneOff) {
        throw new Error('You cannot counter unless there is a one-off pending'); // THIS IS INCORRECT; SHOULD PULL INTO i18Nn
      }

      const player = playedBy ? currentState.p1 : currentState.p0;
      const playedCard = player.hand.find(({ id }) => requestedMove.cardId === id);
      if (!playedCard) {
        throw new Error('game.snackbar.global.playFromHand'); // THIS IS A CORRECT TRANSLATION KEY
      }
```

We should update all these errors to match the bottom example by pulling the raw text into `en.json` at an appropriate key, and putting translated versions of the same error into `fr.json` and `es.json` for French and Spanish, respectively

tcox-coding:
## Issue number

- Resolves #1094 

## Please check the following

- For New Features:


tcox-coding:
Some unit tests not passing, will reopen when fixed.

PR #1178, opened 2025-03-28
[Feature]: Replace the player counts in the Home Page with the player usernames 

### LINKED ISSUE #1103: [Feature]: Replace the player counts in the Home Page with the player usernames

negativeInteger:
## 1103

- Resolves #1103 

## Please check the following

- For New Features:

### DIFF
--- a/src/stores/gameList.js
+++ b/src/stores/gameList.js
@@ -7,7 +7,7 @@ class GameSummary {
   constructor(obj) {
     this.id = obj.id ? obj.id : null;
     this.name = obj.name ? obj.name : null;
-    this.numPlayers = Object.prototype.hasOwnProperty.call(obj, 'players') ? obj.players.length : 0;
+    this.players = Object.prototype.hasOwnProperty.call(obj, 'players') ? obj.players : [];
     this.status = Object.prototype.hasOwnProperty.call(obj, 'status') ? obj.status : GameStatus.ARCHIVED;
     this.isRanked = Object.prototype.hasOwnProperty.call(obj, 'isRanked') ? obj.isRanked : false;
     this.isOver = false;
@@ -52,14 +52,14 @@ export const useGameListStore = defineStore('gameList', {
     joinGame(data) {
       const updatedGame = this.openGames.find((game) => game.id === data.gameId);
       if (updatedGame) {
-        updatedGame.numPlayers = Math.min(2, updatedGame.numPlayers + 1);
+        updatedGame.players.push(data.newPlayer);
         updatedGame.status = data.newStatus;
       }
     },
     otherLeftGame(gameId) {
       const updatedGame = this.openGames.find((game) => game.id === gameId);
       if (updatedGame) {
-        updatedGame.numPlayers = Math.max(0, updatedGame.numPlayers - 1);
+        updatedGame.players.pop();
       }
     },
     setIsRanked({ gameId, isRanked }) {
--- a/tests/e2e/specs/out-of-game/home.spec.js
+++ b/tests/e2e/specs/out-of-game/home.spec.js
@@ -507,14 +507,14 @@ describe('Home - Create Game', () => {
     cy.get('[data-cy=game-list-item]')
       .should('have.length', 1)
       .should('include.text', 'test game')
-      .should('include.text', '1 / 2 players');
+      .should('include.text', `vs ${myUser.username}`);
 
     // Test store
     cy.window()
       .its('cuttle.gameListStore.openGames')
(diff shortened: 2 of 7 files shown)


src/stores/gameList.js
```diff
@@ -52,14 +52,14 @@ export const useGameListStore = defineStore('gameList', {
...
-        updatedGame.numPlayers = Math.min(2, updatedGame.numPlayers + 1);
+        updatedGame.players.push(data.newPlayer);
         updatedGame.status = data.newStatus;
       }
     },
     otherLeftGame(gameId) {
       const updatedGame = this.openGames.find((game) => game.id === gameId);
       if (updatedGame) {
-        updatedGame.numPlayers = Math.max(0, updatedGame.numPlayers - 1);
+        updatedGame.players.pop();
```
itsalaidbacklife, line 62: I think this needs to be updated on the backend and here so that we can tell *which* player left and take the appropriate person out. It probably makes sense to emit their id in the socket event and use that to .filter() the array to exclude them.

The user flow to consider here is that the first player to join the game leaves ie

1. Player 1 joins
2. Player 2 joins
3. Player 1 leaves

In this situation we need to remove players[0] not players[1]. So we need to know who left specifically and then use that here to remove the right person

itsalaidbacklife: This warrants a test case as well

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -673,6 +673,79 @@ describe('Home - Create Game', () => {
       cy.get('[data-cy=game-list-item]').should('have.length', 0);
     });
   });
+  it('Does not remove a game when both players are not ready', () => {
```
itsalaidbacklife, line 676: We should confirm that the join button is disabled

itsalaidbacklife (commented):
Nice work. This is very close. We need to update the leave-game flow to emit which user id left so we can filter that user out instead of .pop

PR #1179, opened 2025-04-14
Feat/stalemate gamestate

### LINKED ISSUE #1146: [Feature]: Stalemate GameState Rebuild
### Feature Summary

We need to rebuild stalemates in the gamestate API

### Detailed Description

We need to rebuild Requesting and rejecting a stalemate to follow the same design as the GameState API, this will require making new move types for `stalemate request`, `stalemate-accept` and `stalemate-reject`

This will require a new `GamePhase` called `stalemate-requested` and we will need to verify that we are in this phase for `stalemate-accept` and `stalemate-reject`

stalemate-request/validate is allowed as long as the current phase is not `GamePhase.STALEMATE_REQUESTED`

stalemate-request/execute should set the phase to stalemate requested and specify played by as the request in player

stalemate-accepted/validate should succeed only if the phase is `GamePhase.STALEMATE_REQUESTED` and the requesting player is not the previous playedBy (ie the other player is accepting, not the one who initially requested the stalemate).

stalemate-accepted/execute should set the moveType to MoveType.SATALEMATE_ACCEPTED` and set the playedBy to the requesting player.

stalemate-rejected/validate should succeed only if the phase is `GamePhase.STALEMATE_REQUESTED` and the requesting player is not the previous playedBy (ie the other player is accepting, not the one who initially requested the stalemate).

stalemate-rejected/execute should return a gamestate that is a copy of the previous gamestate before the current one

The existing method for checking if the game is over, checkWinGameState, should add a case for MoveType.STALEMATE_ACCEPTED that ends the game in a stalemate. 

Additionally, the get-log method should add cases for the new move types to describe which player requested, accepted, or rejected a stalemate

 #1116 should give an idea of way to accomplish this. We will also need to remove and test skippers related to this, specifically in `win-conditions` and `reconnecting`

itsalaidbacklife:
Implements requesting, accepting, and rejecting stalemates in the GameState API.

Normalizes behavior for legacy + GameState API parity
## Issue number

- Resolves #1146 

## Please check the following

- For New Features:


api/helpers/game-states/moves/stalemate-accept/execute.js
```diff
@@ -0,0 +1,52 @@
...
+    },
+    playedBy: {
+      type: 'number',
+      description: 'Player number of player requesting move',
+      required: true,
+    },
+    priorStates: {
+      type: 'ref',
+      description: "List of packed gameStateRows for this game's prior states",
+      required: true,
```
Haviles04, line 32: Does this actually need to be required, since it seems it's only used in a few helpers

itsalaidbacklife: So it's a little funky. The part that's weird is that the place that calls any one of these (currently only the `move` endpoint), has to be able to call _all_ of the `validate` and `execute()` helpers and to be able to do so by passing them the same input, since it grabs the 'right' functions based on the `moveType` and then no matter what they are, gives it the same arguments.

So it seemed more consistent to simply say it's required for all of them. We could make it optional for everything except the `stalemate-reject` stuff where this is actually needed (and I don't feel strongly about it), but I lean towards making them all the same signature for consistency's sake

api/controllers/game/move.js
```diff
@@ -23,8 +23,8 @@ module.exports = async function (req, res) {
         throw new Error('You are not a player in this game!');
     }
 
-    validate(gameState, req.body, playedBy);
-    const updatedState = execute(gameState, req.body, playedBy);
+    validate(gameState, req.body, playedBy, game.gameStates);
+    const updatedState = execute(gameState, req.body, playedBy, game.gameStates);
```
itsalaidbacklife, line 27: Sending prior (packed) `GameStateRow`s for the game as a 4th arg to both validate and execute to accommodate stalemate requests since:
1. Whether you can request a stalemate depends on whether one was requested previously this turn
2. When a stalemate request is rejected, we return to the phase from the previous move

Unfortunately this requires updating every `validate()` and `execute()` helper to accept this input, but I think we ultimately get the most consistent API as a result

api/controllers/game/pass.js
```diff
@@ -9,7 +9,7 @@ module.exports = function (req, res) {
       let gameUpdates = {
         turn: game.turn + 1,
         passes: game.passes + 1,
-        log: [ ...game.log, `${player.username} passes` ],
+        log: [ ...game.log, `${player.username} passed` ],
```
itsalaidbacklife, line 12: Standardizing this between game state and legacy API so tests can consistently assert the same log expectations. Might be a good idea to make a pass through all logs to ensure they're the same tense. Picked past tense; we can see how that feels

api/controllers/game/stalemate-reject-gamestate.js
itsalaidbacklife, line 1: I think this whole file got replaced by an implementation as a move helper; TODO: delete this file

api/controllers/game/stalemate-reject.js
```diff
@@ -5,24 +5,17 @@ module.exports = async function (req, res) {
   try {
     const game = await Game.findOne({ id: req.session.game }).populate('players');
 
-    const gameUpdates = {
-      turnStalemateWasRequestedByP0: -1,
-      turnStalemateWasRequestedByP1: -1,
-    };
```
itsalaidbacklife, line 11: Removed this to disallow making multiple stalemate requests in one turn. That's mainly because in the GameState API, each stalemate request writes a row to the `GameStateRow` table, so we don't want to let you keep cluttering up the table with a bajillion stalemate requests. That isn't a problem for the legacy API, since it only records the 'latest' stalemate requests, but adding this constraint ensures that the legacy API is consistent with the GameStateAPI.

So this block + the `Game.update()` call below would effectively 'reset' the last time a user requested a stalemate to allow them to continuously request them back and forth. Now the requirement is you can only request a stalemate once per turn so I'm removing the reset.

api/controllers/game/stalemate-reject.js
```diff
@@ -5,24 +5,17 @@ module.exports = async function (req, res) {
...
 
-    await Game.updateOne({ id: game.id }).set(gameUpdates);
-
     Game.publish([ game.id ], {
-      change: 'rejectStalemate',
+      change: 'stalemateReject',
       game,
       victory,
-      requestedByPNum: req.session.pNum,
+      playedBy: req.session.pNum,
```
itsalaidbacklife, line 18: Replaced the `requestedByPNum` attribute with `playedBy` for consistency with the GameState API and across moves. Now every move has a `playedBy` instead of different names for the same data (who did the thing) based on which move was made

api/controllers/game/stalemate.js
```diff
@@ -6,12 +6,17 @@ module.exports = async function (req, res) {
...
+    let game = await Game.findOne({ id: gameId }).populate('players', { sort: 'pNum' });
     let gameUpdates = {};
     const keyPrefix = 'turnStalemateWasRequestedByP';
     const playerStalemateKey = `${keyPrefix}${pNum}`;
     const playerStalemateVal = game.turn;
+
+    // Error if this player already requested stalemate this turn
+    if (game[playerStalemateKey] === game.turn) {
+      throw new Error('game.snackbar.stalemate.previousStalemateRejected');
+    }
```
itsalaidbacklife, line 18: Adding this error to match new GameState API requirement as described above. You can no longer request a stalemate more than once per turn.

api/controllers/game/stalemate.js
```diff
@@ -23,7 +28,7 @@ module.exports = async function (req, res) {
       currentMatch: null,
     };
 
-    gameUpdates.lastEvent = { change: 'requestStalemate', requestedByPNum: pNum };
+    gameUpdates.lastEvent = { change: 'stalemateRequest', playedBy: pNum };
```
itsalaidbacklife, line 31: Standardized event names to group all the stalemate actions

api/helpers/game-states/moves/concede/execute.js
```diff
@@ -20,6 +20,11 @@ module.exports = {
       description: '0 or 1 for whether p0 or p1 is making the move',
       required: true,
     },
+    priorStates: {
+      type: 'ref',
+      description: "List of packed gameStateRows for this game's prior states",
+      required: true,
+    }
```
itsalaidbacklife, line 27: Added `priorStates` to all `execute()` and `validate()` helpers so that they can accept this input given that the caller in `move.js` calls all of them with the same arguments and its needed for some of them. Little awkward, but at least it's consistent

api/helpers/game-states/moves/stalemate-accept/execute.js
```diff
@@ -0,0 +1,52 @@
+const GamePhase = require('../../../../../utils/GamePhase.json');
+
+module.exports = {
+  friendlyName: 'Accept offer for Stalemate',
+
+  description: 'Accept opponent\'s stalemate offer',
```
itsalaidbacklife, line 6: TODO: write an issue to standardize string use when `'` characters are in the string. I think `\'` is the cleanest, compared with using `""` or backtick-strings just for '

api/helpers/game-states/moves/stalemate-accept/execute.js
```diff
@@ -0,0 +1,52 @@
...
+
+  inputs: {
+    currentState: {
+      type: 'ref',
+      descriptions: 'Object containing the current game state',
+      required: true,
+    },
+    /**
+     * @param { Object } requestedMove - Object describing the request for stalemate
```
itsalaidbacklife, line 15:
```suggestion
     * @param { Object } requestedMove - Object describing the request to accept a stalemate offer
```

api/helpers/game-states/moves/stalemate-accept/execute.js
```diff
@@ -0,0 +1,52 @@
...
+     * @param { MoveType.STALEMATE_ACCEPT } requestedMove.moveType - Specifies that this accepts a stalemate offer
+     * @param { 1 | 0 } requestedMove.playedBy - Which player is requesting a stalemate
```
itsalaidbacklife, line 17:
```suggestion
     * @param { 1 | 0 } requestedMove.playedBy - Which player is accepting a stalemate
```

api/helpers/game-states/moves/stalemate-reject/execute.js
```diff
@@ -0,0 +1,54 @@
...
+      type: 'ref',
+      description: "List of packed gameStateRows for this game's prior states",
+      required: true,
+    }
+  },
+  sync: true,
+  fn: ({ currentState, requestedMove, playedBy, priorStates }, exits) => {
+    let result = _.cloneDeep(currentState);
+
+    const phase = priorStates.at(-2)?.phase ?? GamePhase.MAIN;
```
itsalaidbacklife, line 39: This line is why we need to include `priorStates` in all the helpers. When a stalemate is rejected, go back to the previous game phase

api/helpers/game-states/moves/stalemate-request/validate.js
```diff
@@ -0,0 +1,57 @@
...
+    priorStates: {
+      type: 'ref',
+      description: "List of packed gameStateRows for this game's prior states",
+      required: true,
+    },
+  },
+  sync: true,
+  fn: ({ currentState, priorStates }, exits) => {
+    try {
+      if (
```
itsalaidbacklife, line 39:
```suggestion
      // Can't request stalemate more than once per turn
      if (
```

api/helpers/game-states/validate-gamestate.js
```diff
@@ -20,16 +22,12 @@ function validateTurn(param) {
...
 function validatePhase(param) {
-  if (
-    param === null ||
-    param === undefined ||
-    Number.isNaN(param) ||
-    param < 0 ||
-    (param > 5 && param !== 7)
-  ) {
-    throw new Error('The phase must be a number in [1, 2, 3, 4, 5, 7]');
+  const phases = Object.values(GamePhase);
```
itsalaidbacklife, line 25: We were previously hard-coding the allowed game phases here, but they're already defined on the `GamePhase` object. It's better to validate them by direct comparison to that object as the source of truth

api/models/GameStateRow.js
```diff
@@ -72,7 +73,7 @@ module.exports = {
     phase: {
       type: 'number',
       required: true,
-      isIn: [ 1, 2, 3, 4, 5, 7 ],
+      isIn: Object.values(GamePhase),
```
itsalaidbacklife, line 76: Same idea here; validate via direct comparison to the `GamePhase` object for centralized source of truth on valid phases

config/routes.js
```diff
@@ -91,7 +91,8 @@ module.exports.routes = {
   '/api/game/over': 'game/game-over',
   '/api/game/concede': 'game/concede',
   '/api/game/stalemate': 'game/stalemate',
-  '/api/game/reject-stalemate': 'game/stalemate-reject',
+  '/api/game/stalemate-accept': 'game/stalemate',
```
itsalaidbacklife, line 94: The legacy API still only uses one `game/stalemate` action for requesting + accepting stalemates, but the GameState API treats it as two moveTypes. Adding this redundant route for the legacy API to point to the same `game/stalemate` action allows consistent url transformation between API's. Should remove this on cleanup though along with all the other `/api/game/X` endpoints

src/plugins/sockets/inGameEvents.js
```diff
@@ -193,7 +193,8 @@ export async function handleInGameEvents(evData) {
       // Show OpponentRequestedStalemateDialog if opponent requested stalemate
       //   and game is not yet over
       gameStore.consideringOpponentStalemateRequest =
-        !evData.victory.gameOver && evData.requestedByPNum !== gameStore.myPNum;
```
itsalaidbacklife, line 196: `requestedByPNum` was only used for the stalemate request action, but we're standardizing it to use `playedBy` everywhere so now we need to additionally check the `moveType`.

src/plugins/sockets/inGameEvents.js
```diff
@@ -193,7 +193,8 @@ export async function handleInGameEvents(evData) {
...
+        !evData.victory.gameOver && evData.playedBy !== gameStore.myPNum;
+      gameStore.waitingForOpponentToStalemate = !evData.victory.gameOver && evData.playedBy === gameStore.myPNum;
```
itsalaidbacklife, line 197: Setting this now so you can see the overlay w/ waiting for your opponent on page refresh (previously it would disappear)

src/routes/game/GameView.vue
```diff
@@ -9,7 +9,7 @@
     <template v-else>
       <div id="game-menu-wrapper" class="d-flex flex-column flex-sm-row align-center" :style="menuWrapperStyle">
         <SpectatorListMenu :spectating-users="spectatingUsers" :vuetify-display="$vuetify" />
-        <GameMenu :is-spectating="isSpectating" />
+        <GameMenu :is-spectating="isSpectating" @handle-error="handleError" />
```
itsalaidbacklife, line 12: Added error handling to the `GameMenu` so we can see snackbar errors when you request an illegal stalemate (e.g. b/c you requested one already this turn)

src/routes/game/components/dialogs/components/OpponentRequestedStalemateDialog.vue
```diff
@@ -88,7 +88,7 @@ export default {
     async acceptStalemate() {
       this.loadingAccept = true;
       try {
-        await this.gameStore.requestStalemate();
+        await this.gameStore.acceptStalemate();
```
itsalaidbacklife, line 91: Created a separate store action for accepting stalemate vs requesting one initially to make legacy API more consistent w/ GameState API, which treats these as two separate actions

src/stores/game.js
```diff
@@ -208,7 +208,10 @@ export const useGameStore = defineStore('game', {
       this.lastEventPlayerChoosing = newGame.lastEvent?.pNum === this.myPNum ?? null;
       this.lastEventDiscardedCards = newGame.lastEvent?.discardedCards ?? null;
       this.waitingForOpponentToStalemate =
-        (newGame.lastEvent?.requestedByPNum === this.myPNum && !newGame.gameIsOver) ?? false;
```
itsalaidbacklife, line 211: swapped the bespoke `requestedByPNum` with the `playedBy` value we use everywhere. Now need to check moveType

tests/e2e/specs/in-game/game-over/rankedMatches.spec.js
```diff
@@ -183,8 +182,9 @@ describe('Creating And Updating Ranked Matches', () => {
       .click();
     cy.get('#waiting-for-opponent-stalemate-scrim').should('be.visible');
     // Opponent confirms
-    cy.stalemateOpponent();
+    cy.acceptStalemateOpponent();
```
itsalaidbacklife, line 185: Commands, like the app, now separate accepting stalemate from requesting one initially (though they resolve to the same endpoint in the legacy API)

tests/e2e/specs/in-game/game-over/rankedMatches.spec.js
```diff
@@ -214,6 +214,8 @@ describe('Creating And Updating Ranked Matches', () => {
       p1Hand: [],
       p1Points: [],
       p1FaceCards: [],
+      topCard: Card.SIX_OF_SPADES,
+      secondCard: Card.FOUR_OF_CLUBS,
```
itsalaidbacklife, line 218: Test didn't care what these cards were, but legacy API defaults to leaving 2 cards in the deck when you don't specify them but you set `deck: []`, whereas the GameState API treats that as scrapping the entire deck. Setting them explicitly here so both API's will behave the same for this test

src/translations/en.json
```diff
@@ -273,6 +273,12 @@
         "mustUsePointCard": "You can only scuttle with an ace through ten",
         "mustTargetPointCard": "You can only scuttle a card your opponent has played for points"
       },
+      "stalemate": {
+        "alreadyConsideringStalemate": "A stalemate has already been offered",
+        "noStalemateOffered": "Your opponent has not offered a stalemate",
+        "opponentConsideringStalemate": "Your opponent is considering your stalemate offer",
+        "previousStalemateRejected": "Stalemate request was already rejecteed this turn"
```
itsalaidbacklife, line 280:
```suggestion
        "previousStalemateRejected": "Stalemate request was already rejected this turn"
```

tests/e2e/specs/in-game/game-over/winConditions.spec.js
```diff
@@ -385,38 +374,120 @@ describe('Stalemates', () => {
...
       cy.get('#waiting-for-opponent-stalemate-scrim').should('be.visible');
+      cy.reload();
+      cy.get('#waiting-for-opponent-stalemate-scrim').should('be.visible');
+      cy.acceptStalemateOpponent();
+      assertStalemate();
+    });
 
-      cy.drawCardOpponent();
-      cy.get('#waiting-for-opponent-stalemate-scrim').should('not.exist');
+    it('Rejects a stalemate while resolving a seven', () => {
```
itsalaidbacklife, line 383: This confirms we can request + reject a stalemate while in the middle of a different phase and then get back to that phase to accept the next real move after its rejected (the payoff of adding `priorStates` to all the move helpers)

api/helpers/game-states/moves/stalemate-request/validate.js
```diff
@@ -0,0 +1,57 @@
...
+        priorStates.some(
+          (state) => state.moveType === MoveType.STALEMATE_REQUEST && state.turn === currentState.turn,
+        )
+      ) {
+        throw new Error('game.snackbar.stalemate.previousStalemateRejected');
+      }
```
itsalaidbacklife, line 45: This is the reason we added `priorStates` to the `validate()` helpers; prevent players from requesting more than one stalemate per turn, since each one writes a new row to the db

api/helpers/game-states/moves/stalemate-accept/execute.js
Haviles04: That's a fair point, I think the consistency is good.

api/helpers/game-states/validate-gamestate.js
Haviles04: Way cleaner!

api/helpers/game-states/moves/stalemate-accept/execute.js
itsalaidbacklife: https://github.com/cuttle-cards/cuttle/issues/1181

PR #1180, opened 2025-04-16
translate error messages in GameState API

casens5:
## Issue number

- Resolves # 1094

## Please check the following

- For New Features:
  - [/] Have tests been added to cover any new features or fixes?
  - [/] Has the documentation been updated accordingly?

- i didn't know how to manually test the changes, so i'm relying on the existing tests
- i used claude to create translations, which i know are not perfect
- some error messages included string interpolation, and i didn't see any examples in the existing API via sails (only in vue), so i'm assuming eg. `return exits.error(sails.__('game.other.errorUnpackCards', {message: err.message} ))` is correct
- in the past with translation keys, i've just used a totally flat document structure.  rather than make assumptions about how the keys should be organized, i mostly dumped them into an "other" category
- line 96 of `api/helpers/game-states/validate-gamestate.js` appears to be a catch-all error handler?  and would be untranslated


.gitignore
```diff
@@ -143,3 +143,5 @@ nbproject
 tests/e2e/screenshots
 tests/e2e/videos
 storybook-static
+*.swo
+*.swp
```
itsalaidbacklife, line 147: Found the vim user ;)

Good QoL catch; thanks for adding these to `.gitignore` so we can keep 'em out of version control

itsalaidbacklife (commented):
Nice work! This is a great start. The biggest thing to work out is the handling of server-side text resolution. Currently we're not configured for sails to pick up the lang files or current locale so we can't rely on sails to resolve localized text. We could investigate upgrading our setup to accommodate this, but my instinct is that we're better off with the simpler approach of keeping the translations client side at the cost of skipping serverside parameterization, but it's worth discussing

casens5:
i got a little hung up on the fact that i didn't actually test the changes i made, i just hoped that the pattern in the codebase makes sense.  but now, revisiting what you're proposing:

i'm not sure how much we'd have to do to pass parameters and keys to `vue-i18n`, but overall i think it's not worth it.  a few strings that are either in english with parameters or are translated but lack parameters is a fine solution until users start complaining.

src/translations/en.json
```diff
@@ -524,5 +550,11 @@
     "heading": "World Championship IV",
     "paragraph": "The Cuttle World Championship IV is today, Saturday March 8th at 12pm EST! Watch the top 8 players from the past YEAR of competitive play battle it out for the title of Cuttle World Champion and their pick from the 8 incredible prize decks donated by player SUBMARINO! Watch it live at ",
     "twitchLink": "twitch.tv/cuttle_cards"
+  },
+  "other": {
```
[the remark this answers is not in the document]
casens5: it wasn't clear to me how/why the keys should be structured.  the bigger commit has keys in `game.snackbar.other`, whereas these 4 keys are simply in `other`.  i'm guessing "snackbar" is the text pane to the right of the game where the record of the moves are listed?  so i'm not sure where/how these strings would appear.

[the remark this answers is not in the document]
itsalaidbacklife: okay cool, thanks for clarifying. `"snackbar"` is actually the little temporary alert (sometimes called a toast) that shows up after you submit some API requests. We mainly use them to display user-facing errors:
image

Here's the [vuetify component docs](https://vuetifyjs.com/en/components/snackbars/) on them for reference. I'll take a look at the structure again and see if it makes sense to adjust how their grouped

src/translations/en.json
```diff
@@ -524,5 +544,10 @@
...
+    "addGameToMatch": "Could not add game to match",
+    "dontWipeDbOnProd": "Error: This action is only permitted in development or staging",
```
itsalaidbacklife, line 550: Note to self: don't wipe db on prod

src/translations/en.json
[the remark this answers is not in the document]
itsalaidbacklife: Okay looking again, I think the split between these two sections is reasonable, though I'd like to slightly adjust it. Generally, backend errors do go in `"snackbar"` sections because that's where we display them, so that's good. 

The way we've organized these so far is essentially by page and then area of function. Some of these areas are specific to pages and some aren't so I'll comment on them individually

src/translations/en.json
itsalaidbacklife: Let's put `"dontWipeDbOnProd"` in the top-level `"error"` block at the very bottom. This one is the weirdest because we really shouldn't be hitting it outside of testing (thank goodness haha) so it doesn't fit super neatly into the normal space of errors. It's the one that by itself might warrant creating an "other" section live you have here, but the other two have more well-defined spots they belong in imo so instead we'll throw this in `"error"` down at the bottom

src/translations/en.json
```diff
@@ -524,5 +544,10 @@
...
+    "seasonDataNotFound": "Could not find requested season data"
```
itsalaidbacklife, line 551: This one is page specific to stats, so let's put it in `"stats.error"`

src/translations/en.json
```diff
@@ -303,6 +309,20 @@
...
+      "waitingForCounter": "Waiting for opponent to counter",
+      "moveTypeMustBeString": "The MoveType cannot be empty and must be a string",
+      "turnMustBeNumber": "The turn cannot be empty and must be a number",
+      "phaseIsInvalid": "The phase must be a number in [1, 2, 3, 4, 5, 7]",
+      "playedByIsInvalid": "The playedBy attribute must be a number in [0,1]",
+      "errorUnpackCards": "Error at unpacking cards",
+      "gameStatusInvalid": "Game has not yet started or is over",
+      "missingDotGameStates": "Game was not populated with .gameStates collection",
+      "cardsAlreadyDealt": "Cards are already dealt",
+      "gameIdMustBeNumber": "The gameId cannot be empty and must be a number"
```
itsalaidbacklife, line 325: It's correct that you have these all inside `"game.snackbar"`. Generally the organization there is that any errors specific to a singular move e.g. drawing a card get their own subsection here e.g. `"game.snackbar.draw.deckIsEmpty"` and anything that is used in multiple places for game logic goes in `"game.snackbar.global"`. So with that in mind:

```
"notInCounteringPhase" // create new "game.snackbar.counter" section for this
"pendingOneOff" // create new "game.snackbar.resolve" section for this
"waitingForCounter" // also goes in "game.snackbar.resolve"
"moveTypeMustBeString" goes in "game.snackbar.global" (because it's generic across moves)
// Same with all the below, which are generic to game logic across moves:
      "turnMustBeNumber",
      "phaseIsInvalid",
      "playedByIsInvalid",
      "errorUnpackCards",
      "gameStatusInvalid",
      "missingDotGameStates",
      "cardsAlreadyDealt",
      "gameIdMustBeNumber"
```

api/helpers/game-states/convert-card-to-str.js
```diff
@@ -71,7 +71,7 @@ module.exports = {
 
       return exits.success(cardId + attachments);
     } catch (err) {
-      return exits.error('Error at unpacking cards: ' + err.message);
+      return exits.error(sails.__('game.other.errorUnpackCards'));
```
itsalaidbacklife, line 74: I still believe we want to avoid `sails.__` because sails doesn't know the selected lang or where to find the json; so let's just throw the key name

itsalaidbacklife (changes requested):
Nice work, this is almost there. Have you taken a crack at testing these? I think the best way to do so would be via unit testing with vitest. I'm happy to walk you through it, just so we can confirm the basics are in place.

Otherwise I'm requesting a bit of reorg of the text files and a tweak to remove one remaining `sails.__` call.

Issue #965, opened 2024-06-16
[Feature]: Clean up and RELEASE the GameState API


itsalaidbacklife:
### Feature Summary

After all other issues in the [GameState API MVP Project](https://github.com/orgs/cuttle-cards/projects/8) are completed, we should make the necessary changes to utilize the GameState API in production.

### Detailed Description

Specifically, this will require:
1. Removing the feature flags scripts from `package.json` for enabling/disabling the GameState API
2. Reverting #963 so that CI the tests only run once in CI (vs with + without the GameState API)
3. Removing all references to the `SAILS_USE_GAMESTATE_API`, `VITE_USE_GAMESTATE_API`, and `CYPRESS_USE_GAMESTATE_API` from frontend and test code such that app code + tests all use the GameState API all the time
4. Remove all the routes and controller files for the legacy api calls that are replaced by the GameState API endpoints
5. Clean up the endpoint callers to remove extraneous data e.g. `opId` everywhere it's used.
6. Update tests so that tests which set `deck: []` don't need to specify topCard + secondCard and draw twice to clear the deck. [See this comment](https://github.com/cuttle-cards/cuttle/pull/1187/files#r2061385938) for context
7. Update the tests so that card ids are read directly from the card fixture, rather than the store, since now every card ID is known in advance of runtime. This includes removing the `cy.playPointsById()` command entirely

**Note: Look for comments with `// FIXME: Remove in #965`** throughout the codebase to clean these up. For example:
`src/stores/game.js`: remove `opId` and `targetId` from `requestResolveSevenDoubleJacks`

Issue #347, opened 2023-03-05
[DevEx]: Remove `isLoggedIn` session variable and update backend auth checks to check that `usr` is not-null


itsalaidbacklife:
### Improvement Summary

When a user authenticates, the backend updates their session data to include both their user id (as `usr`) and the boolean `isLoggedIn: true`. The `isLoggedIn` variable is unnecessary because we could instead check that `usr` is not null

### Detailed Description

We should update all the backend code that references `req.session.isLoggedIn` to avoid setting it, and when reading that variable to instead do a check that `req.session.usr` is truthy. This will include updating `api/policies/isLoggedIn.js` which is the primary consumer of the `isLoggedIn` session variable.

PR #1187, opened 2025-04-26
Test/unskip gamestate tests

### LINKED ISSUE #1127: [Feature]: Unskip all remaining tests in the GameState API
### Feature Summary

Once we've worked through the remaining issues that enable us to unskip tests, we should look through the entire test suite for any remaining `cy.skipOnGameStateApi();` calls and remove them

### Detailed Description

This should allow us to run all the remaining tests. If there are bugs at that point, we can either fix them in this issue, or write follow up issues per bug, depending on their complexity.

itsalaidbacklife:
Unskips all remaining tests for gamestate api and removes `cy.skipOnGamestateApi()` command. All tests looking good!
## Issue number

- Resolves #1127 

## Please check the following

- For New Features:


api/controllers/UserController.js
```diff
@@ -64,7 +64,7 @@ module.exports = {
       // Query for game if user is in one
       const gameId = (user.game ?? req.session.game) ?? null;
 
-      if (process.env.VITE_USE_GAMESTATE_API) {
+      if (process.env.VITE_USE_GAMESTATE_API && gameId) {
```
itsalaidbacklife, line 67: This block is logic for reproducing the latest socket event for a specific game, so it's only needed when the there is a game in question. It was erroring when we hit `reLogin` in the gamestate api without a gameId provided.

api/controllers/game/stalemate.js
```diff
@@ -47,7 +47,7 @@ module.exports = async function (req, res) {
 
       await Game.updateOne({ id: gameId }).set({
         lastEvent: {
-          change: 'stalemateRequest',
+          change: 'stalemateAccept',
```
itsalaidbacklife, line 50: Couldn't previously differentiate between stalemate was requested vs accepted in socket event via the `change` attribute. This makes the difference more explicit

api/helpers/add-game-to-match.js
```diff
@@ -40,7 +40,8 @@ module.exports = {
       let numPlayer1Wins = 0;
       let numPlayer2Wins = 0;
 
-      for (const priorGame of [ game, ...relevantMatch.games ]) {
+      const priorGames = _.uniqBy([ game, ...relevantMatch.games ], 'id');
```
itsalaidbacklife, line 43: Hit a case where we were re-adding the same game to a match and it was counted twice, causing the match to appear over with only one win. Basic flow was hitting reLogin in the gamestate api as a spectator after the game is over.

This ensures each game only appears once in the list (using id as the thing to compare)

src/plugins/sockets/inGameEvents.js
```diff
@@ -196,6 +196,7 @@ export async function handleInGameEvents(evData) {
         !evData.victory.gameOver && evData.playedBy !== gameStore.myPNum;
       gameStore.waitingForOpponentToStalemate = !evData.victory.gameOver && evData.playedBy === gameStore.myPNum;
       break;
+    case SocketEvent.ACCEPT_STALEMATE:
```
itsalaidbacklife, line 199: Added this case to allow clearing both dialog and overlay when stalemate is accepted. They used to both stay open in the legacy API because it looked like a request for a stalemate initiation, rather than to accept one and end the game.

tests/e2e/specs/in-game/game-over/rematch.spec.js
```diff
@@ -567,30 +567,31 @@ describe('Spectating Rematches', () => {
 
       cy.get('[data-cy=opponent-match-result] [data-cy-result-img=lost]').should('be.visible');
 
+      // The stalemate dialog should dissappear
```
itsalaidbacklife, line 570:
```suggestion
      // The stalemate dialog should disappear
```

tests/e2e/specs/in-game/game-over/rematch.spec.js
```diff
@@ -494,6 +492,8 @@ describe('Spectating Rematches', () => {
         p1Points: [],
         p1FaceCards: [],
         deck: [],
+        topCard: Card.FOUR_OF_HEARTS,
+        secondCard: Card.FIVE_OF_CLUBS,
```
itsalaidbacklife, line 496: Specifiying topCard + secondCard is an awkward workaround for the fact that the GameState API's `loadGameFixture` endpoint will clear the entire deck when you set `deck: []` and the legacy api will leave two cards: `topCard` and `secondCard`, so really, most of the tests that are clearing the deck actually want to get straight to the end of the deck and would prefer to scrap the top two cards, but for the legacy API to do that, we instead scrap all but 2 cards then draw twice.

This means the tests would be simpler and more performant when we switch over to full gamestate api to clean these cases up by removing the `topCard` and `secondCard` and drawing two cards from most tests that specify `deck: []`. 

We should update #965 to look for that

tests/e2e/specs/in-game/game-over/rematch.spec.js
```diff
@@ -567,30 +567,31 @@ describe('Spectating Rematches', () => {
 
       cy.get('[data-cy=opponent-match-result] [data-cy-result-img=lost]').should('be.visible');
 
+      // The stalemate dialog should dissappear
+      cy.wait(1000);
+      cy.get('#opponent-requested-stalemate-dialog').should('not.exist');
```
itsalaidbacklife, line 572: When you want to ensure that something disappears, it's sometimes necessary to `cy.wait()` first just to make sure that we are checking that the thing is not present after a delay, instead of in the beginning

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -329,45 +329,43 @@ describe('Home - Game List', () => {
     });
 
     it('Disables spectate button if on home view before game finishes', () => {
-      cy.skipOnGameStateApi();
       cy.signupOpponent(playerOne);
       cy.signupOpponent(playerTwo);
 
       // Navigate to homepage and select spectate tab
-      cy.visit('/');
+      // cy.visit('/');
```
itsalaidbacklife, line 336: Todo: double check this; should we remove it or put it back?

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -329,45 +329,43 @@ describe('Home - Game List', () => {
...
       // make sure button is still enabled after 1 person stalemates
-      cy.get('@stalemateGameId').then(function () {
-        cy.get(`[data-cy-spectate-game=${this.stalemateGameId}]`).should('be.not.disabled');
-        cy.acceptStalemateOpponent();
-        cy.get(`[data-cy-spectate-game=${this.stalemateGameId}]`).should('be.disabled');
+      cy.get('@stalemateGameId').then(function (stalemateGameId) {
+        cy.stalemateOpponent(stalemateGameId);
+        cy.recoverSessionOpponent(playerOne);
+        cy.get(`[data-cy-spectate-game=${stalemateGameId}]`).should('be.not.disabled');
+        cy.acceptStalemateOpponent(stalemateGameId);
```
itsalaidbacklife, line 362: The GameState API gets `gameId` as an explicit request param rather than from the session, so to make a request on behalf of a user when the test-perspective user isn't in the game requires caching the game id beforehand and passing it explicitliy in the command like this.

Here our user is sitting on the home page while two users request + accept a stalemate

tests/e2e/support/commands.js
```diff
@@ -10,12 +10,6 @@ io.sails.url = 'localhost:1337';
 io.sails.useCORSRouteToGetCookie = false;
 const env = Cypress.env('VITE_USE_GAMESTATE_API');
 
-Cypress.Commands.add('skipOnGameStateApi', () => {
-  if (env) {
-    cy.state('runnable').ctx.skip();
-  }
-});
```
itsalaidbacklife, line 17: @Haviles04 the end of an era!

tests/e2e/support/commands.js
```diff
@@ -1095,19 +1089,19 @@ Cypress.Commands.add('concedeOpponent', (gameId = null) => {
   cy.makeSocketRequest('game', 'concede', { moveType: MoveType.CONCEDE }, 'POST', gameId);
 });
 
-Cypress.Commands.add('stalemateOpponent', () => {
+Cypress.Commands.add('stalemateOpponent', (gameId = null) => {
```
itsalaidbacklife, line 1092: Added `gameId` param so that we can call these from the home page and have it apply to a specific game by its id since the game id is not taken from the requesting user's session anymore

types/SocketEvent.js
```diff
@@ -13,6 +13,7 @@ export default {
   SET_IS_RANKED: 'setIsRanked',
   REJECT_STALEMATE: 'stalemateReject',
   REQUEST_STALEMATE: 'stalemateRequest',
+  ACCEPT_STALEMATE: 'stalemateAccept',
```
itsalaidbacklife, line 16: Should probably rename these to match move type:

STALEMATE_REQUEST, STALEMATE_ACCEPT, STALEMATE_REJECT

types/SocketEvent.js
```diff
@@ -11,8 +11,9 @@ export default {
   RE_LOGIN: 'reLogin',
   READY: 'ready',
   SET_IS_RANKED: 'setIsRanked',
-  REJECT_STALEMATE: 'stalemateReject',
-  REQUEST_STALEMATE: 'stalemateRequest',
+  STALEMATE_REJECT: 'stalemateReject',
+  STALEMATE_REQUEST: 'stalemateRequest',
+  STALEMATE_ACCEPT: 'stalemateAccept',
```
itsalaidbacklife, line 16: Renamed keys to match the MoveType names

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -322,52 +322,49 @@ describe('Home - Game List', () => {
         // No open games appear
         cy.contains('[data-cy-join-game]', 'Join Casual').should('not.exist');
         // Existing game is available to spectate
+        cy.wait(500);
```
itsalaidbacklife, line 325: This test was failing because Vuetify's `v-btn-toggle` component  fails to update which option is selected when the button is clicked right after the page loads before it's finished initializing.

We should fix this and remove the need for the wait: #1190

config/policies.js
```diff
@@ -89,5 +89,6 @@ module.exports.policies = {
 
   TestController: {
     '*': 'developmentOrStagingOnly',
+    'wipeDatabase': 'deleteSession'
```
itsalaidbacklife, line 92: This policy was previously unused, maybe because it was implemented incorrectly. My idea here is that whenever we're wiping the database, we should also clear out the cypress user's session. That should prevent cross-test interactions via carryover session data. If it works, this would resolve #1191

itsalaidbacklife: Confirmed: resolves #1191

api/controllers/UserController.js
```diff
@@ -61,26 +61,35 @@ module.exports = {
...
+
+
+        if (!game) {
+          return res.ok({
+            username: user.username
+          });
+        }
+
+        const gameObject = game.gameStates?.length ? await unpackGamestate(game.gameStates.at(-1)) : null;
+        const socketEvent = game?.gameStates?.length ?
```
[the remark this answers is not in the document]
itsalaidbacklife: Fixed this by applying the `deleteSession` policy to the `wipeDatabase` endpoint

api/controllers/UserController.js
```diff
@@ -89,15 +98,15 @@ module.exports = {
...
       }
-      // FIXME: #965
+      // TODO #965
       // Remove everything between here and catch AFTER gamestate is deployed
       const unpopulatedGame = gameId ? await gameService.findGame({ gameId }) : null;
       const populatedGame =
         unpopulatedGame?.status === GameStatus.STARTED
           ? await gameService.populateGame({ gameId })
           : null;
       req.session.loggedIn = true;
```
itsalaidbacklife, line 108: Todo:

1. Move this up with req.session.usr assignment
2. Write issue to remove isLoggedIn

tests/e2e/specs/in-game/game-over/rematch.spec.js
itsalaidbacklife: On second thought I think we can remove this since the element was already present in the previous update/assertion so we don't need to worry about it passing the not exists assertion by checking before the overlay renders since it's already rendered

itsalaidbacklife (commented):
Made another pass

api/controllers/UserController.js
itsalaidbacklife: #347

PR #1192, opened 2025-05-01
Feat/disable stalemate request during other phases

itsalaidbacklife:
This fixes a bug where rejecting a stalemate during phases of the game like countering or seven resolution and then refreshing the page would break players' ability to make the correct move. Root cause is that the client doesn't know or use the `phase` attribute, only `moveType` and so it doesn't know which moves are legal if the moveType was stalemateRejected. 

Long term it'd be best to utilize `phase` directly on the clientside to control which move is expected, but presently the best option is to disable stalemates during phases other than `GamePhase.MAIN`.

## Issue number

- N/A

## Please check the following

- For New Features:

### DIFF
--- a/api/helpers/game-states/moves/stalemate-request/validate.js
+++ b/api/helpers/game-states/moves/stalemate-request/validate.js
@@ -39,7 +39,7 @@ module.exports = {
 
       // Must not already be considering a stalemate
       if (currentState.phase === GamePhase.STALEMATE_REQUEST) {
-        throw new Error('game.snackbar.global.alreadyConsideringStalemate');
+        throw new Error('game.snackbar.stalemate.alreadyConsideringStalemate');
       }
 
       // Must be in main phase
--- a/api/helpers/game-states/moves/stalemate-request/validate.js
+++ b/api/helpers/game-states/moves/stalemate-request/validate.js
@@ -36,6 +36,17 @@ module.exports = {
   sync: true,
   fn: ({ currentState, priorStates }, exits) => {
     try {
+
+      // Must not already be considering a stalemate
+      if (currentState.phase === GamePhase.STALEMATE_REQUEST) {
+        throw new Error('game.snackbar.global.alreadyConsideringStalemate');
+      }
+
+      // Must be in main phase
+      if (currentState.phase !== GamePhase.main) {
+        throw new Error('game.snackbar.stalemate.wrongPhase');
+      }
+
       // Can't request stalemate more than once per turn
       if (
         priorStates.some(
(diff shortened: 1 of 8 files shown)


api/helpers/game-states/moves/stalemate-request/validate.js
```diff
@@ -36,6 +36,17 @@ module.exports = {
...
+
+      // Must not already be considering a stalemate
+      if (currentState.phase === GamePhase.STALEMATE_REQUEST) {
+        throw new Error('game.snackbar.global.alreadyConsideringStalemate');
+      }
+
+      // Must be in main phase
+      if (currentState.phase !== GamePhase.main) {
+        throw new Error('game.snackbar.stalemate.wrongPhase');
+      }
```
itsalaidbacklife, line 48: The bottom error is a superset of the first i.e. would fire every time the phase is "wrong" including when it's a pending stalemate request, but I think explicitly calling out that you can't do that because a stalemate request is already pending is better than the "can't request a stalemate during other moves" catch-all of the second check. 

This was the key change that prevents the bug, which would happen if a stalemate was requested then rejected during non-main phases

api/helpers/game-states/moves/stalemate-request/validate.js
```diff
@@ -36,6 +36,17 @@ module.exports = {
...
+
+      // Must not already be considering a stalemate
+      if (currentState.phase === GamePhase.STALEMATE_REQUEST) {
+        throw new Error('game.snackbar.stalemate.alreadyConsideringStalemate');
+      }
+
+      // Must be in main phase
+      if (currentState.phase !== GamePhase.main) {
+        throw new Error('game.snackbar.stalemate.wrongPhase');
+      }
```
itsalaidbacklife, line 48: The bottom error is a superset of the first i.e. would fire every time the phase is "wrong" including when it's a pending stalemate request, but I think explicitly calling out that you can't do that because a stalemate request is already pending is better than the "can't request a stalemate during other moves" catch-all of the second check.

This was the key change that prevents the bug, which would happen if a stalemate was requested then rejected during non-main phases

api/helpers/game-states/moves/stalemate-request/validate.js
```diff
@@ -45,11 +56,6 @@ module.exports = {
         throw new Error('game.snackbar.stalemate.previousStalemateRejected');
       }
 
-      // Must not already be considering a stalemate
-      if (currentState.phase === GamePhase.STALEMATE_REQUEST) {
-        throw new Error('game.snackbar.global.alreadyConsideringStalemate');
-      }
-
```
itsalaidbacklife, line 52: Moved this error first and swapped the error key from `global` to `stalemate` to ensure it gets top priority and that it's grouped with the other stalemate validation

PR #1197, opened 2025-05-09
Feat/release gamestate api

### LINKED ISSUE #965: [Feature]: Clean up and RELEASE the GameState API

itsalaidbacklife:
RELEASE THE GAMESTATE API

Cleans up and removes the conditional behavior behind the GameState API making it the default and only boot mode.

## Issue number

- Resolves #965 
- Capstone piece constituting the MVP completion of https://github.com/orgs/cuttle-cards/projects/8

## Please check the following

- For New Features:


src/router.js
```diff
@@ -61,6 +61,13 @@ const checkAndSubscribeToLobby = async (to) => {
   }
 };
 
+const getGameState = (to) => {
+  const gameStore = useGameStore();
+  const gameId = parseInt(to.params.gameId);
+  const gameStateIndex = parseInt(to.query.gameStateIndex ?? -1);
+  return gameStore.requestGameState(gameId, gameStateIndex);
+};
```
Haviles04, line 69: I like this idea, can we do it in spectate as well?

api/errors/notFoundError.js
```diff
@@ -0,0 +1,10 @@
+const CustomErrorType = require('./customErrorType');
+
+class NotFoundError extends Error {
+  constructor(message) {
```
itsalaidbacklife, line 4: It might be reasonable for the constructor to input a resource type (e.g. 'Player', 'Game', and then an id and programmatically construct the message for consistent formatting, but this is easy and flexible

itsalaidbacklife:
> This looks really good, great work! Happy this is finally gonna be released.
> 
> One small question, in `router.js` can we use `getGameState()` for spectate routes as well?

The main thing that's different is we want the spectator view to also log that you're spectating by inserting and/or updating your `UserSpectatingGame` row. So it's basically the same, plus that side effect. Maybe it would make sense to combine the se into one parameterized endpoint? Not sure but that difference is why it's in its own endpoint atm, otherwise the `specate/join` endpoint does the same lookup and sending of the latest game state.

Similarly joining a lobby is basically the same, plus logic for preventing you from joining if the game is full and writing on the game that you've joined. My instinct is that the move might be to create a helper for doing the shared formatting of the data i.e. give it the game and it'll return the latest game state as a socket event so all three endpoints can lean up. Thoughts?

Haviles04:
> > This looks really good, great work! Happy this is finally gonna be released.
> > One small question, in `router.js` can we use `getGameState()` for spectate routes as well?
> 
> The main thing that's different is we want the spectator view to also log that you're spectating by inserting and/or updating your `UserSpectatingGame` row. So it's basically the same, plus that side effect. Maybe it would make sense to combine the se into one parameterized endpoint? Not sure but that difference is why it's in its own endpoint atm, otherwise the `specate/join` endpoint does the same lookup and sending of the latest game state.
> 
> Similarly joining a lobby is basically the same, plus logic for preventing you from joining if the game is full and writing on the game that you've joined. My instinct is that the move might be to create a helper for doing the shared formatting of the data i.e. give it the game and it'll return the latest game state as a socket event so all three endpoints can lean up. Thoughts?

A helper for each different join, called in a centralized endpoint like we did with `move.js`? That's an interesting and cool idea but IMO out of scope for this PR

src/router.js
itsalaidbacklife: The main difference is that spectate also should create a `UserSpectatingGame` entry for the requesting user. We currently do that + get the latest game data via the `/spectate` endpoint. It may make sense to combine those to parameterize one endpoint to handle both cases, but for now keeping them separate is simpler

PR #1202, opened 2025-06-05
fix(LobbyView.vue): navigate to game if readying after game already started

itsalaidbacklife:
Fixes bug where hitting ready after game has already started breaks the lobby. Instead, redirects to the game view if hitting ready after the game has started

## Issue number

- Resolves #

## Please check the following

- For New Features:


src/stores/game.js
```diff
@@ -491,6 +493,8 @@ export const useGameStore = defineStore('game', {
         io.socket.post(`/api/game/${this.id}/ready`, (res, jwres) => {
           if (jwres.statusCode === 200) {
             return resolve(res);
+          } else if (jwres.statusCode === 400 && res.code === 'ALREADY_STARTED') {
+            return resolve(this.id);
```
itsalaidbacklife: I don't love this way of propagating the specific error through to the component from the server, but I'm not sure what would be better. Basically when the game is already started, we want a specific handler that navigates you to the game automatically

src/routes/lobby/LobbyView.vue
```diff
@@ -141,7 +142,19 @@ const opponentUsername = computed(() => gameStore.opponentUsername);
 // Methods
 async function ready() {
   readying.value = true;
-  await gameStore.requestReady();
+  try {
+    await gameStore.requestReady();
+  } catch (err) {
+    // If game has already started; navigate to GameView
+    if (err?.code === 'GAME_ALREADY_STARTED') {
```
itsalaidbacklife: I'd like to pull the string `'GAME_ALREADY_STARTED'` off the existing `CustomErrorType` enum, but I think we'll get hosed by the fact that it's in commonjs and used on the backend. I could try switching that file to json but I think that breaks our ability to autocomplete it anyway...

api/errors/gameAlreadyStartedError.js
```diff
@@ -0,0 +1,11 @@
+const CustomErrorType = require('./customErrorType');
+
+class GameAlreadyStartedError extends Error {
+  constructor(gameId) {
+    super(`Game ${gameId} has already started!`);
+    this.code = CustomErrorType.GAME_ALREADY_STARTED;
+    this.gameId = gameId;
+  }
```
itsalaidbacklife: This might be better as a generic `ConflictError` so it's usable in other situations. I'm unsure (conflicted if you will heh) about it because we kindof _want_ to very specifically detect this problem to give it very specific handling, but maybe the context of when a 409 occurs is sufficient to specify it and the error itself can be more generic

itsalaidbacklife: Swapped it for `ConflictError`. We can see whether this becomes something useful in other situations, but it's more consistent with the other errors being associated with specific status codes and otherwise generic to that status

PR #1203, opened 2025-06-05
Feat/rewatch finished games

### LINKED ISSUE #63: Enable Watchable Replays
### Feature Summary

It would be fun to be able to watch games that have already happened as a replay able experience

### Detailed Description

You could watch your own games or tournament games, stepping forward and backward through time

itsalaidbacklife:
Enables rewatching finished matches! Supports:
- Step forward
- Step backward
- Skip forward
- Skip backward
- navigating to the next rematch game

Builds it in as an expansion/generalization of spectating. Intent is to support live + post-hoc spectating similarly.
## Issue number

- Resolves #63 

## Please check the following

- For New Features:


api/controllers/game/move.js
```diff
@@ -18,7 +20,7 @@ module.exports = async function (req, res) {
     // Validate Request //
     //////////////////////
     // Game must be in progress
-    if (!game.gameStates.length || !gameState) {
+    if (!game?.gameStates?.length || !gameState) {
```
itsalaidbacklife, line 30: TODO: It would be a good idea to throw a separate `NotFoundError` if the game isn't found

api/controllers/game/spectate/join.js
```diff
@@ -44,15 +49,31 @@ module.exports = async function (req, res) {
...
+          message: `Could not find game state ${gameStateIndex} for game ${game.id}`
+        });
+    }
+    const gameState = unpackGamestate(game.gameStates.at(gameStateIndex));
     const socketEvent = await createSocketEvent(game, gameState);
     Game.publish([ game.id ], socketEvent);
 
 
-    return res.ok(socketEvent.game);
+    return res.ok(socketEvent);
```
itsalaidbacklife, line 71: EXP: Switched from returning just the `game` to returning the entire socket event so that we can process transitions and overlays etc

api/helpers/game-states/get-log.js
itsalaidbacklife, line 1: EXP: Added log messages for previously missing move types. This removes the issue of having empty log messages and clarifies when things like concede happen more explicitly

src/components/BaseSnackbar.vue
```diff
@@ -10,6 +10,7 @@
     class="base-snackbar"
     position="fixed"
     location="bottom"
+    z-index="2412"
```
itsalaidbacklife, line 13: EXP: I feel awkward about these massive hard-coded z-indices, but this is one bigger than the z-index used in the menu and playback controls in order to have both these appear on top of everything else in vuetify e.g. dialogs and overlays, and that the snackbars appear on top of the playback controls as well

src/plugins/sockets/connectivityEvents.js
```diff
@@ -16,7 +19,6 @@ export async function handleConnect() {
       const response = await gameStore.requestGameState(gameId, gameStateIndex);
       if (response?.victory?.gameOver && response.game.rematchGame) {
         await gameStore.requestGameState(response.game.rematchGame);
-        gameStore.myPNum = (gameStore.myPNum + 1) % 2;
```
itsalaidbacklife, line 19: TODO: consolidate all  references to `myPNum =` to clean up redundancy

Moved the assignment of `gameStore.myPNum` into the actual gameStore methods for `requestGameState()` and `requestSpectate()`.

src/plugins/sockets/connectivityEvents.js
```diff
@@ -1,11 +1,14 @@
 import { useAuthStore } from '@/stores/auth';
 import { useGameStore } from '@/stores/game';
+import { useGameHistoryStore } from '@/stores/gameHistory';
 import router from '@/router.js';
 import { ROUTE_NAME_GAME, ROUTE_NAME_SPECTATE, ROUTE_NAME_HOME } from '@/router';
 
 export async function handleConnect() {
   const authStore = useAuthStore();
   const gameStore = useGameStore();
+  const gameHistoryStore = useGameHistoryStore();
```
itsalaidbacklife, line 10: EXP: Bringing this in to include the current game state index when requesting to spectate so you stay on the correct state

src/plugins/sockets/inGameEvents.js
```diff
@@ -34,22 +35,18 @@ export async function handleInGameEvents(evData, newRoute = null) {
...
     }
-    case SocketEvent.DEAL: {
-      // Update state
-      if (isSpectating) {
-        gameStore.myPNum = 0; // always spectate as p0
-        gameStore.isSpectating = true;
-      }
-      gameStore.resetPNumIfNullThenUpdateGame(evData.game);
-      break;
-    }
```
itsalaidbacklife, line 45: TODO: clean these up to consolidate same-handled states and to ensure they all reset the pNum if null

Now that we trigger inGameEvents for potentially every state state, we should no longer separate deal as its own case. We should update every case to `resetPNumIfNullThenUpdateGame()`.

api/controllers/game/spectate/join.js
```diff
@@ -44,15 +49,31 @@ module.exports = async function (req, res) {
...
 
     const { unpackGamestate, createSocketEvent } = sails.helpers.gameStates;
-    const gameState = unpackGamestate(game.gameStates.at(-1));
+    // Default to first gamestate for finished games, last for live ones
+    const gameStateIndex = Number(
+      req.query.gameStateIndex ?? (
+        [ GameStatus.FINISHED, GameStatus.ARCHIVED ]
+          .includes(game.status) ? 0 : -1
+      )
+    );
```
itsalaidbacklife: DECIDE: This is kindof a chonker of an expression. Not sure how clearly it reads. Intent is to take the query param value if present, then otherwise coalesce to 0 if the game is finished and -1 if it's ongoing

api/helpers/game-states/create-socket-event.js
```diff
@@ -56,6 +56,19 @@ module.exports = {
...
+        return _.pick(priorGameState, [
+          'createdAt',
+          'moveType',
+          'playedBy',
+          'playedCard',
+          'targetCard',
+          'discardedCards',
+          'resolved',
+        ]);
+      });
```
itsalaidbacklife, line 70: EXP: Including these is how we tell if we're on the last state. Eventually it'll also be used to move the game log client side

src/plugins/sockets/inGameEvents.js
```diff
@@ -160,20 +157,24 @@ export async function handleInGameEvents(evData, newRoute = null) {
 
       const { gameId: oldGameId } = targetRoute.params;
 
+      const route = {
+        name: targetRoute.name,
+        params: {
+          gameId: evData.gameId,
+        },
+      };
```
itsalaidbacklife, line 164: EXP: Pulling the route into a var like this before we push lets us request to the server to spectate before we navigate to the new route. That lets us figure out what the gameStateIndex should be i.e. is the new game finished or ongoing?

src/router.js
```diff
@@ -83,6 +83,46 @@ const getGameState = async (to) => {
...
 
+const setupSpectate = async (to) => {
+  const gameStore = useGameStore();
+  let { gameId } = to.params;
+  gameId = Number(gameId);
+  const { gameStateIndex } = to.query;
+  const isValidGameStateIndex = 
+    gameStateIndex !== undefined && 
+    Number.isInteger(Number(gameStateIndex)) && 
+    (Number(gameStateIndex) === -1 || Number(gameStateIndex) >= 0);
```
itsalaidbacklife: TODO: Should investigate the places we're cleaning/processing a gameStateIndex to centralize and standardize them. I feel like I've sloppily just put these checks all over the place inconsistently

src/router.js
```diff
@@ -83,6 +83,46 @@ const getGameState = async (to) => {
...
+  try {
+    await gameStore.requestSpectate(gameId, gameStateIndex, to);
+    gameStore.id = gameId;
```
itsalaidbacklife: TODO: I hope this isn't necessary anymore since we should be setting this during requestSpectate
```suggestion
```

src/router.js
```diff
@@ -198,11 +239,11 @@ const router = createRouter({
   },
 });
 
-router.beforeEach(async (to, _from, next) => {
+router.beforeEach(async (_to, _from, next) => {
   const authStore = useAuthStore();
   // Make sure we try and reestablish a player's session if one exists
   // We do this before the route resolves to preempt the reauth/logout logic
-  await authStore.requestStatus(to);
```
itsalaidbacklife, line 205: EXP: Removed the route-specific handling of `requestStatus()` so it can now focus entirely on auth and let the page-specific nav guards decide on any page-specific logic

src/routes/game/GameView.vue
```diff
@@ -342,7 +342,7 @@
             :class="{ 'my-turn': gameStore.isPlayersTurn }"
           >
             <UsernameToolTip
-              v-if="$vuetify.display.smAndUp"
+              v-if="$vuetify.display.smAndUp && !gameHistoryStore.showPlaybackControls"
```
itsalaidbacklife, line 345: EXP: I added a `UsernameToolTip` to the playback controls themselves so that the username can appear where it normally does but on top of the playback controls while they're being used. To avoid duplicates, we remove this one when that's the case

src/routes/game/GameView.vue
```diff
@@ -417,6 +417,7 @@
         @target="beginTargeting"
       />
       <GameDialogs @clear-selection="clearSelection" @handle-error="handleError" />
+      <PlaybackControls v-if="gameHistoryStore.showPlaybackControls" />
```
itsalaidbacklife, line 420: EXP: Instead of emitting events, the `PlaybackControls` fully manage transitions between the states themselves. This way it keeps the amount of code necessary in the already-beefy `GameView` to a minimum.

src/routes/game/GameView.vue
```diff
@@ -502,17 +505,17 @@ export default {
...
+    ////////////////////////////////////////////
     deck() {
       return this.gameStore.deck;
     },
     scrap() {
       return this.gameStore.scrap;
     },
     logs() {
-      return this.gameStore.log;
+      return this.gameHistoryStore.log;
```
itsalaidbacklife, line 518: EXP: Moved the log into the `gameHistoryStore` so that we could compute how many logs to show based on the current gameStateIndex. This will also make it easier to move the log clientside in the future

src/routes/game/components/PlaybackControls.vue
```diff
@@ -0,0 +1,101 @@
...
+    query: {
+      ...route.query,
+      gameStateIndex,
+    },
+  });  
+}
+
+watch(() => gameHistoryStore.currentGameStateIndex, async (newVal) => {
+  await gameStore.requestGameState(gameStore.id, newVal);
+  gameStore.myPNum = 0;
```
itsalaidbacklife: TODO: Double check if we need this; I think we do since requestGameState doesn't set your pNum if you aren't in the game

itsalaidbacklife: But maybe it will if `resetPNumIfNull` correctly sets it for spectators

src/routes/game/components/PlaybackControls.vue
itsalaidbacklife, line 1: EXP: Overall I'm really pleased with how this component came out. 100 lines (including css) for a pretty robust feature set, and I think it's fairly easy to read

src/routes/game/components/dialogs/components/gameOver/GameOverDialog.vue
```diff
@@ -13,7 +13,7 @@
         :stalemates="stalemates"
         :latest-result="latestResult"
         :is-ranked="isRanked"
-        :is-spectating="gameStore.isSpectating"
+        :is-spectating="gameHistoryStore.isSpectating"
```
itsalaidbacklife, line 16: DECIDE: This change is maybe the most awkward in this PR. I wanted `isSpectating` to be computed using the current route so that we don't have to keep setting it manually. You're spectating if and only if the current route is the spectate view, so it should logically be computed from the route.

But the `gameStore` doesn't allow the use of the router atm because it's created as an options API store instead of a setup store. So in order to make this computed property and put it into the store, I put it inside the `gameHistoryStore` instead of the `gameStore` simply because it's a new store I could write in the setup store syntax to use composables i.e. `useRoute`.

Another alternative could be to move the `isSpectating` computed into its own composable at the top level. I don't see a strong reason for one over the other, but that might be less 'random' than having it in the `gameHistoryStore`

src/routes/game/components/dialogs/components/gameOver/GameOverDialog.vue
```diff
@@ -291,14 +293,29 @@ export default {
...
+            ...route,
+            query: {
+              gameStateIndex,
+            }
+          });
+        } catch (err) {
+          this.$router.push({
+            name: 'Home',
+            query: { gameId: rematchGameId, error: err?.message ?? err ?? `Could not spectate game ${rematchGameId}`
+            } });
```
itsalaidbacklife: TODO: Whitespace is a little whack here

src/routes/home/components/GameListItem.vue
```diff
@@ -143,17 +143,7 @@ export default {
...
-        .then(() => {
-          this.joiningGame = false;
-          this.$router.push(`/spectate/${this.gameId}`);
-        })
-        .catch((error) => {
-          this.joiningGame = false;
-          this.$emit('error', this.t(error?.message ?? error));
-          this.gameListStore.gameFinished(this.gameId);
-        });
+      this.$router.push(`/spectate/${this.gameId}?gameStateIndex=-1`);
```
itsalaidbacklife, line 146: EXP: All this logic to set up spectating is now handled in the router nav guard so we don't need it in the component; just push to the new route

src/stores/auth.js
```diff
@@ -99,12 +94,6 @@ export const useAuthStore = defineStore('auth', {
           this.authSuccess(username);
         }
 
-        const gameStore = useGameStore();
-        if (isSpectating) {
-          const { gameId } = route.params;
-          gameStore.requestSpectate(Number(gameId));
-        }
```
itsalaidbacklife, line 106: EXP: Finally removed this route-specific clutter so `requestStatus()` can be purely auth

src/stores/game.js
```diff
@@ -223,9 +225,9 @@ export const useGameStore = defineStore('game', {
       this.p1Ready = newGame.p1Ready ?? this.p1Ready;
       this.passes = newGame.passes ?? this.passes;
       this.players =
-        newGame.players?.map((player) =>
-          setPlayers(player, this.myPNum, this.hasGlassesEight, this.isSpectating),
-        ) ?? this.players;
+      newGame.players?.map((player) =>
+        setPlayers(player, this.myPNum, this.hasGlassesEight, this.isSpectating),
+      ) ?? this.players;
```
itsalaidbacklife: TODO: Looks like this should indent another level

src/stores/game.js
```diff
@@ -236,7 +238,12 @@ export const useGameStore = defineStore('game', {
       this.currentMatch = newGame.currentMatch ?? this.currentMatch;
       this.p0Rematch = newGame.p0Rematch ?? null;
       this.p1Rematch = newGame.p1Rematch ?? null;
+      this.rematchGameId = newGame.rematchGame ?? null;
       this.gameIsOver = newGame.gameIsOver ?? false;
+      this.status = newGame.status ?? GameStatus.ARCHIVED;
+    
+      const gameHistoryStore = useGameHistoryStore();
+      gameHistoryStore.gameStates = newGame.gameStates ?? [];
```
itsalaidbacklife, line 246: EXP: setting `rematchGameId`, `status`, and `gameStates` on every call to `updateGame()` to keep them up to date with each update when navigating across states during replays, instead of relying on only specific socket events to update those properties

src/stores/game.js
```diff
@@ -277,12 +284,16 @@ export const useGameStore = defineStore('game', {
...
+      const gameHistoryStore = useGameHistoryStore();
       // Set my pNum if it is null
       if (this.myPNum === null) {
         let myPNum = game.players.findIndex(({ username }) => username === authStore.username);
         if (myPNum === -1) {
           myPNum = null;
         }
+        if (gameHistoryStore.isSpectating) {
+          myPNum = 0;
+        }
```
itsalaidbacklife, line 296: TODO: Reinvestigate places we set `myPNum` so that we can consolidate them. I _think_ this should do for most cases and we may not need other assignments

src/stores/game.js
```diff
@@ -436,6 +447,7 @@ export const useGameStore = defineStore('game', {
         io.socket.get(`/api/game/${gameId}?gameStateIndex=${gameStateIndex}`, (res, jwres) => {
           switch (jwres.statusCode) {
             case 200:
+              this.resetState();
```
itsalaidbacklife, line 450: EXP: resetting state clears out dialogs and overlays. Need to be careful to ensure id, and myPNum always get set

src/stores/gameHistory.js
```diff
@@ -0,0 +1,81 @@
+// stores/gameHistoryStore.js
```
itsalaidbacklife:
```suggestion
```

src/stores/gameHistory.js
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  const log = computed(() => {
+    return currentGameStateIndex.value === -1 ? 
+      [ ...gameStore.log ] : 
+      gameStore.log.slice(0, currentGameStateIndex.value + 1);
+  });
+
+  const showPlaybackControls = computed(() => {
+    return isSpectating.value && [ GameStatus.FINISHED, GameStatus.ARCHIVED ].includes(gameStore.status);
```
itsalaidbacklife, line 57: WRITE UP: issue for adding playback controls to live spectating

Currently it seemed easier to only allow playback controls during finished games. It might be workable for live ones in the future, but I figure we can try that out once it's up and running for finished games

tests/e2e/specs/out-of-game/home.spec.js
```diff
@@ -291,13 +281,9 @@ describe('Home - Game List', () => {
...
         cy.get(`[data-cy-spectate-game=${gameId}]`).click();
-        assertSnackbar('Unable to spectate game', 'error', 'newgame');
-        // Spectate button should now be disabled
-        cy.get(`[data-cy-spectate-game=${gameId}]`).should('be.disabled');
+        cy.url().should('contain', `/spectate/${gameId}?gameStateIndex=-1`);
+        cy.get('[data-player-hand-card]').should('have.length', 5);
       });
-      // Refresh page -- no games available to spectate
-      cy.reload();
-      cy.get('[data-cy=no-spectate-game-text]').should('contain', 'No Games Available to Spectate');
```
itsalaidbacklife, line 300: TODO: Should replace this with navagating to spectate-list then making the same assertion

tests/e2e/support/commands.js
```diff
@@ -209,7 +210,7 @@ Cypress.Commands.add('signupOpponent', (opponent) => {
   cy.makeSocketRequest('user', 'signup', {
     username: opponent.username,
     password: opponent.password,
-  });
+  }).as(`${opponent.username}Id`);
```
itsalaidbacklife, line 213: EXP: This lets us get the id of the players we sign up

tests/e2e/support/commands.js
```diff
@@ -773,12 +778,19 @@ Cypress.Commands.add('vueRoute', (route) => {
  *   deck?: {suit: number, rank: number}[] deletes all cards except these from the deck
  * }
  */
-Cypress.Commands.add('loadGameFixture', (pNum, fixture) => {
-  return cy
+Cypress.Commands.add('loadGameFixture', (pNum, fixture, gameId = null) => {
```
itsalaidbacklife, line 781: EXP: Added `gameId` param for loading fixtures in games you're not watching atm (to join as spectator later)

tests/unit/fixtures/gameStates/points.js
```diff
@@ -200,6 +200,16 @@ export const points = {
...
+      gameStates:  [
+        {
+          discardedCards:  [],
+          moveType: MoveType.POINTS,
+          playedBy: 0,
+          playedCard: 'TS',
+          resolved: null,
+          targetCard: null,
+        },
+      ],
```
itsalaidbacklife, line 212: EXP: Added `.gameStates` to the socketEvent; need to expect it in the unit tests

tests/e2e/specs/in-game/spectating/replays.spec.js
```diff
@@ -0,0 +1,504 @@
+import dayjs from 'dayjs';
+import utc from 'dayjs/plugin/utc';
+
```
itsalaidbacklife:
```suggestion
```

src/stores/gameHistory.js
```diff
@@ -0,0 +1,80 @@
...
+    return gameStates.value.at(currentGameStateIndex.value) ?? null;
+  });
+
+  const priorGameStates = computed(() => {
+    return currentGameStateIndex.value === -1 ? 
+      [ ...gameStates.value ] : 
+      gameStates.value.slice(0, currentGameStateIndex.value + 1);
+  });
+
+  const log = computed(() => {
```
itsalaidbacklife, line 50: WRITE UP: issue to translate log and move it clientside

src/stores/game.js
```diff
@@ -451,15 +463,16 @@ export const useGameStore = defineStore('game', {
...
+    async requestSpectate(gameId, gameStateIndex = 0, route = null) {
+      const slug = `${gameId}/spectate?gameStateIndex=${gameStateIndex}`;
       try {
-        const res = await this.makeSocketRequest(slug, { gameId });
+        this.resetState();
+        const res = await this.makeSocketRequest(slug, {});
         this.myPNum = 0;
-        this.isSpectating = true;
-        this.updateGame(res.body);
+        this.id = gameId;
```
itsalaidbacklife: TODO: is this necessary? Likewise do we need to set myPNum = 0 or is it enough to handle it in `updateGame` via `resetPNumIfNull`?

src/stores/gameHistory.js
itsalaidbacklife: #1213

src/stores/gameHistory.js
itsalaidbacklife: #1214

api/errors/conflictError.js
```diff
@@ -1,8 +1,8 @@
 const CustomErrorType = require('./customErrorType');
 
 class ConflictError extends Error {
-  constructor(resourceType, resourceId) {
-    super(`${resourceType} ${resourceId} can't be updated like that right now`);
+  constructor(resourceType, resourceId, message = '') {
+    super(message ?? `${resourceType} ${resourceId} can't be updated like that right now`);
```
itsalaidbacklife, line 5: EXP: Now allowing consumers to specify message to overwrite default

api/controllers/game/spectate/join.js
Haviles04: Idk if this is any better, The original isn't terrible to read but it is a chunky boi

```suggestion
    const gameStateIndex = Number(req.query.gameStateIndex)?? game.status >= GameStatus.finished ? 0 : -1  
```

src/router.js
```diff
@@ -83,6 +83,46 @@ const getGameState = async (to) => {
   return;
 };
 
+const setupSpectate = async (to) => {
+  const gameStore = useGameStore();
+  let { gameId } = to.params;
+  gameId = Number(gameId);
```
Haviles04:
```suggestion
const gameId = Number(to.params.gameId)
```

Haviles04: Is there some sort of blocker for this?

src/router.js
Haviles04: Could we just call `const gameStateIndex = Number(to.query.gameStateIndex)` and then `if(!gameStateIndex)` since `Nan` would be falsey... or is 0 a valid gamestateindex

itsalaidbacklife: 0 is a valid gameState index; it means go to the beginning. That's where we want to land when going to a finished game and it's where the skip-backward button brings you. That said there might still be a way to simplify this nicely

src/router.js
itsalaidbacklife: Can't think of a reason not to. I was hung up on using destructuring but this is better.

src/router.js
Haviles04:
```
const gameStateIndex  = Number(to.query.gameStateIndex);
const isValidGameStateIndex = Number.isInteger(gameStateIndex) && gameStateIndex >= -1 
```

Maybe something like this?

src/stores/game.js
itsalaidbacklife: Deleted all of them except `resetPNum()` here and in `requestSpectate()` itself. Need them in both to ensure it works when requesting the next game state with `requestGameState()` (that will trigger this conditional) and in `requestSpectate()` for when you first load the page

src/router.js
itsalaidbacklife: Yeah that looks equivalent and more concise

src/plugins/sockets/connectivityEvents.js
itsalaidbacklife: Done. Had to keep it both in `requestSpectate()` for initial page load and in resetPNum for updates via requestGameState()

src/router.js
itsalaidbacklife: Rolled this out to the front and backend places using this query param

api/controllers/game/spectate/join.js
itsalaidbacklife: Refactored to take same approach of separate `isGameStateIndexValid` variable. Also adjusted to prevent getting `NaN`

src/routes/game/components/dialogs/components/gameOver/GameOverDialog.vue
itsalaidbacklife: Going to just keep it in gameHistoryStore as the path of least resistance

itsalaidbacklife:
Failure is confirmed to be the flaky 5's spec; merging!

PR #1217, opened 2025-06-14
Fix/request spectate after nine

### LINKED ISSUE #1212: [Bug]: Cards become unfrozen after requesting a stalemate
### Bug Summary

Requesting a stalemate will unfreeze any frozen cards

### Detailed Description

Currently whether a card is frozen by a 9 (and unable to be played next turn) is determined by whether the current gamestate's latest move was to resolve a 9 and target that card. 

But this is inadequate because requesting a stalemate creates a new game state (so does rejecting one) which will therefore cause `unpack-gamestate.js` to compute that no cards are frozen. We need to update how each card's `isFrozen` attribute is computed in order to account for the possibility that a stalemate may have been requested and/or rejected as the previous move type.

### Current Behavior

Requesting a stalemate after a 9 resolves will cancel out the card being frozen

### Expected Behavior

Requesting and/or rejecting a stalemate after resolving a 9 preserves the fact that the 9 is frozen and can't be played next turn

### Steps to Reproduce

1. Start a game
2. Play a 9
3. Resolve the 9
4. Request a stalemate
5. The card that was targeted by the 9 is no longer frozen and can now be played even though this should not be allowed

itsalaidbacklife:
## Issue number

- Resolves #1212

## Please check the following

- For New Features:

### DIFF
--- a/api/helpers/game-states/moves/stalemate-reject/execute.js
+++ b/api/helpers/game-states/moves/stalemate-reject/execute.js
@@ -42,11 +42,7 @@ module.exports = {
       ...result,
       ...requestedMove,
       playedBy,
-      phase,
-      playedCard: null,
-      targetCard: null,
-      discardedCards: [],
-      resolved: null,
+      phase
     };
 
     return exits.success(result);
--- a/api/helpers/game-states/unpack-gamestate.js
+++ b/api/helpers/game-states/unpack-gamestate.js
@@ -1,5 +1,3 @@
-const MoveType = require('../../../utils/MoveType.json');
-
 module.exports = {
   friendlyName: 'unpack GameState',
 
--- a/api/helpers/game-states/unpack-gamestate.js
+++ b/api/helpers/game-states/unpack-gamestate.js
@@ -44,7 +44,7 @@ module.exports = {
             ? value.map((cardStr) => {
               // Is the card frozen
               const freezeCard =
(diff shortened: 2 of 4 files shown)


api/helpers/game-states/moves/stalemate-reject/execute.js
```diff
@@ -42,11 +42,7 @@ module.exports = {
       ...result,
       ...requestedMove,
       playedBy,
-      phase,
-      playedCard: null,
-      targetCard: null,
-      discardedCards: [],
-      resolved: null,
```
itsalaidbacklife, line 49: Removing these so that they can keep their same value from the previous move. This lets the check on whether a card is frozen read from these properties to identify if they should still be frozen after a resolve 9

api/helpers/game-states/unpack-gamestate.js
```diff
@@ -44,7 +44,7 @@ module.exports = {
             ? value.map((cardStr) => {
               // Is the card frozen
               const freezeCard =
-                  gameStateRow.moveType === MoveType.RESOLVE &&
+                  !!gameStateRow.resolved &&
```
itsalaidbacklife, line 45: This was the key fix. The moveType might be REQUEST_STALEMATE or REQUEST_STALEMATE but still have just resolved a 9

PR #1223, opened 2025-06-17
Refactor/use phase clientside

### LINKED ISSUE #1222: [DevEx]:  refactor the logic for displaying dialogs and overlays to use the game phase
### Improvement Summary

We should refactor the way the modal elements are controlled on/off to utilize the game's `phase` ref as a central source of truth.

### Detailed Description

Currently all the modal elements in the GameView display based on individual refs which are toggled on/off whenever events are processed. This means the same piece of data e.g. `waitingForOpponentToCounter` need to be set and unset based on different events, causing the opportunity for desync.

The GameState API uses a `phase` attribute to designate which state we're in and therefore which moves are allowed on the backend. If we send this to the client, we can compute which if any modal elements to display based on that data as well as who's turn it is, the current number of two's, etc. 

This will pave the way for additional improvements such as showing the relevant player's username instead of the word 'Opponent'

itsalaidbacklife:
## Issue number

- Resolves #1222

## Please check the following

- For New Features:

### DIFF
--- a/tests/unit/fixtures/gameStates/points.js
+++ b/tests/unit/fixtures/gameStates/points.js
@@ -134,6 +134,7 @@ export const points = {
       p1Ready: true,
       p0Rematch: null,
       p1Rematch: null,
+      phase: GamePhase.MAIN,
       turnStalemateWasRequestedByP0: -1,
       turnStalemateWasRequestedByP1: -1,
       lock: null,
--- a/tests/unit/fixtures/gameStates/resolveNine.js
+++ b/tests/unit/fixtures/gameStates/resolveNine.js
@@ -124,6 +124,7 @@ export const resolveNine = {
       p0Rematch: null,
       p1Ready: true,
       p1Rematch: null,
+      phase: GamePhase.MAIN,
       passes: 0,
       players: [
         {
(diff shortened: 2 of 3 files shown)


src/plugins/sockets/inGameEvents.js
```diff
@@ -46,92 +46,36 @@ export async function handleInGameEvents(evData, newRoute = null) {
...
-      gameStore.resetPNumIfNullThenUpdateGame(evData.game);
-      if (evData.pNum !== gameStore.myPNum) {
-        gameStore.waitingForOpponentToCounter = false;
-        gameStore.myTurnToCounter = true;
-      } else {
-        gameStore.waitingForOpponentToCounter = true;
-        gameStore.myTurnToCounter = false;
-      }
-      break;
-    // Sevens
```
itsalaidbacklife, line 115: Consolidating all of these into a single case that simply updates the game was a primary motivator for this refactor. When all these variables were separate pieces of state, they each have to be set and unset at the right times abased on the specific event and who did the most recent thing. 



Now that all the modal-controls are defined as computed properties derived from the singular game phase, the possibility for desync is reduced and the handling is more consistent with fewer gotchas like "oh missed a place where X needs to be set". Generally, if a piece of data _can_ be computed, it probably should be computed so that you don't have to ensure it's properly set to the right value all the time with separate side effects

PR #1228, opened 2025-06-18
Feat/log in as spectator

### LINKED ISSUE #1221: [Feature]: Clean up the spectator and replay experience
### Feature Summary

We should make a number of improvements to the spectator experience with a particular focus on making the experience of rewatching finished games as smooth as possible

### Detailed Description

- refactor the logic for displaying dialogs and overlays to use the game phase

- enable logging in on the spectator view when navigating directly to a spectate link while unauthenticated 

- revise the overlay text to show the opponent’s username 

- update the spectator experience to replace all dialogs with the corresponding overlays e.g. “aleph_one may counter”

- update the logic for setting spectator pNum to allow you to spectate as p1

- update the logic for setting spectator pNum to default to whichever player you were in that game if it isn’t specified

itsalaidbacklife:
Enables users to navigate directly to a highlight clip/spectate link and log in from the GameView if not already authenticated
## Issue number

- Resolves #1221

## Please check the following

- For New Features:

### DIFF
--- a/src/stores/game.js
+++ b/src/stores/game.js
@@ -506,6 +506,7 @@ export const useGameStore = defineStore('game', {
         this.updateGame(res.body.game);
         return handleInGameEvents(res.body, route);
       } catch (err) {
+        // Swallow 401 error so we can authenticate from GameView
         if (authStore.mustReauthenticate) {
           this.id = gameId;
           return;
--- a/src/stores/game.js
+++ b/src/stores/game.js
@@ -497,6 +497,7 @@ export const useGameStore = defineStore('game', {
     },
 
     async requestSpectate(gameId, gameStateIndex = 0, route = null) {
+      const authStore = useAuthStore();
       const slug = `${gameId}/spectate?gameStateIndex=${gameStateIndex}`;
       try {
         this.resetState();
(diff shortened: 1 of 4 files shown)


src/stores/game.js
```diff
@@ -505,6 +506,11 @@ export const useGameStore = defineStore('game', {
         this.updateGame(res.body.game);
         return handleInGameEvents(res.body, route);
       } catch (err) {
+        // Swallow 401 error so we can authenticate from GameView
+        if (authStore.mustReauthenticate) {
+          this.id = gameId;
+          return;
+        }
```
itsalaidbacklife, line 513: I tried to refactor `requestGameState()` to work the same way but it failed some of the reconnecting tests. I wrote up #1229 to make that func consistent w/ this one so we can get this improvement up and running

src/stores/gameHistory.js
```diff
@@ -28,8 +28,7 @@ export const useGameHistoryStore = defineStore('gameHistory', () => {
     if (
       isNaN(index) || 
       !isFinite(index) || 
-      index < 0 || 
-      index >= gameStates.value.length
```
itsalaidbacklife, line 32: Annoyingly I removed this constraint to deal with the reauthenticating case, because you'll have the query param with `gameStateIndex` but not have any gameStates yet since your request was denied. We want to be able to grab the gameStateIndex from the url (ideally via this computed prop) before we've actually got the gameStates in place. That means we could have weirdness with out-of-bounds indices, but let's cross that bridge later

PR #1232, opened 2025-06-23
Fix/p0 leaves first

itsalaidbacklife:
Fix bug where p0 leaving a lobby before p1 does would cause players joining the lobby to see themselves as the opponent:

image

Root cause was that the backend was failing to sort the players in the game by pNum

## Issue number

- Resolves #

## Please check the following

- For New Features:

### DIFF
--- a/tests/e2e/specs/out-of-game/lobby.spec.js
+++ b/tests/e2e/specs/out-of-game/lobby.spec.js
@@ -185,7 +185,7 @@ describe('Lobby - P0 Perspective', () => {
       });
   });
 
-  it.only('readying, leaving and joining back', () => {
+  it('readying, leaving and joining back', () => {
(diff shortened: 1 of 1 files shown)


api/controllers/game/join.js
```diff
@@ -58,6 +58,7 @@ module.exports = async function (req, res) {
     }
     
     game.players.push({ username: user.username, pNum });
+    game.players.sort((p1, p2) => p1.pNum - p2.pNum);
```
itsalaidbacklife, line 61: Previously I assumed that if the player joining was p0 there would not be a p1, but that's not the case if p0 leaves the game and then someone joins again. In that case, p1 will be added first on line 30, then p0 will be added on line 60, so we need to sort after we have both players in the array

PR #1240, opened 2025-07-01
Fix/pass stalemate replay

itsalaidbacklife:
Fixes a bug where watching a replay of a finished game that ended in a stalemate via passes would always show the game over dialog for every turn, instead of waiting until the last state.

[Reported here](https://discord.com/channels/833694928684187669/833700472774852628/1387643073961656461)

## Issue number

N/A

## Please check the following

- For New Features:


config/env/staging.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
    ***************************************************************************/
 
   models: {
-    migrate: 'alter',
+    migrate: 'safe',
```
itsalaidbacklife, line 26: Switching to migrate: safe to make it easier to preserve data on boot

api/helpers/game-states/unpack-gamestate.js
```diff
@@ -64,6 +64,7 @@ module.exports = {
       };
 
       const data = {
+        id: gameStateRow.id,
```
itsalaidbacklife, line 67: Added the id of from the packed GameStateRow obj onto the unpacked GameState obj so that we could compare 'which' gameState we're talking about in `create-socket-event`. This lets us identify the correct squence of gamestates to check when counting consecutive passes

api/helpers/game-states/pack-gamestate.js
```diff
@@ -74,6 +74,9 @@ module.exports = {
 
       const combinedData = { ...gameState, ...convertedData };
 
+      // Remove the id field to prevent database duplicate id errors
+      delete combinedData.id;
```
itsalaidbacklife, line 78: Need to delete the `id` from the unpacked GameState object so that the backend doesn't attempt to write the id field into the newly inserted GameStateRow when saving

api/helpers/game-states/create-socket-event.js
```diff
@@ -29,7 +29,10 @@ module.exports = {
 
       const countPasses = (function () {
         let numPasses = 0;
-        for (const gameState of game.gameStates.slice(-3)) {
+        const currentIndex = game.gameStates.findIndex(gs => gs.id === gameState.id);
+        const threeUpToCurrent = game.gameStates.slice(Math.max(0, currentIndex - 2), currentIndex + 1);
+
+        for (const gameState of threeUpToCurrent) {
```
itsalaidbacklife, line 35: This was the crux of the fix. Previously, we always looked at the 'latest three' game states to count passes, but for replays, we need to look at the 'three states leading up to the one we're looking at'. Doing _that_ requires the ability to identify which gamestate is which, so I've added the id from the gamestate row onto the game state obj via `pack-gamestate`

api/controllers/game/move.js
```diff
@@ -57,7 +57,7 @@ module.exports = async function (req, res) {
     const updatedState = execute(gameState, req.body, playedBy, game.gameStates);
     const gameStateRow = await saveGamestate(updatedState);
     game.gameStates.push(gameStateRow);
-    const socketEvent = await createSocketEvent(game, updatedState);
+    const socketEvent = await createSocketEvent(game, { ...updatedState, id: gameStateRow.id });
```
itsalaidbacklife, line 60: Need to get the id for the just-created gamestate from the obj saved to the db, so we spread the in-memory obj with the id from the db

tests/e2e/specs/in-game/spectating/replayHelpers.js
itsalaidbacklife, line 1: replays.spec.js was getting big, so I pulled these into their own file. It might make sense to move them into fixtures because they mainly set up data, but they also make assertions.

It might make sense to put them in the global /helpers file, but they're all limited to specifically be used in `replays.spec.js` so I've colocated them as sibling files

itsalaidbacklife, line 1: Pulled these helpers into their own file

tests/e2e/specs/in-game/spectating/replays.spec.js
```diff
@@ -803,6 +386,65 @@ describe('Rewatching finished games', () => {
         cy.get('[data-cy=clip-highlight]').should('be.visible');
       });
     });
+
+    describe('Replaying other game states', () => {
+      it('Watches a replay of a game that ends via passes', () => {
```
itsalaidbacklife, line 391: New test case to verify the fix

PR #1241, opened 2025-07-02
Haviles04/discord auth

Haviles04:
## Issue number

- Resolves #

## Please check the following

- For New Features:

TODO: 
- ~~UI update~~
- ~~Fix Unit Test~~
- ~~Form Validation/Rules~~
- ~~More translations~~
- ~~HTTP/HTTPS ENV normalization~~


api/models/User.js
```diff
@@ -13,7 +13,6 @@ module.exports = {
     },
     encryptedPassword: {
       type: 'string',
-      required: true,
```
Haviles04, line 16: @itsalaidbacklife This is one thing I was unsure of. Since we don't really have a password for discord. I'm unsure of what to do in this situation

package.json
```diff
@@ -77,7 +78,7 @@
     "docker:stop": "docker compose -f ./docker/docker-compose.yml down",
     "start:serve": "vite preview",
     "start:client": "vite --host",
-    "start:server": "node app.js",
+    "start:server": "VITE_JWT_SECRET=cuttle69420 VITE_DISCORD_CLIENT_ID=1385321376889438238 VITE_DISCORD_CLIENT_SECRET=BflVlmWGk_6FnCsAkzr5aHQaJmn_RoUa node app.js",
```
itsalaidbacklife: We need a system for pulling these out of version control and then to regenerate them. It likely makes sense for them to be set via `.env.local` which is already `.gitignore`'d. The trick is we'd need:
1. a way of getting sails to pickup dotenv values
2. a way of picking them up in CI if we test this feature e2e i.e. defining them as secrets on the github repo and pulling them into the github action
3. To set them in prod via heroku env vars

api/controllers/UserController.js
```diff
@@ -88,16 +87,92 @@ module.exports = {
...
+
+  discordRedirect: async function(_, res) {
+    const url = new URL('https://discord.com/api/oauth2/authorize');
+    const { generateSecret } = sails.helpers.oauth;
+    const state = generateSecret();
+
+    url.searchParams.set('response_type', 'code');
+    url.searchParams.set('client_id', process.env.VITE_DISCORD_CLIENT_ID);
+    url.searchParams.set('scope', 'identify email guilds.members.read');
+    url.searchParams.set('redirect_uri', 'http://localhost:1337/api/user/discord/callback');
```
itsalaidbacklife: We need to parameterize this. Similarly to the credentials issue, we'll need a way of managing env vars for the backend. Currently vite is already picking up and setting `VITE_API_URL` which has the value we'd need to pull out here, but alternatively we could bake it into sails custom config as recommended in the [sails docs](https://sailsjs.com/documentation/reference/application/advanced-usage/sails-get-base-url) (note the warning on `.getBaseUrl()` which says not to use it but instead recommends setting it as a config val.

api/controllers/UserController.js
```diff
@@ -88,16 +87,92 @@ module.exports = {
...
+    if (!code) {
+      return res.badRequest('Missing code');
+    }
+
+    const params = {
+      client_id: String(process.env.VITE_DISCORD_CLIENT_ID),
+      client_secret: String(process.env.VITE_DISCORD_CLIENT_SECRET),
+      grant_type: 'authorization_code',
+      code,
+      redirect_uri: `http://localhost:1337/api/user/discord/callback`,
```
itsalaidbacklife: Another place we need to parameterize the base url

api/controllers/UserController.js
```diff
@@ -88,16 +87,92 @@ module.exports = {
...
     } catch (err) {
       // Something happened and we couldn't verify the user, log them out
       await sails.helpers.logout(req);
       return res.badRequest(err);
     }
   },
+
+  discordRedirect: async function(_, res) {
+    const url = new URL('https://discord.com/api/oauth2/authorize');
+    const { generateSecret } = sails.helpers.oauth;
```
itsalaidbacklife: Do you know if `generateSecret()` will be generic across oauth providers, or if it's discord-specific?

I'm wondering about namespacing the helpers under `sails.helpers.oauth.discord` for the ones that are only for discord, and leaving any helpers generic to all oauth under `sails.helpers.oauth` itself.

Relatedly, and I imagine it's too early to say with just one oauth provider, but I'm wondering about whether it would be practical or even desirable to take an analogous approach to how we've set up the validate/execute helper namespacing to develop a generic redirect endpoint which figures out the identity provider and then grabs the relevant provider-specific helpers by using it's name like

```
const { providerName } = req.query;
const { generateSecret } = sails.helpers.oauth[providerName];
```

Probably too early to say and idk enough about oauth to clearly identify how much generalizes across providers, but the thought came to mind

api/controllers/UserController.js
```diff
@@ -88,16 +87,92 @@ module.exports = {
...
+
+      req.session.loggedIn = true;
+      req.session.usr = updatedUser.id;
+
+      return res.redirect('http://localhost:8080');
+
+
+
+    }catch (err) {
+      return res.redirect('http://localhost:8080/logout');
```
itsalaidbacklife: I'd like to review the flow with you to ensure I better understand it and also to discuss whether there should be any more particular handling. Do we care about 404 for user not found vs no code presented?

Also is this only for signup or also login? If there's no user found should we make one?

itsalaidbacklife: Need client's base url parameterization here

api/helpers/oauth/fetch-discord-identity.js
```diff
@@ -0,0 +1,66 @@
...
+  fn: async function ({ token, user }, exits) {
+    try {
+      const response = await fetch('https://discord.com/api/users/@me', {
+        headers: {
+          'Authorization': `Bearer ${token.access_token}`,
+        },
+      });
+
+      if (!response.ok) {
+        throw new Error(`Failed to fetch user info: ${response.status}`);
```
itsalaidbacklife: I'm wondering about the appropriate error handling. Does this case cover a not found and also a server error coming from discord?

We now have a somewhat fancy custom error system for defining more specific errors we can sniff out via codes. You can see the error definitions in `api/errors` and how they're used in most of the gamestate endpoints e.g. `/move`

Wondering if we should be returning different status codes etc

api/helpers/oauth/fetch-discord-identity.js
```diff
@@ -0,0 +1,66 @@
...
+
+      const providerId = data.id;
+      const provider = 'discord';
+      const { email } = data;
+      let { username } = data;
+
+      const prevIdentity = await Identity.findOne({ providerId }).populate('user');
+
+      if (prevIdentity && prevIdentity.user) {
+        if(user && prevIdentity.user.id !== user){
```
itsalaidbacklife:
```suggestion
        if (user && prevIdentity.user.id !== user) {
```

We should add a lint rule around this kind of whitespace

api/helpers/oauth/fetch-discord-identity.js
```diff
@@ -0,0 +1,66 @@
...
+          return exits.error('Account already linked to another profile');
```
itsalaidbacklife: This looks like something we'll want a custom error for so that the client can display a particular message. Need to bake what the UX should be

api/helpers/oauth/fetch-discord-identity.js
```diff
@@ -0,0 +1,66 @@
...
+        }
+        return exits.success(prevIdentity.user);
+      }
+      const existingUserName = await User.findOne({ username });
+
+      if (existingUserName) {
+        username = `${username}${Math.floor(Math.random() * 1000)}`;
+      }
+      const updatedUser = user ? await User.findOne({ id: user }) : await User.create({ username }).fetch();
```
itsalaidbacklife: So if their username from the oauth provider is already available, we automatically assign it to them, and if that username already exists in our database, we append a random number to it?

It's intriguing, but let's bake the UX of it. I think we're going to want to give users the ability to change their username from their discord one. The question is whether we should just make changing your username a thing, generally, or specifically make it a part of this flow

api/helpers/oauth/fetch-discord-identity.js
```diff
@@ -0,0 +1,66 @@
...
+        provider,
+        email,
+        user: updatedUser.id,
+      });
+
+      return exits.success(updatedUser);
+
+    } catch (e) {
+      console.error('Error fetching identity from Discord:', e);
+      return exits.error(e);
```
itsalaidbacklife: Want to bake the expected errors and their UX

api/models/User.js
itsalaidbacklife: Yeah let's discuss. We want to make sure you either have a discord identity or a password, but there's a chicken and egg problem because the discord identity requires a user id which then won't have the password. So what do we do to prevent users from being created with no way to log in?

api/hooks/customUserHook/index.js
```diff
@@ -21,15 +21,16 @@ module.exports = function userHook() {
...
-        User.findOne(id).exec(function (error, user) {
-          if (error) {
-            reject(error);
-          } else if (!user) {
-            reject({ message: 'User does not exist' });
-          } else {
-            return resolve(user);
-          }
-        });
+        User.findOne(id).populate('identities')
```
itsalaidbacklife, line 24: I want to re-examine where this hook is used. It doesn't really add any value over a raw .find() imo. I initially wrote it before sails supported promises on its queries (or else I didn't realize that it did) so this was just intended as a promise-wrapper for a basic query. That shouldn't really be necessary any more. 

Do we actually want/need the User records to populate their identities?

src/components/DiscordLink.vue
```diff
@@ -0,0 +1,20 @@
...
+<script setup>
+import { computed } from 'vue';
+import { useAuthStore } from '@/stores/auth';
+const authStore = useAuthStore();
+
+const hasDiscord = computed(() => {return authStore.identities.some(({ provider }) => provider === 'discord');});
+const title = hasDiscord.value ? 'Linked to Discord' : 'Link Discord';
+
+const handleClick = () => {
+  if(hasDiscord.value){
```
itsalaidbacklife:
```suggestion
  if (hasDiscord.value) {
```

src/components/DiscordLink.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+  <v-list-item :title="title" prepend-icon="mdi-link" @click="handleClick" />
+</template>
+
+<script setup>
+import { computed } from 'vue';
+import { useAuthStore } from '@/stores/auth';
+const authStore = useAuthStore();
+
+const hasDiscord = computed(() => {return authStore.identities.some(({ provider }) => provider === 'discord');});
```
itsalaidbacklife: Okay so this is why we're populating identies on users. I want to ensure we don't accidentally leak those out for anyone except when you request your own user, so this will warrant an inventory of requests for users

src/stores/auth.js
```diff
@@ -146,5 +151,10 @@ export const useAuthStore = defineStore('auth', {
         throw new Error(err);
       }
     },
-  },
+    async oAuth(provider) {
+      const url = `http://${import.meta.env.VITE_API_URL}/api/user/${provider}/redirect`;
```
itsalaidbacklife: We'll want to ensure we use https in prod, so we'll need to adjust things to include that in the API url I think i.e. set https:// in the value itself for prod and http:// for development

src/routes/login/LoginView.vue
```diff
@@ -130,6 +131,17 @@
...
+            <v-btn
+              class="w-100 my-10 text-subtitle-2 text-sm-subtitle-1 text-md-h6 h-auto py-2"
+              size="large"
+              color="newSecondary"
+              type="button"
+              @click="oAuth('discord')"
+            >
+              <img class="discord" src="/img/loginView/logo-discord.svg">
+              <span class="discordButton"> Login With Discord </span>
+            </v-btn>
```
itsalaidbacklife: Need to do some design work, but I'm thinking that we'll basically replace the discord button that already exists with this and then add icon buttons for all our social media accounts e.g. discord server, youtube, patreon

api/controllers/UserController.js
Haviles04: The secret is generic

api/controllers/UserController.js
Haviles04: this is one of those things where I was like, I'll hard code it now to just make sure it all works, and then completely forgot about it heh

api/hooks/customUserHook/index.js
```diff
@@ -21,15 +21,16 @@ module.exports = function userHook() {
...
+          .exec(function (error, user) {
```
itsalaidbacklife, line 25: Let's use await instead of .exec()

Haviles04:
--- Migration
```
CREATE TABLE identity (
    id SERIAL PRIMARY KEY,
    user_id INTEGER NOT NULL REFERENCES "user"(id) ON DELETE CASCADE,
    provider_id VARCHAR(255) NOT NULL,
    provider VARCHAR(255) NOT NULL,
    user_name VARCHAR(255),
    created_at TIMESTAMPTZ DEFAULT now(),
    updated_at TIMESTAMPTZ DEFAULT now()
);
```

src/routes/login/LoginView.vue
```diff
@@ -117,19 +117,33 @@
...
+              <div class="my-4 d-flex align-center">
+                <v-btn variant="text" href="https://discord.com/invite/9vrAZ8xGyh">
+                  <img src="/img/loginView/logo-discord-blue.svg">
+                </v-btn>
+                <v-btn variant="text" href="https://www.youtube.com/@cuttle-cards2245">
+                  <img src="/img/loginView/logo-youtube.svg">
+                </v-btn>
+                <v-btn class="py-1" variant="text" href="https://www.patreon.com/cuttle">
+                  <img src="/img/loginView/logo-patreon.svg">
+                </v-btn>
```
itsalaidbacklife: Let's make these all open in new tabs. And can you also add a link to the github repo?

src/routes/login/LoginView.vue
```diff
@@ -251,6 +266,17 @@ export default {
...
+    $route: {
+      immediate: true,
+      handler() {
+        if (this.$route.query?.error) {
+          this.handleError(this.t(this.$route.query.error));
+          this.$router.replace('/');
+        }
+      }
+    }
+  },
```
itsalaidbacklife, line 302: What does this do?

tests/unit/vitest-sails.config.mjs
```diff
@@ -16,13 +16,16 @@ export default defineConfig({
...
     globals: true,
     environment: 'node',
-    setupFiles: ['tests/unit/setup-sails.vitest.js'],
+    setupFiles: [ 'tests/unit/setup-sails.vitest.js' ],
     // Sails tests can not be parallelized
     async: false,
     threads: false,
     isolate: false,
+    env:{
+      VITE_JWT_SECRET: 'cuttle69420'
```
itsalaidbacklife: lol

src/routes/login/LoginView.vue
Haviles04: displays errors passed through query params

.env
```diff
@@ -1,3 +1,4 @@
-VITE_API_URL=localhost:1337
+VITE_FRONTEND_URL=http://localhost:8080
+VITE_API_URL=http://localhost:1337
```
itsalaidbacklife, line 2: Which variables are needed clientside vs serverside to boot and to use the discord integration?

I think we might be able to use a sails specific setup for environment variables (e.g. something in config/env/development.js) to manage setting defaults that make things work out of the box

api/controllers/UserController.js
```diff
@@ -88,16 +87,130 @@ module.exports = {
 
     try {
       // If the user is logged in, see if we can find them first to verify they exist
-      const { username } = await userAPI.findUser(id);
+      const { username, identities } = await userAPI.findUser(id);
```
itsalaidbacklife: I want to deprecate the entire userAPI in favor of namespaces helpers in a new helpers/users folder. That said I think this could just be a raw User.findOne().populate() call

api/controllers/UserController.js
```diff
@@ -88,16 +87,130 @@ module.exports = {
...
+        headers: {
+          'Content-Type': 'application/x-www-form-urlencoded',
+        },
+        body: new URLSearchParams(params),
+      });
+
+      const tokenData = await tokenRes.json();
+
+      if (!tokenData) {
+        throw new Error();
```
itsalaidbacklife: What would cause this? I'm assuming we should respond 500 for it

api/controllers/UserController.js
itsalaidbacklife: Or are we also filtering out sensitive properties here? That'd be a reason for a single helper which could arguably be left as-is for this update

api/controllers/UserController.js
```diff
@@ -88,16 +87,130 @@ module.exports = {
...
+      if (!updatedUser) {
+        throw new Error();
+      }
+
+      req.session.loggedIn = true;
+      req.session.usr = updatedUser.id;
+
+      return res.redirect(`${process.env.VITE_FRONTEND_URL}/`);
+
+    }catch (err) {
```
itsalaidbacklife: Whitespace

api/controllers/UserController.js
```diff
@@ -88,16 +87,130 @@ module.exports = {
...
+
+      if (!tokenData) {
+        throw new Error();
+      }
+      const user = req.session.usr ?? null;
+
+      if(!user){
+        req.session.tokenData = tokenData;
+        return res.redirect(`${process.env.VITE_FRONTEND_URL}/?oauthsignup=discord`);
+      }
```
itsalaidbacklife: I believe that this works but I don't understand the flow. Am I correct that the endpoint flow is:

1. Client hits discordRedirect which sends them to discord with url params that tell discord to redirect back to the discord callback on our server
2. Client goes through the discord  auth and redirects back to our discordCallback 
3. discordCallback fires and we hit this block 
4. If they have an existing user we proceed and redirect them home and if not we redirect them to homeview with the param to open the new dialog 

Where in that flow before this block would req.session.usr get set? I don't see how we'd hit the situation where they already have a user atm.

My expectation would be that we make some call to discord that then gives us an id from discord which we look up in the identities table to verify which user id is the current user's and then look them up by *that* instead of req.session.usr. What am I missing?

api/helpers/oauth/fetch-discord-identity.js
```diff
@@ -0,0 +1,71 @@
+module.exports = {
+  friendlyName: 'Generate Secret',
+
+  description: 'Generates a secret for OAuth state',
+
```
itsalaidbacklife: This looks incorrectly copied from generateSecret

api/controllers/UserController.js
itsalaidbacklife: Okay I think I've got it:

What I described happens in completeDiscord() via fetchDiscordIdentity() and this is actually handling the link accounts case where you're already logged in and now linking discord auth?

itsalaidbacklife: Let's add comments. I want to make it easier to understand the sequence of endpoints, the role of each one, and a bit of explanation of the constituent parts.

I'm thinking maybe a comment block above all the discord endpoints that explains how they relate to one another, then one above each function describing its purpose, and the main within-the-function thing I'm thinking we should explain is the account sync case/why you might have req.session.usr

itsalaidbacklife (commented):
Review checkpoint

api/controllers/UserController.js
```diff
@@ -88,16 +87,130 @@ module.exports = {
...
+        return res.redirect(`${process.env.VITE_FRONTEND_URL}/login?error=${err}`);
+      }
+    }
+
+    const { tokenData } = req.session;
+    const updatedUser = await fetchDiscordIdentity(tokenData, user , username);
+
+    if (!updatedUser) {
+      throw new Error();
+    }
```
itsalaidbacklife: I think uncaught errors crash the server rather than 500 and I don't see a try/catch. We could res.serverError() instead or set up a try catch that does

api/helpers/oauth/fetch-discord-identity.js
```diff
@@ -0,0 +1,72 @@
...
+        throw new Error();
+      }
+
+      const data = await response.json();
+
+      const providerId = data.id;
+      const provider = 'discord';
+      const { username: discordUsername } = data;
+
+      const prevIdentity = await Identity.findOne({ providerId }).populate('user');
```
itsalaidbacklife: I think we should also be checking that the provider matches

api/controllers/UserController.js
```diff
@@ -88,16 +87,129 @@ module.exports = {
...
+      const loggedInUser = req.session.usr ?? null;
+      if ( prevIdentity && loggedInUser && prevIdentity.user.id !== loggedInUser ){
+        // Fail if identity is already attached to someone elses account
+        throw new Error('login.snackbar.discord.alreadyLinked');
+      }
+
+      if (!prevIdentity && loggedInUser) {
+        // If no identity but user is already log in, create Identity
+        await Identity.create({
+          provider: 'discord',
```
Haviles04:
```suggestion
          provider,
```

api/controllers/UserController.js
```diff
@@ -88,16 +87,129 @@ module.exports = {
...
+
+      const tokenData = await retrieveTokenData(code);
+
+      const providerIdentity = await fetchIdentity(tokenData);
+      const prevIdentity = await Identity.findOne({ providerId: providerIdentity.id }).populate('user');
+
+      const loggedInUser = req.session.usr ?? null;
+      if ( prevIdentity && loggedInUser && prevIdentity.user.id !== loggedInUser ){
+        // Fail if identity is already attached to someone elses account
+        throw new Error('login.snackbar.discord.alreadyLinked');
```
Haviles04:
```suggestion
        throw new Error('login.snackbar.oAuth.alreadyLinked');
```

api/controllers/UserController.js
```diff
@@ -88,16 +87,129 @@ module.exports = {
...
+        throw new Error();
+      }
+
+      const tokenData = await retrieveTokenData(code);
+
+      const providerIdentity = await fetchIdentity(tokenData);
+      const prevIdentity = await Identity.findOne({ providerId: providerIdentity.id }).populate('user');
+
+      const loggedInUser = req.session.usr ?? null;
+      if ( prevIdentity && loggedInUser && prevIdentity.user.id !== loggedInUser ){
```
Haviles04:
```suggestion
      if ( prevIdentity && loggedInUser && prevIdentity.user?.id !== loggedInUser ){
```

src/routes/home/components/OauthSignupDialog.vue
```diff
@@ -0,0 +1,115 @@
...
+
+const show = computed({
+  get: () => props.modelValue,
+  set: (value) => {
+    emit('update:modelValue', value);
+  }
+});
+
+const switchModes = () => {
+  noAccount.value = !noAccount.value;
```
Haviles04, line 92:
```suggestion
  noAccount.value = !noAccount.value;
  password.value = ""
```

api/controllers/UserController.js
```diff
@@ -88,16 +87,126 @@ module.exports = {
...
+        req.session.tokenData = tokenData;
+        return res.redirect(`${process.env.VITE_FRONTEND_URL}/?oauthsignup=${provider}`);
+      }
+
+      req.session.loggedIn = true;
+      req.session.usr = loggedInUser ?? prevIdentity.user.id;
+
+      return res.redirect(`${process.env.VITE_FRONTEND_URL}/`);
+
+    }catch (err) {
```
Haviles04:
```suggestion
    } catch (err) {
```

.env
Haviles04: client needs `VITE_API_URL` and the backend needs them all

src/routes/home/HomeView.vue
```diff
@@ -208,18 +208,27 @@ export default {
...
+    },
   },
   async created() {
-    await this.gameListStore.requestGameList();
-    this.loadingData = false;
-    this.tab = this.$route.path;
+    try {
+      await this.gameListStore.requestGameList();
+    } catch (err) {
+      console.log(err);
```
Haviles04, line 222: We have to swallow the error so the oauthsignup can call requestGameList

api/controllers/UserController.js
```diff
@@ -88,16 +87,129 @@ module.exports = {
...
+    const { createRedirectParams } = sails.helpers.oauth[req.params.provider];
+    const url = createRedirectParams();
+
+    res.set('Cache-Control', 'private, no-cache');
+    return res.redirect(url.href);
+  },
+
+  oAuthCallBack: async function(req, res) {
+    const { code } = req.query;
+    const { state } = req.query;
```
itsalaidbacklife:
```suggestion
    const { code, state } = req.query;
```

config/policies.js
```diff
@@ -29,6 +29,9 @@ module.exports.policies = {
     // reLogin does not require password intentionally-- if you are already logged in it will
     // not require a password to validate the session
     reLogin: [ 'hasValidUsername' ],
+    oAuthRedirect: true,
+    oAuthCallback: true,
```
itsalaidbacklife: I think we should create a `hasValidAuthProvider` policy to apply to all three of the oAuth endpoints. It will start out with only accepting `'discord'` and we can add to it if and when we add providers

api/controllers/UserController.js
itsalaidbacklife: Looking at `.findUser()`'s definition, I don't think it adds any value over replacing this with a native call
```suggestion
      const { username, identities } = await User.findOne({ id }).populate('identities');
```

api/controllers/UserController.js
itsalaidbacklife: We should also have a policy to validate that `code` and `state` are set as query params and require that for `oAuthCallback`

api/controllers/UserController.js
```diff
@@ -88,16 +87,129 @@ module.exports = {
...
+
+      if (!verified || !code) {
+        // throw generic error
+        throw new Error();
+      }
+
+      const tokenData = await retrieveTokenData(code);
+
+      const providerIdentity = await fetchIdentity(tokenData);
+      const prevIdentity = await Identity.findOne({ providerId: providerIdentity.id }).populate('user');
```
itsalaidbacklife: Need to verify the provider as well so we don't look up a different user who had the same providerId for a different provider
```suggestion
      const prevIdentity = await Identity.findOne({ provider, providerId: providerIdentity.id }).populate('user');
```

api/models/Identity.js
```diff
@@ -0,0 +1,25 @@
...
+    },
+    providerId:{
+      type:'string',
+      required: true
+    },
+    provider: {
+      type: 'string',
+      required: true
+    },
+    userName:{
```
itsalaidbacklife: let's rename this to `providerUserName`. Reading some of the logic I got confused thinking we were using our username when it was really the discord one
```suggestion
    providerUserName:{
```

api/controllers/UserController.js
```diff
@@ -88,16 +87,129 @@ module.exports = {
...
+          provider,
+          providerId: providerIdentity.id,
+          user: loggedInUser,
+          username: providerIdentity.username,
+        });
+      } else if (!prevIdentity) {
+        // If no user, and no identity, redirect to identity signup
+        req.session.tokenData = tokenData;
+        return res.redirect(`${process.env.VITE_FRONTEND_URL}/?oauthsignup=${provider}`);
+      }
```
itsalaidbacklife: Let's move the comments above the conditionals; easier to scan what a condition does before you read it imo

api/controllers/UserController.js
```diff
@@ -88,16 +87,129 @@ module.exports = {
...
+      const prevIdentity = await Identity.findOne({ providerId: providerIdentity.id }).populate('user');
+
+      const loggedInUser = req.session.usr ?? null;
+      if ( prevIdentity && loggedInUser && prevIdentity.user?.id !== loggedInUser ){
+        // Fail if identity is already attached to someone elses account
+        throw new Error('login.snackbar.oAuth.alreadyLinked');
+      }
+
+      if (!prevIdentity && loggedInUser) {
+        // If no identity but user is already log in, create Identity
```
itsalaidbacklife:
```suggestion
        // If no identity but user is already logged in, create Identity
```

api/controllers/UserController.js
```diff
@@ -88,16 +87,129 @@ module.exports = {
...
+        await Identity.create({
+          provider,
+          providerId: providerIdentity.id,
+          user: loggedInUser,
+          username: providerIdentity.username,
+        });
```
itsalaidbacklife: In this case (user is logged in and linking their account; they don't have a previous identity), should we redirect home without `?oauthsignup=${provider}`? I'd think we're done at that point and just logging in?
```suggestion
      if (!prevIdentity && loggedInUser) {
        // If no identity but user is already log in, create Identity
        await Identity.create({
          provider,
          providerId: providerIdentity.id,
          user: loggedInUser,
          username: providerIdentity.username,
        });

        // redirect home
        return res.redirect(`${process.env.VITE_FRONTEND_URL}`);
```

itsalaidbacklife (commented):
Review checkpoint

api/controllers/UserController.js
Haviles04: we currently reuse the logic from 152-155 that is used if you have an account and are logging in

api/controllers/UserController.js
Haviles04: It would be anything on discord ends, which will throw a generic error from provider message for us

api/controllers/OAuthController.js
```diff
@@ -0,0 +1,122 @@
...
+      }
+      return res.redirect(`${process.env.VITE_FRONTEND_URL}/login?error=${message}`);
+    }
+  },
+
+  // Final step in Oauth Registration,
+  oAuthCompleteRegistration: async function(req, res) {
+    const { username, password } = req.body;
+    const { fetchIdentity } = sails.helpers.oauth[req.params.provider];
+    try{
```
itsalaidbacklife:
```suggestion
    try {
```

api/controllers/OAuthController.js
```diff
@@ -0,0 +1,122 @@
...
+      // Find existing user
+      let user = null;
+      if ( password ) {
+        try {
+          user = await User.findOne({ username });
+          if (!user) {
+            throw {
+              message: 'login.snackbar.userNotFound',
+            };
```
itsalaidbacklife: I think we can make these regular errors; those will have a `message` property by default
```suggestion
            throw new Error('login.snackbar.userNotFound');
```

api/controllers/OAuthController.js
```diff
@@ -0,0 +1,122 @@
...
+      }
+
+      const { tokenData } = req.session;
+      const providerIdentity = await fetchIdentity(tokenData);
+
+      if (!user) {
+        // If no existing user, check username is not a duplicate and if not, create new user
+        const foundUsername = await User.findOne({ username });
+        if (foundUsername) {
+          throw ({ message: 'login.snackbar.usernameIsTaken' });
```
itsalaidbacklife:
```suggestion
          throw new Error('login.snackbar.usernameIsTaken');
```

api/controllers/OAuthController.js
```diff
@@ -0,0 +1,122 @@
+
+/**
+ * OAuthController
+ *
+ * @description :: Server-side logic for managing Users via Oauth authentication
+ * @help        :: See http://sailsjs.org/#!/documentation/concepts/Controllers
+ */
```
itsalaidbacklife, line 16:
```suggestion
/**
 * OAuthController
 *
 * @description :: Server-side logic for managing Users via Oauth authentication
 *
 * Flow is: 
 *   1) Client navigates to oAuthRedirect(), which redirects to identity provider's auth
 *   2) Identity provider redirects back to our oAuthCallback(), logs user in or preps 1st-time registration
 *   3) New users decide whether to create a new account or link an old one, then hit oAuthCompleteRegistration()
 *
 * Covers three cases:
 *   Returning user login, New user signup, Linking account before login, link account after login
 *
 * @help        :: See http://sailsjs.org/#!/documentation/concepts/Controllers
 */
```

api/helpers/oauth/discord/fetch-identity.js
```diff
@@ -0,0 +1,34 @@
...
+    const response = await fetch('https://discord.com/api/users/@me', {
+      headers: {
+        'Authorization': `Bearer ${token.access_token}`,
+      },
+    });
+
+    if (!response.ok) {
+      // throw generic error
+      return exits.error(
+      );
```
itsalaidbacklife, line 25: Any reason not to make this an explicit error?
```suggestion
      return exits.error(
        new Error('Could not authenticate with discord')
      );
```

api/hooks/customUserHook/index.js
itsalaidbacklife: Updated #188 to call out removing the entire customUserHook instead

src/stores/auth.js
```diff
@@ -82,6 +84,9 @@ export const useAuthStore = defineStore('auth', {
         if (username) {
           this.authSuccess(username);
         }
+        if(identities) {
```
itsalaidbacklife:
```suggestion
        if (identities) {
```

src/stores/auth.js
```diff
@@ -135,5 +140,29 @@ export const useAuthStore = defineStore('auth', {
         throw new Error(err);
       }
     },
+    async oAuth(provider) {
+      const url = `${import.meta.env.VITE_API_URL}/api/user/${provider}/redirect`;
+      window.location.href = url;
+    },
+    async completeOAuth(provider, credentials) {
+      const { username, password } = credentials;
+      try{
```
itsalaidbacklife:
```suggestion
      try {
```

itsalaidbacklife: We should set up a lint rule for this

itsalaidbacklife: Created #1259 to make this consistent

src/stores/auth.js
```diff
@@ -135,5 +140,29 @@ export const useAuthStore = defineStore('auth', {
...
+          }),
+          body: JSON.stringify({
+            provider,
+            username,
+            password
+          }),
+          credentials: 'include',
+        });
+        this.requestStatus(true);
+      }catch(e){
```
itsalaidbacklife:
```suggestion
      } catch(e) {
```

api/models/Identity.js
```diff
@@ -0,0 +1,25 @@
...
+ * @description :: Represents a user's profile from Oauth provider.
+ * @docs        :: http://sailsjs.org/#!documentation/models
+ */
+
+module.exports = {
+  attributes: {
+    user: {
+      model: 'user',
+    },
+    providerId:{
```
itsalaidbacklife:
```suggestion
    providerId: {
```

api/models/Identity.js
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @description :: Represents a user's profile from Oauth provider.
+ * @docs        :: http://sailsjs.org/#!documentation/models
+ */
+
+module.exports = {
+  attributes: {
+    user: {
+      model: 'user',
+    },
```
itsalaidbacklife, line 12:
```suggestion
    },
    // id of user in 3rd party identity provider
```

api/models/Identity.js
```diff
@@ -0,0 +1,25 @@
...
+    providerId:{
+      type:'string',
+      required: true
+    },
+    provider: {
+      type: 'string',
+      required: true
+    },
+    providerUsername:{
```
itsalaidbacklife:
```suggestion
    providerUsername: {
```

src/components/DiscordLink.vue
```diff
@@ -0,0 +1,28 @@
...
+
+<script setup>
+import { computed } from 'vue';
+import { useAuthStore } from '@/stores/auth';
+import { useI18n } from 'vue-i18n';
+
+const { t } = useI18n();
+const authStore = useAuthStore();
+
+const hasDiscord = computed(() => {return authStore.identities.some(({ provider }) => provider === 'discord');});
```
itsalaidbacklife:
```suggestion
const hasDiscord = computed(() => authStore.identities.some(({ provider }) => provider === 'discord'));
```

api/controllers/OAuthController.js
```diff
@@ -0,0 +1,129 @@
...
+ * OAuthController
+ *
+ * @description :: Server-side logic for managing Users via Oauth authentication
+ *
+ * Flow is: 
+ *   1) Client navigates to oAuthRedirect(), which redirects to identity provider's auth
+ *   2) Identity provider redirects back to our oAuthCallback(), logs user in or preps 1st-time registration
+ *   3) New users decide whether to create a new account or link an old one, then hit oAuthCompleteRegistration()
+ *
+ * Covers three cases:
```
itsalaidbacklife:
```suggestion
 * Covers four cases:
```

PR #1248, opened 2025-07-16
Allow spectating from player 1's perspective via ?pNum query param

### LINKED ISSUE #1227: [DevEx]:  Allow users to spectate as p1
### Improvement Summary

Currently whenever you spectate a game, you view the game from player 0's perspective. We should update this logic to allow you to spectate as player 1 by including a `?pNum=1` request parameter

### Detailed Description

Currently the spectator always views the game from the perspective of p0. This can make for an awkward experience when a player is watching their own game after the fact if they played that game as p0, because then they will view the game as p1.

We should update the router logic to support links passing a `?pNum` query param (either equal to 0 or 1) and taking that value as `gameStore.myPNum` if it is passed. If a valid pNum is not passed in the query params, then `gameStore.myPNum` should default to:

- the user's `pNum` if they were a player in that game and the game is completed
- otherwise 0

Mang0codes:
## #1227

Relevant #1227 (https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
- Resolves #1227

## Please check the following

- For New Features:

?pNum=0 → View as Player 0

Screenshot 2025-07-16 212858
?pNum=1 → View as Player 1

Screenshot 2025-07-16 212926


Mang0codes:
@Haviles04 Could you please take a look at this one when you have time, one of the tests is failing and I think it's because previously, the spectate route worked without needing pNum now after the update the code always sets myPNum, but the logic is relying on userId, p0, and p1, which don’t exist before login, because of that, something in the render logic fails to trigger #reauthenticate-dialog, likely due to a render short-circuit or an uncaught error.

src/stores/game.js
```diff
@@ -782,5 +782,13 @@ export const useGameStore = defineStore('game', {
       }
     },
 
+    isUserAPlayer() {
+      return this.userId === this.p0?.id || this.userId === this.p1?.id;
+    },
+
+    getMyOriginalPlayerNumber() {
+      return this.userId === this.p0?.id ? 0 : 1;
+    },
```
itsalaidbacklife, line 791: `userId` isn't defined on the game store. In fact we don't currently store the user's id on the client at all, though perhaps we should.

The `game` store has `resetPNumIfNull()` which is how we currently set the pNum when it's missing. That function currently sets your pNum to the matching player's index by username if you're a player in the game, and it sets it to 0 if you're spectating. We should update that function to account for the pNum query parameter if it is set if you're also spectating. Then I think we could rely on that function instead these new ones.

itsalaidbacklife: I'm refactoring myPNum as a computed property so that it can't desync from the rest of the state, given that it should be dependent entirely on the url + crosss-referencing the username in the auth store with the usernames in gameStore.players.

Once #1256 is merged, than I think we can centralize all the logic for consuming the new `pNum` query param (grabbed by calling the `useRoute()` composable) inside the `gameStore.myPNum` computed property. Basically if you're spectating we should take that query param and if it's either 0 or 1 use that, and otherwise default to 0. Then the non-specatting cases can stay as they are, looking up which username from the array of players matches the one from the auth store

Issue #1225, opened 2025-06-17
[Feature]: Show spectators game overlays and never game dialogs


itsalaidbacklife:
### Improvement Summary

We should update the spectator experience so that whenever a modal element is shown for a mid-turn phase e.g. countering or picking a card to discard, that the spectator sees this via the corresponding game overlay and not the corresponding game dialog.

### Detailed Description

Currently the spectator always views the game from the perspective of p0 and sees whatever dialogs they say e.g.

![Image](https://github.com/user-attachments/assets/c74202f7-a346-402d-b1d5-5fa534292d7c)

This is misleading because it looks to the spectator like they're actually being asked to make a choice that they aren't being asked to make. When someone is spectating, we should switch to only showing them the corresponding overlay for the "inactive" player:

![Image](https://github.com/user-attachments/assets/1911e733-6cb6-4ac4-8b97-84760e2ae947)

This would better communicate that the spectator does not have a choice to make, but instead is watching a player make a decision. 

**Note: This should be a follow up to #1224 and needs to ensure that the relevant player's username shows up i.e. if aleph_one is faced with the decision to counter, it should read "aleph_one may counter" regardless of whether aleph_one was p0 or p1.**

PR #1249, opened 2025-07-17
Feature/Revise game overlays to show player usernames instead of the …

### LINKED ISSUE #1224: [Feature]: Revise game overlays to show player usernames instead of the word 'Opponent'
### Improvement Summary

We should update all the in-game overlays that say 'Opponent' to instead say the player username of the relevant player

### Detailed Description

Currently the in-game overlays all show text like "Opponent may counter". We should switch this to say the username of the player who may counter e.g. "aleph_one may counter".

Mang0codes:
Shows opponent's name instead of the word 'Opponent'

## #1224

- Resolves #1224

## Please check the following

- For New Features:

Screenshot 2025-07-17 205232

Screenshot 2025-07-17 205208

### DIFF
--- a/src/routes/game/components/GameOverlays.vue
+++ b/src/routes/game/components/GameOverlays.vue
@@ -71,7 +71,7 @@
       class="game-overlay"
     >
       <h1 :class="[$vuetify.display.xs === true ? 'text-h5' : 'text-h3', 'overlay-header']">
-        {{ t('game.overlays.choosingFromScrap', { opponentUsername }) }}
+        {{ choosingFromScrapMessage }}
       </h1>
     </v-overlay>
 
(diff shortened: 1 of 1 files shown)


src/routes/game/components/GameOverlays.vue
```diff
@@ -71,7 +71,7 @@
       class="game-overlay"
     >
       <h1 :class="[$vuetify.display.xs === true ? 'text-h5' : 'text-h3', 'overlay-header']">
-        {{ t('game.overlays.opponentChoosingFromScrap') }}
+        {{ `${opponentUsername} ${t('game.overlays.choosingFromScrap')}` }}
```
itsalaidbacklife: Can we pull this string into a computed property similar to what's done above with `opponentDiscardingText`? 
That will make it easier to execute #1225 where we'll start showing spectators the overlays for every mid-turn phase instead of the current behavior where we show spectators an overlay when it's the opponent's chance to do something and a corresponding dialog when it's the player's chance to do the thing.

In that update, it'll be nice to have these as computed props already so we can dynamically select which username to use since spectators would see either the 'player' or the 'opponent' in these overlays

Mang0codes: Sure ..I'll commit the changes by today

itsalaidbacklife: Can we additionally refactor these  strings with the username so that they're managed as parameters to the t() call? It's called [interpolation](https://vue-i18n.intlify.dev/guide/essentials/syntax.html#named-interpolation) and basically that would let different language files format the username itself inside the string in case there are languages where the translation works better with the username somewhere other than the beginning of the string

Haviles04:
@Mang0codes Are you still interested in finishing this issue :)

PR #1254, opened 2025-07-21
Prevent signing up with a username that matches another case-insensitively 

### LINKED ISSUE #1234: [Feature]: Prevent signing up with a username that matches another case-insensitively
### Feature Summary

myUsername and myusername should not both be allowed; we should only accept one version of a username and ignore case

### Detailed Description

We had an issue where a player accidentally signed up with two different capitalizations of the same username without realizing it. We should prevent this kind of issue by updating `api/UserController`'s `signup` method to perform a case-insensitive search for users with that username and error if one is found to prevent users from having case-insensitive duplicate usernames

Mang0codes:
## #1234

- Resolves #1234

## Please check the following

- For New Features:

Now the api/UserController's `signup` method: normalizes the incoming username to lowercase and performs a case-insensitive search to make usernames case-insensitive and prevent duplicates like `MyName` vs `myname`


api/controllers/UserController.js
```diff
@@ -15,7 +15,10 @@ module.exports = {
   signup: async function (req, res) {
     try {
       const { username, password } = req.body;
-      const foundUser = await User.findOne({ username: username });
+      const userName = username.toLowerCase();
+      const foundUser = await User.findOne({ 
+        username: { $regex: new RegExp(`^${userName}$`, 'i') }
+      });
```
itsalaidbacklife, line 21: So I've done further digging on this issue and unfortunately it isn't a simple one. TL;DR sails doesn't support a [regex where clause](https://sailsjs.com/documentation/concepts/models-and-orm/query-language#?contains) and so we can't just pass a regex like this. This one requires further thought. 

Here are three options that occur to me:

1. Update the columnType of `username` for the `User` model to be `citext`. This would require updating all the existing duplicate username rows to ensure their uniqueness
2. Create a postgresql computed column e.g. `usernameLowerCase` which we instantiate via a migration and add to the `User` model with comments clarifying that it's read only. Then query based on this column on signup, comparing against `username.toLowerCase()` and erroring if found. This would avoid the need to fix the existing duplicates while preventing creating new ones, but it would introduce the possibility of telling sails to write to that column and causing an error. We would also need to find a way to ensure that the development db (which uses sails-disk in memory rather than a real postgres db) can correctly function.
3. Same idea as above except instead of having postgres create the column, we could add it as a 'regular' attribute and manually set the lower case version of the username on signup. Similarly, we'd avoid the need to backfill the previous records, but we'd need to ensure any code that can modify username also sets the new `usernameLowerCase` column too. This would avoid the issue of database adapter specific idiosyncrosies.
4. Compute the list of all permutations of different capitalizations of the username here and query for users with a username in that list. This could get long as the length of usernames increases, but it avoids the need to add and manage an extra column

Truthfully, I'm not sure which is better. Penny for your thoughts; this one is a bigger lift than I originally thought

itsalaidbacklife: Here's a SO on the issue: https://stackoverflow.com/questions/55475718/how-can-i-use-the-find-method-to-perform-a-case-insensitive-search-on-the-databa


5. We could use a raw SQL query to use postgres' `ilike` operator, but we'd need to manage a way for it to work in sails-disk too

PR #1256, opened 2025-07-26
Refactor/rewrite my p num as computed

itsalaidbacklife:
Refactors `gameStore.myPNum` into a computed property (from a ref). This should prevent opportunities for descynchronization and set up #1248 so that it can centralize logic for utilizing the `pNum` query param to set myPNum just inside the computed prop.

Generally if a piece of data can be computed from other data you're already storing, than it _should_ be a computed prop,, to avoid these opportunities for desync. And which pnum the current player is should be determined by their url (whether they're spectating) and which player's username matches the one in the auth store
## Issue number

N/A

## Please check the following

- For New Features:

### DIFF
--- a/src/routes/game/GameView.vue
+++ b/src/routes/game/GameView.vue
@@ -1,7 +1,7 @@
 <template>
   <div id="game-view-wrapper">
     <!-- Unauthenticated/Must re-log in/ Unavailable game -->
-    <template v-if="gameStore.myPNum === null">
+    <template v-if="gameStore.myPNum === null || !gameStore.player || !gameStore.opponent">
       <GameUnavailableView />
     </template>
 
--- a/src/routes/game/GameView.vue
+++ b/src/routes/game/GameView.vue
@@ -842,7 +842,7 @@ export default {
      * @param player is the player object
      */
     kingCount(player) {
-      return player.faceCards.reduce((kingCount, card) => kingCount + (card.rank === 13 ? 1 : 0), 0);
+      return player?.faceCards?.reduce((kingCount, card) => kingCount + (card.rank === 13 ? 1 : 0), 0) ?? 0;
     },
     /**
      * Returns the number of points to win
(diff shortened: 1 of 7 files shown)


src/routes/game/GameView.vue
```diff
@@ -842,7 +842,7 @@ export default {
      * @param player is the player object
      */
     kingCount(player) {
-      return player.faceCards.reduce((kingCount, card) => kingCount + (card.rank === 13 ? 1 : 0), 0);
+      return player?.faceCards?.reduce((kingCount, card) => kingCount + (card.rank === 13 ? 1 : 0), 0) ?? 0;
```
itsalaidbacklife, line 845: This wasn't necessary for the scope, but it's been a common source of type errors for this computed to barf if `player` is undefined. Better to coalesce nicely to 0

src/stores/auth.js
itsalaidbacklife, line 1: This no longer needs to set myPNum at all given that it's computed

src/stores/game.js
```diff
@@ -18,7 +18,7 @@ function queenCount(player) {
   if (!player?.faceCards) {
     return 0;
   }
-  return player.faceCards.reduce((queenCount, card) => queenCount + (card.rank === 12 ? 1 : 0), 0);
+  return player?.faceCards?.reduce((queenCount, card) => queenCount + (card.rank === 12 ? 1 : 0), 0) ?? 0;
```
itsalaidbacklife, line 21: Similarly to `kingCount`, this is a bit of scope creep for this PR but it should stop this function from type erroring when `player` is undefined

src/stores/game.js
```diff
@@ -117,6 +116,13 @@ export const useGameStore = defineStore('game', () => {
   const gameHistoryStore = useGameHistoryStore();
 
   // Computed (getters)
+  const myPNum = computed(() => {
+    if (gameHistoryStore.isSpectating) {
+      return 0;
```
itsalaidbacklife, line 121: @Mang0codes This will be the part to change in #1248 to read from the `pNum` query param and return the right value for `myPNum` when spectating

PR #1262, opened 2025-09-28
Feat/asymmetric client events

### LINKED ISSUE #187: [DevEx]: Customize socket messages so each player gets asymmetric information
### Improvement Summary

Currently both clients receive the same socket update with the full game object. This is theoretically a security concern because it means each player has info on which cards are in the opponent hand and which are on top of the deck.

### Detailed Description

We should instead update the app to make separate socket rooms for each player in each game, and to separately parse the game object into two complementary data transfer objects that scrub all the info that a player isn't allowed to see (e.g. the opponent's hand). 

This will also entail rework of the frontend which currently assumes full access to the entire game object

itsalaidbacklife:
## Issue number

- Resolves #187 

## Please check the following

- For New Features:


api/controllers/game/join-rematch.js
itsalaidbacklife, line 1: I discovered `sails.sockets.addRoomMembersToRooms()` which totally removes the need for this endpoint. We'd been forcing both players to make a separate request to subscribe their socket to the new game after it's created (since the first player to agree to a rematch doesn't know when the new game is created), but we can actually just add all members of the old room to the new one. And having separate rooms for the players vs the spectators means we can do this nicely for just the players letting the spectators opt-in independently of the game creation

api/controllers/game/get-game.js
```diff
@@ -38,7 +65,10 @@ module.exports = async function(req, res) {
     }
 
     const gameState = unpackGamestate(packedGameState);
-    const response = await createSocketEvent(game, gameState);
+    const socketEvents = await createSocketEvents(game, gameState);
+    
+    // Return the appropriate state based on user relationship
+    let response = socketEvents[`${userRelationship}State`];
```
itsalaidbacklife: Looks like this should be `const`
```suggestion
    const response = socketEvents[`${userRelationship}State`];
```

plan.md
itsalaidbacklife, line 1: Todo: delete

tests/e2e/support/commands.js
```diff
@@ -680,7 +680,6 @@ Cypress.Commands.add('reconnectOpponent', (opponent) => {
 
 Cypress.Commands.add('rematchAndJoinRematchOpponent', ({ gameId }) => {
```
itsalaidbacklife: Todo: delete this and replace calls with regular old rematchOpponent

tests/e2e/support/commands.js
```diff
@@ -678,11 +678,6 @@ Cypress.Commands.add('reconnectOpponent', (opponent) => {
   });
 });
 
-Cypress.Commands.add('rematchAndJoinRematchOpponent', ({ gameId }) => {
-  cy.makeSocketRequest('game', 'rematch', { gameId, rematch: true });
-  cy.makeSocketRequest('game', 'join-rematch', { oldGameId: gameId });
-});
```
itsalaidbacklife, line 684: Since the join-rematch endpoint was removed, we don't need a command to differentiate between rematch and join-rematch. All calls to this command were replaced with `rematchOpponent()`

tests/e2e/support/commands.js
```diff
@@ -726,10 +721,6 @@ Cypress.Commands.add('rematchOpponent', ({ gameId, rematch, whichPlayer, skipDom
   }
 });
 
-Cypress.Commands.add('joinRematchOpponent', ({ oldGameId = null }) => {
-  cy.makeSocketRequest('game', 'join-rematch', { oldGameId });
-});
```
itsalaidbacklife, line 731: This endpoint was removed and so all calls to it have been tidied up

tests/e2e/support/helpers.js
```diff
@@ -358,48 +362,59 @@ export function setupSeasons() {
...
+
+        // Check Hand
+        const checkingCurrentPlayer = i === pNum;
+        const handShouldBeRevealed = checkingCurrentPlayer || playerShouldHaveGlasses || deckShouldBeEmpty || isSpectating;
+        if (handShouldBeRevealed) {
+          expect(cardListsMatch(gamePlayer.hand, fixturePlayer.hand)).to.eq(
+            true,
+            `P${i} Hand should match fixture, but actual: ${printCardList(
+              gamePlayer.hand,
+            )} did not match ficture: ${printCardList(fixturePlayer.hand)}`,
```
Haviles04:
```suggestion
            )} did not match fixture: ${printCardList(fixturePlayer.hand)}`,
```

tests/e2e/support/helpers.js
```diff
@@ -358,48 +362,59 @@ export function setupSeasons() {
...
+          expect(allHidden).to.eq(true, `All cards in P${i} Hand should be hidden, 
+            but actual hand included known cards: ${printCardList(gamePlayer.hand)}`);
+        }
+
+        // Points
+        expect(cardListsMatch(gamePlayer.points, fixturePlayer.points)).to.eq(
+          true,
+          `P${i} Points should match fixture, but actual: ${printCardList(
+            gamePlayer.points,
+          )} did not match ficture: ${printCardList(fixturePlayer.points)}`,
```
Haviles04:
```suggestion
          )} did not match fixture: ${printCardList(fixturePlayer.points)}`,
```

tests/e2e/support/helpers.js
```diff
@@ -358,48 +362,59 @@ export function setupSeasons() {
...
+        );
+
+        // Face Cards
+        expect(cardListsMatch(gamePlayer.faceCards, fixturePlayer.faceCards)).to.eq(
+          true,
+          `P${i} Face Cards should match fixture, but actual: ${printCardList(
+            gamePlayer.faceCards,
+          )} did not match ficture: ${printCardList(fixturePlayer.faceCards)}`,
```
Haviles04:
```suggestion
          )} did not match fixture: ${printCardList(fixturePlayer.faceCards)}`,
```

Haviles04 (approved):
LGTM, Nice work :)

I think we could refactor the FE to not need any cards in the deck, just a length but might be worthy of a different PR

PR #1265, opened 2025-10-05
refactored lifecycle hooks

iamritikiit:
## Issue number

- Resolves #

## Please check the following

- For New Features:

### DIFF
--- a/src/routes/home/components/announcementDialog/AnnouncementDialog.vue
+++ b/src/routes/home/components/announcementDialog/AnnouncementDialog.vue
@@ -52,7 +52,7 @@
 </template>
 
 <script setup>
-import { ref, computed, onMounted } from 'vue';
+import { ref, computed } from 'vue';
 import { useI18n } from 'vue-i18n';
 import dayjs from 'dayjs';
 import { getLocalStorage, setLocalStorage } from '_/utils/local-storage-utils.js';
@@ -80,11 +80,12 @@ const close = () => {
   preferenceSaved.value = true;
 };
 
-onMounted(() => { 
-  if (getLocalStorage('announcement') !== announcementData.id) {
-    show.value = true;
-  }
-});
+// Initialisation logic moved from onMounted
+// onMounted(() => { 
+if (getLocalStorage('announcement') !== announcementData.id) {
+  show.value = true;
+}
+// });
 </script>
 
 <style scoped>
--- a/src/routes/lobby/LobbyView.vue
+++ b/src/routes/lobby/LobbyView.vue
@@ -99,7 +99,7 @@
 </template>
 
 <script setup>
-import { onMounted, ref, computed, watch } from 'vue';
+import { ref, computed, watch } from 'vue';
 import { useRouter, onBeforeRouteLeave } from 'vue-router';
 import { useI18n } from 'vue-i18n';
 import { useGameStore } from '@/stores/game';
@@ -179,9 +179,9 @@ watch(opponentUsername, (newVal) => {
 });
 
 // Lifecycle
-onMounted(() => {
-  playAudio(joinAudio);
-});
+// onMounted(() => {
+playAudio(joinAudio);
+// });
 
 // Router
 onBeforeRouteLeave((to, from, next) => {
--- a/src/routes/login/LoginView.vue
+++ b/src/routes/login/LoginView.vue
@@ -300,7 +300,7 @@ export default {
       }
     }
   },
-  mounted() {
+  created() {
     if (this.goingToForm) {
       this.scrollAndFocusLogin();
     }
--- a/src/routes/game/GameView.vue
+++ b/src/routes/game/GameView.vue
@@ -753,10 +753,12 @@ export default {
       }
     }
   },
-  async mounted() {
+  created(){
     if (!this.authStore.authenticated) {
       this.authStore.mustReauthenticate = true;
     }
+  },
+  async mounted() {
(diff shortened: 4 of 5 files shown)


src/routes/login/LoginView.vue
```diff
@@ -300,7 +300,7 @@ export default {
       }
     }
   },
-  mounted() {
+  created() {
```
Haviles04: I think this one should stay on mounted since it's scrolling to something in the DOM

src/routes/home/components/announcementDialog/AnnouncementDialog.vue
```diff
@@ -80,11 +80,12 @@ const close = () => {
   preferenceSaved.value = true;
 };
 
-onMounted(() => { 
-  if (getLocalStorage('announcement') !== announcementData.id) {
-    show.value = true;
-  }
-});
+// Initialisation logic moved from onMounted
```
Haviles04: No need for comments, if we're going to remove it we can just remove it

Haviles04 (changes requested):
Thanks for your contribution! I left a few small points of feedback :)

src/routes/home/components/announcementDialog/AnnouncementDialog.vue
```diff
@@ -80,11 +80,12 @@ const close = () => {
...
+// onMounted(() => {
```
itsalaidbacklife: Our general philosophy on comments is that if it describes a change to the code and not why the code is doing what it's doing now, that it shouldn't go in source code. 

It can still be helpful to explain things like this to a reviewer, but the best place to put comments describing why you're making a change is by commenting on your own PR in GitHub. That way you can communicate to reviewers about your reasoning or whatever they need to know about the changes without cluttering the code with comments that aren't needed to understand the code once it's merged 
```suggestion
```

src/routes/home/components/announcementDialog/AnnouncementDialog.vue
```diff
@@ -80,11 +80,12 @@ const close = () => {
...
 };
 
-onMounted(() => { 
-  if (getLocalStorage('announcement') !== announcementData.id) {
-    show.value = true;
-  }
-});
+// Initialisation logic moved from onMounted
+// onMounted(() => { 
+if (getLocalStorage('announcement') !== announcementData.id) {
```
itsalaidbacklife, line 83: While we're in here we could clarify the why 
```suggestion
// show announcement by default if it hasn't previously been dismissed
if (getLocalStorage('announcement') !== announcementData.id) {
```

src/routes/login/LoginView.vue
itsalaidbacklife: This needs to stay mounted because the function is interacting with the DOM (scrolling)
```suggestion
  mounted() {
```

src/routes/lobby/LobbyView.vue
```diff
@@ -179,9 +179,9 @@ watch(opponentUsername, (newVal) => {
 });
 
 // Lifecycle
-onMounted(() => {
-  playAudio(joinAudio);
-});
+// onMounted(() => {
+playAudio(joinAudio);
+// });
```
itsalaidbacklife: This is a sneaky one but it needs to stay in onMounted because playAudio() is interacting with the DOM 
```suggestion
 onMounted(() => {
playAudio(joinAudio);
 });
```

itsalaidbacklife (changes requested):
Thank you for your contribution! This is a good start. I've requested several changes. I haven't yet scanned the rest of the code base to verify if this is comprehensive; have you searched for all the mounted and onMounted calls throughout the project?

src/routes/login/LoginView.vue
iamritikiit: Thanks for catching that! I’ve moved the scrolling logic back to mounted so it runs after the DOM is available.

src/routes/home/components/announcementDialog/AnnouncementDialog.vue
iamritikiit: Sure, I have now removed the unnecessary comments.

src/routes/home/components/announcementDialog/AnnouncementDialog.vue
iamritikiit: Yes, I have searched for all the mounted and onMounted calls throughout the project and made the required changes, also have removed the unnecessary comments.

src/routes/login/LoginView.vue
iamritikiit: Thanks for pointing that out, I have now moved the logic back to mounted so that it runs after DOM is available.

src/routes/lobby/LobbyView.vue
iamritikiit: Yes, we should keep playAudio() in onMounted as this will be interacting with the browser's DOM and media APIs.
Now, I have moved back the logic inside of the onMounted() again.

src/routes/home/components/announcementDialog/AnnouncementDialog.vue
iamritikiit: Understood, we should use a comment only if it explains intention—not change history, I will keep that in mind.

iamritikiit:
I’ve addressed all the feedback and resolved the previous comments. Please let me know if any further adjustments are needed. Looking forward to your review!

Haviles04 (approved):
This lgtm, I searched for all mounted and onMounted calls and I think all the ones we need to change have been addressed

PR #1269, opened 2025-10-09
Fix/scuttle animation

itsalaidbacklife:
Fixes bug where the scuttle animation doesn't show when your opponent scuttles you. Bug was due to the lookup of which card your opponent's using to scuttle not working when the opponent hand is hidden

## Please check the following

- For New Features:

### DIFF
--- a/tests/e2e/specs/in-game/basicMoves.spec.js
+++ b/tests/e2e/specs/in-game/basicMoves.spec.js
@@ -83,7 +83,9 @@ describe('Game Basic Moves - P0 Perspective', () => {
     // Player scuttles 6 of diamonds with 7 of clubs
     cy.get('[data-player-hand-card=7-0]').click(); // 7 of clubs
     cy.get('[data-move-choice=scuttle]').click();
-    cy.get('[data-opponent-point-card=6-2]').click(); // 6 of hearts
+    cy.get('[data-opponent-point-card=6-2]').click(); // Scuttles the 6 of hearts
+    cy.get('[data-opponent-point-card=6-2] [data-scuttled-by=7-0]').should('be.visible'); // Scuttle transition
+  
     assertGameState(0, {
       p0Hand: [ Card.ACE_OF_CLUBS, Card.ACE_OF_SPADES ],
       p0Points: [ Card.TWO_OF_CLUBS, Card.TEN_OF_HEARTS ],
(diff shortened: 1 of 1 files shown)


src/routes/game/components/GameCard.vue
```diff
@@ -25,7 +25,11 @@
     />
     <Transition :name="scuttledByTransition">
       <template v-if="scuttledBy">
-        <img :class="scuttledByClass" :src="`/img/cards/card-${scuttledBy.suit}-${scuttledBy.rank}.svg`">
+        <img
+          :class="scuttledByClass"
+          :src="`/img/cards/card-${scuttledBy.suit}-${scuttledBy.rank}.svg`"
+          :data-scuttled-by="`${scuttledBy.rank}-${scuttledBy.suit}`"
```
itsalaidbacklife, line 31: Added a selector here to facilitate testing that the transition card appears. This made me notice that the image sources go suit-rank while the data selectors go rank-suit. I don't love the inconsistency but also it doesn't seem worthwhile to change at the moment

src/stores/game.js
```diff
@@ -290,22 +290,25 @@ export const useGameStore = defineStore('game', () => {
...
+
+    if (targetCardIndex === -1) {
       updateGame(game);
       return;
     }
-    const [ playedCard ] = scuttlingPlayer.hand.splice(playedCardIndex, 1);
-    const targetCard = scuttledPlayer.points[targetCardIndex];
-    targetCard.scuttledBy = playedCard;
+
+    scuttlingPlayer.hand.splice(playedCardIndex, 1);
```
itsalaidbacklife, line 308: Here I'm relying on the fact that splicing at the index -1 pulls from the end of the list, so if the opponent scuttles while their hand is hidden, it'll splice the last card from their hand. Otherwise it picks the 'right' one

src/stores/game.js
```diff
@@ -290,22 +290,25 @@ export const useGameStore = defineStore('game', () => {
...
+  async function processScuttle({ game, playedCard, targetCard, playedBy }) {
     if (!player.value) {
       updateGame(game);
       return;
     }
     const scuttlingPlayer = players.value[playedBy];
     const scuttledPlayer = players.value[(playedBy + 1) % 2];
-    const playedCardIndex = scuttlingPlayer.hand.findIndex((card) => card.id === playedCardId);
-    const targetCardIndex = scuttledPlayer.points.findIndex((card) => card.id === targetCardId);
-    if (playedCardIndex === -1 || targetCardIndex === -1) {
```
itsalaidbacklife, line 302: The root cause of the issue was the change #1262 to hide the opponent's hand. The lookup for the playedCard was missing since none of the opponent's hand cards were revealed, and then with the playedCardIndex === -1, we'd skip straight to updateGame().

Basic approach to the fix was to send the entire playedCard instead of just playedCardId and to use that object to send the scuttledBy instead of the object from the scuttling player's hand (which we may not have if their hand is hidden).

PR #1271, opened 2025-10-23
[DevEx]: Replace .then() with await in handleInGameEvents and ensure proper async handling #923

### LINKED ISSUE #923: [DevEx]: Replace .then() withal await in inGameEvents and troubleshoot
### Improvement Summary

The src/plugins/inGameEvents.js file has a use of `.then()` where `await` would be cleaner, but `await` doesn’t seem to work. We should troubleshoot why.

### Detailed Description

As of #900, we are using an async `sleep()` function to force the app to wait before updating certain variables in order to allow animations to complete in stages. Generally we call this with `await sleep(1000)` (or whatever duration in ms), but in one spot in the inGameEvents.js file, we instead call `sleep(1000).then()`. This is because for whatever reason, the DOM failed to update when we tried `await sleep(1000)` in that spot. 

We should investigate why and refactor to enable us to use `await` consistently

ryuga001:
Fixes #923
Resolves [DevEx]: Replace .then() with await in handleInGameEvents and ensure proper async handling #923

`handleInGameEvents` now uses `await` instead of `.then()` for cleaner async handling.  
`handleGameOver` waits for animations (via `sleep`) before updating game state with `gameStore.setGameOver(evData.victory)`.  

Ensures proper order of event handling and improves code readability.

### DIFF
--- a/src/plugins/sockets/inGameEvents.js
+++ b/src/plugins/sockets/inGameEvents.js
@@ -52,7 +52,6 @@ export async function handleInGameEvents(evData, newRoute = null) {
     case SocketEvent.SEVEN_FACE_CARD:
     case SocketEvent.SEVEN_JACK:
     case SocketEvent.SEVEN_DISCARD:
-    case SocketEvent.SEVEN_SCUTTLE:
     case SocketEvent.SEVEN_ONE_OFF:
     case SocketEvent.SEVEN_TARGETED_ONE_OFF:
     case SocketEvent.STALEMATE_REQUEST:
@@ -61,6 +60,7 @@ export async function handleInGameEvents(evData, newRoute = null) {
       gameStore.updateGame(evData.game);
       break;
     case SocketEvent.SCUTTLE:
+    case SocketEvent.SEVEN_SCUTTLE:
       gameStore.processScuttle(evData);
       break;
     case SocketEvent.RESOLVE_THREE:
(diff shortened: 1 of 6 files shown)


src/plugins/sockets/inGameEvents.js
```diff
@@ -6,6 +6,13 @@ import { ROUTE_NAME_GAME, ROUTE_NAME_SPECTATE, ROUTE_NAME_LOBBY } from '@/router
 import SocketEvent from '_/types/SocketEvent';
 import { sleep } from '@/util/sleep';
 
+// Handles Game Over event asynchronously
+// This ensures any end-of-game animations complete first
```
Haviles04: I'm not sure these comments are necessary

Haviles04 (commented):
Nice job this looks good, 1 small nitpick :)

src/plugins/sockets/inGameEvents.js
ryuga001: @Haviles04  I have updated the PR, removed these comments

Haviles04:
> Hi @Haviles04 ,
> Could you please assign me another issue to work on? Thank you!

You technically don't need to be assigned and you can just pick something and work on it! If you pick something I can always assign it to you after.

Haviles04:
After further review I think it makes more sense to move `sleep` to `gameStore.setGameOver` I think this is more consistent with how we handle card transitions for 3s,4s,and 5s,

Haviles04:
> Hi @Haviles04, I’ve updated the PR and moved the sleep logic into the setGameOver function


This looks great! Thank you for your contribution!

Haviles04 (approved):

PR #1276, opened 2025-10-28
Feat/dependency updates 20251027, resolves #1277

### LINKED ISSUE #1277: [DevEx]: Update dependencies (October, 2025)
### Improvement Summary

Update all deps

### Detailed Description

We need to check deps -- start with low hanging easy to update fruit and move on to the larger dependencies to get the app fully up to date.

seriouslysean:
@itsalaidbacklife you've lets all these dependencies languish. It's like you were waiting for me to come update them all again, hahaha. This is Cypress 15, and a bunch of smaller deps that won't break anything. I'll do some of the larger ones in follow up PRs.

## Issue number

- Resolves #1277

## Please check the following

- For New Features:

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "vitest": "^0.34.6"
   },
   "engines": {
-    "node": "22"
+    "node": ">=22.11.0"
   },
   "main": "app.js",
   "private": true,


src/components/MarkdownContent.vue
```diff
@@ -9,10 +9,7 @@ const props = defineProps({
   },
 });
 
-const text = computed(() => parse(props.markdown, {
-  headerIds: false,
-  mangle: false,
-}));
+const text = computed(() => parse(props.markdown));
```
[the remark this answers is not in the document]
seriouslysean: This is no longer supported by the markdown plugin, so should be safe to remove.

package.json
```diff
@@ -38,31 +38,31 @@
...
     "supertest": "^7.1.4",
     "vite": "5.4.8",
     "vite-plugin-vue-devtools": "^7.7.7",
-    "vite-plugin-vuetify": "^2.0.4",
+    "vite-plugin-vuetify": "^2.1.2",
     "vitest": "^0.34.6"
   },
   "engines": {
-    "node": "22.11"
+    "node": "22"
```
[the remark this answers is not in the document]
seriouslysean: Good suggestion 👍🏼

tests/e2e/support/index.js
```diff
@@ -23,6 +23,14 @@ Cypress.on('uncaught:exception', (err) => {
   }
 });
 
+// Hide Vue DevTools panel during tests to prevent it from covering UI elements
+Cypress.on('window:before:load', (win) => {
+  win.document.head.insertAdjacentHTML(
+    'beforeend',
+    '<style>.vue-devtools__panel { display: none !important; }</style>'
+  );
+});
```
itsalaidbacklife, line 32: Interesting move. I actually like having the devtools open in the Cypress GUI so that I can troubleshoot with cypress + the devtools. I'd think we could either do this programmatically based on whether Cypress is in GUI mode, or otherwise we could configure it with a build time flag? But that introduces complexity in the build process. What do you think?

seriouslysean: We can't do this, because it's breaking the is visible test in Cypress. We can look in to enabling it again in the future, but that's out of scope for this change. Additionally, because we use dev to test locally, a build time flag won't work. We always want it on for the engineers.

I think we need to change how Cypress checks if something is visible. I think it's just seeing if ANY ELEMENT is on top of the visible element. We need a different check imo.

itsalaidbacklife: So is the update to cypress 15 what causes the change in test behavior?

seriouslysean: No, the update to the vue dev tools adds a floating island element and when it's over the snack bar the Cypress is visible check. Cypress 15 update was totally fine 🤘🏼

seriouslysean: In any event, if this is good, we can solve that Cypress issue in this issue https://github.com/cuttle-cards/cuttle/issues/1278

PR #1286, opened 2025-11-10
Fix/discard animations

### LINKED ISSUE #1285: [Bug]: Opponent discarding with 4 or 5 doesn't reveal the discarded cards
### Bug Summary

When your opponent discards cards from their hand with a 4 or 5, you should first see those cards flip over to reveal which cards are being discarded before they translate down and fade out. This is no longer working and instead the card backs move down and fade out w/o showing you which cards were selected/discarded

### Detailed Description

You can see this in the tests in 4_fours.spec.js and 5_fives.spec.js where the player plays a 4 or the opponent plays a 5. We should update these tests (and selectors on the cards if necessary) to assert that the discarded cards were revealed before being removed from the opponent hand

### Current Behavior

Discarded cards in the opponent hand show the card back and not the card front, so you can't see which cards were discarded.

### Expected Behavior

The discarded cards flip to face-up before they move downwards and fade out

### Steps to Reproduce

1. Play a 4 against your opponent
2. They resolve
3. Opponent picks two cards to discard
Should see the cards flip over to reveal their faces before they're scrapped

Instead, the cards are not revealed as they're discarded.

itsalaidbacklife:
Fixes animation issues with opponent discards
## Issue number

- Resolves #1285 

## Please check the following

- For New Features:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cuttle",
-  "version": "13.2.0",
+  "version": "13.2.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "cuttle",
-      "version": "13.2.0",
+      "version": "13.2.1",
       "dependencies": {
         "@unhead/vue": "^1.11.14",
         "chart.js": "^4.5.1",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "13.2.0",
+  "version": "13.2.1",
   "dependencies": {
     "@unhead/vue": "^1.11.14",
     "chart.js": "^4.5.1",
--- a/src/plugins/sails.js
+++ b/src/plugins/sails.js
@@ -15,8 +15,16 @@ export const io = sails(socketIoClient);
 
 export const reconnectSockets = () => {
   return new Promise((resolve, reject) => {
-    io.socket.disconnect();
-    io.socket.reconnect();
+    try {
+      if (io.socket.isConnecting() || io.socket.isConnected()) {
+        io.socket.disconnect();
+      }
+    } catch (_) {/* empty */}
+    try {
+      if (!io.socket.isConnected()) {
+        io.socket.reconnect();
+      }
+    } catch (_) {/* empty */}
     const MAX_TRIES = 10; // 10 seconds
     const INTERVAL = 500; // 10 * 500 = 5000 (5 seconds)
     let tries = 1;


src/routes/game/GameView.vue
```diff
@@ -108,11 +108,11 @@
...
-                      v-for="(card) in gameStore.opponent.hand"
-                      :key="card.id"
-                      :suit="isBeingDiscarded(card) ? card.suit : undefined"
-                      :rank="isBeingDiscarded(card) ? card.rank : undefined"
-                      data-opponent-hand-card
+                      v-for="(card, index) in gameStore.opponent.hand"
+                      :key="`opponent_hand_card_${index}`"
+                      :suit="card.suit ?? undefined"
+                      :rank="card.rank ?? undefined"
+                      :data-opponent-hand-card="card.suit && card.rank ? `${card.rank}-${card.suit}` : ''"
```
itsalaidbacklife, line 115: TODO: We should refactor all the GameCard instances for the opponent hand to consolidate the versions where you see the cards vs don't so that we can nicely flip the cards over to reveal them

src/routes/game/GameView.vue
```diff
@@ -1098,9 +1098,6 @@ export default {
         this.$refs.logsContainer.scrollTop = this.$refs.logsContainer.scrollHeight;
       }
     },
-    isBeingDiscarded(card) {
-      return this.gameStore.lastEventDiscardedCards?.some(discardedCard => discardedCard.id === card.id); 
-    }
```
itsalaidbacklife, line 1103: EXP: pulled this into the game store to use it in the animation helpers

src/stores/game.js
```diff
@@ -230,7 +230,6 @@ export const useGameStore = defineStore('game', () => {
       typeof lastEventPlayer === 'number' && myPNum.value !== null
         ? lastEventPlayer === myPNum.value
         : null;
-    lastEventDiscardedCards.value = newGame.lastEvent?.discardedCards ?? null;
```
itsalaidbacklife, line 233: Removed lastEventDiscardedCards from state as we now pull them directly from the socket event

src/stores/game.js
```diff
@@ -359,17 +357,23 @@ export const useGameStore = defineStore('game', () => {
   }
   async function processFours(discardedCards, game) {
     phase.value = GamePhase.MAIN;
-    lastEventDiscardedCards.value = discardedCards;
-    await sleep(1000);
+    if (opponent.value.hand[0]?.isHidden) {
+      opponent.value.hand = [
+        ...opponent.value.hand.slice(0, opponent.value.hand.length - discardedCards.length),
+        ...discardedCards
+      ];
```
itsalaidbacklife, line 366: This is a little sloppy, but the discarded cards always go to the end of the opponent's hand atm. It could be could to do it randomly

tests/e2e/specs/in-game/one-offs/4_fours.spec.js
```diff
@@ -37,6 +39,37 @@ describe('FOURS', () => {
       });
     });
 
+    it('Plays a 4 to make opponent discard two cards of their choice while player has glasses', () => {
```
itsalaidbacklife, line 42: EXP: I added this test because the logic bifurcates whether to call the transition logic based on whether the opponent hand is revealed. This doesn't really increase confidence significantly, but I added it to manually view the way the transitions looked. If and when we improve the transitions of opponent hand cards while they have glasses, we can use this tests to view them

tests/e2e/specs/in-game/one-offs/5_fives.spec.js
```diff
@@ -532,6 +532,8 @@ describe('FIVES', () => {
       cy.resolveFiveOpponent(Card.FIVE_OF_HEARTS);
       cy.get('#waiting-for-opponent-discard-scrim').should('not.exist');
 
+      cy.get('[data-opponent-hand-card=5-2]').should('be.visible');
```
itsalaidbacklife, line 535: EXP: this verifies that the opponent's discarded card is revealed before the full state change

tests/e2e/specs/in-game/one-offs/4_fours.spec.js
```diff
@@ -25,6 +25,8 @@ describe('FOURS', () => {
       // Opponent chooses two cards to discard
       cy.discardOpponent(Card.ACE_OF_HEARTS, Card.TEN_OF_HEARTS);
       cy.get('#waiting-for-opponent-discard-scrim').should('not.exist');
+      cy.get('[data-opponent-hand-card=1-2]').should('be.visible');
+      cy.get('[data-opponent-hand-card=10-2]').should('be.visible');
```
itsalaidbacklife, line 29: EXP: Verifies that discarded cards are revealed

src/stores/game.js
itsalaidbacklife: I took a crack at making it random by defining this utility function, but it becomes more complicated than just replacing random values in the array because we're using the indices as keys atm. So this would take further baking. Not sure how to manage it rn


Meanwhile I thought this helper was neat
```

function replaceElementsAtRandom(original, replacements) {
  const randomIndices = sampleSize(range(original.length), replacements.length);

  const result = [ ...original ];
  randomIndices.forEach((originalIndex, replacementNumber) => {
    result[originalIndex] = replacements[replacementNumber];
  });

  return result;
}
```

src/routes/game/GameView.vue
itsalaidbacklife: Created #1290 to resolve this

PR #1289, opened 2025-11-11
[Feature]: Create a 'My Profile' page #237

### LINKED ISSUE #237: [Feature]: Create a 'My Profile' page
### Feature Summary

We should create a page to view your account and configure various account settings.

### Detailed Description

Initial implementation of this page should display your username and the option to set an email address (which is private). We can update this with additional functionality later. This should be accessible from the lefthand nav drawer once you log in.

This will entail updating the "User" model to have an additional, optional `email` attribute, and running a migration to add that column to the "user" table of the database

Stasgysk:
## Issue number

- Resolves #237 

## Please check the following

- For New Features:

### DIFF
--- a/tests/e2e/specs/out-of-game/profile.spec.js
+++ b/tests/e2e/specs/out-of-game/profile.spec.js
@@ -206,11 +206,6 @@ describe('Profile Page', () => {
       cy.contains('[data-cy="game-list-item"]', 'Stalemate Game').should('be.visible');
 
       cy.contains('[data-cy="game-list-item"]', 'Player Won Game')
-        .find('button')
-        .contains('Replay')
-        .click();
-
-      cy.contains('[data-cy="game-list-item"]', 'Player Won Game')
         .find('[data-cy="replay-link"] a, [data-cy="replay-link"]')
         .should('have.attr', 'href')
         .and('include', '/spectate/');


src/routes/profile/ProfileView.vue
```diff
@@ -0,0 +1,104 @@
...
+        style="background-color: rgba(var(--v-theme-surface-2)); color: rgba(var(--v-theme-surface-1))"
+      >
+        <h2>My Games</h2>
+
+        <v-virtual-scroll
+          v-if="games.length > 0"
+          :items="games"
+          height="600"
+          item-height="100"
+        >
```
Haviles04: I think a table might work better here, especially if we implement pagination.

@itsalaidbacklife penny for your thoughts

api/controllers/game/get-history.js
```diff
@@ -0,0 +1,36 @@
+module.exports = async function getHistory(req, res) {
+  const userId = req.session.usr;
+  if (!userId) {
+    return res.status(403).json({ message: 'Forbidden' });
+  }
```
Haviles04: I think this is covered under `isLoggedIn` policy?

Stasgysk: Added this endpoint to policy

api/controllers/game/get-history.js
```diff
@@ -0,0 +1,33 @@
+module.exports = async function getHistory(req, res) {
+  const userId = req.session.usr;
+
+  const sortBy = req.query.sortBy || 'createdAt';
+  const sortDirection = req.query.sortDirection === 'asc' ? 'ASC' : 'DESC';
+
+  const finishedGames = await Game.find({
+    status: 3,
+    or: [ { p0: userId }, { p1: userId } ]
+  }).sort(`${sortBy} ${sortDirection}`);
```
itsalaidbacklife: I think that @Haviles04 is right; we need to implement server side pagination to avoid this blowing up either server or client memory. I think that loading 20 at a time would be plenty by default. This will also afford us flexibility for other potential use cases. 

Apologies for suggesting to refrain from using it; it'll be safer to manage pagination here than clientside

api/controllers/game/get-history.js
```diff
@@ -0,0 +1,33 @@
...
+  const opponentMap = Object.fromEntries(opponents.map(u => [ u.id, u.username ]));
+
+  const composedGames = finishedGames.map(game => {
+    const opponentId = game.p0 === userId ? game.p1 : game.p0;
+    return {
+      id: game.id,
+      name: game.name,
+      isRanked: game.isRanked,
+      createdAt: game.createdAt,
+      winnerLabel: game.winner === userId ? 'You' : 'Opponent',
```
itsalaidbacklife: My instinct is to make this the more 'raw' data of either the username or the id of the winner and then to compute clientside which user that actually is. imo id makes the most sense, though then we hit the snag of not currently storing the user's id clientside. We could add that to the authentication logic or just rely on username.

My preference is to have a robust API response with 'everything we need' that we hopefully don't often need to change and then to make it up to the client to reformat etc. In that spirit, we might also want to include the `match` value of the game, which would let us group them up by match if desired later.

api/controllers/game/get-history.js
```diff
@@ -0,0 +1,33 @@
+module.exports = async function getHistory(req, res) {
+  const userId = req.session.usr;
+
+  const sortBy = req.query.sortBy || 'createdAt';
+  const sortDirection = req.query.sortDirection === 'asc' ? 'ASC' : 'DESC';
+
+  const finishedGames = await Game.find({
+    status: 3,
```
itsalaidbacklife: Let's import the `GameStatus` enum from `util/GameStatus.json` and then this can be
```suggestion
    status: GameStatus.FINISHED,
```

src/routes/profile/ProfileView.vue
```diff
@@ -0,0 +1,104 @@
...
+
+      <!-- Games List -->
+      <v-card
+        flat
+        class="pa-4 mt-4"
+        style="background-color: rgba(var(--v-theme-surface-2)); color: rgba(var(--v-theme-surface-1))"
+      >
+        <h2>My Games</h2>
+
+        <v-virtual-scroll
```
itsalaidbacklife, line 37: To implement serverside pagination, I think we'll want to switch this to a `v-infinite-scroll` (https://vuetifyjs.com/en/components/infinite-scroller/) and then pull more data on the `@load` event

The ideal would be to virtualize it so that we also only render a fixed number of entries at a time (minimizing clientside memory for the DOM), but that could be a future improvement I think

src/stores/gameHistory.js
itsalaidbacklife, line 1: This store is really used for tracking the history of moves within a given game, rather than the history of one's games over time. I think we could either create a new store for the list of historical games (maybe `myGames`?) or use the existing `gameList` store which is for the list of games to play and to specate. I lean towards using `gameList`; what do you think?

tests/e2e/specs/out-of-game/profile.spec.js
```diff
@@ -0,0 +1,122 @@
...
+    cy.contains('span', 'Not connected');
+  });
+
+  it('Shows Discord connected state', () => {
+    cy.vueRoute('/my-profile');
+
+    // Patch auth store to simulate Discord connection
+    cy.window().then((win) => {
+      const pinia = win.cuttle.app.config.globalProperties.$pinia;
+      const authStore = pinia._s.get('auth');
```
itsalaidbacklife:
```suggestion
    cy.window().its(('cuttle.authStore').then((authStore) => {
```

This is because we more directly expose the pinia stores to cypress on the window object in addition to the app itself

tests/e2e/specs/out-of-game/profile.spec.js
```diff
@@ -0,0 +1,122 @@
...
+            provider: 'discord',
+            username: 'TestDiscordUser#1234'
+          }
+        ]
+      });
+    });
+
+    cy.contains('h2', 'Discord').should('be.visible');
+    cy.contains('span', 'Connected as TestDiscordUser#1234').should('be.visible');
+    cy.contains('span', 'Not connected').should('not.exist');
```
itsalaidbacklife: Let's add some data-cy selectors to the template code to make these assertions less brittle to style changes. Something like
```suggestion
    cy.get('[data-cy=discord-username]').should('be.visible').contain;
    cy.get('[data-cy=not-connected]').should('not.exist').should('contain', 'TestDiscordUser#1234');
```

tests/e2e/specs/out-of-game/profile.spec.js
```diff
@@ -0,0 +1,122 @@
...
+    cy.vueRoute('/my-profile');
+    cy.contains('h1', 'My Profile').should('be.visible');
+    cy.contains('No games found').should('be.visible');
+  });
+
+  it('Lists mocked finished games and can replay', () => {
+    cy.vueRoute('/my-profile');
+    cy.contains('h1', 'My Profile').should('be.visible');
+    cy.contains('No games found').should('be.visible');
+    cy.wait(500);
```
itsalaidbacklife: Why the wait here? We generally want to avoid them if possible. Is this to wait for games to load?

tests/e2e/specs/out-of-game/profile.spec.js
```diff
@@ -0,0 +1,122 @@
...
+            name: 'Second Game',
+            isRanked: false,
+            winnerLabel: 'Opponent',
+            opponentName: 'AnotherUser',
+            createdAt: '2025-02-01T00:00:00Z',
+          },
+        ],
+        loading: false,
+      });
+    });
```
itsalaidbacklife: Our preference is to skip mocking and instead to test the feature end to end by seeding data with an API call. 

You should be able to use the existing `cy.loadFinishedGameFixtures()` to programmatically create games with the specified shape. The trick will be setting the `p0`, `p1`, and `winner` attributes, which are the integer ids of users. The nuance here is that those user ids aren't known until test runtime since they're created via api calls during test time. But we can use the `cy.signupOpponent()` command to create a new user and this will create a cypress alias of their id called `${opponent.username}Id`. 

Using the alias be cleaner if we do the setup in a beforeEach() hook, and use the `function` keyword in the test itself. Then we'll be able to use `this.playerOneId` etc to refer to each player's id and can use it here. That could look something like this

```
// can nest in existing describe
describe('Populated game list content', () => {
  beforeEach(() => {
    cy.signupOpponent(playerOne); // these can be imported from fixtures/userFixtures.js
    cy.signupOpponent(playerTwo);
  });

  it('Lists finished games and can replay', function() {
    cy.loadFinishedGameFixtures([
          {
            name: 'Mocked Game',
            isRanked: true,
            p0: this.Player1Id, // could instead use this[`${playerOne.username}Id`]
            p1: this.myUsernameId, // or this[`${myUser.username}Id`]
            winner: this.myUsernameId, // player won
            createdAt: '2025-01-01T00:00:00Z',
          },
          // etc
    ]);
  });
});

```

itsalaidbacklife: This way we're testing that the backend works as well

src/routes/profile/ProfileGameListItem.vue
```diff
@@ -0,0 +1,71 @@
...
+            icon="mdi-trophy"
+            color="yellow-darken-2"
+            class="mr-2"
+          />
+          <v-icon
+            v-if="winnerLabel === 'Opponent'"
+            icon="mdi-trophy-broken"
+            color="grey-darken-2"
+            class="mr-2"
+          />
```
itsalaidbacklife, line 26: Another thing I'm realizing is that we need a state for stalemates. Those happen if neither player can win once the deck is exhausted. At the db level, it's designated by `winner: null` for any game that's finished `status: 3`

api/controllers/game/get-history.js
Stasgysk: Added .limit and .skip

api/controllers/game/get-history.js
Stasgysk: added match, made it so we send winnerId and not label

api/controllers/game/get-history.js
Stasgysk: changed it

src/routes/profile/ProfileView.vue
Stasgysk: Added loading additional games when scrolling

src/stores/gameHistory.js
Stasgysk: Created a new store "myGames". Didnt want to change "gameList" as it is a little bit different functionality I think

tests/e2e/specs/out-of-game/profile.spec.js
Stasgysk: Changed it

tests/e2e/specs/out-of-game/profile.spec.js
Stasgysk: Also tried to change it everywhere

tests/e2e/specs/out-of-game/profile.spec.js
Stasgysk: I removed wait here, just added wait when scrolling

tests/e2e/specs/out-of-game/profile.spec.js
Stasgysk: Used this one cy.loadFinishedGameFixtures([

src/routes/profile/ProfileGameListItem.vue
Stasgysk: Added this too

src/components/DiscordLink.vue
```diff
@@ -16,7 +16,6 @@ const { t } = useI18n();
 const authStore = useAuthStore();
 
 const hasDiscord = computed(() => authStore.identities.some(({ provider }) => provider === 'discord'));
-const title = t(hasDiscord.value ? 'login.discordLinked' : 'login.linkDiscord');
```
itsalaidbacklife, line 19: Instead of moving the logic into the template can we just make this a computed property?

api/controllers/game/get-history.js
```diff
@@ -0,0 +1,52 @@
+const GameStatus = require('../../../utils/GameStatus.json');
+
+module.exports = async function getHistory(req, res) {
+  const userId = req.session.usr;
+
+  const sortBy = req.query.sortBy || 'createdAt';
```
itsalaidbacklife: We should sanitize the input to prevent SQL injection shenanigans
```suggestion
  const allowedSortColumns = [ 'createdAt', 'winner', 'isRanked' ];
  const sortBy = allowedSortColumns.find((col) => col === req.query.sortBy) ?? 'createdAt';
```

api/controllers/game/get-history.js
```diff
@@ -0,0 +1,52 @@
...
+    const opponentId = game.p0 === userId ? game.p1 : game.p0;
+    return {
+      id: game.id,
+      name: game.name,
+      isRanked: game.isRanked,
+      createdAt: game.createdAt,
+      match: game.match,
+      winnerId: game.winner,
+      opponentId: opponentId,
+      opponentName: opponentMap[opponentId] || 'Unknown',
```
itsalaidbacklife: Penny for your thoughts on this one. In #1227  we'll add the ability to spectate as p1 by setting a query param in the spectate link.

Currently when you watch. a replay you always do so as p0 **even if you're watching a game where you were p1**. So it would be nice for the client to be able to determine which player number you were in a game so as to correctly format the link accordingly (that actual formatting can be added in #1227). One way we could do this would be to have the server p0 and p1 and have the client figure it out. They could either be objects or ids. Or the server could send a `myPNum` attribute to designate which person was the requesting player. 

What do you think is best?

I lean towards having the server send:
```
// ...other attributes
p0: { username: string, id: int}
p1: { username: string, id: int}
```

so that it matches the underlying data structure and for the flexibility of potentially requesting this info for games you aren't yourself in at a later time. Then `winner` could just be the id of the winner as well

src/routes/profile/ProfileGameListItem.vue
```diff
@@ -0,0 +1,86 @@
...
+          <span class="game-name">
+            {{ winnerLabel === null ?
+              t('profile.stalemate') :
+              winnerLabel === true ?
+                t('profile.win') :
+                t('profile.lose') }} —
+            {{ name }}
```
itsalaidbacklife: Let's pull this into a computed prop

src/routes/profile/ProfileGameListItem.vue
```diff
@@ -0,0 +1,86 @@
...
+          </span>
+        </div>
+        <p class="text-surface-1">
+          <v-icon
+            class="mr-4"
+            size="medium"
+            :icon="isRanked ? 'mdi-sword-cross' : 'mdi-coffee-outline'"
+            aria-hidden="true"
+          />
+          {{ isRanked ? t('global.ranked') : t('global.casual') }} • {{ t('profile.opponent') }}: {{ opponentName }}
```
itsalaidbacklife: Same here. We lean towards minimal computation in template code

src/routes/profile/ProfileView.vue
```diff
@@ -0,0 +1,155 @@
...
+}
+
+onMounted(fetchGames);
+
+onUnmounted(() => {
+  myGamesStore.resetGames();
+});
+</script>
+
+<style>
```
itsalaidbacklife: It's probably worth adding a lint rule to prevent unscoped styles. I'll write up an issue about it
```suggestion
<style scoped>
```

src/routes/profile/ProfileGameListItem.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+  <div class="history-item" data-test="game-list-item">
```
itsalaidbacklife: I like how generic data-test is, but we've been using `data-cy` so let's stick with that
```suggestion
  <div class="history-item" data-cy="game-list-item">
```

tests/e2e/specs/out-of-game/profile.spec.js
```diff
@@ -0,0 +1,344 @@
...
+
+      cy.loadFinishedGameFixtures(games);
+      cy.loginPlayer(myUser);
+      cy.vueRoute('/my-profile');
+
+      cy.get('[data-test="game-list-item"]').should('have.length.at.least', 5);
+
+      cy.get('.v-virtual-scroll').scrollTo('bottom');
+      cy.wait(500);
+      cy.get('.v-virtual-scroll').scrollTo('bottom');
```
itsalaidbacklife: The idea here is to wait until more games load right? To avoid a hard-coded wait I'd prefer to use an assertion that a new game shows up. We could use a `.get().contains()` to grab the latest item by its name perhaps?
```suggestion
      cy.get('.v-virtual-scroll').scrollTo('bottom');
      cy.get('[data-cy=game-list-item]').contains('game 30 name'); // whatever the expected bottom option is.
      cy.get('.v-virtual-scroll').scrollTo('bottom');
```

tests/e2e/specs/out-of-game/profile.spec.js
```diff
@@ -0,0 +1,344 @@
...
+
+      cy.window().then(win => {
+        const pinia = win.cuttle.app.config.globalProperties.$pinia;
+        const myGamesStore = pinia._s.get('myGames');
+        expect(myGamesStore.games.length).to.eq(30);
+      });
+
+      cy.contains('[data-test="game-list-item"]', 'Game 30').should('exist');
```
itsalaidbacklife: let's go a step further and assert it's visible
```suggestion
      cy.contains('[data-test="game-list-item"]', 'Game 30').should('be.visible');
```

itsalaidbacklife: Based on my comment above, I think we could move this line up and replace the cy.wait() call with this

tests/e2e/specs/out-of-game/profile.spec.js
itsalaidbacklife: Actually you've already got the assertion that the final game appears down below. I think we just show move that here instead of the cy.wait()

tests/e2e/specs/out-of-game/profile.spec.js
```diff
@@ -0,0 +1,344 @@
...
+        p0: this[`${playerOne.username}Id`],
+        p1: this[`${myUser.username}Id`],
+        winner: this[`${myUser.username}Id`],
+      }));
+
+      cy.loadFinishedGameFixtures(games);
+      cy.loginPlayer(myUser);
+      cy.vueRoute('/my-profile');
+
+      cy.get('[data-test="game-list-item"]').should('have.length.at.least', 5);
```
itsalaidbacklife: Maybe also reasonable to assert that it's below the page limit?
```suggestion
      cy.get('[data-test="game-list-item"]').should('have.length.below', 21);
```

tests/e2e/specs/out-of-game/profile.spec.js
```diff
@@ -0,0 +1,344 @@
...
+        winner: this[`${myUser.username}Id`],
+      }));
+
+      cy.loadFinishedGameFixtures(games);
+      cy.loginPlayer(myUser);
+      cy.vueRoute('/my-profile');
+
+      cy.window().then(win => {
+        const pinia = win.cuttle.app.config.globalProperties.$pinia;
+        const myGamesStore = pinia._s.get('myGames');
```
itsalaidbacklife: Same as above; our syntactical pattern for grabbing stores off the window is `cy.window().its()`. To do this, we need to update `App.vue`'s created hook to add the `myGameStore()` to the window object during testing
```suggestion
      cy.window().its('cuttle.myGamesStore').then(myGamesStore => {
```

tests/e2e/specs/out-of-game/profile.spec.js
```diff
@@ -0,0 +1,344 @@
...
+
+      cy.loginPlayer(myUser);
+      cy.vueRoute('/my-profile');
+      cy.get('[data-test="game-list-item"]').should('exist');
+
+      cy.vueRoute('/');
+
+      cy.window().then(win => {
+        const pinia = win.cuttle.app.config.globalProperties.$pinia;
+        const myGamesStore = pinia._s.get('myGames');
```
itsalaidbacklife: Same as above

src/routes/profile/ProfileView.vue
itsalaidbacklife: Created #1293 to standardize this later

src/components/DiscordLink.vue
Stasgysk: Moved it to computed property

api/controllers/game/get-history.js
Stasgysk: Changed it

api/controllers/game/get-history.js
Stasgysk: added p0 and p1 with "isOpponent" property

src/routes/profile/ProfileGameListItem.vue
Stasgysk: Done

src/routes/profile/ProfileGameListItem.vue
Stasgysk: Done

tests/e2e/specs/out-of-game/profile.spec.js
Stasgysk: Made it with assert

tests/e2e/specs/out-of-game/profile.spec.js
Stasgysk: Done

tests/e2e/specs/out-of-game/profile.spec.js
Stasgysk: Done

tests/e2e/specs/out-of-game/profile.spec.js
Stasgysk: Changed it everywhere

tests/e2e/specs/out-of-game/profile.spec.js
Stasgysk: Done

src/routes/profile/ProfileGameListItem.vue
```diff
@@ -0,0 +1,98 @@
...
+  return _props.winnerLabel ? t('profile.win') : t('profile.lose');
+});
+
+const gameModeIcon = computed(() =>
+  _props.isRanked ? 'mdi-sword-cross' : 'mdi-coffee-outline'
+);
+
+const gameModeText = computed(() =>
+  _props.isRanked ? t('global.ranked') : t('global.casual')
+);
```
itsalaidbacklife: Now that `props` is referenced we can remove the _ from its name
```suggestion
const props = defineProps({
  name: {
    type: String,
    required: true
  },
  isRanked: Boolean,
  winnerLabel: {
    type: [ Boolean, null ],
    default: null
  },
  opponentName: {
    type: String,
    required: true
  }
});

const gameResultText = computed(() => {
  if (props.winnerLabel === null) {return t('profile.stalemate');}
  return props.winnerLabel ? t('profile.win') : t('profile.lose');
});

const gameModeIcon = computed(() =>
  props.isRanked ? 'mdi-sword-cross' : 'mdi-coffee-outline'
);

const gameModeText = computed(() =>
  props.isRanked ? t('global.ranked') : t('global.casual')
);
```

Stasgysk: Done

src/routes/profile/ProfileGameListItem.vue
Stasgysk: done

src/routes/profile/ProfileView.vue
```diff
@@ -0,0 +1,161 @@
...
+      sortBy: 'createdAt',
+      sortDirection: 'desc'
+    });
+  } catch (error) {
+    console.error('Failed to load more games:', error);
+  }
+}
+
+function goToReplay(gameId) {
+  window.location.href = `/spectate/${gameId}`;
```
itsalaidbacklife: Let's use the Vue router for navigation. We can import useRouter and instantiate it at the top then this becomes a router.push() call

src/routes/profile/ProfileView.vue
```diff
@@ -0,0 +1,161 @@
...
+}
+
+onMounted(fetchGames);
+
+onUnmounted(() => {
+  myGamesStore.resetGames();
+});
+</script>
+
+<style>
```
itsalaidbacklife:
```suggestion
<style scoped>
```

src/routes/profile/ProfileView.vue
Stasgysk: Added useRouter

src/routes/profile/ProfileView.vue
Stasgysk: Changed

itsalaidbacklife (approved):
Excellent work! This is such a cool feature; I'm delighted to share it with the players

PR #1292, opened 2025-11-11
[Feature]: Error handling UX for Lobby

### LINKED ISSUE #1008: [Feature]: Error handling UX for Lobby
### Feature Summary

We should render errors from the `ready` endpoint in a snackbar in the `LobbyView` page

### Detailed Description

To do this, we should update the errors thrown in `api/controllers/game/ready` itself and in its delegated helper `api/helpers/deal-cards.js` so that they throw translation key strings that can be used in the `i18n` translation system.

We should add the text currently thrown in these errors to `src/translations/en.json` in the `lobby` section in a new `lobby.error` block, and add translations for all other language files in the `src/translations` in their respective languages, in the same place.

There is already a `BaseSnackbar` instance in `src/routes/lobby/LobbyView.vue` that is currently used to notify users when the game switches between ranked and casual mode. This should have its color and message set dynamically so that it keeps the `'surface-2` color and the existing message for the existing use case, and it changes to the 'error' color and shows translated text based on the error's key when the request to 'ready' throws an error.

OliverJZ:
- Resolves #1008 

## Please check the following

- For New Features:

### DIFF
--- a/config/session.js
+++ b/config/session.js
@@ -37,7 +37,7 @@ module.exports.session = {
   // https://sailsjs.com/documentation/reference/configuration/sails-config-session#?properties
   name: 'cuttle.sid',
   cookie: {
-    // 7d * 24hrs * 60min * 60s * 1000ms
-    maxAge: 7 * 24 * 60 * 60 * 1000, // 7 days
+    // 365d * 24hrs * 60min * 60s * 1000ms
+    maxAge: 365 * 24 * 60 * 60 * 1000, // 365 days
   },
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cuttle",
-  "version": "14.1.1",
+  "version": "14.1.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "cuttle",
-      "version": "14.1.1",
+      "version": "14.1.2",
       "dependencies": {
         "@unhead/vue": "^1.11.14",
         "chart.js": "^4.5.1",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "cuttle",
   "description": "The deepest card game under the sea",
-  "version": "14.1.1",
+  "version": "14.1.2",
   "dependencies": {
     "@unhead/vue": "^1.11.14",
     "chart.js": "^4.5.1",


src/routes/lobby/LobbyView.vue
```diff
@@ -87,13 +87,12 @@
       </v-row>
     </v-container>
     <BaseSnackbar
-      v-model="gameStore.showIsRankedChangedAlert"
+      v-model="snackbarOpen"
```
itsalaidbacklife: I'm sorry to have left your questions hanging in the issue until now; I see you asked about this there.

The reason this ref is in the store is because it can be triggered by the opponent making a request to toggle ranked, which comes in as a socket event that's handled in `src/plugins/sockets/gameListEvents.js` in the `handleIsRanked()` function. I'm open to other ways of organizing it if there's a reasonable alternative but the trick is that this change could happen due to another user's actions and we need to be able to open the snackbar in the socket event handler

src/routes/lobby/LobbyView.vue
```diff
@@ -153,15 +157,34 @@ async function ready() {
...
+
 async function setIsRanked() {
   await gameStore.requestSetIsRanked({
     isRanked: gameStore.isRanked,
   });
+
+  showSnackbar(
+    `${t('lobby.rankedChangedAlert')} ${gameStore.isRanked ? t('global.ranked') : t('global.casual')}`,
+    'surface-2'
+  );
```
itsalaidbacklife: If the socket event needs to handle the snackbar for the case where the opponent makes the change and that same event is fired when the current user makes the request here, we can handle it solely in the socket handler and it will show up regardless of which user makes the request

src/stores/game.js
```diff
@@ -520,7 +520,7 @@ export const useGameStore = defineStore('game', () => {
         if (jwres.statusCode === 409) {
           return reject({ message: res.message, gameId: id.value, code: res.code });
         }
-        return reject(new Error('Error readying for game'));
+        return reject(new Error(res.message));
```
itsalaidbacklife: Probably good to have a fallback for malformed errors
```suggestion
        return reject(new Error(res.message ?? 'Error readying for game'));
```

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -112,8 +112,47 @@ describe('Lobby - Page Content', () => {
       cy.get('[data-cy=opponent-indicator] [data-cy="lobby-card-container"]').should('not.have.class', 'ready');
     });
   });
+
+  it('shows error snackbar when ready() returns a translated error key', () => {
+    cy.window().its('cuttle.gameStore')
+      .then((store) => {
+        cy.stub(store, 'requestReady').rejects({ message: 'lobby.error.fallback' })
+          .as('readyStub');
```
itsalaidbacklife, line 120: Nice trick here; that's handy

tests/e2e/specs/out-of-game/lobby.spec.js
```diff
@@ -160,6 +199,21 @@ describe('Lobby - Page Content (Ranked)', () => {
...
+    cy.get('[data-cy=lobby-snackbar] .v-snackbar__content')
+      .should('contain', 'Game Mode changed to'); // existing English string
+
+    // color (Vuetify): wrapper gets bg-surface-2
+    cy.get('[data-cy=lobby-snackbar] .v-snackbar__wrapper')
+      .should('have.class', 'bg-surface-2');
+
+    cy.get('[data-cy="close-snackbar"]').click();
+    cy.get('[data-cy=lobby-snackbar] .v-snackbar__wrapper').should('not.exist');
+  });
```
itsalaidbacklife: Maybe this should be converted to a test that checks that switching between errors and ranked mode updates shows the correct snackbar color? Otherwise this is mainly the same as the other tests of switching from ranked to casual

itsalaidbacklife:
Added #1299 to refactor snackbars into consolidated global component + store

PR #1295, opened 2025-11-15
Allow users to spectate as p1

### LINKED ISSUE #1227: [DevEx]:  Allow users to spectate as p1

dingraham:
## Issue number

- Resolves #1227 

- Frontend - Added pNum validation in setupSpectate function 
- Game Store - Added useRoute import to access query parameters, when spectating, checks route.query.pNum first
Falls back to user's original pNum if they played in the game
- Game History Store - Now includes current pNum in generated spectate URLs
- GameOverDialog - Preserves pNum when navigating to rematch game to ensures spectator perspective is maintained across game transitions
- Adds a new Cypress test that validates both p0 and p1 for a replay

## Please check the following

- For New Features:
  - [X] Have tests been added to cover any new features or fixes?


src/routes/game/components/dialogs/components/gameOver/GameOverDialog.vue
```diff
@@ -308,6 +308,7 @@ export default {
             ...route,
             query: {
               gameStateIndex,
+              pNum: this.gameStore.myPNum,
```
itsalaidbacklife, line 311: My leaning is to skip the prop and just have the dialog read directly from the store. That's generally considered not best practice to make child components "think" like that, and instead to make most components "dumb" and just render their props, but for Cuttle, we tend to lean heavily and directly in the store across the app.

src/stores/game.js
```diff
@@ -116,17 +117,36 @@ export const useGameStore = defineStore('game', () => {
...
+
+      // Check if a complete game is viewed by a player
+      if (userWasPlayer && gameIsComplete) {
+        return userPNum;
+      }
+
       return 0;
     }
     const pNum = players.value.findIndex(({ username }) => username === authStore.username);
     return pNum > -1 ? pNum : null;
```
itsalaidbacklife: Logic is correct. I'm wondering if we can refactor this to DRY up the variables for looking up the pNum by username in the players array, given that we do that in both spectate and non spectate cases

tests/e2e/specs/in-game/spectating/replays.spec.js
```diff
@@ -26,17 +26,28 @@ describe('Rewatching finished games', () => {
 
     beforeEach(createAndFinishCasualMatch);
 
-    it('Rewatches someone else\'s casual match', function () {  
+    it('Rewatches another players casual match as pNum0', function () {
```
itsalaidbacklife: Nitpick: we call them p0 and p1 for the players 
```suggestion
    it('Rewatches another players casual match as p0', function () {
```

tests/e2e/specs/in-game/spectating/replays.spec.js
```diff
@@ -26,17 +26,28 @@ describe('Rewatching finished games', () => {
...
       cy.visit('/');
       cy.signupPlayer(myUser);
+      const pNumToSpectate = 0;
 
-      rewatchCasualMatch(this.replayGameId);
+      rewatchCasualMatch(this.replayGameId, pNumToSpectate);
+    }); // end it('Watches a finished game clicking through the moves one at a time')
+
+    it('Rewatches another players casual match as pNum1', function () {
```
itsalaidbacklife:
```suggestion
    it('Rewatches another players casual match as p1', function () {
```

tests/e2e/specs/in-game/spectating/replays.spec.js
itsalaidbacklife, line 1: These cases are great. For completeness, can we add two more?
- Spectating your own game (in which you played as p0) from the perspectives of p1
- Spectating a game as which you played as p1 without specifying a pNum query param and verifying that you default to p1

itsalaidbacklife (approved):

PR #1302, opened 2025-12-04
feat(game.js): remove state reset and update on requestGameState

### LINKED ISSUE #1236: [Feature]: Enable transition animations for replays
### Feature Summary

Currently replays 'jump' to the selected gamestate when you step forward and backward without showing the full animations for moves like scuttling, resolving 3's, 4's and 5's. We should adjust the update flow to allow transitions when stepping forward/backward

### Detailed Description

Stepping between moves in replays is managed by the `PlaybackControls.vue`, which makes a call to `gameStore.requestGameState()` when the user selects a next state.

This gets the new game state from the server, and updates the gameStore with the latest data, but it does so in a way that doesn't properly trigger transitions.

Here's the function:
```
    requestGameState(gameId, gameStateIndex = -1, route = null) {
        io.socket.get(`/api/game/${gameId}?gameStateIndex=${gameStateIndex}`, (res, jwres) => {
          switch (jwres.statusCode) {
            case 200:
              this.resetState(); // this clears out old data to ensure we don't keep anything stale
              this.resetPNumIfNullThenUpdateGame(res.game); // this fully updates the game
              return handleInGameEvents(res, route).then(() => { // this should handle transitions
                return resolve(res);
              });
            case 401:
              authStore.mustReauthenticate = true;
              // resolve so we can navigate to gameview & login there
              return resolve(jwres.body.message);
            default:
              return reject(jwres.body.message);
          }
        });
      });
    },
```

The crux of the issue is that we're calling:
1. `resetState()` which fully clears out the game store including game id and myPNum
2. `resetPNumIfNullThenUpdateGame()` which fully updates the game without transitions to the new set (this is currently needed after the reset to ensure we get at minimum the correct game id, and myPNum value
3. Then last we call `handleInGameEvents()` which would fire the transitions if the game state were the previous one instead of the new one.

We need to sort out this flow so that we properly clear out the stale data at the right time, ensure that the necessary data for the game to load is also set (e.g. id and myPNum) and that the transitions can fire when appropriate.

itsalaidbacklife:
## Issue number

- Resolves #1236 

## Please check the following

- For New Features:

### DIFF
--- a/src/plugins/sockets/connectivityEvents.js
+++ b/src/plugins/sockets/connectivityEvents.js
@@ -20,7 +20,7 @@ export async function handleConnect() {
       const gameStateIndex = gameHistoryStore.currentGameStateIndex;
       const response = await gameStore.requestGameState(gameId, gameStateIndex);
       if (response?.victory?.gameOver && response.game.rematchGame) {
-        await gameStore.requestGameState(response.game.rematchGame);
+        await gameStore.requestGameState(response.game.rematchGame, -1, null, true);
         router.push({ name: ROUTE_NAME_GAME, params: { gameId: response.game.rematchGame } });
       }
       return;
--- a/src/stores/game.js
+++ b/src/stores/game.js
@@ -347,7 +347,6 @@ export const useGameStore = defineStore('game', () => {
     }
   }
   async function processScuttle({ game, playedCard, targetCard, playedBy }) {
-    debugger;
     if (!player.value) {
       updateGame(game);
       return;
@@ -469,11 +468,15 @@ export const useGameStore = defineStore('game', () => {
       });
     });
   }
-  function requestGameState(gameId, gameStateIndex = -1, route = null) {
+  function requestGameState(gameId, gameStateIndex = -1, route = null, resetStateBeforeUpdate = false) {
     return new Promise((resolve, reject) => {
       io.socket.get(`/api/game/${gameId}?gameStateIndex=${gameStateIndex}`, (res, jwres) => {
         switch (jwres.statusCode) {
           case 200:
+            if (resetStateBeforeUpdate) {
+              resetState();
+              updateGame(res.game);
+            }
             return handleInGameEvents(res, route).then(() => {
               return resolve(res);
             });


src/stores/game.js
```diff
@@ -473,8 +474,6 @@ export const useGameStore = defineStore('game', () => {
       io.socket.get(`/api/game/${gameId}?gameStateIndex=${gameStateIndex}`, (res, jwres) => {
         switch (jwres.statusCode) {
           case 200:
-            resetState();
-            updateGame(res.game);
```
itsalaidbacklife, line 477: Sooooo I think this is likely to break other things, but it does make the transition animations work, so I'm seeing if this fails tests

src/plugins/sockets/connectivityEvents.js
```diff
@@ -20,7 +20,7 @@ export async function handleConnect() {
       const gameStateIndex = gameHistoryStore.currentGameStateIndex;
       const response = await gameStore.requestGameState(gameId, gameStateIndex);
       if (response?.victory?.gameOver && response.game.rematchGame) {
-        await gameStore.requestGameState(response.game.rematchGame);
+        await gameStore.requestGameState(response.game.rematchGame, -1, null, true);
```
itsalaidbacklife, line 23: I don't love the complicated call signature of this function or the addition of this new arg for this specific case, but ultimately it's not that bad and this will fix the UX issue

src/stores/game.js
```diff
@@ -468,13 +468,15 @@ export const useGameStore = defineStore('game', () => {
...
     return new Promise((resolve, reject) => {
       io.socket.get(`/api/game/${gameId}?gameStateIndex=${gameStateIndex}`, (res, jwres) => {
         switch (jwres.statusCode) {
           case 200:
-            resetState();
-            updateGame(res.game);
+            if (resetStateBeforeUpdate) {
+              resetState();
+              updateGame(res.game);
+            }
```
itsalaidbacklife, line 479: We basically don't want this most of the time but in the spcific case of handling a socket reconnect after a new game has started from a rematch, doing so will keep straight which player the current user is

PR #1304, opened 2025-12-08
fix(leave-lobby): add check that game status is CREATED

itsalaidbacklife:
## Issue number
Games are somewhat frequently losing reference to who's p0 and/or p1 (it's null even when game is finished). This looks to be due to accidental use of the leave-lobby endpoint after a game has finished (eg by pressing the back button to get to the lobby, then hitting the leave lobby button). Adding a check to error if that happens when it's not supposed to, and switching to router.replace() when navigating from lobby to game view to prevent back button from bringing you to the lobby after game completes

- Resolves #

## Please check the following

- For New Features:

### DIFF
--- a/api/controllers/game/leave-lobby.js
+++ b/api/controllers/game/leave-lobby.js
@@ -7,7 +7,6 @@ module.exports = async function (req, res) {
     const promisePlayer = userService.findUser({ userId: req.session.usr });
     const [ game, player ] = await Promise.all([ promiseGame, promisePlayer ]);
   
-    // no-op if the game is complete
     if (game.status !== GameStatus.CREATED) {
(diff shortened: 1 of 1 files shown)


api/controllers/game/leave-lobby.js
```diff
@@ -1,10 +1,16 @@
+const GameStatus = require('../../../utils/GameStatus.json');
```
Haviles04, line 1: nitpick: non blocking ....can we update vite or something to alias this?

itsalaidbacklife: I don't think we can on the backend since it's not using vite and I don't think we have a BE mechanism for this kind of aliasing atm

Haviles04: I figured as much since it wasn't already done 👍

PR #1305, opened 2025-12-09
Fix/subscribe and spectate cleanup

itsalaidbacklife:
- Uses router.replace() when moving into new games to prevent players from using back button to get into lobby or into an old game that's finished
- Revises logic which for redirecting to /game/:gameId from /spectate/:gameId based on player usernames to use server response instead of pinia store to determine when to redirect 
## Issue number

- Resolves #

## Please check the following

- For New Features:

### DIFF
--- a/src/stores/game.js
+++ b/src/stores/game.js
@@ -505,7 +505,7 @@ export const useGameStore = defineStore('game', () => {
       if (err?.statusCode === 409) {
         throw err;
       }
-      const message = err?.message ?? err ?? `Unable to spectate game ${gameId}`;
+      const message = err?.message ?? err?.body?.message ?? err ?? `Unable to spectate game ${gameId}`;
       throw new Error(message);
     }
   }


src/routes/home/components/GameListItem.vue
```diff
@@ -139,10 +139,6 @@ export default {
     },
     spectateGame() {
       this.joiningGame = true;
-      if (this.gameStore.players.some(({ username }) => username === this.authStore.username)) {
-        this.$router.push(`/game/${this.gameId}`);
-        return;
-      }
```
itsalaidbacklife, line 145: This was intended to prevent users from spectating their own games, but the endpoint to spectate already errors if you're in the game and this would sometimes be out of date and direct users to /game instead of /spectate. I think it happened when a user had previously joined a lobby and left and later two other players join and start the game and then the original player tries to spectate that game

Haviles04 (dismissed):

api/controllers/game/spectate/join.js
```diff
@@ -24,7 +24,7 @@ module.exports = async function (req, res) {
     // Can't spectate if you're a player in an ongoing game
     const { p0, p1 } = game;
     if (game.status === GameStatus.STARTED && [ p0?.id, p1?.id ].includes(spectator.id)) {
-      return res.forbidden({ message: 'home.snackbar.cannotSpectate' });
+      return res.status(409).json({ message: 'home.snackbar.cannotSpectate' });
```
itsalaidbacklife, line 27: Switched from forbidden to conflict; I think this better captures the spirit of the issue since it's really "you can't do that right now".

src/plugins/sockets/inGameEvents.js
```diff
@@ -117,7 +117,7 @@ export async function handleInGameEvents(evData, newRoute = null) {
         gameStore.updateGame(evData.newGame);
       }
 
-      router.push(route);
+      router.replace(route);
```
itsalaidbacklife, line 120: Swapped .push() with .replace() when going between games and from lobby into game view. Hoping to prevent users from hitting 'back' to land in the lobby of an ongoing or finished game

src/routes/home/components/GameListItem.vue
itsalaidbacklife: To prevent this from breaking the ability to 'nicely' redirect you to /game/:gameId, I've updated the handling of the request to spectate so that we navigate to /spectate/:gameId but let the router logic handle a 409 error from the BE when requesting to spectate by navigating to /game/:gameId

src/stores/game.js
```diff
@@ -397,15 +397,15 @@ export const useGameStore = defineStore('game', () => {
...
     switch (jwres.statusCode) {
       case 200:
         return resolve(jwres);
       case 401:
         authStore.mustReauthenticate = true;
-        return reject(jwres.body.message);
+        return reject(returnFullResponse ? jwres : jwres.body.message);
       default:
-        return reject(jwres.body.message);
+        return reject(returnFullResponse ? jwres : jwres.body.message);
```
itsalaidbacklife, line 408: This is clunky. I think what would be better than sometimes returning `jwres` and sometimes returning `jwres.body.message` would be to always return `jwres` here and update the callers to expect the error to have `.body.message` and use that for the snackbars. That way we'd have more flexibility to use this function but still let the caller sniff out specific errors and decide how to handle them

src/stores/game.js
```diff
@@ -493,15 +493,19 @@ export const useGameStore = defineStore('game', () => {
...
       if (authStore.mustReauthenticate) {
         id.value = gameId;
         return;
       }
-      const message = err?.message ?? err ?? `Unable to spectate game ${gameId}`;
+      // Failed to join as spectator becuase currently playing
+      if (err?.statusCode === 409) {
+        throw err;
+      }
+      const message = err?.message ?? err?.body?.message ?? err ?? `Unable to spectate game ${gameId}`;
```
itsalaidbacklife, line 508: The change to include `err?.body?.message` is my least favorite part of this PR atm and it really highlights the inconsistency of our error shapes. Basically for the 409 case I wanted to have the actual status code available to the FE for determining whether to trigger the redirect

I _think_ it'd be better to normalize them to expect `jwres` as the rejected content and therefore look for `err?.body?.message` as the source of the error but I think this will take some assessment and app-wide house cleaning so I just jammed this in there to handle the error when a user makes a move in a finished game gracefully

tests/e2e/specs/in-game/spectating/live-spectating.spec.js
```diff
@@ -30,10 +30,26 @@ describe('Spectating Games', () => {
...
+      p0Points: [ Card.TEN_OF_SPADES ],
+      p0FaceCards: [ Card.KING_OF_SPADES ],
+      p1Hand: [ Card.ACE_OF_HEARTS, Card.ACE_OF_DIAMONDS, Card.EIGHT_OF_DIAMONDS ],
+      p1Points: [ Card.TEN_OF_HEARTS ],
+      p1FaceCards: [ Card.KING_OF_HEARTS ],
+    });
     cy.get('@gameId').then((gameId) => {
       cy.visit(`/spectate/${gameId}`);
-      cy.url().should('not.include', `/spectate/${gameId}`);
-      assertSnackbar(`Cannot spectate game because you are playing`, 'error', 'newgame');
```
itsalaidbacklife, line 36: Instead of going home and getting the snackbar you now go to the game without an error

PR #1306, opened 2025-12-11
Style/color cleanup

itsalaidbacklife:
- Removes the unnecessary old `primary` color and renames the current pink `newPrimary` to `primary`.
- Revises colors and layout of rulesview section on tournaments to match new stats view
- Changes the switch color in CreateGameDialog to primary from surface-2

## Please check the following

- For New Features:


src/components/StatsScoringDialog.vue
```diff
@@ -100,7 +100,7 @@ export default {
     },
     activatorColor:{
       type: String,
-      default: 'primary',
+      default: 'surface-2',
```
itsalaidbacklife, line 103: There were two consumers of this component and both set the `activatorColor` to `surface-2` so I made that the default

src/routes/home/components/CreateGameDialog.vue
```diff
@@ -31,13 +31,13 @@
           {{ t('home.playAiContent2') }}
         </h4>
         <div class="d-flex align-center">
-          <StatsScoringDialog activator-color="surface-2" :show-button-text="false" />
```
itsalaidbacklife, line 34: no longer need activator-color since surface-2 is the default prop value

PR #1309, opened 2025-12-18
Feat/minimize dialogs

### LINKED ISSUE #88: [Feature]: Minimize Dialog to view board state
### Feature Summary

Want to be able to minimize the current persistent dialogs to see the game state

### Detailed Description

As a player who is faced with a decision in a dialog (e.g. whether to counter an opponent's one-off), it would be helpful to be able to minimize the dialog in order to see the game board/cards in hands to inform the decision making. 

Should be a client-side feature and refreshing the page should re-open the dialog unminimized.

itsalaidbacklife:
## Issue number

- Resolves #88 

Adds a minimize button to in-game dialogs
image

Which then displays a FAB in the center of the screen to re-open the dialog
image

## Please check the following

- For New Features:

### DIFF
--- a/src/routes/game/components/dialogs/components/ThreeDialog.vue
+++ b/src/routes/game/components/dialogs/components/ThreeDialog.vue
@@ -3,8 +3,8 @@
     v-if="oneOff"
     id="three-dialog"
     v-model="show"
-    scrollable
     :title="t('game.dialogs.threeDialog.title')"
+    scrollable
     minimizable
   >
     <template #body>
@@ -33,7 +33,6 @@
 </template>
 
 <script>
-import { ref, computed, watch } from 'vue';
 import { useI18n } from 'vue-i18n';
 import BaseDialog from '@/components/BaseDialog.vue';
 import CardListSortable from '@/routes/game/components/CardListSortable.vue';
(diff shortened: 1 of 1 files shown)


src/components/BaseDialog.vue
```diff
@@ -9,7 +9,21 @@
   >
     <template #activator="{ props }">
       <span v-bind="props">
-        <slot name="activator" />
+        <template v-if="minimizable && isMinimized">
```
Haviles04: Does this need to be wrapped with a template tag?

itsalaidbacklife: I'm not sure. Currently I'm trying to leverage the :active prop to animate it, so my rough idea is that the component will only exist if the dialog is minimizable and will only be visible if it's active. But that introduced the annoyance of every dialog's activator existing at the same time even when the dialog is unused, making selection during testing annoying. 

I'll try putting v-if and :active on the v-fab itself here and see if that works and still animates

itsalaidbacklife: I've checked and it does not, so long as it's instantiated before we attempt to toggle visibility. So I've removed the `<template>` tag and applied `v-if="minimizable"` but kept `:active="isMinimized" so it will exist if the dialog allows minimization and it will display if the dialog is currently minimized

PR #1320, opened 2026-01-14
Feat/prevent three for three

### LINKED ISSUE #1318: [Feature]: Disallow picking a 3 with a 3
### Feature Summary

We should prevent players from selecting a three from the scrap with their three

### Detailed Description

Playing a three for it’s one-off lets a player choose a card from the scrap pile and put it in their hand. Selecting a three as your target stalls the game by making a non-move which doesn’t meaningfully progress the game. Its only purpose is to avoid making other moves.

To make gameplay more enjoyable, we should disallow this possibility:
- `api/helpers/game-states/resolve-three/validate.js` should throw a new error if the selected card is a three 
  - this error should be be translated in all language files
- `api/helpers/game-states/one-off/validate.js` should error if the playedCard is a three and all cards in the scrap are threes
  - this error should be translated in all language files
- the `ThreeDialog` should remove threes from the displayed cards
- The tests should be adjusted to reflect the above changes
- Update the move descriptions used in the RulesView and MoveChoiceCards in each translation file to specify that you take a "non-three" card to your hand

itsalaidbacklife:
Prevents players from threeing for a three

## Issue number

- Resolves #1318

## Please check the following

- For New Features:


tests/e2e/specs/in-game/one-offs/3_threes.spec.js
```diff
@@ -14,22 +14,81 @@ describe('Playing THREEs', () => {
...
+          try {
+            await store.requestResolveThree(Card.THREE_OF_DIAMONDS.id);
+            cy.then(() => {
+              // Fail test if backend allows request
+              assert.fail('Expected 400 error when requesting to target 3 when resolving three, but came back 200');
+            });
+          } catch (err) {
+            expect(err).to.eq('game.snackbar.oneOffs.three.cannotTargetThree');
+          }
+        });
```
itsalaidbacklife, line 69: TODO: write issue to standardize approach to validating backend errors for bad requests. I found that without the `cy.then()`, cypress would not correctly fail the test on a failed first assertion, though it would sortof work anyway becasue the failure would throw an error which would kick us into the catch block with a different error than the expected error string and _that_ would fail the test. But wrapping this in cy.request seems to make it work correctly. It would be nice to generalize this into a command for calling arbitrary store functions and asserting that they error, then rolling this out across usages

itsalaidbacklife: Created #1321

PR #1354, opened 2026-04-28
 make start:preview script cross-platform by replacing ';' with '&&'

dorothy122:
## Issue number

- Resolves # 

## Please check the following

- For New Features:

- Fixes a cross-platform issue in the 'start:preview' npm script
- It previously used ; - `npm run build;npm run start:server` but this failed on Windows. So I have changed the ; to && so that it should run on all platforms. I have tested this on Windows but not on any other platforms.


package.json
```diff
@@ -12,6 +12,7 @@
     "lodash": "4.17.21",
     "machinepack-passwords": "2.3.0",
     "marked": "^16.4.1",
+    "node": "^24.15.0",
```
itsalaidbacklife, line 15: Node shouldn't be required as a dependency of the project. It's instead listed under "engines" below. 
```suggestion
```

package.json
```diff
@@ -74,7 +75,7 @@
     "start:serve": "vite preview",
     "start:client": "vite --host",
     "start:server": "node app.js",
-    "start:preview": "npm run build;npm run start:server",
+    "start:preview": "npm run build && npm run start:server",
```
itsalaidbacklife, line 78: Interesting catch. Do I take it you use cmd? I'd only used powershell when working with windows and that uses ; the same way bash does, so I'd never encountered this issue. It seems a noninvasive improvement to cross compatibility

itsalaidbacklife (commented):
Looks good. Requesting to remove "node" from dependencies and then this is ready

itsalaidbacklife:
As a last change, since you've updated a dependency, we need to update package-lock.json as well, which contains the exact versions of the full dependency tree. To do this, after removing node from package.json, please run
`npm install` then commit and push the changes that makes to `package-lock.json`

itsalaidbacklife:
@dorothy122 the need to update package-lock.json looks to be what caused the tests to fail since `npm ci` errors when the lock file is out of date. 

First, remove `'node'` from dependencies, and then you can either update the lock file with `npm install` and committing the changes that creates, or you can roll back the update to Vue devtools in package.json

itsalaidbacklife:
@dorothy122 are you still interested in completing this improvement?

itsalaidbacklife:
Closing as stale

PR #1360, opened 2026-05-18
feat(agents): agentic infrastructure - skills, subagents, hooks, and cross-tool config

seriouslysean:
## Issue number

No linked issue — developer experience infrastructure.

## Please check the following

- [x] Do the tests still pass? (`npm run test:unit` 3/3, `npm run test:agents` 29/29)
- [x] Is the code formatted properly? (`npm run lint` clean)
- For New Features:
  - [x] Have tests been added to cover any new features or fixes? (29 vitest tests for git-guardrails)
  - [ ] Has the documentation been updated accordingly? (`.agents/README.md` and `.agents/docs/` are the documentation)

This PR adds agentic infrastructure — it has no effect on game behavior. Verification steps:

1. Symlinks resolve: `readlink .claude/skills` → `../.agents/skills`; same for `.claude/agents`, `.gemini/skills`, `.gemini/agents`
2. Hook fires on `main`: `git commit --allow-empty -m "test"` while on `main` should be denied with a clear message
3. Hook passes on a feature branch: same command on any non-`main` branch should be allowed
4. `AGENTS.md` / `CLAUDE.md` / `GEMINI.md` unchanged: `git diff main -- AGENTS.md CLAUDE.md GEMINI.md` is empty

**What's here vs what's next:** The reference skills (`ctl-sails-patterns`, `ctl-vue-patterns`, `ctl-test-patterns`, `ctl-game-domain`) are currently opt-in. A follow-up will convert these to `.claude/rules/` path-specific rules so they load automatically when a matching file is touched, without relying on keyword routing.


.agents/agents/ctl-docs-reviewer.md
itsalaidbacklife, line 1: I love this. If it can keep docs in sync, that will be a real win

.agents/docs/game-domain.md
```diff
@@ -0,0 +1,98 @@
...
+
+```js
+// In game store / API responses
+{ rank: 7, suit: 0 }   // 7 of Clubs
+
+// In GameCard class (src/stores/game.js)
+card.name  // "7♣️"
+
+// In Cypress selectors
+'[data-player-hand-card=7_0]'   // rank_suit
```
itsalaidbacklife, line 54: An aside: I've been wanting to refactor these selectors to replace 7_0 with 7H (the id) and then refactor the test references to use functions to get cards by id.

It would be exciting to see if the docs review can ensure that this reference is updated accordingly

.agents/docs/github.md
```diff
@@ -0,0 +1,55 @@
...
+| `minor-version` | `x.y.z → x.Y+1.0` |
+| `major-version` | `x.y.z → X+1.0.0` |
+
+The label is applied by the core team, not the contributor. However, the PR author should suggest the appropriate level in the PR description.
+
+Do not manually bump `package.json` — CI commits the bump after the PR merges.
+
+### PR template
+
+Fill the template completely. Link the PR to its issue with `Closes #N` in the body.
```
itsalaidbacklife, line 28:
```suggestion
Fill the template completely. Link the PR to its issue with `Resolves #N` in the body.
```
(Just to match the template)

.agents/agents/ctl-performance-reviewer.md
```diff
@@ -0,0 +1,100 @@
...
+allowed-tools: Read, Grep, Glob, Bash
+---
+
+# Performance Reviewer
+
+Focused performance review for Cuttle changes. Called by `ctl-code-review`.
+
+## Scope
+
+- Socket listener cleanup (listeners added in setup must be removed in teardown)
```
itsalaidbacklife, line 14: Maybe mention events in general?
```suggestion
- Socket and event listener cleanup (listeners added in setup must be removed in teardown)
```

.agents/agents/ctl-performance-reviewer.md
```diff
@@ -0,0 +1,100 @@
...
+grep -n "ref({" src/stores/*.js
+```
+
+Flag large object refs in hot paths (game state, hand, points).
+
+### 4. Database query efficiency
+
+In Sails controllers, `populate` calls are expensive. Flag:
+- Unnecessary `populate` on fields not used in the response
+- Missing `populate` that causes N+1 (accessing `game.p0.username` without populating `p0`)
```
itsalaidbacklife, line 64: This is neat and the poor man's typscript lol

.agents/agents/ctl-performance-reviewer.md
```diff
@@ -0,0 +1,100 @@
...
+
+```bash
+grep -n "\.populate\|\.find\|\.findOne" <changed files> --include="*.js"
+```
+
+### 5. Vue component re-render scope
+
+Components that subscribe to large portions of the game store will re-render on any game event. Flag components that import the entire `useGameStore` but only use one or two fields.
```
itsalaidbacklife, line 72: What's the correct way to do consume only a portion of a store?

.agents/agents/ctl-performance-reviewer.md
```diff
@@ -0,0 +1,100 @@
...
+
+### 6. Vite build impact
+
+```bash
+# Check for large static imports
+grep -n "import.*from.*node_modules" <changed Vue/JS files>
+```
+
+Flag imports of large libraries where a smaller alternative exists.
```
itsalaidbacklife, line 81:
```suggestion
Flag imports of large libraries where a simple vanilla js, or smaller alternative exists.
```

.agents/docs/sails-patterns.md
```diff
@@ -0,0 +1,168 @@
...
+    } catch (_) {}
+
+    const message = err.raw?.message ?? err;
+    switch (err?.code) {
+      case CustomErrorType.FORBIDDEN:
+        return res.forbidden({ message });
+      default:
+        return res.serverError({ message });
+    }
+  }
```
itsalaidbacklife, line 36: I'm wondering if we should call out that the pattern of locking and unlocking the game is only used for mutating updates to a specific game. Or is this more to document that controller actions are a thing and this is their general syntax?

.agents/docs/vue-patterns.md
```diff
@@ -0,0 +1,135 @@
...
+
+Cuttle uses `vue-i18n`. All user-visible strings go through `$t('key')`:
+
+```vue
+<template>
+  <span>{{ $t('game.draw') }}</span>
+</template>
+```
+
+Translation files are in `src/i18n/`.
```
itsalaidbacklife, line 106:
```suggestion
Translation files are in `src/i18n/`. All translation files should have matching keys in matching order, at all times. Otherwise the unit tests for translation would fail and users viewing the site in particular languages would not see the appropriate text at the appropriate time
```

.agents/skills/ctl-code-review/SKILL.md
itsalaidbacklife, line 1: This whole agent structure is really interesting. I don't understand how it fits into the lifestyle. Do you invoke it like a skill? Is this somehow picked up by copilot etc when reviewing PR's in github? How do you envision this being used?

.agents/skills/ctl-git-commit/SKILL.md
```diff
@@ -0,0 +1,91 @@
...
+# Git Commit
+
+Stage and commit changes using Cuttle's conventional commit format.
+
+## Hard rules
+
+- Never `git add -A` or `git add .` — stage specific files only.
+- Never `--no-verify` or any hook bypass.
+- Never bump `package.json` version — CI handles this automatically.
+- Commits on `main` are blocked by the git-guardrails hook. Create a feature branch first.
```
itsalaidbacklife, line 17: So far I haven't installed the gh cli personally or configured an agent to use it to interact with PRs, and I've been doing my commits by hand, so I'm green in this area. 

In order to be comfortable using such a set up on my machine personally, I would want to use a scoped access token that explicitly prevented the agent from pushing to main. Do you know if that's possible?

.agents/skills/ctl-sails-patterns/SKILL.md
itsalaidbacklife, line 1: There looks to be a lot of overlap between this file and the docs/sails-patterns.md reference. Should this simply import that file for discoverabiliry?

.agents/skills/ctl-test-patterns/SKILL.md
```diff
@@ -0,0 +1,176 @@
...
+  it('does something', () => {
+    cy.loadGameFixture(0, {
+      p0Hand: [Card.SEVEN_OF_CLUBS],
+      p0Points: [Card.TWO_OF_CLUBS],
+      p1Points: [Card.SIX_OF_HEARTS],
+    });
+
+    cy.get('[data-player-hand-card=7_0]').click();
+    cy.get('[data-move-choice=untargetedOneOff]').click();
+
```
itsalaidbacklife, line 52: Maybe we should add the general pattern of interweaving DOM assertions before next moves? the common approach is 

- user plays one-off
- verify overlay appears confirming the opponent can respond
- take opponent action
- assert final result

Something like
```suggestion

// Verify DOM update before taking opponent action
cy.get('#waiting-for-opponent-counter-scrim').should('be.visible');

cy.resolveOpponent();
```

itsalaidbacklife: It may also make sense to use the canonical example of an ace scrapping points to showcase the cards moving from loadFixture to assertGameState

.agents/skills/ctl-test-patterns/SKILL.md
```diff
@@ -0,0 +1,176 @@
...
+  p1Hand: [Card.TEN_OF_CLUBS],
+  p1Points: [Card.SIX_OF_HEARTS],
+  p1FaceCards: [Card.QUEEN_OF_HEARTS],
+  topCard: Card.FIVE_OF_DIAMONDS,
+  secondCard: Card.EIGHT_OF_SPADES,
+  scrap: [Card.TWO_OF_HEARTS],
+});
+```
+
+All fields are optional.
```
itsalaidbacklife, line 88:
```suggestion
All fields are optional.

`deck` may be specified as an alternative to `topCard` and `secondCard` to fully specify the order of all cards remaining in the deck. When `deck` is specified, any cards missing from the fixture are placed into the `scrap` (otherwise missing cards are placed in the `deck`).
```

.agents/skills/ctl-test-patterns/SKILL.md
```diff
@@ -0,0 +1,176 @@
...
+});
+```
+
+## Card data selectors
+
+Format: `data-<player>-<location>-card=<rank>_<suit>`
+
+```js
+'[data-player-hand-card=7_0]'        // 7 of Clubs in player's hand
+'[data-opponent-point-card=6_2]'     // opponent's 6 of Hearts on field
```
itsalaidbacklife, line 108:
```suggestion
'[data-opponent-point-card=6_2]'     // opponent's 6 of Hearts on field
'[data-opponent-hand-card=]'        // card in opponent's hand (regardless of known/unknown rank)
'[data-opponent-hand-card=6_2]'     // opponent's 6 of Hearts in hand

```

.agents/skills/ctl-test-patterns/SKILL.md
```diff
@@ -0,0 +1,176 @@
...
+
+File: `.agents/tools/__tests__/*.spec.mjs`
+
+Config: `.agents/tools/vitest.config.mjs`
+
+Run with: `npm run test:agents`
+
+These test pure Node.js logic (no Sails, no Vue). Use ESM imports.
+
+## TDD workflow (from CONTRIBUTING.md)
```
itsalaidbacklife, line 166: @seriouslysean im fine to leave this as-is atm, but I'd love to pick your brain on your workflow. Heretofore, I've been booting the server and client and cypress outside my agent, then having the agent modify test and app code, running the tests myself, and then telling the agent the issues and guiding its iteration. 

This looks like it aspires to a more fully independent loop where the agent runs the relevant tests, analyzes output, and iterates without dev input. Is that what you envision, and have you worked that way? I'm intrigued by it but have trouble conceptualizing the flow concretely. Would love to put heads together

.agents/skills/ctl-test-patterns/SKILL.md
```diff
@@ -0,0 +1,176 @@
...
+
+1. Write or update the test first.
+2. Run `npm run start:server` in background.
+3. Run `npm run e2e:client` (or `e2e:gui` for interactive).
+4. Implement until tests pass.
+5. Run `npm run lint` and `npm run test:unit`.
+
+## Full reference
+
+`.agents/docs/testing.md`
```
itsalaidbacklife, line 176: Generally the overlap between the skill and docs versions of these files makes me think it might be better to have the skill simply @import the doc(s) to centralize the source of truth and reduce drift. That way the point of the skill is essentially frontmatter for scoping and managing discoverabiliry of interrelated docs.

The downside might be a reduced capacity to provide summarized docs for reduced token usage, but it's not clear to me how the agent would know when to escalate, and if it does read the full docs, it's parsing much of the same content redundantly. Penny for your thoughts

.agents/tools/__tests__/git-guardrails.spec.mjs
itsalaidbacklife, line 1: This is cool af

.agents/tools/__tests__/git-guardrails.spec.mjs
```diff
@@ -0,0 +1,131 @@
+import { describe, it, expect } from 'vitest';
+import { checkDestructive, isCommitOnMain, isPushOnMain, evaluate } from '../git-guardrails.mjs';
+
+describe('git-guardrails', () => {
+  describe('checkDestructive', () => {
+    it('allows a clean commit command', () => {
+      expect(checkDestructive('git commit -m "fix: thing"')).toBeUndefined();
```
itsalaidbacklife, line 7: So you he guardrails return undefined when the action is allowed and otherwise a string describing the reason it's disallowed? Is that built into agent specs or do we have logic that informs the agents how to interpret the results of the guardrails?

.claude/agents
```diff
@@ -0,0 +1 @@
+../.agents/agents
```
itsalaidbacklife, line 1: Does this need '@import`?

.claude/skills
itsalaidbacklife, line 1: Unfortunately, I don't think this will work as intended. My understanding is that Claude code is specifically hard coded to look for skills in .claude and that it ignores the contents of the folder which are not in the official skill format, meaning this file will be ignored and the skills won't be properly discoverable. 

It looks like the options are to
1. Forgo generic skills and simply put these in .claude (do you use another agent personally? For myself I'd be content with just getting Claude configured well, though as an open source tool, the flexibility to be generic makes sense. This is my current leaning for simplicity's sake, but especially if you're interested in  using another agent yourself, then we can look for better solutions
2. We define symlinks from the specific agent directories to the genetic .agents directory using a post install script in package.json. This might be the most robust but also finicky depending on OS nonsense ie windows will be a PITA. 
3. We define some automation to duplicate whole files based on .agents as a source of truth, overwriting the contents of .claude etc on commit or some other hook. This feels the clunkiest but may be the most reliable method for actual cross agent, is agnostic doc sharing.

I lean towards ditching the agnostic approach and sticking with .claude. What do you think @seriouslysean ?

itsalaidbacklife (commented):
Very cool! Finally made it through a full review. Looking forward to discussing

PR #1366, opened 2026-08-01
feat(LoginView): add confirm-password field to signup form

itsalaidbacklife:
## Summary

Adds a **confirm password** field to the signup form (`src/routes/login/LoginView.vue`). It renders only during signup (not login), mirrors the visibility/type of the existing password field, and blocks submission with a visible error state when the two passwords don't match or the confirm field is empty.

This should help both to prevent signups with typo'd passwords, and to make the signup view more clearly differentiated from login

This is **purely frontend validation** — the backend still receives a single `username` + `password`; `submitLogin` is unchanged.

image

## Changes

- **`LoginView.vue`**: New `confirm-password` field wrapped in `<template v-if="!isLoggingIn">` so its validation rule only participates in the signup form (Vuetify drops unmounted inputs from `v-form` validation, so login is unaffected). Adds `confirmPw` data, a `passwordsMatch` rule ("Passwords must match", matching the existing hardcoded rule-message style), and clears the field in `switchMode`/`handleLogin`. The field shares the existing eye-toggle via `showPass`.
- **`en.json`**: Adds `login.confirmPassword` label (other locales fall back to `en`).
- **`authentication.spec.js`**: Fills the new field in all signup-mode UI submissions, and adds two new cases:
  - Rejects signup when confirm password does not match (asserts the "Passwords must match" message, disabled submit, stays on `/signup`).
  - Rejects signup when confirm password is empty (disabled submit, stays on `/signup`).

  Both new cases deliberately avoid the `forceFormSubmit` helper, since force-submitting would send a valid username/password the backend would accept.

## Verification

- `npm run lint` passes.
- Full check needs the app/db running: `npx cypress run --spec tests/e2e/specs/out-of-game/authentication.spec.js`


Haviles04 (approved):

Haviles04 (changes requested):
I think more translation files need updates

Haviles04 (dismissed):
Lgtm

src/routes/login/LoginView.vue
```diff
@@ -322,6 +360,7 @@ export default {
     },
     switchMode() {
       this.pw = '';
+      this.confirmPw = '';
```
itsalaidbacklife: It's actually annoying that we clear out the pw on switch; let's remove these both

tests/e2e/specs/out-of-game/authentication.spec.js
```diff
@@ -163,17 +170,33 @@ describe('Signing Up', () => {
...
+    cy.get('[data-cy=confirm-password]').type(myUser.password + 'mismatch');
+    cy.get('#confirm-password-messages').should('contain', 'Passwords must match');
+    cy.get('[data-cy=submit]').should('be.disabled');
+    assertFailedAuth('/signup');
+  });
+  it('Rejects signup when confirm password is empty', () => {
+    cy.get('[data-cy=username]').type(myUser.username);
+    cy.get('[data-cy=password]').type(myUser.password);
+    cy.get('[data-cy=submit]').should('be.disabled');
+    assertFailedAuth('/signup');
```
itsalaidbacklife, line 201: For both cases, let's add checks that hitting enter also fails the auth

itsalaidbacklife (commented):
Two todos

PR #1370, opened 2026-08-02
resolved issue #1293 modified es-lint config file

### LINKED ISSUE #1293: [DevEx]: Prevent unscoped `<style>` blocks with eslint rule enforce-style-attribute
### Improvement Summary

We should configure the lint rule https://eslint.vuejs.org/rules/enforce-style-attribute to force all `<style>` blocks to be `scoped` except for the one in `App.vue`

### Detailed Description

We generally require all style blocks to be scoped except for `App.vue`, where we put global styles. But we don't have a consistent way to enforce this. We should configure the above lint rule to require style blocks to be `<style scoped>` but add an override for `App.vue` that prevent the rule from applying to that file. Roughly this

```

  rules: {
    'vue-scoped-css/enforce-style-type': ['error', { allows: ['scoped', 'module'] }],
  },
  overrides: [
    {
      files: ['src/App.vue'],
      rules: {
        'vue-scoped-css/enforce-style-type': 'off',
      },
    },
  ],
```

princekamariya:
## #1293

- Resolves #1293 

## Please check the following

- [✓] Do the tests still pass? (see [Run the Tests](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#run-the-tests))
- [✓] Is the code formatted properly? (see [Linting (Formatting)](https://github.com/cuttle-cards/cuttle/blob/main/docs/setup-development.md#linting-formatting))
- For New Features:

Adds the `vue/enforce-style-attribute` ESLint rule so every `<style>` block must be `scoped`, with an override disabling the rule for `src/App.vue` (which holds global styles).

One other file in the repo, `src/components/MarkdownContent.vue`, also had a plain `<style>` block. Since its markdown content is rendered via `v-html` (and scoped CSS attributes aren't applied to `v-html`-injected DOM), it was converted to `<style scoped>` using `:deep(p)` so the padding rule still reaches the dynamically rendered markdown.

To verify:
1. Run `npm run lint` ,should pass with no errors.
2. Try adding a plain (non-scoped) `<style>` block to any `.vue` file other than `App.vue` , `npm run lint` should now fail with `vue/enforce-style-attribute`.
3. Confirm `src/App.vue`'s existing global `<style>` block still lints clean (override working).
4. Confirm `src/components/MarkdownContent.vue`'s markdown-rendered padding still displays correctly in the browser


itsalaidbacklife (approved):
Thanks for your contribution!

PR #1379, opened 2026-09-17
feat: rework nines to topdeck the target card

itsalaidbacklife:
## Issue number

- Resolves #

## What this does

A Nine one-off now puts its target **on top of the deck** instead of returning it to its controller's hand. The opponent's next draw is their own card coming back, so the Nine costs them a whole turn rather than one card-play.

Nines are one of the highest ranks in the deck, which makes them excellent scuttle fuel — so the one-off was almost never the right line. This is a deliberate buff, shipped as a **beta for the Spades season** with a home-page announcement and a community poll toward the end of the year.

This is an **alternative** to the double-bounce rework in #1378, branched fresh from `main` so the two can be compared and one chosen.

### The new rule

- A Nine one-off puts **one** target card on top of the deck.
- Legal targets: point cards, Royals/glasses, and the **top Jack** of a stack. Buried jacks are not targetable.
- A targeted jack's host point card still reverts to its owner.
- Non-targeted attachments on a targeted point card are still scrapped.
- Queen rules are **unchanged** (1 Queen → only the Queen is targetable; 2+ → no Nine one-off).
- Nothing is frozen any more.
- Playing a 9 for points or as a scuttle is unchanged.

**No schema change, so no migration is required.** The deck stays face down — `create-socket-events.js` already hides it — and the game log names the topdecked card, so both players have the information.

## Notable findings along the way

**`isFrozen` was never persisted.** It was re-derived on unpack from *"resolved is a 9 AND this card equals `targetCard`"*. That derivation runs over every converted array **including `deck`**, so it had to be deleted — otherwise it would have stamped `isFrozen` on the card the nine had just topdecked. Since `resolve/nine.js` was its only producer, the freeze mechanic is now dormant. The rest of the plumbing is deliberately left in place so a revert stays a clean `git revert`; a follow-up issue should remove it if the poll keeps the change.

**Pre-existing bug fixed:** `findTargetCard` matches *any* jack in *any* stack, while the resolver blindly `pop()`ed the top one — so a raw socket request naming a buried jack scrapped the wrong card. New `is-top-jack.js` helper, called from both `one-off/validate.js` and its `seven-one-off` twin.

**Pre-existing AI bug fixed:** jack move bodies omitted `targetType`, so `findTargetCard` returned `null` and the validator rejected every one the bot attempted. Nines now also enumerate point-card targets, which topdecking makes worth playing.

**Pre-existing leak fixed:** `resolve/execute.js` never cleared `oneOffTarget` / `oneOffTargetType` on the fizzle path (masked because the next one-off overwrote them).

**Dead animation code removed:** the four nine transition computeds branched on `lastEventTargetType`, which is always `null` — `resolve/execute.js` nulls `oneOffTargetType` before the socket event is built, and `create-socket-events.js` never puts it on `lastEvent`. Reviving it would need a new column, so instead each row gets one direction that suits both things a Nine can do to it.

**Hand limit interaction is gone entirely.** The nine was the only effect that handed a card to the inactive player, so after this change you can never start a turn over the hand limit. Two `handLimit.spec.js` tests covering that state are removed — their draw→discard and discard-2 paths are already covered by the Drawing and Fives blocks — and the pass-triggers-discard test (the only coverage of that path) now seeds the over-limit hand directly, with a comment explaining why.

## Please check the following

- For New Features:

### Local results

| Check | Result |
|---|---|
| `npm run lint` | clean |
| `npm run test:unit` | 86 / 86 |
| Full `npx cypress run` (34 specs) | **372 passing, 1 failing, 5 pending** |

The single failure is `out-of-game/stats.spec.js`, which is **unrelated and pre-existing**. **CI is green on this branch** (all 5 cypress containers) — the failure only reproduces locally, and only on a fast, idle machine.

Root cause, worth its own issue: `statsFixtures.js` computes the Current Season `startTime` (line 30) and the Week-1 game `updatedAt` (line 571) as two separate `dayjs.utc().subtract(2, 'week')` calls in the same module, while `StatsController` filters with a strict `updatedAt > startTime`. If both land in the same millisecond the game is excluded and week 1 counts 0.

The counter-intuitive part is that **a faster machine makes it more likely to fail**. Measuring the real fixture module, the delta between the two timestamps is 0 or 1 ms:

| Condition | Deltas observed | Result |
|---|---|---|
| Idle CPU | `1 0 0 1 0 1 0 0` | fails ~half the time |
| CPU under load | `1 1 1 1 1 1 1 1` | always passes |

So GitHub's shared ubuntu-22.04 runners are slow enough that the fixture always straddles a millisecond tick and the test passes, which is why this has never been seen in CI. Both the fixture and the controller logic predate this PR by over a year (`statsFixtures.js` relative dates since before #1097, Oct 2024; `StatsController` filter since #796, Nov 2023).

The fix would be a one-liner — capture the timestamp once and derive both from it, or make the filter `>=`.

### Test coverage

`9_nines.spec.js` (16 tests) is rewritten around **draw-back proofs** — after the nine resolves, the next draw returns exactly the topdecked card, which demonstrates both top-of-deck placement and the absence of freezing in one assertion. New coverage for:

- the **empty-deck endgame**: a Nine played at deck-out puts a card back, which re-enables drawing and removes the pass affordance (`pass/validate.js` throws on a non-empty deck, and it breaks the consecutive-pass run `countPasses` uses for the 3-pass stalemate). This is allowed deliberately — it's a real endgame tactic.
- **top-jack-only** targeting, asserted through the UI on a triple-jack stack
- the hand-limit **non**-interaction from both perspectives

Also updated: `targeting_cleanup.spec.js`, `opponent_sevens.spec.js` (extended with the seven → nine → draw-back path for both faceCard and jack targets), `player_sevens.spec.js`, `handLimit.spec.js`, and the `resolveNine` unit fixture.

### Manual pass

Home-page announcement verified rendering (title, both 9s as display cards, both sections, no unresolved i18n keys). All five locales carry the new keys at full parity with real translations.

### Follow-ups

- If the poll **keeps** the change, open an issue to delete the dormant freeze plumbing (`convert-str-to-card`'s `isFrozen` input, the `playedCard.isFrozen` validator guards, `GameCard.vue`'s snowflake/`.frozen` styles, `MoveChoiceOverlay` handling, the two i18n keys ×5, `SnackBarError.FROZEN_CARD`).
- If the poll **reverts** it, `git revert` needs zero data migration, but bump `CURRENT_RULES_VERSION` *forward* to `3.0.0` rather than reusing `1.0.2` — games are stamped with it.
- The rules-page art is now stale and can't be fixed here: `/img/rulesView/oneoffs_nine.png` and the animated GIF in the external `cuttle-cards/cuttle-assets` repo.


api/helpers/game-states/ai/get-move-bodies-for-move-type.js
```diff
@@ -224,6 +247,17 @@ module.exports = {
...
+            // Only nines can target the point card itself
+            if (targetedOneOff.rank === 9) {
+              res.push({
+                moveType,
+                playedBy,
+                cardId: targetedOneOff.id,
+                targetId: opponentPoint.id,
+                targetType: TargetType.point,
+              });
+            }
```
itsalaidbacklife, line 259: Was the ability to play a nine one-off on a point card when resolving a seven entirely missing before?

api/helpers/game-states/is-top-jack.js
```diff
@@ -0,0 +1,32 @@
...
+      type: 'ref',
+      descriptions: 'Player object from the current game state',
+      required: true,
+    },
+  },
+  sync: true,
+  fn: ({ targetId, player }, exits) => {
+    const host = player.points.find(({ attachments }) => attachments.some(({ id }) => id === targetId));
+
+    return exits.success(host?.attachments.at(-1).id === targetId);
```
itsalaidbacklife:
```suggestion
    return exits.success(player.points.some((point) => point.attachments.at(-1)?.id === targetId);
```

api/helpers/game-states/moves/one-off/validate.js
```diff
@@ -86,6 +86,15 @@ module.exports = {
             throw new BadRequestError('Twos can only target royals or glasses');
           }
 
+          // Nines put the target on top of the deck, so a buried jack is never a legal target
+          if (
+            playedCard.rank === 9 &&
```
itsalaidbacklife: This should apply to both twos and nines with the comment saying something like "Only the top attachment can be targeted"

api/helpers/game-states/moves/resolve/execute.js
```diff
@@ -61,6 +61,9 @@ module.exports = {
       result.moveType = MoveType.FIZZLE;
       result.scrap.push(result.oneOff);
       result.oneOff = null;
+      // Nothing is pending anymore, so the target slots have to clear too
+      result.oneOffTarget = null;
+      result.oneOffTargetType = null;
```
itsalaidbacklife, line 66: Does the log ever use these? This looks right but we should ensure the log can report on which card was affected

api/helpers/game-states/moves/resolve/nine.js
```diff
@@ -23,36 +23,47 @@ module.exports = {
...
+      return exits.success(result);
+    }
     const [ targetCard ] = opponent.points.splice(targetIndex, 1);
-    
-    // remove last attachment
-    targetCard.attachments.pop();
+
+    // Remove the targeted jack, which the validator guarantees is the top of the stack
+    const jackIndex = targetCard.attachments.findIndex(({ id }) => id === result.oneOffTarget.id);
+    const [ jack ] = targetCard.attachments.splice(jackIndex, 1);
```
itsalaidbacklife: Since the validator guarantees it's the top jack, should we just assume so here?

api/helpers/game-states/moves/seven-one-off/validate.js
```diff
@@ -76,6 +76,15 @@ module.exports = {
             throw new BadRequestError('Twos can only target royals or glasses');
           }
 
+          // Nines put the target on top of the deck, so a buried jack is never a legal target
+          if (
+            playedCard.rank === 9 &&
```
itsalaidbacklife: Same as above; this restriction applies to twos as well

src/routes/game/GameView.vue
```diff
@@ -553,18 +551,10 @@ export default {
...
-                // Nine on jack causes points to swap control
-                case 'jack':
-                  return Transitions.SLIDE_UP;
-                // Everything else expect cards to move back to hand
-                case 'point':
-                case 'faceCard':
-                default:
-                  return Transitions.SLIDE_DOWN;
-              }
+              return Transitions.SLIDE_UP;
```
itsalaidbacklife, line 580: This value determines both enter and leave transitions. Instead of always making card both enter and leave upwards, we should create a two-stage animation, similar to processThree() in game.js, which should first flip the targeted card over face down, and then transition it towards the deck on its way out. 

Towards the deck depends on the breakpoint: at xs the deck is downwards from the points and down + left from the faceCards. Above xs, for the player points + faceCards, the deck is up + left, and for the opponent points + faceCards, it's straight to the left.

Then the player point card list should have its enter animation from above still, so points coming in after a jack was topdecked enter from the opponent's side. Jacks should transition in the same direction as a their respective point area does, towards the deck. 

The flip animation to hide the targeted card should be doable using the existing visibility check perhaps via the isHidden field

src/routes/game/GameView.vue
```diff
@@ -610,16 +591,10 @@ export default {
...
+            // Nines move the target off the board entirely (to the deck), or hand a
+            // jacked point card back down to the player
             case 9:
-              switch (this.gameStore.lastEventTargetType) {
-                // Nine on a jack switches point card control
-                case 'jack':
-                  return Transitions.SLIDE_DOWN;
-                // Everything else returns cards to hand
-                default:
-                  return Transitions.SLIDE_UP;
```
itsalaidbacklife, line 621: See above re: transitions

src/routes/game/GameView.vue
```diff
@@ -712,18 +679,6 @@ export default {
...
-        case 'point':
-          return this.nineTargetIndex !== null ? this.gameStore.opponent.points[this.nineTargetIndex] : null;
-        case 'faceCard':
-          return this.nineTargetIndex !== null
-            ? this.gameStore.opponent.faceCards[this.nineTargetIndex]
-            : null;
-        default:
-          return null;
-      }
-    },
```
itsalaidbacklife, line 726: Is this actually unused? Was it already unused?

src/routes/home/components/announcementDialog/data/announcementData.js
```diff
@@ -1,23 +1,23 @@
...
-        { text: 'announcement.paragraph' },
-        { text: 'announcement.twitchLink', url: 'https://www.twitch.tv/cuttle_cards' },
-        { text: 'announcement.linkSeparator' },
-        { text: 'announcement.youtubeLink', url: 'https://www.youtube.com/live/mMYULGUXrfs' },
-      ],
+      paragraph: [ { text: 'announcement.paragraph' } ],
+    },
+    {
+      heading: 'announcement.heading2',
+      paragraph: [ { text: 'announcement.paragraph2' } ],
```
itsalaidbacklife: The announcement should contextualize this change as the first of several variations that we will trial during the spades season (two nine buffs and two four buffs), and inviting the user to read the full context and explanation here: https://www.patreon.com/cuttle/posts/help-refine-4-9-169762967

src/stores/game.js
```diff
@@ -104,7 +104,6 @@ export const useGameStore = defineStore('game', () => {
   // Last Event
   const lastEventChange = ref(null);
   const lastEventOneOffRank = ref(null);
-  const lastEventTargetType = ref(null);
```
itsalaidbacklife, line 107: Did we really remove all consumers of this?

tests/e2e/specs/in-game/one-offs/9_nines.spec.js
```diff
@@ -52,37 +41,62 @@ describe('Playing NINES', () => {
...
         p0Hand: [ Card.NINE_OF_HEARTS ],
         p0Points: [ Card.TEN_OF_HEARTS ],
         p0FaceCards: [],
-        p1Hand: [ Card.SIX_OF_HEARTS, Card.QUEEN_OF_HEARTS, Card.ACE_OF_DIAMONDS ],
+        p1Hand: [ Card.SIX_OF_HEARTS, Card.QUEEN_OF_HEARTS ],
         p1Points: [],
         p1FaceCards: [],
         scrap: [ Card.NINE_OF_SPADES ],
       });
+      assertDeckCount(47);
```
itsalaidbacklife: Let's update assertGameState to accept an optional topCard and secondCard and to make the appropriate assertions, then add topCard assertions to each of these

api/helpers/game-states/ai/get-move-bodies-for-move-type.js
itsalaidbacklife: Yes — entirely missing for point cards in the \`SEVEN_ONE_OFF\` branch on \`main\`; it only ever enumerated opponent face cards and top jacks. For twos that omission is correct (they can only target Royals and glasses), so the gap was nine-specific.

The plain \`ONE_OFF\` branch was worse: it also skipped point cards, *and* omitted \`targetType\` on the jack bodies. Without it \`findTargetCard\` falls through its \`switch\` to \`default\` and returns \`null\`, so the validator rejected every nine-on-jack the bot attempted. Both are fixed here.

api/helpers/game-states/is-top-jack.js
itsalaidbacklife: Applied in 19de2d55 — the optional chaining makes the empty-attachments case fall out for free, so the intermediate \`host\` lookup goes away:

```js
return exits.success(player.points.some(({ attachments }) => attachments.at(-1)?.id === targetId));
```

(your snippet was missing a closing paren on \`some(...)\`)

api/helpers/game-states/moves/one-off/validate.js
itsalaidbacklife: Done in 19de2d55, and it turns out to be a real fix rather than just tidier wording.

Twos were already removing the *exact* targeted jack (\`resolve/two.js\` tracks \`jackIndex\` through the inner loop), so at first glance buried jacks looked handled. But a jack's owner is never stored — \`convert-card-to-str.js\` derives it from the jack's position in the stack:

```js
let playerNum = card.attachments.length % 2 === 0 ? cardOwner : opponent;
```

Attachments alternate owners by construction, so removing a buried jack renumbers every jack above it and can flip control of the point card. So the restriction belongs on both ranks.

Renamed the key to \`game.snackbar.oneOffs.onlyTopJack\` (dropped the \`nine\` nesting) across all five locales. Worth noting the UI already offered only the top jack for both ranks — \`validFaceCardTargetIds\` pushes \`attachments[attachments.length - 1].id\` — so this closes a socket-level hole rather than changing what players can click. All 5 twos tests still pass.

api/helpers/game-states/moves/resolve/nine.js
itsalaidbacklife: Agreed — done in 19de2d55, now just \`targetCard.attachments.pop()\` with a comment naming the invariant and where it's enforced.

The guarantee holds through the counter window too: between validate and resolve the only legal moves are counter and resolve, neither of which touches jack stacks. And it is now enforced in both \`one-off/validate.js\` and \`seven-one-off/validate.js\`, for twos and nines alike.

api/helpers/game-states/moves/seven-one-off/validate.js
itsalaidbacklife: Same change applied here in 19de2d55 — see the reply on \`one-off/validate.js\` for why this is a correctness fix for twos and not just a wording change.

src/routes/game/GameView.vue
itsalaidbacklife: Already unused on \`main\` — I checked before deleting. The whole cluster was dead:

- \`nineTarget\` (the computed) is referenced nowhere, in this file or any other.
- \`nineTargetIndex\` is only ever *written* — \`null\` at init and \`null\` in \`clearOverlays\` — and only ever *read* inside \`nineTarget\`.
- \`targetType\` likewise: assigned \`null\` in both places, read only by \`nineTarget\`'s switch, so the switch could never take a branch.

\`git grep nineTarget main -- src/ tests/\` returns only those GameView lines, nothing else.

src/stores/game.js
itsalaidbacklife: Yes — on \`main\` the only consumers were the four transition computeds in this PR's diff:

```
main:src/routes/game/GameView.vue:558   switch (this.gameStore.lastEventTargetType) {
main:src/routes/game/GameView.vue:585   ... === 'faceCard'
main:src/routes/game/GameView.vue:615   switch (this.gameStore.lastEventTargetType) {
main:src/routes/game/GameView.vue:636   ... === 'faceCard'
```

Nothing in any other component, store, or test.

It was also permanently \`null\`, so those branches were unreachable: \`resolve/execute.js\` nulls \`oneOffTargetType\` before the socket event is built, and \`create-socket-events.js\` only puts that field on the game object, never inside \`lastEvent\` (which is what the store reads).

Flagging one interaction though: your transitions comment needs the opponent-points leave direction to differ between a topdecked point card and a jack's host reverting. I have a staging approach that avoids reintroducing this field — asking about it on that thread.

src/routes/home/components/announcementDialog/data/announcementData.js
itsalaidbacklife: Done in 19de2d55. Second section is now "The first of four Spades season trials":

> This is a beta, and the first of four rules variations we're trialling over the Spades season — two Nine buffs and two Four buffs. We'll put them to a community poll toward the end of the year to decide what stays. For the full context and reasoning behind all four, see **Help refine the 4 and the 9**

The link label is a new \`announcement.patreonLink\` key with the href in \`announcementData.js\`, translated in all five locales. Verified rendering end-to-end — dialog opens, both sections show, and the anchor carries the right \`href\`.

api/helpers/game-states/moves/resolve/execute.js
itsalaidbacklife: Checked — the log never referenced these on the fizzle path, so clearing them is safe.

\`MoveType.FIZZLE\` has its own message and it does not name the target:

```js
case MoveType.FIZZLE:
  return `The ${getFullCardName(resolved)} is countered, and all cards played this turn are scrapped.`;
```

The target only reaches the log via \`targetCard\`, which is set from \`oneOffTarget\` at the *end* of the successful path (\`targetCard: result.oneOffTarget\`). The fizzle branch returns before that, so \`targetCard\` stays whatever the pending-one-off row had — \`null\` — both before and after my change.

So nothing regressed. But it does mean a countered targeted one-off is logged without saying what it was aimed at, which is arguably a gap worth closing: "The 9♣ targeting the 10♦ is countered…". That needs both a copy change and storing \`targetCard\` on the fizzle row (where it currently means "card affected", and on a fizzle nothing was). Happy to do it here or split it out — asking which you'd prefer.

src/routes/game/GameView.vue
itsalaidbacklife: Built it — `processNines` in the store, mirroring `processThrees`.

One thing your spec surfaced that I want to flag, because it drove the shape: **a single transition name per render can't express both leave directions a nine needs.** In one rank-9 resolve, opponent points can lose a topdecked point card (should travel toward the deck) *and* lose a jack's host reverting to its owner (should travel down to the player). Same list, same event, opposite directions.

So it's staged, which also avoids reintroducing `lastEventTargetType`:

| stage | what changes | transition |
|---|---|---|
| 1 (1000ms) | target flips face down in place | `CARD_FLIP` |
| 2 (1000ms) | target leaves the field; deck count grows | `to-deck-*` |
| 3 | `updateGame` — a jack's host lands in player points | normal `SLIDE_UP` |

Each render now has one kind of change in it, so each gets the right direction.

The flip needed no `GameCard` changes at all — exactly as you guessed. Withholding `suit`/`rank` is enough for its existing `isBack` check (`!this.suit && !this.rank`), and the `CARD_FLIP` transition already wrapping those ``s does the rotation. Stage 1 is 1000ms to match `--duration-slow`; at my first attempt (600ms) the flip was visibly cut off mid-rotation. Data attributes still use the real card so selectors stay stable.

Directions per your spec, in `toDeckTransition(side, area)`:

- xs → points `to-deck-down`, faceCards `to-deck-down-left`
- above xs → player `to-deck-up-left`, opponent `to-deck-left`

Jacks follow their own point row — that meant wrapping `.jacks-container` in a `TransitionGroup` bound to the same name, since it was a plain `<div>` before and jacks never animated at all.

Verified by sampling the DOM every 120ms through a real topdeck:

```
 121ms  ace present, to-deck-left-move          (layout settling)
1080ms  to-deck-left-leave-active  x=-4.8px   opacity 0.95
1440ms                             x=-68.8px  opacity 0.28
1922ms                             x=-95.5px  opacity 0.006
2041ms  removed
```

Two regression tests added to `9_nines.spec.js` covering the flip plus the direction at each breakpoint.

src/routes/game/GameView.vue
itsalaidbacklife: Covered by the `processNines` staging — see the reply on the `playerPointsTransition` thread for the full approach and the measurements.

tests/e2e/specs/in-game/one-offs/9_nines.spec.js
itsalaidbacklife: Added `deckLength` to `assertGameState` (asserted in both the store and the DOM badge) and threaded it through 13 of the nines fixtures.

I did **not** add `topCard`/`secondCard` assertions, because they can't work from a player's perspective — flagging in case that changes your view. `create-socket-events.js` hides the deck:

```js
const hideDeck = () => {
  const isResolvingSeven = gameState.phase === GamePhase.RESOLVING_SEVEN;
  if (isResolvingSeven) {
    return [ ...gameState.deck.slice(0, 2), ...gameState.deck.slice(2).map(createHiddenCard) ];
  }
  return gameState.deck.map(createHiddenCard);
};
```

So `gameStore.topCard` is `{ isHidden: true }` outside a seven, and the existing `fixture.topCard` branch only passes today during `RESOLVING_SEVEN` or when the deck is empty.

What the spec asserts instead: the deck size, plus that **the next draw is exactly the topdecked card**. Since a draw is `deck.shift()`, drawing that specific card proves it was in position 0 — which is the property we actually care about, and it also demonstrates the tempo cost to the opponent.

If you'd rather have a literal top-card assertion, the clean way is a test-only `GET /api/test/gamestate` returning the unhidden state for `assertGameState` to read. Happy to add that — just didn't want to widen the test API without asking.

api/helpers/game-states/moves/seven-one-off/validate.js
```diff
@@ -76,6 +76,16 @@ module.exports = {
             throw new BadRequestError('Twos can only target royals or glasses');
           }
 
+          // Only the top attachment can be targeted. Each jack's owner is derived from its
+          // position in the stack, so removing a buried one silently reassigns the jacks above
+          // it -- and with them, control of the point card.
+          if (
```
itsalaidbacklife:
```suggestion
          // Only the top attachment can be targeted
          if (
```

tests/e2e/specs/in-game/one-offs/9_nines.spec.js
itsalaidbacklife: Confirmed — skipping the top card assertion. `assertGameState` takes an optional `deckLength` (checked against both the store and the deck badge) and it is threaded through the nines fixtures; deck *contents* stay unasserted, with the draw-back assertions carrying the top-of-deck proof. No test-only endpoint added.

src/plugins/sockets/inGameEvents.js
```diff
@@ -64,6 +64,8 @@ export async function handleInGameEvents(evData, newRoute = null) {
     case SocketEvent.RESOLVE:
       if (evData.game.lastEvent?.oneOff?.rank === 7) {
         await gameStore.processSevens(evData.game);
+      } else if (evData.game.lastEvent?.oneOff?.rank === 9) {
+        await gameStore.processNines(evData.targetCard, evData.game);
```
itsalaidbacklife: Let's switch() on the rank and use default for the else case

src/routes/game/GameView.vue
```diff
@@ -783,12 +770,21 @@ export default {
...
+     * to render its back (see its isBack check) and flips it via the existing CARD_FLIP
+     * transition. Data attributes keep using the real card so selectors stay stable.
+     */
+    faceDownWhenTopdecked(card) {
+      return this.isTopdeckedCard(card) ? { suit: undefined, rank: undefined } : card;
+    },
     showCustomSnackbarMessage(messageKey) {
       this.snackbarStore.alert(this.t(messageKey), 'base-dark');
     },
     clearOverlays() {
```
itsalaidbacklife, line 801: Is this used anywhere other than the one function below it? Seems like it'd be better as an inline variable in that function if so

src/sass/transitions.scss
```diff
@@ -138,3 +138,46 @@
...
+  position: absolute;
+  transition: opacity var(--duration-slow), transform var(--duration-slow);
+}
+
+.to-deck-down-move,
+.to-deck-down-left-move,
+.to-deck-up-left-move,
+.to-deck-left-move {
+  transition: transform var(--duration-slow);
+}
```
itsalaidbacklife: Why are these two blocks separate?  Wouldn’t it be simpler to combine them?

src/stores/game.js
```diff
@@ -374,6 +375,71 @@ export const useGameStore = defineStore('game', () => {
...
+
+    // Stage 2: it leaves the board toward the deck, and the deck grows to match
+    removeCardFromField(targetCard.id);
+    deck.value.unshift({ isHidden: true });
+    await sleep(1000);
+
+    // Stage 3: authoritative state. A topdecked jack returns its point card here, which is
+    // why this is a separate render from the exit above.
+    topdeckedCard.value = null;
+    updateGame(game);
```
itsalaidbacklife: Can the last two phases be combined and how would that look? Only one card is targeted so if that's a jack, then the jacks should exit towards the deck and the points should exit up/down towards the opponent. And if the targeted card is a point card than it should exit towards the deck and would be the only card to move.

If there is a reason that's not feasible, then please explain, and otherwise test it out locally, create videos of both two and three stage transitions on both jacks and point cards, and comment them here for comparison

api/helpers/game-states/moves/seven-one-off/validate.js
itsalaidbacklife: Already in — you committed this yourself in 9e365c30. The full rationale lives on the `one-off/validate.js` copy so it is stated once rather than twice.

src/plugins/sockets/inGameEvents.js
itsalaidbacklife: Done in e52a68b9:

```js
case SocketEvent.RESOLVE:
  switch (evData.game.lastEvent?.oneOff?.rank) {
    case 7:
      await gameStore.processSevens(evData.game);
      break;
    case 9:
      await gameStore.processNines(evData.targetCard, evData.game);
      break;
    default:
      gameStore.updateGame(evData.game);
  }
  break;
```

src/routes/game/GameView.vue
itsalaidbacklife: Two things here, so splitting them:

- `topdeckedCardId` was only ever read by `isTopdeckedCard` — inlined in e52a68b9, computed gone.
- `isTopdeckedCard` itself has to stay a method: the template calls it in the `is-glasses` bindings on both face-card rows, so a topdecked glasses eight falls through to `GameCard`'s `isBack` check and flips to a card back instead of holding the goggles art.

src/sass/transitions.scss
itsalaidbacklife: Combined in e52a68b9 — the four directions now come out of one `@each` instead of six repeated blocks:

```scss
$to-deck-offsets: (
  'down': translateY(96px),
  'down-left': translate(-96px, 96px),
  'up-left': translate(-96px, -96px),
  'left': translateX(-96px),
);

@each $direction, $offset in $to-deck-offsets {
  .to-deck-#{$direction}-leave-active { … }
  .to-deck-#{$direction}-move { … }
  .to-deck-#{$direction}-leave-to { opacity: 0; transform: $offset; }
}
```

To answer the literal question: `-leave-active` and `-move` cannot share one rule, because only `-leave-active` may be `position: absolute` — that is what lets the remaining cards reflow into the gap, whereas a `-move` element has to keep its place in the flow to slide there. They now sit next to each other per direction with a comment saying so, which I think was the real ask.

src/stores/game.js
itsalaidbacklife: You are right, and the reason I gave for three stages does not hold up. Collapsed to two in e52a68b9.

**Why the conflict was not real.** I claimed opponent points needed two leave directions at once — a topdecked point card toward the deck, a topdecked jack's point card back to its owner. But those leave through *different* TransitionGroups: the jack through `.jacks-container`, the point card through `.field-points`. I had bound the jacks container to `opponentPointsTransition`, which is what forced them to share a direction and manufactured the conflict. Giving the jacks container its own transition dissolves it, exactly as you described.

**What it is now.**

1. the target flips face down where it sits
2. the server state lands — the target leaves for the deck while a topdecked jack's point card heads back to its owner, in the same render

A second thing fell out of this. The old code asked "is a topdeck running?" and pointed *every* row at the deck. That was already wrong for the jack case: the point card entering the player's row got a to-deck name, and those classes are leave-only, so it had no enter animation to play. Each row now resolves its own direction from the zone the target sat in (`opponentJacks`, `playerPoints`, …), captured before the state lands. More precise, and it is what makes one render able to carry both directions.

It also deletes code: `removeCardFromField` existed only to give the old stage two something to animate, and the deck count now updates with the rest of the state instead of being nudged early. Net **-96 / +106** across four files, most of the additions being the per-zone computeds and comments.

**Verification.** `9_nines.spec.js` 18/18, including the two animation tests that assert the flip and the to-deck direction at each breakpoint.

Here is the frame that settles it — 2-stage, jack target, mid-transition. The point card and the face-down jack are both fading out of the opponent's row while the Ten has already arrived in the player's row, all in one render:

| | 3-stage (before) | 2-stage (after) |
|---|---|---|
| renders | flip → exit → state | flip → exit + state |
| jack target | jack leaves, *then* the point card moves a beat later | both move together |
| point target | identical to the eye | identical to the eye |
| local board mutation | yes (`removeCardFromField`) | none |

For the point-card target the two are visually indistinguishable, which is expected — only one card moves either way. The difference is only visible on a jack, where three stages inserted a beat between the jack leaving and the point card coming home; two stages has them move at once, which reads better and matches what actually happens in the rules.

I recorded all four for comparison — `jack-2stage`, `jack-3stage`, `point-2stage`, `point-3stage`, same scenarios in each pair — but GitHub has no API for attaching media to a comment, so I cannot post them inline. Sending them over chat; happy for you to drop them in here if useful.

itsalaidbacklife: Back to three stages in 420d7f2d — your call on how it reads, and the extra beat before the point card comes home is the whole point of the animation.

I reverted only the **timing**, not the whole commit, so the rest of that round survives: per-row direction from the zone the target sat in, the jacks lists on their own transitions, the `switch` in `inGameEvents`, the inlined `topdeckedCardId`, and the `@each` in the SCSS. `removeCardFromField` comes back, since stage two needs something to animate before the server state lands.

**A correction to my previous reply on this thread.** I wrote that pointing every row at the deck was "already wrong for the jack case — the point card entering the player's row got a to-deck name … so it had no enter animation to play." That is not true of the three-stage code. Stage three clears `topdeckedCard` *before* `updateGame`, so the point card always entered under the normal transition and animated fine. That bug only existed in the two-stage version I built, where the flag stayed set through `updateGame`. Sorry — I overstated it.

So per-zone is a tidiness win here, not a bug fix: only the row actually losing a card points at the deck, instead of all four. The behaviour is the same as what you saw in the recorded clips.

What I have **not** changed is your underlying point, which stands: the staging is not forced. A jack and its point card leave through separate TransitionGroups, so one render could carry both. The comment above `processNines` now says that plainly and frames three stages as a pacing choice rather than a constraint, so nobody later "fixes" it back.

`9_nines.spec.js` 18/18 and lint clean on the restore.

src/routes/game/GameView.vue
```diff
@@ -544,7 +542,32 @@ export default {
...
+     * across. Jacks follow their own point row.
+     */
+    toDeckTransition() {
+      return (side, area) => {
+        if (this.$vuetify.display.xs) {
+          return area === 'points' ? Transitions.TO_DECK_DOWN : Transitions.TO_DECK_DOWN_LEFT;
+        }
+        return side === 'player' ? Transitions.TO_DECK_UP_LEFT : Transitions.TO_DECK_LEFT;
+      };
+    },
```
itsalaidbacklife: Why does this return an anonymous function instead of inputting the args directly and returning what's currently the inner func's return value? How can that even work since the callers pass args directly to this func?

src/stores/game.js
```diff
@@ -374,6 +380,95 @@ export const useGameStore = defineStore('game', () => {
...
+      }
+
+      fieldPlayer.points?.forEach((pointCard) => {
+        const jackIndex = pointCard.attachments?.findIndex(({ id }) => id === cardId) ?? -1;
+        if (jackIndex > -1) {
+          pointCard.attachments.splice(jackIndex, 1);
+        }
+      });
+    });
+  }
```
itsalaidbacklife: Move this func def above processNines()

tests/e2e/specs/in-game/one-offs/9_nines.spec.js
```diff
@@ -579,39 +645,63 @@ describe('Playing NINES', () => {
...
-      // opponent plays nine to return seven to player's hand
+      // opponent plays nine to topdeck the player's seven
       cy.playTargetedOneOffOpponent(Card.NINE_OF_CLUBS, Card.SEVEN_OF_CLUBS, 'point');
 
       // Player resolves
       cy.get('#cannot-counter-dialog').should('be.visible')
         .get('[data-cy=cannot-counter-resolve]')
         .click();
 
+      // The seven did not come back to the player's hand
```
itsalaidbacklife:
```suggestion
      // The seven is back on top of the deck
```

src/routes/game/GameView.vue
itsalaidbacklife: No good reason — it was a plain mistake. Fixed in d1b373e1; it is a method taking `(side, area)` directly now.

To answer how it worked: `this.toDeckTransition` read the computed, which evaluated to the arrow function, and `('player', 'points')` then invoked *that*. So the call was really `(computedValue)('player', 'points')` — legal, but only because I had wrapped it. The wrapper existed purely to make an argument-taking helper callable from the `computed` block, which is the wrong block for it. Moving it to `methods` removes the need entirely.

Reactivity is unaffected: the four transition computeds call the method while they evaluate, so reading `$vuetify.display.xs` inside it still registers as their dependency. The two animation tests cover both breakpoints (`to-deck-left` above xs, `to-deck-down` at xs) and both pass, which is what I wanted to confirm on this one.

src/stores/game.js
itsalaidbacklife: Moved in d1b373e1 — `removeCardFromField` now sits directly after `findFieldZone`, so both helpers come before `processNines` rather than one above and one below it.

tests/e2e/specs/in-game/one-offs/9_nines.spec.js
itsalaidbacklife: Applied in d1b373e1. Better framing — it states the outcome instead of ruling one out, which is also what the assertion below it is really establishing.

PR #1382, opened 2026-09-18
fix(tests): assert the oauth identity link without comparing ids

### LINKED ISSUE #1381: [Bug]: OAuth unit tests compare User.id to Identity.id and fail after any Cypress run
### Bug Summary

`google.spec.js` and `discord.spec.js` assert that a user's id equals their identity's id — two unrelated auto-increment sequences. The assertion only holds while those sequences happen to coincide, so the six of them fail locally after anything creates users without identities (any Cypress run). CI never sees it.

### Detailed Description

Six assertions across the two OAuth specs take this form:

```js
expect(status.id).toEqual(status.identities[0].id);
```

`google.spec.js:57,76,114` and `discord.spec.js:57,76,114`.

`status.id` is the session's **`User`** id. `status.identities[0].id` is the **`Identity`** row's own primary key. They live in different tables with independent sequences and have no reason to match — the field that actually carries the linkage is `identities[0].user`, the foreign key.

**Why it usually passes.** Dev configures no datastore adapter, so `config/datastores.js` falls back to `sails-disk`, backed by NeDB files in `.tmp/localDiskDb/`. The dev server, Cypress and the Sails unit tests all lift with `environment: 'development'`, so all three share one on-disk database. `sails-disk` seeds each table's sequence at lift time from the highest id on disk, then holds it in memory (`node_modules/sails-disk/index.js`):

```js
// If there's a sequence for this table, then load the records in reverse PK order
// to get the last sequence value.
db.find({}).sort(sortObj).limit(1).exec(function(err, records) {
  if (records.length === 0) { return next(err); }   // no rows -> sequence stays 0
  datastore.sequences[sequenceName] = records[0][primaryKeyCol];
});
```

On a clean checkout both sequences start at 0 and advance in lockstep, because every user the OAuth specs create also gets exactly one identity. The assertion is then coincidentally true.

**Why it breaks.** Cypress creates dozens of users via username/password signup and essentially no identities, so `user.db`'s max id races ahead of `identity.db`'s. The next unit-test lift seeds the two sequences from those maxes and inherits the gap. `wipeDatabase()` in `beforeEach` deletes the rows but never lowers the in-memory counters, so the gap survives.

It does **not** self-heal: each failing run leaves its own users on disk to seed the next lift, preserving the offset.

**Why CI is green.** Each CI job starts from a fresh checkout with no `.tmp/localDiskDb/`, so both sequences begin at 0 and stay aligned. This only bites locally, where Cypress and the unit tests share a datastore.

### Current Behavior

Six failures in `npm run test:unit`, e.g. after a Cypress run that left 53 users on disk:

```
AssertionError: expected 53 to deeply equal 1
AssertionError: expected 54 to deeply equal 2
AssertionError: expected 56 to deeply equal 4
```

`User.id` resumes from Cypress's high-water mark while `Identity.id` starts from 1. Re-running does not help — the next run showed 57/58/60 vs 5/6/8, the same offset of 52.

The only workaround today is `rm -rf .tmp/localDiskDb`, which lasts until the next Cypress run.

### Expected Behavior

The OAuth specs assert that an identity was linked to the account, in a way that does not depend on id values, and pass regardless of what else has touched the shared datastore.

### Steps to Reproduce

1. `rm -rf .tmp/localDiskDb && npm run start:dev`
2. `npx cypress run --spec "tests/e2e/specs/out-of-game/home.spec.js"` (leaves ~53 users, 0 identities)
3. Stop the dev server
4. `npm run test:unit` → 6 failures in `google.spec.js` / `discord.spec.js`
5. Run it again → still fails, with both ids shifted up by 4

### Proposed fix

Replace the id comparison with assertions that express what those three tests are actually checking — that an OAuth identity got linked to the account:

```js
expect(status.identities).toHaveLength(1);
expect(status.identities[0].provider).toEqual('google');   // 'discord' in the other spec
```

Worth noting the obvious fix — swapping `.id` for `.user` — is correct but **vacuous**: `User.findOne({ id }).populate('identities')` already selects only rows whose `user` matches, so it can never fail. It would trade a fragile assertion for one that tests nothing.

The version above covers two things the original could not:

- **exactly one** identity. The current assertion only inspects `[0]`, so a duplicate-identity bug on repeat login passes silently — a real risk given the linking branch in `OAuthController`:
  ```js
  // If no identity but user is already logged in, create Identity
  if (!prevIdentity && loggedInUser) {
    await Identity.create({ provider, providerId: providerIdentity.id, user: loggedInUser, ... });
  ```
- **the right provider**, which matters now that Google and Discord share this code path; a cross-provider mix-up is currently invisible.

Leave `expect(status.id).toEqual(userRes.body)` (`:97`, `:113`) alone — comparing the session user to the id returned by signup is meaningful and correct.

### Secondary: `wipeDatabase` swallows errors

While in here, `api/helpers/wipe-database.js` reports success on failure:

```js
} catch (err) {
  return exits.success(false);
}
```

A partial or failed wipe returns `false` rather than erroring, and no caller checks the return value — so tests would run against a dirty database with no signal.

itsalaidbacklife:
## Issue number

- Resolves #1381

## What this fixes

Six assertions in `google.spec.js` and `discord.spec.js` compared a **`User` id** to an **`Identity` id**:

```js
expect(status.id).toEqual(status.identities[0].id);
```

Separate tables, independent sequences, no reason to match — the field carrying the link is `identities[0].user`. The assertion only held while the two sequences happened to coincide, so it fails locally after anything creates users without identities.

Full mechanism (shared `sails-disk` datastore, sequences seeded from the on-disk max at lift, `wipeDatabase` not lowering them) is written up in #1381.

## Why not just swap `.id` for `.user`

That stops the failures but asserts nothing. `User.findOne({ id }).populate('identities')` already selects only rows whose `user` matches the id being queried, so the comparison can never fail — it would trade a fragile assertion for a vacuous one.

Instead, assert what those three tests are actually checking, which is that an OAuth identity got linked to the account:

```js
expect(status.identities).toHaveLength(1);
expect(status.identities[0].provider).toEqual('google');   // 'discord' in the other spec
```

This also covers two things the original could not:

- **exactly one** identity — the old assertion only inspected `[0]`, so a duplicate-identity bug on repeat login passed silently, which is a live risk given the linking branch in `OAuthController`
- **the right provider** — a Google/Discord mix-up is currently invisible now that both share this code path

`expect(status.id).toEqual(userRes.body)` (4 call sites) is deliberately **left alone**: comparing the session user to the id returned by signup is meaningful and correct.

## Also included

`api/helpers/wipe-database.js` reported success when the wipe threw:

```js
} catch (err) {
  return exits.success(false);   // -> now exits.error(err)
}
```

No caller checks that return value, so a partial or failed wipe left every downstream test running against a dirty database with no signal. Flagged as secondary in #1381.

## Please check the following

- For New Features:

Test-only change plus one helper; the existing specs are the coverage.

### Verified against a live failing baseline

My `.tmp/localDiskDb/` was still carrying the drift from reproducing #1381, so there was a genuinely failing starting state rather than a synthetic one:

| | User max id | Identity max id | Result |
|---|---|---|---|
| `main`, drifted DB | 60 | 8 | **6 failed** — `65≠13`, `66≠14`, `68≠16`, `61≠9`, `62≠10`, `64≠12` |
| This branch, same DB | 78 | 8 | **10 passed** |

The gap *widened* to 70 during the run and the tests passed anyway, which is the point — the assertions no longer read id values. `npm run test:unit` then passed **86/86 on a first run** in the state that previously guaranteed 6 failures.

### Mutation-tested

Since my objection to the `.user` alternative was that it can't fail, I checked the replacement can:

| Mutation | Result |
|---|---|
| `provider` → `'discord'` in the google spec | fails ✓ |
| `toHaveLength(2)` | fails ✓ |

### Full suite

`wipeDatabase` runs at the top of nearly every spec, so the error-propagation change needed broad coverage rather than a spot check:

- `npm run lint` — clean
- `npm run test:unit` — 86 / 86
- Full `npx cypress run` — **34 specs, 375 passing, 0 failing**, 5 pending (pre-existing skips)

### Reproducing the original failure

1. `rm -rf .tmp/localDiskDb && npm run start:dev`
2. `npx cypress run --spec "tests/e2e/specs/out-of-game/home.spec.js"`
3. Stop the dev server
4. `npm run test:unit` — fails on `main`, passes on this branch


api/helpers/wipe-database.js
```diff
@@ -21,7 +21,9 @@ module.exports = {
       ]);
 
     } catch (err) {
-      return exits.success(false);
+      // Surface the failure. Reporting success here left callers running against a database
+      // that was only partially wiped, with no signal that anything had gone wrong.
```
itsalaidbacklife: Code comments shouldn't describe why _changes_ are made in a PR; they should describe why the code does what it does. No need for this comment post-merge, so we shouldn't keep it
```suggestion
```

